PR #556, opened 2023-09-05
Compile and test all crates by default.

chriseth:
Note that the halo2 dependency is still disabled through a feature.

The idea here is that the CLI depends on all crates anyway. So the only effect of the `default-members` property here is that `cargo test` only runs the cli tests, which is hardly what we want.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,10 +23,6 @@ members = [
     "type_check",
 ]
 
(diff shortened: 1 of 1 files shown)


Schaeff:
oops, updated main by merging by mistake, my bad

Schaeff (dismissed):
I think this is better as well

PR #558, opened 2023-09-05
Module system

Schaeff:
Similar to rust
- modules can contain modules, machines and imports, which each define a symbol (in the same namespace)
- `use` allows to import symbols from anywhere to anywhere
- `use super::foo` is supported to go up in the module tree
- paths inside of machines are relative to the module the machine is in:
```
mod A {
   use super::path::to::Machine as Submachine;
   machine Machine {
      Submachine sub; // `Submachine` is seen in the context of `A`, not in the context of this machine
   }
}
```

Implement as a separate `importer` crate with two passes on the parse tree
-  one pass by value, replacing `mod abc;` by `mod abc { // content of ./abc.asm or ./abc/mod.asm }`. This means that the importer currently needs fs access. This could be abstracted away later to an `ImportResolver` trait for no-fs use cases.
- two passes, the first one by reference to build a map of all relative paths to their canonical absolute path, and one by value to remove all imports and replace the relative paths by their canonical absolute path in the map


analysis/src/vm/batcher.rs
```diff
@@ -1,15 +1,17 @@
 //! Batch compatible statements in each function of each machine
```
Schaeff, line 1: A lot of these changes are just moving to using a visitor. I could potentially move this to a PR before the module system.

Schaeff:
@chriseth there are quite a few things to clean up but I'd appreciate a preliminary review

analysis/src/vm/batcher.rs
chriseth: Yes, please do that :)

Schaeff: I ended up reverting everything after type checking, as we can just keep the same structure as before, using absolute paths instead of names.

ast/src/parsed/asm.rs
```diff
@@ -1,21 +1,164 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct SymbolDefinition<T> {
     pub name: String,
+    pub symbol: Symbol<T>,
+}
+
+#[derive(Debug, PartialEq, Eq, From, Clone)]
+pub enum Symbol<T> {
```
chriseth, line 39: Shouldn't this be the symbol's value and just the name be the symbol?

ast/src/parsed/asm.rs
```diff
@@ -1,21 +1,164 @@
...
+
+#[derive(Debug, PartialEq, Eq, From, Clone)]
+pub enum Module<T> {
+    External(String),
+    Local(ASMModule<T>),
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Import {
+    // the path imported in the source
```
chriseth, line 53:
```suggestion
    /// the path imported in the source
```

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+    let paths = generate_path_map(&program)?;
+
+    let program = ASMProgram {
+        main: canonicalize_paths_in_module(AbsoluteSymbolPath::default(), program.main, &paths),
+    };
+
+    Ok(program)
+}
+
+// replace references to symbols with absolute paths. This removes the import statements. This always succeeds if the symbol table was generated correctly.
```
chriseth, line 23:
```suggestion
/// replace references to symbols with absolute paths. This removes the import statements. This always succeeds if the symbol table was generated correctly.
```

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+fn canonicalize_paths_in_module<T: FieldElement>(
+    absolute_path: AbsoluteSymbolPath,
```
chriseth, line 25: Is the the absolute path to the module? Then maybe `module_path` would be more suitable, we already know that it is an absolute path from it's type.

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    m
+}
+
+/// Answers the question: When at absolute path `p`, refering to relative path `r`, what is the absolute path to the canonical symbol imported?
+pub type PathMap = BTreeMap<(AbsoluteSymbolPath, SymbolPath), AbsoluteSymbolPath>;
+
+#[derive(PartialEq, Debug)]
+pub struct State<'a, T> {
```
chriseth, line 82: Can you document this? The name is quite generic.

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+}
+
+/// Checks a module
+///
+/// # Errors
+///
+/// This function will return an error if a name is not unique, or if any path in this module does not resolve to anything
+fn check_module<'a, T: Clone>(
+    location: AbsoluteSymbolPath,
+    module: &ASMModule<T>,
```
chriseth, line 202: Can we replace this by a trait so it can be used more generally?

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+    state: State<'a, T>,
+) -> Result<State<'a, T>, String> {
+    let _ = module.symbol_definitions().try_fold(
+        BTreeMap::<String, Option<Symbol<T>>>::default(),
+        |mut acc, SymbolDefinition { name, .. }| match acc.insert(name.clone(), None) {
+            Some(_) => Err(format!("duplicate name {name}")),
```
chriseth, line 208: Source locations would be nice :)

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+        },
+    )?;
+
+    module
+        .symbol_definitions()
+        // start with the initial state
+        .try_fold(state, |state, SymbolDefinition { name, symbol }| {
+            // update the state
+            match symbol {
+                Symbol::Machine(m) => check_machine(location.clone().join(name.clone()), m, state),
```
chriseth, line 219:
```suggestion
                Symbol::Machine(machine) => check_machine(location.clone().join(name.clone()), machine, state),
```

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+                Symbol::Module(m) => {
+                    let m = match m {
```
chriseth, line 221:
```suggestion
                    let module = match module {
```

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+///
+/// # Errors
+///
+/// This function will return an error if any of the paths does not resolve to anything
+fn check_machine<'a, T: Clone>(
+    location: AbsoluteSymbolPath,
+    m: &Machine<T>,
+    state: State<'a, T>,
+) -> Result<State<'a, T>, String> {
+    // we check the path in the context of the parent module
```
chriseth, line 242: Because the machine does not create a namespace?

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+        .iter()
+        .try_fold(
+            (
+                state,
+                AbsoluteSymbolPath::default(),
+                Symbol::Module(Module::Local(root)),
+            ),
+            |(state, mut location, symbol), member| {
+                match symbol {
+                    // machines do not expose symbols
```
chriseth, line 121: So I cannot even access a symbol within the same machine?

I think we should abstract that to some more generic terms like namespaces, scopes, etc.

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+
+#[derive(PartialEq, Debug)]
+pub struct State<'a, T> {
+    /// the root module of this program
+    root: &'a ASMModule<T>,
+    /// the paths to replace
+    pub paths: PathMap,
+}
+
+/// Checks a relative path in the context of an absolute path
```
chriseth, line 89: And especially, builds up the state. What does "check" mean here?

chriseth: So it's more like "resolves relative paths and stores the absolute paths in the state", right?

importer/src/path_canonicalizer.rs
Schaeff: exactly

ast/src/parsed/asm.rs
```diff
@@ -1,21 +1,164 @@
...
+pub struct SymbolDefinition<T> {
     pub name: String,
+    pub symbol: Symbol<T>,
+}
+
+#[derive(Debug, PartialEq, Eq, From, Clone)]
+pub enum Symbol<T> {
+    Machine(Machine<T>),
+    Import(Import),
+    Module(Module<T>),
```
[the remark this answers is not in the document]
Schaeff: Added, it's basically any module except the root module
```
mod A {
    mod B {

    }
}

"a" => SymbolValue::Module(Module { "b" => SymbolValue::Module(Module {}) })
```

importer/src/path_canonicalizer.rs
Schaeff: changed doc

importer/src/path_canonicalizer.rs
Schaeff: added docs

importer/src/path_canonicalizer.rs
Schaeff: What do you mean by a trait? Like a visitor for the parse tree?

importer/src/path_canonicalizer.rs
Schaeff: Added in the string now, eventually we should have a more structure error type.

importer/src/path_canonicalizer.rs
```diff
@@ -0,0 +1,378 @@
...
+                    Symbol::Machine(_) => {
+                        Err(format!("symbol not found in `{location}`: `{member}`"))
+                    }
+                    // modules expose symbols
+                    Symbol::Module(Module::Local(module)) => module
+                        .symbol_definitions()
+                        .find_map(|SymbolDefinition { name, symbol }| {
+                            (name == member).then_some(symbol.clone())
+                        })
+                        .ok_or_else(|| format!("symbol not found in `{location}`: `{member}`"))
```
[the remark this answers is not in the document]
Schaeff: Are you referring to `Option::ok_or_else`? It already wraps the error in `Err`:

```rust
fn ok_or_else<E, F>(self, err: F) where F: FnOnce() -> E -> Result<T, E>
```

analysis/src/macro_expansion.rs
```diff
@@ -35,29 +40,57 @@ impl<T> MacroExpander<T>
...
+        let statements = module
+            .statements
             .into_iter()
-            .map(|mut m| {
-                m.statements.iter_mut().for_each(|s| match s {
-                    MachineStatement::InstructionDeclaration(_, _, Instruction { body, .. }) => {
-                        match body {
-                            InstructionBody::Local(body) => {
-                                *body = self.expand_macros(std::mem::take(body))
+            .map(|mut s| {
```
chriseth, line 47: This would probably benefit from a visitor

importer/src/module_loader.rs
```diff
@@ -0,0 +1,151 @@
...
+            .map(|s| match s {
+                ModuleStatement::SymbolDefinition(SymbolDefinition { name, value }) => {
+                    let value = match value {
+                        SymbolValue::Module(m) => {
+                            let m = match m {
+                                Module::External(name) => path
+                                    .clone()
+                                    .map(|path| {
+                                        // for this, we skip the last part of the current location as if we are at `a::b::c` and declare `d`, we are looking as `a/b/d`
+                                        let path = path.parent().unwrap().join(name);
```
chriseth, line 37: This is a bit too far indented for my taste.

asm_to_pil/src/romgen.rs
```diff
@@ -391,7 +392,7 @@ _loop;
         let res = generate_rom_str::<Bn254Field>(vm);
 
         assert_eq!(
-            res.get("VM")
+            res.get(&AbsoluteSymbolPath::default().join("VM"))
```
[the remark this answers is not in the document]
Schaeff: done

Schaeff:
@chriseth I introduced a folder which simplifies some things. We could push it a bit further (folding statements should return an optional statement, so that we can implement filtering) but I think it's good enough for now

ast/src/parsed/folder.rs
```diff
@@ -0,0 +1,45 @@
+use super::asm::{
+    ASMModule, ASMProgram, Import, Machine, Module, ModuleStatement, SymbolDefinition, SymbolValue,
+};
+
+pub trait Folder<T> {
```
chriseth, line 5: Isn't this more like a `try_map`? For fold I would assume that you accumulate a list into a single item using a function that takes two arguments.

Anyway, could you please document a little?

chriseth: I'm wondering if a mutable visitor would not be better here.

chriseth: Ok, it's not a unified type like for expression, so a visitor would not make too much sense.

chriseth:
Could change the name of the Folder but I'm also fine with it.

Please squash.

chriseth (approved):

chriseth:
Can also squash merge.

PR #571, opened 2023-09-06
Add comparisons and logical operators.

chriseth:


georgwiese (approved):
Awesome! I tried applying this on #570 and it simplifies the code a lot, because I can implement byte comparison as a lookup and also get rid of the macros.

I have a few small suggestions, but no blockers. Also, it could make sense to go through `test_data` and simplify some examples (e.g. remove the `is_zero` macro in `block_lookup_or`). But that could also be done in a future PR.

chriseth:
Updated.

georgwiese (approved):
LGTM!

PR #572, opened 2023-09-06
Move FunctionCall.

chriseth:


ast/src/parsed/asm.rs
```diff
@@ -132,15 +132,3 @@ pub struct Param {
...
 }
-
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
-pub struct FunctionCall<T> {
-    pub id: String,
-    pub arguments: Vec<Expression<T>>,
-}
-
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
-pub enum PlookupOperator {
```
leonardoalt, line 143: So this isn't needed at all anymore?

chriseth: Yes, it was not used.

Schaeff (approved):

PR #585, opened 2023-09-09
Allow binding instructions to local functions

Schaeff:
For completeness and for the lolz


leonardoalt:
How does this actually work?

leonardoalt:
Do we still want this?

PR #586, opened 2023-09-11
Add CNAME file.

chriseth:


Schaeff:
sorry about confusion but seems like #587 is cleaner?

PR #589, opened 2023-09-11
Allow different path for vs code.

chriseth:


Schaeff:
What's the rationale for this change?

chriseth:
I think vs code runs the tests not from the crate dir but from the workspace.

Schaeff:
Is this the only occurrence of this? Also could it be maybe fixed by having a symlink to the test_data folder? Or is it dirty?

leonardoalt:
feels like a hack tbh

chriseth:
Seems to be this issue: https://github.com/rust-lang/rust-analyzer/issues/4705

leonardoalt (approved):

PR #592, opened 2023-09-11
Implement witgen for intermediate polynomials

Schaeff:
This is done by inlining the intermediate polynomials at the beginning of witgen.

Todo:
- [ ] more tests

### DIFF
--- a/compiler/tests/asm.rs
+++ b/compiler/tests/asm.rs
@@ -164,7 +164,6 @@ fn full_pil_constant() {
 }
 
 #[test]
-#[should_panic = "assertion failed: poly.is_fixed() || poly.is_witness()"]
 fn intermediate() {
     let f = "intermediate.asm";
     verify_asm::<GoldilocksField>(f, Default::default());
@@ -172,7 +171,6 @@ fn intermediate() {
 }
 
 #[test]
-#[should_panic = "assertion failed: poly.is_fixed() || poly.is_witness()"]
 fn intermediate_nested() {
     let f = "intermediate_nested.asm";
     verify_asm::<GoldilocksField>(f, Default::default());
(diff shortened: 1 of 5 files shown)


Schaeff:
Yeah that's what I meant by more tests, not sure the visitor works on nested

leonardoalt:
yea it was wrong, it needs to be post-visit

executor/src/witgen/util.rs
```diff
@@ -66,3 +69,71 @@ pub fn substitute_constants<T: Copy>(
...
+                }
+            }
+        }
+        std::ops::ControlFlow::Continue::<()>(())
+    });
+    expr
+}
+
+pub fn substitute_intermediate<T: Copy>(
+    identities: &[Identity<T>],
```
Schaeff, line 107: This could be passed as value to avoid cloning inside this function

executor/src/witgen/util.rs
```diff
@@ -66,3 +69,71 @@ pub fn substitute_constants<T: Copy>(
         })
         .collect()
 }
+
+fn inline_intermediates_in_expression<T: Copy>(
+    expr: &Expression<T>,
```
Schaeff, line 74: This could be passed as value to avoid cloning inside this function

Schaeff:
Need to do the same in halo2 and we're done

leonardoalt:
Implemented inlining before halo2 and squashed

leonardoalt:
@Schaeff I moved the inlining stuff into pil_analyzer now so that it can be used by both witgen and halo2 directly. pil_analyzer was already a dependency of these 2

leonardoalt (approved):

PR #553, opened 2023-09-01
Implementing the atomic instructions needed for revm

lvella:

PR #598, opened 2023-09-12
Make reg offset optional

chriseth:
Extracted from https://github.com/powdr-labs/powdr/pull/553


asm_utils/src/ast.rs
```diff
@@ -12,25 +12,25 @@ pub enum Statement<R: Register, F: FunctionOpKind> {
 #[derive(Clone, Debug)]
 pub enum Argument<R: Register, F: FunctionOpKind> {
     Register(R),
-    RegOffset(R, Expression<F>),
+    RegOffset(Option<Expression<F>>, R),
```
chriseth, line 15: I also swapped them so they match the source order.

leonardoalt (approved):

PR #614, opened 2023-09-19
Support types for declarations

chriseth:
TODO:

- [x] See if we can do anything with the types. Especially clarify a bit more about implicit conversions. Check that it really only happens for columns (i.e. not for lambda functions that happen to have type `col`).
Maybe we should instead add an explicit conversion call during analysis if the type turns out to be int instead of fe.

Whenever functions are called, we probably want to check the types. At least until we have a proper type checker, then we don't need that any more.

- [x] also document.
- [x] check if we can use this already to define arrays of fixed colmuns (and if yes, update arith_improved.pil)
- [x] let's try to even remove the implicit conversion of constants to field elements. It's just way too confusing.

### DIFF
--- a/ast/src/analyzed/types.rs
+++ b/ast/src/analyzed/types.rs
@@ -1,0 +1,138 @@
+use std::fmt::Display;
+
+use powdr_number::FieldElement;
+
+use crate::parsed::{ArrayTypeName, Expression, FunctionTypeName, TupleTypeName, TypeName};
+
+use super::Reference;
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct TypedExpression<T, Ref = Reference> {
+    pub e: Expression<T, Ref>,
+    pub ty: Option<Type>,
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub enum Type {
+    /// Boolean
+    Bool,
+    /// Integer (arbitrary precision)
+    Int,
+    /// Field element (unspecified field)
+    Fe,
+    /// String
+    String,
+    /// Algebraic expression
+    Expr,
+    /// Polynomial identity or lookup (not yet supported)
+    Constr,
+    Array(ArrayType),
+    Tuple(TupleType),
+    Function(FunctionType),
+}
+
+impl Type {
+    /// Returns the column type `int -> fe`.
+    pub fn col() -> Self {
+        Type::Function(FunctionType::col())
+    }
+
+    /// Returns true if the type name needs parentheses around it during formatting
+    /// when used inside a complex expression.
+    pub fn needs_parentheses(&self) -> bool {
+        match self {
+            Type::Bool
+            | Type::Int
+            | Type::Fe
+            | Type::String
+            | Type::Expr
+            | Type::Constr
+            | Type::Array(_)
+            | Type::Tuple(_) => false,
+            Type::Function(fun) => fun.needs_parentheses(),
+        }
+    }
+}
+
+impl<T: FieldElement, Ref: Display> From<TypeName<Expression<T, Ref>>> for Type {
+    fn from(value: TypeName<Expression<T, Ref>>) -> Self {
+        match value {
+            TypeName::Bool => Type::Bool,
+            TypeName::Int => Type::Int,
+            TypeName::Fe => Type::Fe,
+            TypeName::String => Type::String,
+            TypeName::Expr => Type::Expr,
+            TypeName::Constr => Type::Constr,
+            TypeName::Col => Type::col(),
+            TypeName::Array(ar) => Type::Array(ar.into()),
+            TypeName::Tuple(tu) => Type::Tuple(tu.into()),
+            TypeName::Function(fun) => Type::Function(fun.into()),
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct ArrayType {
+    pub base: Box<Type>,
+    pub length: Option<u64>,
+}
+
+impl<T: FieldElement, Ref: Display> From<ArrayTypeName<Expression<T, Ref>>> for ArrayType {
+    fn from(name: ArrayTypeName<Expression<T, Ref>>) -> Self {
+        let length = name.length.as_ref().map(|l| {
+            if let Expression::Number(n) = l {
+                n.to_degree()
+            } else {
+                panic!(
+                    "Array length expression not resolved in type name prior to conversion: {name}"
+                );
+            }
+        });
+        ArrayType {
+            base: Box::new(Type::from(*name.base)),
+            length,
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct TupleType {
+    pub items: Vec<Type>,
+}
+
+impl<T: FieldElement, Ref: Display> From<TupleTypeName<Expression<T, Ref>>> for TupleType {
+    fn from(value: TupleTypeName<Expression<T, Ref>>) -> Self {
+        TupleType {
+            items: value.items.into_iter().map(Into::into).collect(),
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct FunctionType {
+    pub params: Vec<Type>,
+    pub value: Box<Type>,
+}
+
+impl FunctionType {
+    /// Returns the column type `int -> fe`.
+    pub fn col() -> Self {
+        FunctionType {
+            params: vec![Type::Int],
+            value: Box::new(Type::Fe),
+        }
+    }
+
+    pub fn needs_parentheses(&self) -> bool {
+        *self != Self::col()
+    }
+}
+
+impl<T: FieldElement, Ref: Display> From<FunctionTypeName<Expression<T, Ref>>> for FunctionType {
+    fn from(name: FunctionTypeName<Expression<T, Ref>>) -> Self {
+        FunctionType {
+            params: name.params.into_iter().map(Into::into).collect(),
+            value: Box::new(Type::from(*name.value)),
+        }
+    }
+}
--- a/ast/src/parsed/mod.rs
+++ b/ast/src/parsed/mod.rs
@@ -24,7 +24,12 @@ pub enum PilStatement<T> {
     Include(SourceRef, String),
     /// Name of namespace and polynomial degree (constant)
     Namespace(SourceRef, SymbolPath, Expression<T>),
-    LetStatement(SourceRef, String, Option<Expression<T>>),
+    LetStatement(
+        SourceRef,
+        String,
+        Option<TypeName<Expression<T>>>,
+        Option<Expression<T>>,
+    ),
     PolynomialDefinition(SourceRef, String, Expression<T>),
     PublicDeclaration(
         SourceRef,
@@ -68,7 +73,7 @@ impl<T> PilStatement<T> {
             | PilStatement::PolynomialConstantDefinition(_, name, _)
             | PilStatement::ConstantDefinition(_, name, _)
             | PilStatement::PublicDeclaration(_, name, _, _, _)
-            | PilStatement::LetStatement(_, name, _) => Box::new(once(name)),
+            | PilStatement::LetStatement(_, name, _, _) => Box::new(once(name)),
             PilStatement::PolynomialConstantDeclaration(_, polynomials)
             | PilStatement::PolynomialCommitDeclaration(_, polynomials, _) => {
                 Box::new(polynomials.iter().map(|p| &p.name))
@@ -98,8 +103,11 @@ impl<T> PilStatement<T> {
             | PilStatement::Namespace(_, _, e)
             | PilStatement::PolynomialDefinition(_, _, e)
             | PilStatement::PolynomialIdentity(_, e)
-            | PilStatement::ConstantDefinition(_, _, e)
-            | PilStatement::LetStatement(_, _, Some(e)) => Box::new(once(e)),
+            | PilStatement::ConstantDefinition(_, _, e) => Box::new(once(e)),
+
+            PilStatement::LetStatement(_, _, type_name, value) => {
+                Box::new(type_name.iter().flat_map(|t| t.expressions()).chain(value))
+            }
 
             PilStatement::PublicDeclaration(_, _, _, i, e) => Box::new(i.iter().chain(once(e))),
 
@@ -107,8 +115,7 @@ impl<T> PilStatement<T> {
             | PilStatement::PolynomialCommitDeclaration(_, _, Some(fundef)) => fundef.expressions(),
             PilStatement::PolynomialCommitDeclaration(_, _, None)
             | PilStatement::Include(_, _)
-            | PilStatement::PolynomialConstantDeclaration(_, _)
-            | PilStatement::LetStatement(_, _, None) => Box::new(empty()),
+            | PilStatement::PolynomialConstantDeclaration(_, _) => Box::new(empty()),
         }
     }
 
@@ -126,8 +133,14 @@ impl<T> PilStatement<T> {
             | PilStatement::Namespace(_, _, e)
             | PilStatement::PolynomialDefinition(_, _, e)
             | PilStatement::PolynomialIdentity(_, e)
-            | PilStatement::ConstantDefinition(_, _, e)
-            | PilStatement::LetStatement(_, _, Some(e)) => Box::new(once(e)),
+            | PilStatement::ConstantDefinition(_, _, e) => Box::new(once(e)),
+
+            PilStatement::LetStatement(_, _, type_name, value) => Box::new(
+                type_name
+                    .iter_mut()
+                    .flat_map(|t| t.expressions_mut())
+                    .chain(value),
+            ),
 
             PilStatement::PublicDeclaration(_, _, _, i, e) => Box::new(i.iter_mut().chain(once(e))),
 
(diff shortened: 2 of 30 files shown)


ast/src/parsed/utils.rs
```diff
@@ -50,8 +51,10 @@ where
         | PilStatement::PolynomialDefinition(_, _, e)
         | PilStatement::PolynomialIdentity(_, e)
         | PilStatement::PublicDeclaration(_, _, _, e)
-        | PilStatement::ConstantDefinition(_, _, e)
-        | PilStatement::LetStatement(_, _, Some(e)) => postvisit_expression_mut(e, f),
+        | PilStatement::ConstantDefinition(_, _, e) => postvisit_expression_mut(e, f),
+        // TODO we are not visiting the expressions inside the type name here.
```
chriseth, line 55: still need to fix this

ast/src/parsed/mod.rs
```diff
@@ -413,3 +424,45 @@ impl<T> ArrayExpression<T> {
         }
     }
 }
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub enum TypeName<E> {
+    Int,
+    Fe,
+    Col,
```
lvella, line 432: What is Fe? What is Int? What is col as a type?

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -191,7 +191,7 @@ impl<T: FieldElement> PILContext<T> {
             PilStatement::ConstantDefinition(_, name, value) => {
                 self.handle_constant_definition(name, value)
             }
-            PilStatement::LetStatement(start, name, None) => {
+            PilStatement::LetStatement(start, name, _type_name, None) => {
```
lvella, line 194: Maybe better to use just '_'? I remember there is a difference (but I am sure the optimizer makes it irrelevant).

ast/src/parsed/utils.rs
chriseth: I'm tempted to implement @Schaeff 's version of the visitor before finishing this pr...

chriseth:
Reminder to myself: Once we have types for columns, we can simplify a lot of tests in constant_evaluator again so that we don't need the conversion to fe.

std/array.asm
```diff
@@ -13,4 +13,5 @@ let map = |arr, f| new(len(arr), |i| f(arr[i]));
 let fold = |arr, initial, folder| std::utils::fold(len(arr), |i| arr[i], initial, folder);
 
 /// Returns the sum of the array elements.
+/// TODO we need generic types in order to remove the [..][0]-workaround.
```
leonardoalt, line 16: Can you write a comment somewhere saying why it's not possible atm?

std/debug.asm
```diff
@@ -2,6 +2,6 @@
 /// when evaluated.
 /// It returns an empty array so that it can be used at constraint level.
 /// This symbol is not an empty array, the actual semantics are overridden.
-let print = [];
+let print: string -> int[] = [];
 
-let println = [|msg| print(msg + "\n")][0];
\ No newline at end of file
+let println: string -> int[] = |msg| print(msg + "\n");
```
leonardoalt, line 7: Why is the return type `int[]`?

test_data/pil/book/generic_to_algebraic.pil
```diff
@@ -6,20 +6,17 @@ namespace Main(16);
...
+    // This function takes an algebraic expression (a column or expression
+    // involving columns) and returns an identity that forces this expression
+    // to equal 20.
+    let equals_twenty: expr -> identity = |x| x == 20;
+    // This declares an array of 16 witness columns.
     col witness wit[16];
-    // This expression has to evaluate to a constraint, but we can still use
+    // This expression has to evaluate to an identity, but we can still use
     // higher order functions and all the flexibility of the language.
     // The sub-expression "sum(16, |i| wit[i]" evaluates to the algebraic
```
leonardoalt, line 17:
```suggestion
    // The sub-expression `sum(16, |i| wit[i])` evaluates to the algebraic
```

test_data/pil/book/generic_to_algebraic.pil
```diff
@@ -6,20 +6,17 @@ namespace Main(16);
...
     // expression "wit[0] + wit[1] + ... + wit[15]", which is then
-    // turned by "equals_twenty" into the constraint
+    // turned by "equals_twenty" into the identity
```
leonardoalt, line 19:
```suggestion
    // turned into the identity by `equals_twenty`
```

leonardoalt (commented):
Some initial comments on the resulting syntax

ast/src/parsed/mod.rs
```diff
@@ -496,3 +508,142 @@ impl<T> ArrayExpression<T> {
...
+                .chain(once(&self.value).flat_map(|v| v.expressions())),
+        )
+    }
+    /// Returns an iterator over all (top-level) expressions in this type name.
+    pub fn expressions_mut(&mut self) -> Box<dyn Iterator<Item = &mut E> + '_> {
+        Box::new(
+            self.params
+                .iter_mut()
+                .flat_map(|t| t.expressions_mut())
+                .chain(once(&mut self.value).flat_map(|v| v.expressions_mut())),
```
chriseth, line 640: fix

pacheco (commented):
left some suggestions, comments and questions

std/debug.asm
```diff
@@ -2,6 +2,6 @@
 /// when evaluated.
 /// It returns an empty array so that it can be used at constraint level.
 /// This symbol is not an empty array, the actual semantics are overridden.
-let print = [];
+let print: string -> constr[] = [];
```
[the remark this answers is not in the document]
chriseth: Not without changing the code.

ast/src/parsed/mod.rs
```diff
@@ -496,3 +508,141 @@ impl<T> ArrayExpression<T> {
...
+    Bool,
+    /// Integer (arbitrary precision)
+    Int,
+    /// Field element (unspecified field)
+    Fe,
+    /// String
+    String,
+    /// Column, shorthand for "int -> fe"
+    Col,
+    /// (Algebraic) expression
```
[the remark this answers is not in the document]
chriseth: Because the algebraic part is not represented in the name. I can remove the parentheses.

ast/src/analyzed/types.rs
```diff
@@ -0,0 +1,139 @@
...
+    Expr,
+    /// Polynomial identity or lookup (not yet supported)
+    Constr,
+    Array(ArrayType),
+    Tuple(TupleType),
+    Function(FunctionType),
+}
+
+impl Type {
+    pub fn col() -> Self {
```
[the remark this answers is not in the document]
chriseth: I actually don't think that these really do allocate if they are only used with comparisons. I tried with `is_col`, but I actually think `== Type::col()` reads better and we need this constructor anyway.

std/array.asm
```diff
@@ -13,4 +13,5 @@ let map = |arr, f| new(len(arr), |i| f(arr[i]));
 let fold = |arr, initial, folder| std::utils::fold(len(arr), |i| arr[i], initial, folder);
 
 /// Returns the sum of the array elements.
-let sum = [|arr| fold(arr, 0, |a, b| a + b)][0];
\ No newline at end of file
+/// This actually also works on field elements, so the type is currently too restrictive.
+let sum: int[] -> int = [|arr| fold(arr, 0, |a, b| a + b)][0];
```
leonardoalt, line 17: which PR removes this workaround?

test_data/pil/arith_improved.pil
```diff
@@ -29,8 +29,7 @@ namespace Arith(N);
     /// returns f(0) + f(1) + ... + f(length - 1)
     let sum = |length, f| fold(length, f, 0, |acc, e| acc + e);
 
-    // TODO the weird syntax is needed so that this is not classified as a constant column
-    let force_boolean = (|| |x| x * (1 - x) == 0)();
+    let force_boolean: expr -> constr = |x| x * (1 - x) == 0;
```
leonardoalt, line 32: actually, why is a `constr` not just an `expr`?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -505,6 +505,58 @@ namespace N(16);
...
+        let expected = r#"namespace N(16);
+    let f: expr, expr -> constr[] = (|x, y| [(x == y)]);
+    let g: expr -> constr[1] = (|x| [(x == 0)]);
+    col witness x;
+    col witness y;
+    N.x = N.y;
+    N.x = 0;
 "#;
         let formatted = analyze_string::<GoldilocksField>(input).to_string();
         assert_eq!(formatted, expected);
```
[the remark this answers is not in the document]
chriseth: Bullshit.

std/array.asm
chriseth: I think we could remove it here.

test_data/pil/arith_improved.pil
chriseth: because it compares to zero. If you want to keep everything an expression (implicitly assuming it is equated to zero in the end) you can also return an expression, but I think it is cleaner like this.

leonardoalt: so a `constr` is an expr that = 0?

ast/src/analyzed/types.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub enum Type {
+    /// Boolean
+    Bool,
+    /// Integer (arbitrary precision)
+    Int,
+    /// Field element (unspecified field)
+    Fe,
```
leonardoalt, line 22: eventually we could also support specifying the field in the type, like `(BV 256), (BV 32)` in smtlib2

ast/src/analyzed/types.rs
```diff
@@ -0,0 +1,141 @@
...
+        match self {
+            Type::Bool
+            | Type::Int
+            | Type::Fe
+            | Type::String
+            | Type::Expr
+            | Type::Constr
+            | Type::Array(_)
+            | Type::Tuple(_) => false,
+            Type::Function(fun) => fun.needs_parantheses(),
```
leonardoalt, line 52:
```suggestion
            Type::Function(fun) => fun.needs_parentheses(),
```

ast/src/analyzed/types.rs
```diff
@@ -0,0 +1,141 @@
...
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct TupleType {
+    pub items: Vec<Type>,
+}
+
+impl<T: FieldElement, Ref: Display> From<TupleTypeName<Expression<T, Ref>>> for TupleType {
+    fn from(value: TupleTypeName<Expression<T, Ref>>) -> Self {
+        TupleType {
+            items: value.items.into_iter().map(Into::into).collect(),
```
leonardoalt, line 109: empty allowed?

ast/src/analyzed/types.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct FunctionType {
+    pub params: Vec<Type>,
+    pub value: Box<Type>,
```
leonardoalt, line 117: what about multi return?

leonardoalt: ah I guess that's just a tuple

ast/src/analyzed/types.rs
```diff
@@ -0,0 +1,141 @@
...
+impl FunctionType {
+    /// Returns the column type `int -> fe`.
+    pub fn col() -> Self {
+        FunctionType {
+            params: vec![Type::Int],
+            value: Box::new(Type::Fe),
+        }
+    }
+
+    pub fn needs_parantheses(&self) -> bool {
```
leonardoalt, line 129:
```suggestion
    pub fn needs_parentheses(&self) -> bool {
```

ast/src/analyzed/types.rs
```diff
@@ -0,0 +1,141 @@
...
+        match value {
+            TypeName::Bool => Type::Bool,
+            TypeName::Int => Type::Int,
+            TypeName::Fe => Type::Fe,
+            TypeName::String => Type::String,
+            TypeName::Expr => Type::Expr,
+            TypeName::Constr => Type::Constr,
+            TypeName::Col => Type::Function(FunctionType {
+                params: vec![Type::Int],
+                value: Box::new(Type::Fe),
```
leonardoalt, line 68: couldn't you use `FunctionType::col()` here? Looks duplicated

test_data/pil/arith_improved.pil
chriseth: a constr is a `expr  = expr`.

chriseth: https://docs.powdr.org/pil/types.html#constraints

ast/src/parsed/mod.rs
```diff
@@ -496,3 +508,141 @@ impl<T> ArrayExpression<T> {
...
+    /// Column, shorthand for "int -> fe"
+    Col,
+    /// Algebraic expression
+    Expr,
+    /// Polynomial identity
+    Constr,
+    Array(ArrayTypeName<E>),
+    Tuple(TupleTypeName<E>),
+    Function(FunctionTypeName<E>),
+}
```
leonardoalt, line 531: Why can't we use the other Type type?

ast/src/parsed/mod.rs
```diff
@@ -496,3 +508,141 @@ impl<T> ArrayExpression<T> {
...
+            TypeName::Function(f) => f.expressions_mut(),
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub struct ArrayTypeName<E> {
+    pub base: Box<TypeName<E>>,
+    pub length: Option<E>,
+}
```
leonardoalt, line 588: Same here

importer/src/path_canonicalizer.rs
```diff
@@ -79,9 +80,14 @@ impl<'a, T> Folder<T> for Canonicalizer<'a> {
...
-                            SymbolValue::Expression(mut e) => {
-                                canonicalize_inside_expression(&mut e, &self.path, self.paths);
-                                Some(Ok(SymbolValue::Expression(e)))
+                            SymbolValue::Expression(mut exp) => {
+                                for tne in
+                                    exp.type_name.iter_mut().flat_map(|tn| tn.expressions_mut())
+                                {
+                                    canonicalize_inside_expression(tne, &self.path, self.paths);
+                                }
+                                canonicalize_inside_expression(&mut exp.e, &self.path, self.paths);
```
leonardoalt, line 89: What is this doing?

pil-analyzer/src/evaluator.rs
```diff
@@ -686,7 +689,7 @@ mod test {
     #[test]
     pub fn capturing() {
         let src = r#"namespace Main(16);
-            let f = |n, g| match n { 99 => |i| n, 1 => g(3) };
+            let f: int, int -> int = |n, g| match n { 99 => |i| n, 1 => g(3) };
```
leonardoalt, line 692: I don't understand the type of the RHS. Aren't the types of the match arms different? `T -> T` and `T`

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -505,6 +505,58 @@ namespace N(16);
...
+        assert_eq!(formatted, expected);
+    }
+
+    #[test]
+    fn complex_type_resolution() {
+        let input = r#"namespace N(16);
+    let f: int -> int = |i| i + 10;
+    let x: (int -> int), int -> int = |k, i| k(2**i);
+    let y: (int -> fe)[x(f, 2)];
+    let z: (((int -> int), int -> int)[x(|i| i, 3)], col) = ([x, x, x], y[0]);
```
leonardoalt, line 531: so an array is also a function?

ast/src/analyzed/types.rs
chriseth: True, but I kind of like this, since we want to be generic. Of course, it could be `Fe<n>` as well or even `Fe<n: n >= 2**32>`.

ast/src/analyzed/types.rs
chriseth: Yes, on purpose.

ast/src/parsed/mod.rs
chriseth: You mean why do we have to distinguish between TypeName and Type? The main reason is that TypeName has an Expression as array length while Type has a concrete number. Also when we will have user-defined type, TypeName for a struct will just be a string and the Type will be the full struct type.

importer/src/path_canonicalizer.rs
chriseth: It turns relative name paths to absolute using the import/use statements.

pil-analyzer/src/evaluator.rs
chriseth: true, this does not type check. Will fix.

pil-analyzer/src/pil_analyzer.rs
chriseth: no, it should not. will recheck the example. This test is just about parsing the type names, evaluating the array legths and printing it again. the values are not touched (except by the array length evaluator).

ast/src/analyzed/types.rs
leonardoalt: it could be a const generic, but yea future stuff

leonardoalt (approved):

leonardoalt (approved):

PR #617, opened 2023-09-19
Parsing .insn assembly directive.

lvella:
Dismembered from PR #553.

### DIFF
--- a/riscv/src/lib.rs
+++ b/riscv/src/lib.rs
@@ -329,3 +329,74 @@ fn output_files_from_cargo_build_plan(
 
     assemblies
 }
+
+/// Maps an instruction in .insn syntax to Statement::Instruction() in the expected format.
+///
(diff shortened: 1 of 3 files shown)


riscv/src/lib.rs
```diff
@@ -329,3 +329,62 @@ fn output_files_from_cargo_build_plan(
...
+        }
+        (0b1110011, 0b001) => "csrrw",
+        (0b1110011, 0b010) => "csrrs",
+        (0b1110011, 0b011) => "csrrc",
+        // won't interpret "csr*i" instructions because it is too weird to
+        // encode an immediate as a register
+        _ => panic!("unsupported .insn instruction"),
+    };
+
+    let args = vec![
```
chriseth, line 383: At least for ecall and ebreak, those should be empty. Maybe also for some others?

lvella: Yeah... also both fences. Fixed.

chriseth (approved):

PR #619, opened 2023-09-19
lh instruction

lvella:
Dismembered from PR #553.

Built upon PR #617 because it changes `riscv_asm.larlpop` and #617 fixes the larlpop build.

### DIFF
--- a/riscv/src/compiler.rs
+++ b/riscv/src/compiler.rs
@@ -70,9 +70,8 @@ impl Architecture for RiscvArchitecture {
             | "srli" | "srl" | "srai" | "seqz" | "snez" | "slt" | "slti" | "sltu" | "sltiu"
             | "sgtz" | "beq" | "beqz" | "bgeu" | "bltu" | "blt" | "bge" | "bltz" | "blez"
             | "bgtz" | "bgez" | "bne" | "bnez" | "jal" | "jalr" | "call" | "ecall" | "ebreak"
-            | "lw" | "lb" | "lbu" | "lhu" | "sw" | "sh" | "sb" | "nop" | "fence" | "fence.i" => {
-                false
-            }
+            | "lw" | "lb" | "lbu" | "lh" | "lhu" | "sw" | "sh" | "sb" | "nop" | "fence"
+            | "fence.i" => false,
             "j" | "jr" | "tail" | "ret" | "unimp" => true,
             _ => {
                 panic!("Unknown instruction: {instr}");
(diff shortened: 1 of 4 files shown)


riscv/src/riscv_asm.lalrpop
```diff
@@ -189,6 +189,6 @@ Symbol: String = {
 }
 
 Number: i64 = {
-    r"-?[0-9][0-9_]*" => i64::from_str(<>).unwrap().into(),
-    r"0x[0-9A-Fa-f][0-9A-Fa-f_]*" => i64::from_str_radix(&<>[2..].replace('_', ""), 16).unwrap().into(),
+    r"-?[0-9][0-9_]*" => i64::from_str(<>).unwrap(),
+    r"0x[0-9A-Fa-f][0-9A-Fa-f_]*" => u64::from_str_radix(&<>[2..].replace('_', ""), 16).unwrap() as i64,
```
chriseth, line 193: I still don't understand this, why not parse as i64 right away? As it is now, we parse a big 64 bit number and make it negative without noticing. Does this happen?

riscv/src/compiler.rs
```diff
@@ -1277,6 +1289,23 @@ fn process_instruction(instr: &str, args: &[Argument]) -> Vec<String> {
...
+            // Assumes the address is a multiple of two.
+            let (rd, rs, off) = rro(args);
+            only_if_no_write_to_zero_vec(
+                vec![
+                    format!("tmp1 <== wrap({rs} + {off});"),
+                    "addr <== and(tmp1, 0xfffffffc);".to_string(),
+                    "tmp2 <== and(tmp1, 0x3);".to_string(),
+                    format!("{rd} <== mload();"),
+                    format!("{rd} <== shr({rd}, 8 * tmp2);"),
+                    format!("{rd} <== and({rd}, 0x0000ffff);"),
```
chriseth, line 1303: I think this can be removed.

lvella: In lhu too?

lvella: Oh, right!

riscv/src/riscv_asm.lalrpop
lvella: No, currently `unwrap()` panics if number is > 2**63.

lvella: (what makes sense, as I am parsing i64 and the number is beyond max)

riscv/src/compiler.rs
lvella: fixed

chriseth (approved):

PR #620, opened 2023-09-19
amoadd.w instruction

lvella:
Dismembered from PR #553.

Built upon PR #619 because the test file contains 64-bit hex literals, and #619 fixes parsing of 64-bit hex literals.


riscv/src/compiler.rs
```diff
@@ -1391,6 +1403,24 @@ fn process_instruction(instr: &str, args: &[Argument]) -> Vec<String> {
         // atomic and synchronization
         "fence" | "fence.i" => vec![],
 
+        insn if insn.starts_with("amoadd.w") => {
+            let (rd, rs2, rs1, off) = rrro(args);
```
leonardoalt, line 1407: image

Not sure the order is correct here? It would help to keep the register names in the same order as the spec, if I understand the spec correctly, that being `(rs1, rs2, rd, off)`.

lvella: The order in the binary spec has nothing to do with the order in asm. Got it wrong in another instruction because it seems there is no authoritative documentation, but it was caught by the RISC-V test suite.

This one passes, and the one source I found says it is the way I did, so I am confident:
https://msyksphinz-self.github.io/riscv-isadoc/html/rva.html#amoadd-w

leonardoalt: ok, makes sense

leonardoalt (approved):

PR #621, opened 2023-09-19
lr/sc instructions

lvella:
Dismembered from PR #553.

Built upon PRs #615 and #620, because the test file relies on instructions introduced in these PRs.

### DIFF
--- a/riscv/src/compiler.rs
+++ b/riscv/src/compiler.rs
@@ -72,7 +72,9 @@ impl Architecture for RiscvArchitecture {
             | "sltiu" | "sgtz" | "beq" | "beqz" | "bgeu" | "bltu" | "blt" | "bge" | "bltz"
             | "blez" | "bgtz" | "bgez" | "bne" | "bnez" | "jal" | "jalr" | "call" | "ecall"
             | "ebreak" | "lw" | "lb" | "lbu" | "lh" | "lhu" | "sw" | "sh" | "sb" | "nop"
-            | "fence" | "fence.i" | "amoadd.w.rl" | "amoadd.w" => false,
+            | "fence" | "fence.i" | "amoadd.w" | "amoadd.w.aq" | "amoadd.w.rl"
+            | "amoadd.w.aqrl" | "lr.w" | "lr.w.aq" | "lr.w.rl" | "lr.w.aqrl" | "sc.w"
+            | "sc.w.aq" | "sc.w.rl" | "sc.w.aqrl" => false,
             "j" | "jr" | "tail" | "ret" | "unimp" => true,
             _ => {
                 panic!("Unknown instruction: {instr}");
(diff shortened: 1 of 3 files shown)


riscv/src/compiler.rs
```diff
@@ -1428,6 +1433,40 @@ fn process_instruction(instr: &str, args: &[Argument]) -> Vec<String> {
             ]
         }
 
+        insn if insn.starts_with("lr.w") => {
+            // Very similar to "lw":
```
chriseth, line 1437: Can we somehow add an assertion about what characters come after `"lr.w"` in `insn`?

riscv/src/compiler.rs
```diff
@@ -1428,6 +1433,40 @@ fn process_instruction(instr: &str, args: &[Argument]) -> Vec<String> {
...
+            let (rd, rs, off) = rro(args);
+            assert_eq!(off, 0);
+            // TODO misaligned access should raise misaligned address exceptions
+            only_if_no_write_to_zero_vec(
+                vec![
+                    format!("addr <=X= {rs};"),
+                    format!("{rd} <== mload();"),
```
chriseth, line 1444: The standard says you have to sign-extend the value, or did I read that wrong?

riscv/src/compiler.rs
```diff
@@ -1428,6 +1433,40 @@ fn process_instruction(instr: &str, args: &[Argument]) -> Vec<String> {
...
+                    // For consistency sake, I tried to use format!() here, but
+                    // the lint doesn't like it
+                    "lr_sc_reservation <=X= 1;".into(),
```
chriseth, line 1447: Strictly, we should have one reservation bit per memory address (if I read the standard correctly). If we don't have that, maybe we could scrap the reservation altogether, or would some code break in that case?

chriseth: If we keep this, we should do it even if `rd` is zero.

riscv/src/compiler.rs
lvella: I think that part is only relevant for 64-bits when using `lr.w`. On 32 bits, `lr.w` is already the size of the register.

riscv/src/compiler.rs
lvella: > Strictly, we should have one reservation bit per memory address (if I read the standard correctly). If we don't have that, maybe we could scrap the reservation altogether, or would some code break in that case?

The reservation can be for any memory size. It seems that a typical implementation is <= 1Kb, but the ISA specs says to assume it to be of any size, so ours is for the entire memory.

> If we keep this, we should do it even if `rd` is zero.

Indeed, makes sense.

riscv/src/compiler.rs
lvella: Do you have an idea of where we can find the exhaustive list?

lvella: Found in source: https://github.com/bminor/binutils-gdb/blob/e030ce2c79feee6a35665a6580a23dcf937ea46f/opcodes/riscv-opc.c#L332

I've enumerated all possibilities in "instruction_ends_control_flow", will that be enough?

chriseth (approved):

PR #624, opened 2023-09-21
Using a custom cargo profile in the per PR tests.

lvella:
Should reduce the storage usage, because there is much less debug information generated.


leonardoalt:
looks good to me, @chriseth ?

leonardoalt (approved):

chriseth (approved):

PR #631, opened 2023-09-25
Use itertools::format.

chriseth:


chriseth:
Reviewers, please double-check that I used the same separators.

backend/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
+default = []
```
chriseth, line 7: Will undo this.

util/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
chriseth, line 1: will remove.

lvella (approved):

PR #634, opened 2023-09-26
Visitors via traits.

chriseth:


lvella:
Maybe the trait name should be `ExpressionVisitable` or something in this line? Because visitor is usually the name of the function that is passed, and the trait belongs to the data structure that is "visited".

lvella (commented):
Well, there certainly is much more code in the new `ast/src/parsed/visitor.rs` than in the old `ast/src/parsed/utils.rs`, and I find the old easier to follow.

Schaeff:
I need to read through this again, but after seeing functions like `post_visit_expressions_mut`, I had a go at using a second trait in order to encode the order in which to visit rather than encoding it in the visitor functions. Then implementing a pass is very minimal. Not 100% happy with it yet but maybe you have some thoughts [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d56d5616e50b70fc77e99e3565850b6b)

chriseth:
I had a similar thought - if the visitation orders are different traits, it is much harder to mistakenly "switch" visitation order in the implementation.

Furthermore, the actual order really only matters for the implementation on `Expression`, for the rest it is exactly the same.

ast/src/parsed/visitor.rs
```diff
@@ -0,0 +1,385 @@
...
+        F: FnMut(&mut Expression<T, Ref>) -> ControlFlow<B>,
+    {
+        self.arguments
+            .iter_mut()
+            .try_for_each(|item| item.visit_expressions_mut(f, o))
+    }
+
+    fn visit_expressions<F, B>(&self, f: &mut F, o: VisitOrder) -> ControlFlow<B>
+    where
+        F: FnMut(&Expression<T, Ref>) -> ControlFlow<B>,
```
chriseth, line 379: If we could somehow shorten the signature here...

chriseth:
Was able to make it simpler by making the visitation order a runtime parameter. Also we now have all four combinations (pre,post x mut,ref).

Schaeff:
Yes, I thought about a runtime one as well. What made you go with that despite [this comment](https://github.com/powdr-labs/powdr/pull/634#issuecomment-1736187584) above? I'm unsure what's best now but I think we should optimise for usage of the trait to be as simple as possible even if it means the trait machinery is complex.

chriseth:
I also agree that usage is most important (plus maybe ease of adding a new ast node type). I found it more important to avoid duplicate pre- and post- implementations for non-expression nodes and I was not able to achieve that with separate traits for pre and post (mainly because of the recursive calls).

Schaeff:
@chriseth I don't have a better solution so I think we should go ahead with this for now. It could be nice to unify these with `ast/parsed/folder`. It's not blocking this though.

ast/src/parsed/visitor.rs
chriseth: I really prefer passing lambda functions instead of actual proper traits. I see that this is needed for something like the Folder, but since the signature is always the same for the simple recursive type of `Expression`, we can use lambda functions instead.

Again, this is the trade-off between usability from the outside and how complicated the mechanism is in the inside.

We can probably get away with removing the `B` type altogether (and replace it by `()`), because we can also just pass a return value through a closure instead of through the `ControlFlow` mechanism.

Would you prefer that?

asm_to_pil/src/romgen.rs
```diff
@@ -29,23 +27,16 @@ use crate::{
...
-    ) -> ControlFlow<()> {
-        previsit_expression_mut(e, &mut |e| {
-            if let Expression::Reference(r) = e {
-                if let Some(v) = substitution.get(r.name()).cloned() {
-                    *r.name_mut() = v;
-                }
-            };
-            ControlFlow::Continue::<()>(())
-        });
-        ControlFlow::Continue(())
```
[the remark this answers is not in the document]
chriseth: Yes it was, it re-started a full tree visit at each expression.

Schaeff (dismissed):
LGTM, as long as the recently found issues in visitors are fixed here as well

chriseth:
Yes, cross-checked them. Rebased.

Schaeff (approved):

PR #636, opened 2023-09-26
Complicated attempt of implementing mulhsu

lvella:
~This can be improved a little if #648 is accepted.~

EDIT: Now uses #648.

### DIFF
--- a/riscv/src/compiler.rs
+++ b/riscv/src/compiler.rs
@@ -67,12 +67,12 @@ impl Architecture for RiscvArchitecture {
     fn instruction_ends_control_flow(instr: &str) -> bool {
         match instr {
             "li" | "lui" | "la" | "mv" | "add" | "addi" | "sub" | "neg" | "mul" | "mulhu"
-            | "divu" | "remu" | "xor" | "xori" | "and" | "andi" | "or" | "ori" | "not" | "slli"
-            | "sll" | "srli" | "srl" | "srai" | "seqz" | "snez" | "slt" | "slti" | "sltu"
-            | "sltiu" | "sgtz" | "beq" | "beqz" | "bgeu" | "bltu" | "blt" | "bge" | "bltz"
-            | "blez" | "bgtz" | "bgez" | "bne" | "bnez" | "jal" | "jalr" | "call" | "ecall"
-            | "ebreak" | "lw" | "lb" | "lbu" | "lh" | "lhu" | "sw" | "sh" | "sb" | "nop"
-            | "fence" | "fence.i" | "amoadd.w" | "amoadd.w.aq" | "amoadd.w.rl"
+            | "mulhsu" | "divu" | "remu" | "xor" | "xori" | "and" | "andi" | "or" | "ori"
+            | "not" | "slli" | "sll" | "srli" | "srl" | "srai" | "seqz" | "snez" | "slt"
+            | "slti" | "sltu" | "sltiu" | "sgtz" | "beq" | "beqz" | "bgeu" | "bltu" | "blt"
+            | "bge" | "bltz" | "blez" | "bgtz" | "bgez" | "bne" | "bnez" | "jal" | "jalr"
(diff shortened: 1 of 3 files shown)


riscv/src/compiler.rs
```diff
@@ -904,6 +904,27 @@ fn process_instruction(instr: &str, args: &[Argument]) -> Vec<String> {
...
         }
+        "mulhsu" => {
+            let (rd, r1, r2) = rrr(args);
+            only_if_no_write_to_zero_vec(
+                vec![
+                    format!("tmp1 <== to_signed({r1});"),
+                    // tmp2 is 1 if tmp1 is non-negative
+                    "tmp2 <== is_positive(tmp1 + 1);".into(),
+                    // If negative, convert to positive
+                    "skip_if_zero 0, tmp2;".into(),
```
chriseth, line 915: I know this was not introduced here, but I'm wondering if the skip instruction is actually compatible with the batching.

riscv/src/compiler.rs
```diff
@@ -904,6 +904,27 @@ fn process_instruction(instr: &str, args: &[Argument]) -> Vec<String> {
...
+                    "tmp1 <=X= 0 - tmp1;".into(),
+                    format!("tmp1, {rd} <== mul(tmp1, {r2});"),
```
chriseth, line 917: Is there a test with tmp1 being the most negative number? I'm asking because if you negate it, it does not fit 16 bits any more and this could create problems when using the `mul` operation.

riscv/src/compiler.rs
```diff
@@ -904,6 +904,27 @@ fn process_instruction(instr: &str, args: &[Argument]) -> Vec<String> {
...
+                    // If was negative before, convert back to negative
+                    "skip_if_zero (1-tmp2), 2;".into(),
+                    "tmp1 <== is_equal_zero(tmp1);".into(),
+                    // If the lower bits are zero, return the two's complement,
+                    // otherwise return one's complement.
+                    format!("{rd} <== wrap_signed(-{rd} - 1 + tmp1);"),
```
chriseth, line 923: I don't understand, doesn't this always negate `rd`?

riscv/src/compiler.rs
lvella: But with smarter batching, if the second argument is always a constant, I think this can be handled.

Alternatively, instead of this relative jump instruction, we can add new labels an jump there.

riscv/src/compiler.rs
lvella: Assuming you meant 32 bits, it does fit. The most negative number is -2**31, the mul operands range in 0 to (2**32-1), so 2**31 is within range.

riscv/src/compiler.rs
lvella: Yes, this always negate. Also, it either adds 0 or 1 to the negation (the value of `tmp1), because that is the "carry" from the two's complement of the lower bits.

But depending whether the value was originally negative, this instruction and the previous are skipped entirely by the `skip_if_zero`.

riscv/src/compiler.rs
chriseth: Sure, it is compatible with the current batching, we just need to be careful that future batching does not destroy it (and it shouldn't since people can do what they want with `pc`).

riscv/src/compiler.rs
chriseth: And is there a test?

riscv/src/compiler.rs
chriseth: So `skip_if_zero 1` acutally skips 2 instructions?

chriseth: oh sorry, the second argument is the number of instructions to skip!

chriseth (approved):

riscv/src/compiler.rs
lvella: `test_7` covers the most negative value

PR #637, opened 2023-09-26
Skip diff computation if not needed.

lvella:


ast/src/lib.rs
```diff
@@ -25,8 +26,8 @@ impl DiffMonitor {
         std::mem::swap(&mut self.previous, &mut self.current);
         self.current = Some(s.to_string());
 
-        if let Some(current) = &self.current {
-            if let Some(previous) = &self.previous {
+        if log_enabled!(log::Level::Trace) {
```
chriseth, line 29: Oh wow! We should certainly have that check!

chriseth (approved):

PR #645, opened 2023-09-29
Inline constants during PIL analysis phase.

chriseth:


pil_analyzer/src/pil_analyzer.rs
```diff
@@ -84,7 +89,11 @@ impl<T> From<PILAnalyzer<T>> for Analyzed<T> {
         };
         previsit_expressions_in_pil_file_mut(&mut result, &mut |e| {
             if let Expression::Reference(Reference::Poly(reference)) = e {
-                assign_id(reference);
+                if let Some(value) = lookup_constant(reference) {
+                    *e = Expression::Number(*value)
```
chriseth, line 93: I'm not sure anymore if we should do substitution right here. This way, if we print the analyzed pil file, the constants will already be inlined.

PR #648, opened 2023-09-29
Multiplication returning both high and low bits.

lvella:
Makes use of multi-return.

### DIFF
--- a/riscv/src/compiler.rs
+++ b/riscv/src/compiler.rs
@@ -535,6 +535,7 @@ fn preamble() -> String {
     reg X[<=];
     reg Y[<=];
     reg Z[<=];
+    reg W[<=];
     reg tmp1;
     reg tmp2;
     reg tmp3;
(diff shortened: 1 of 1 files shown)


leonardoalt (approved):

leonardoalt:
Actually @georgwiese pointed out these constraints are not sounds due to overflow

leonardoalt:
or actually, it is sounds as long as the assumption that Z and W are 32 bits is correct

lvella:
> Actually @georgwiese pointed out these constraints are not sounds due to overflow

How? If both inputs are 0 <= x < 2**32, the maximum possible result is 0xfffffffe00000001, which is smaller than the Goldilocks field characteristic of 0xffffffff00000001.

leonardoalt:
yea, we should eventually do a pass checking whether everything is properly wrapped

lvella:
> or actually, it is sounds as long as the assumption that Z and W are 32 bits is correct

Well, then we have to prove that every instruction always leave something u32 in the registers.

leonardoalt:
Groebner Basis activated

PR #652, opened 2023-10-02
Remove constraints blocks

Schaeff:
```
machine {
   constraints {
      col witness x;
   }
   constraints {
      col witness y;
      x = y;
   }
}
```

Becomes 

```
machine {
   col witness x;
   col witness y;
   x = y;
}
```

I did not change `analysis/README.md` as the diff looks quite different now.
I also tried to update the book in a way that made sense, pointing to the docs for powdr-pil when mentioning pil code in machines.


analysis/src/macro_expansion.rs
```diff
@@ -33,22 +33,38 @@ impl<T: FieldElement> Folder<T> for MacroExpander<T> {
...
 
     fn fold_machine(&mut self, mut machine: Machine<T>) -> Result<Machine<T>, Self::Error> {
-        machine.statements.iter_mut().for_each(|s| match s {
-            MachineStatement::InstructionDeclaration(_, _, Instruction { body, .. }) => {
-                match body {
-                    InstructionBody::Local(body) => {
-                        *body = self.expand_macros(std::mem::take(body))
-                    }
-                    InstructionBody::CallableRef(..) => {
+        machine.statements = machine
```
Schaeff, line 36: This was changed to a functional approach, as a statement can yield many, and `iter_mut` does not allow that. Hence the usage of flat_map here.

analysis/src/macro_expansion.rs
```diff
@@ -33,22 +33,38 @@ impl<T: FieldElement> Folder<T> for MacroExpander<T> {
...
+                    )]
                 }
-            }
-            MachineStatement::InlinePil(_, statements) => {
-                *statements = self.expand_macros(std::mem::take(statements));
-            }
-            _ => {}
-        });
+                MachineStatement::Pil(start, statement) => self
+                    .expand_macros(vec![statement])
```
Schaeff, line 59: It would make more sense to me for `expand_macros` to take a single statement roather than a list, but I tried to keep this diff small.

ast/src/asm_analysis/display.rs
```diff
@@ -40,7 +40,7 @@ impl<T: Display> Display for Machine<T> {
         write_items_indented(f, &self.registers)?;
         write_items_indented(f, &self.instructions)?;
         write_items_indented(f, &self.callable)?;
-        write_items_indented(f, &self.constraints)?;
+        write_items_indented(f, &self.pil)?;
```
chriseth, line 43: Is it a problem that these might be ordered differently? I mean can we always re-parse stuff that is printed here?

chriseth (approved):

ast/src/asm_analysis/display.rs
Schaeff: Can you elaborate here? Within constraints, the order is preserved throughout the compiler, but eventually this should not be a requirement, right? There are no tests which parse outputs of this display afaik, this could make sense to add but it seems unrelated to this change.

chriseth:
still conflicts.

ast/src/asm_analysis/display.rs
chriseth: I'm just wondering if it might make sense to move the pil a bit further up or what happens if you mix constraints and instruction declarations and such, but yes, unrelated to this PR.

chriseth (approved):

PR #653, opened 2023-10-02
Fix column removal

chriseth:


ast/src/analyzed/mod.rs
```diff
@@ -114,14 +114,13 @@ impl<T> Analyzed<T> {
         .map(|polys| {
             polys
                 .iter()
-                .enumerate()
                 .fold(
                     (0, BTreeMap::new()),
-                    |(shift, mut replacements), (i, (poly, _def))| {
-                        assert_eq!(i as u64, poly.id);
```
chriseth, line 121: This assertion was wrong for arrays: There is one ID per array element, so the IDs are not contiguous on the names, but just on the actual columns.

leonardoalt (approved):

lvella (approved):

georgwiese (approved):
LGTM

PR #655, opened 2023-10-02
Fix visitors.

chriseth:


ast/src/analyzed/util.rs
```diff
@@ -36,21 +36,6 @@ where
         .try_for_each(|i| previsit_expressions_in_identity_mut(i, f))
 }
 
-pub fn postvisit_expressions_in_identity_mut<T, F, B>(
```
chriseth, line 39: This function did not visit the actual expresions in the lookup, just the selectors. Maybe we should have a proper visitor system like in #634 ...

georgwiese (approved):
LGTM!

PR #657, opened 2023-10-03
Fix more visitors.

chriseth:
Would prefer to merge https://github.com/powdr-labs/powdr/pull/634 instead of this.


chriseth:
#634 will be merged instead.

PR #658, opened 2023-10-03
Fix constant namespaces.

chriseth:


pil_analyzer/src/pil_analyzer.rs
```diff
@@ -489,7 +489,12 @@ impl<T: FieldElement> PILAnalyzer<T> {
     }
 
     fn namespaced_ref(&self, namespace: &Option<String>, name: &str) -> String {
-        format!("{}.{name}", namespace.as_ref().unwrap_or(&self.namespace))
+        if name.starts_with('%') || (namespace.is_none() && self.constants.contains_key(name)) {
+            // Constants are not namespaced
```
chriseth, line 493: I would like to keep it like that until we agree on a scope and namespace access pattern that works for both pil and powdr.

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -489,7 +489,12 @@ impl<T: FieldElement> PILAnalyzer<T> {
     }
 
     fn namespaced_ref(&self, namespace: &Option<String>, name: &str) -> String {
-        format!("{}.{name}", namespace.as_ref().unwrap_or(&self.namespace))
+        if name.starts_with('%') || (namespace.is_none() && self.constants.contains_key(name)) {
```
Schaeff, line 492: In the case where `name` starts with `%`, seems like `namespace` should be `None`? If so, it would be good to assert that.

chriseth: Added.

Schaeff (approved):

Issue #662, opened 2023-10-04
infinite loop in compilation


leonardoalt:
```
machine Main {
	reg pc[@pc];
	Main main;
}
```

leonardoalt:
this still happens

PR #667, opened 2023-10-05
Detect circular dependency in use statements

Schaeff:
Detect cases of the form

```
use module::Machine;

mod module {
   use super::Machine;
}
```

Note that this does not cover #662, which is unrelated to the import system, but rather related to recursive data structures.

I simplified the canonicalizer to rely on single absolute paths instead of tuples.

### DIFF
--- a/importer/src/path_canonicalizer.rs
+++ b/importer/src/path_canonicalizer.rs
@@ -9,7 +9,7 @@ use std::{
 use ast::parsed::{
     asm::{
         ASMModule, ASMProgram, AbsoluteSymbolPath, Import, Machine, MachineStatement, Module,
-        ModuleRef, ModuleStatement, SymbolDefinition, SymbolPath, SymbolValue, SymbolValueRef,
+        ModuleRef, ModuleStatement, SymbolDefinition, SymbolValue, SymbolValueRef,
     },
     folder::Folder,
 };
@@ -81,7 +81,7 @@ impl<'a, T> Folder<T> for Canonicalizer<'a> {
             if let MachineStatement::Submachine(_, path, _) = s {
                 *path = self
                     .paths
-                    .get(&(self.path.clone(), std::mem::take(path)))
+                    .get(&self.path.clone().join(std::mem::take(path)))
                     .cloned()
                     .unwrap()
                     .into();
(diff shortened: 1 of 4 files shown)


importer/src/path_canonicalizer.rs
```diff
@@ -104,6 +104,33 @@ pub struct State<'a, T> {
     pub paths: PathMap,
 }
 
+#[derive(Default)]
+struct PathChain {
```
chriseth, line 108:
```suggestion
struct PathDependencyChain {
```
?

importer/src/path_canonicalizer.rs
```diff
@@ -104,6 +104,33 @@ pub struct State<'a, T> {
...
+    paths: Vec<AbsoluteSymbolPath>,
+}
+
+impl PathChain {
+    fn push(&mut self, path: AbsoluteSymbolPath) -> Result<(), String> {
+        match self.paths.iter().position(|p| p == &path) {
+            Some(index) => {
+                self.paths.push(path);
+                Err(format!(
+                    "Cycle detected: {}",
```
chriseth, line 118:
```suggestion
                    "Cycle detected in `use` statements: {}",
```

importer/test_data/import_after_usage.expected.asm
```diff
@@ -0,0 +1,7 @@
+machine Foo {
+}
+mod module {
+    machine Bar {
+        Foo foo;
```
chriseth, line 5: Unrelated to this PR, but as is, this would not compile, would it, because `Foo` is not available like that. Maybe we should print (and also parse) that as `::Foo`? Or should we just live with the fact that it doesn't work as input?

chriseth:
Looks good! Can you squash?

importer/test_data/import_after_usage.expected.asm
chriseth: I'm not sure I understand. At the point of that Display trait, we are already in a situation where everything is inlined, right? So even if we had crates, they would not really exist, everything is turned into a single file.

So I guess the question is if we want to have a special symbol name for the top level "module" (or meta-module, meaning the module containing all top-level module) that is available everywhere or just have a special syntax like `::...`.

I kind of would like to have the ideal that everything we display should also be parseable, even though I know that it migth not always be achievable. But at least we should introduce something that distinguishes relative from absolute paths in the Display trait to make it harder to trip over such things.

Schaeff:
```
machine Machine {}
mod module {
   use ::Machine;
}
```

Modules are compilation units. Without absolute paths, you can compile `module` as a standalone thing. With absolute paths, you cannot. Not the end of the world and probably still worth introducing absolute paths as I agree with the idea that we should be able to parse everything, but still a bit odd to me. If you make "crates" the compilation unit, you don't have that problem because you drop the idea that you can compile each module, and always start from the root of the crate.

chriseth (approved):

PR #669, opened 2023-10-05
adjust readme

leonardoalt:


README.md
```diff
@@ -1,34 +1,33 @@
...
-## Basic Concept
+For detailed documentation please visit [the powdr book](https://docs.powdr.org/).
 
 *powdr* is a toolkit that helps build zkVMs and similar proof frameworks.
 
 It has two main components:
 
-- A polynomial identity language that allows you to define polynomial constraints, lookups, etc.
-- An extensible assembly language to perform dynamic executions.
+- powdr-asm: an extensible assembly IR language to perform dynamic executions.
```
chriseth, line 18: would the word `self-defining` be fitting?

leonardoalt: where exactly?

If for the language itself I don't really like it. The language exists with very few builtins, but they exist: register datatypes, register assignments, machines, inline constraints, instructions, labels.

How about:

- powdr-asm: an extensible assembly IR language to define any VM / instruction set

?

chriseth: I'm mainly thinking about the fact that the language is not just `extensible`, but much more. But I'm also fine with it as is.

chriseth (approved):

README.md
leonardoalt: ok, this sentence is mostly the same, we can re-visit in the future

PR #672, opened 2023-10-05
Serdefying the analysys result.

lvella:
So it can be saved and restored.

### DIFF
--- a/ast/Cargo.toml
+++ b/ast/Cargo.toml
@@ -7,10 +7,9 @@ edition = "2021"
 
 [dependencies]
 itertools = "0.11.0"
-num-bigint = "0.4.3"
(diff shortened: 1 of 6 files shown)


chriseth:
What would you want to use it for?
If we allow exporting the internal data structure representation, this kind of defines an interface we have to maintain.
On the other hand, the source code representation already defines a representation we can import and export, and we also have the polygon json export as well.

lvella:
> What would you want to use it for? If we allow exporting the internal data structure representation, this kind of defines an interface we have to maintain. On the other hand, the source code representation already defines a representation we can import and export, and we also have the polygon json export as well.

I just wanted to load the analyzed results faster than running the analysis again, when developing the powdr-riscv executor.

lvella:
This is useful for debug and development.

PR #678, opened 2023-10-06
poseidon gl rust

leonardoalt:
Another attempt at supporting the low level GL Poseidon in Rust.

There are 2 contentious parts in this PR that I'm not sure yet how to solve before suggesting to actually merge this:

### Lots of new registers

This causes lots of new columns, even with the optimizer. I guess this is fine in general, but if we do add it to the RISCV machine it will make even cases that do not use Poseidon more complicated. We could have a simple optimizer pass that removes (i) unused instructions; (ii) unused registers; (iii) unused machines. This would at least make it equivalent if Poseidon is not used. For the case where it is used, moving registers to memory would likely be the best option for a big optimization.

### Argument layout in the co-processor

RISCV uses the registers a0-a7 for arguments (x10-x17) if the arguments fit. In our case, we have 12 u64 which require 24 RISCV words and therefore, doesn't fit. The compiler then puts the input structs (there are 3 structs of 4 u64s each) in memory, and assigns the argument registers to each struct. The memory layout of each struct is 8 consecutive words, 2 for each u64, where the first word is the lower 32-bit half of the u64.

The signature of our co-processor in Rust/C is
```rust
fn poseidon_gl_coprocessor(a: PoseidonGLU256, b: PoseidonGLU256, c: PoseidonGLU256) -> PoseidonGLU256;

#[repr(C)]
pub struct PoseidonGLU256(pub u64, pub u64, pub u64, pub u64);
```

As far as I understood, the generated RISCV assembly uses `a1`, `a2` and `a3`, respectively, to point to `a`, `b` and `c` in memory. It requires the output to be written to the memory address in `a0`. This is odd, because usually it would use `a0` for the first argument and so on. Since the tests do work using the suggested layout I think it is correct for this case. The main problem is deciding whether this is reproducible, that is, is the compiler gonna use a different set of registers at some point?


riscv/runtime/src/coprocessors.rs
```diff
@@ -5,8 +5,16 @@
 // during the reachability analysis.
 extern "C" {
     fn poseidon_coprocessor(a: u32, b: u32) -> u32;
+    fn poseidon_gl_coprocessor(a: PoseidonGLU256, b: PoseidonGLU256, c: PoseidonGLU256) -> PoseidonGLU256;
 }
 
+#[repr(C)]
+pub struct PoseidonGLU256(pub u64, pub u64, pub u64, pub u64);
```
leonardoalt, line 12: We needed the struct and the repr(C) directive because otherwise the compiler complains this is FFI unsafe

riscv/runtime/src/coprocessors.rs
```diff
@@ -5,8 +5,16 @@
 // during the reachability analysis.
 extern "C" {
     fn poseidon_coprocessor(a: u32, b: u32) -> u32;
+    fn poseidon_gl_coprocessor(a: PoseidonGLU256, b: PoseidonGLU256, c: PoseidonGLU256) -> PoseidonGLU256;
```
leonardoalt, line 8: these are the same arguments as the Poseidon co-processor: 8 field elements as input and 4 cap field elements

riscv/src/compiler.rs
```diff
@@ -444,6 +456,30 @@ fn preamble() -> String {
...
+    reg P2;
+    reg P3;
+    reg P4;
+    reg P5;
+    reg P6;
+    reg P7;
+    reg C0;
+    reg C1;
+    reg C2;
+    reg C3;
```
leonardoalt, line 482: Yes this is not ideal. We need 12 extra state registers to store the data that goes into the machine, and 12 extra assignment registers for the instruction. If we find a way to do this in a lower level we could save columns.

riscv/tests/riscv_data/poseidon_gl_via_coprocessor.rs
```diff
@@ -0,0 +1,57 @@
+#![no_std]
+
+extern crate alloc;
+use alloc::vec::Vec;
+
+use runtime::coprocessors::{poseidon_gl, PoseidonGLU256};
+
```
leonardoalt, line 7: This test is the same as test_data/std/poseidon_gl_test.asm, but in Rust

riscv/runtime/src/coprocessors.rs
gballet: What if instead, the list of values was passed as two parameters: `*mut u64` and a `usize` ? The first parameter points to the buffer or a `[u64]` memory slice. That wouldn't fix the register stuff just yet, but it would ensure that it only uses 2 arguments (which I believe is more stable, register wise, than using more arguments) and it's future proof in the sense that you could pass more than 2 field elements in the future (i.e. with a `usize` parameter > 24).

Why 24 ? because in the current API, I assume the first 8 bytes are for field element `a`, the next 8 are for field element `b` and the last 8 are for the output (hence the `mut` pointer).

riscv/runtime/src/coprocessors.rs
gballet: right, hence my suggestion to use `[u64]` instead, so that we are less prone to this kind of ABI restrictions.

gballet (commented):
This branch doesn't compile, although it doesn't seem to be caused directly by the change.

```
error[E0425]: cannot find function `compile_solidity` in module `evm`
   --> halo2/src/aggregation.rs:390:10
    |
390 |     evm::compile_solidity(&loader.solidity_code())
    |          ^^^^^^^^^^^^^^^^ not found in `evm`

error[E0599]: no method named `solidity_code` found for struct `Rc<EvmLoader>` in the current scope
   --> halo2/src/aggregation.rs:390:35
    |
390 |     evm::compile_solidity(&loader.solidity_code())
    |                                   ^^^^^^^^^^^^^ method not found in `Rc<EvmLoader>`
 ```
 
 So I could not verify the asm output. This being said, I proposed an alternative encoding. We can discuss that this week.

**Comment by leonardoalt** [comment]
@gballet You need to cargo update or delete cargo lock because one dep was updated

File: riscv/runtime/src/coprocessors.rs (hunk 2 above)
Comment (leonardoalt) [reply riscv/runtime/src/coprocessors.rs:8]: Re last assumption, it's actually 12 u64 input elements, and 4 output.

But yea, the buffer idea sounds good. Wouldn't it be even better if we use a `mut [u64; 12]` (if possible) on the Rust side? Because the PIL co-processor takes exactly 12 inputs, so I think we should ensure the same API. I'll try that next

**Comment by leonardoalt** [comment]
@gballet I updated the PR with your idea, it seems to work nicely!

**Comment by leonardoalt** [comment]
This PR is now ready for review. I've just

- added @gballet as an author
- cleaned up removing old poseidon stuff
- used @gballet 's idea of buffer instead of many args

the "many new registers" problem can be solved by https://github.com/powdr-labs/powdr/issues/682

File: riscv/runtime/src/coprocessors.rs (hunk 2 above)
Comment (gballet) [reply riscv/runtime/src/coprocessors.rs]: right, 12 and not 24.

File: riscv/tests/riscv_data/poseidon_gl_via_coprocessor.rs [hunk 5]
```diff
@@ -0,0 +1,55 @@
...
+    let minus_one = 0xffffffff00000001 - 1;
+    let mut i: [u64; 12] = [minus_one; 12];
+    poseidon_gl(&mut i);
+    assert_eq!(i[0], 13691089994624172887);
+    assert_eq!(i[1], 15662102337790434313);
+    assert_eq!(i[2], 14940024623104903507);
+    assert_eq!(i[3], 10772674582659927682);
+
+    let mut i: [u64; 12] = [18446744069414584321; 12];
```
georgwiese, line 29: The original test sets the capacity (last 4 elements) to 0. How can the test still pass?

If I change set the capacity to 0 the test still passes, so looks like the capacity is always set to 0.

riscv/runtime/src/coprocessors.rs
```diff
@@ -4,9 +4,9 @@
 // The function itself will be removed by the compiler
 // during the reachability analysis.
 extern "C" {
-    fn poseidon_coprocessor(a: u32, b: u32) -> u32;
+    fn poseidon_gl_coprocessor(data: *mut [u64; 12]);
 }
 
-pub fn poseidon_hash(a: u32, b: u32) -> u32 {
-    unsafe { poseidon_coprocessor(a, b) }
+pub fn poseidon_gl(data: &mut [u64; 12]) {
```
georgwiese, line 10: Should we add some docstring? If I understand right, it replaces the first 4 elements with the hash, right?

georgwiese (commented):
I'll have a closer look after lunch, but the fact that the test passes is strange. Also wondering how much slower the Keccak benchmark is now; if it's significant, perhaps we'll want to include the new machines only when needed?

riscv/tests/riscv_data/poseidon_gl_via_coprocessor.rs
leonardoalt: that's indeed weird, need to take a look

georgwiese (commented):
These are the benchmark results on my machine:

```
                        time:   [37.447 s 38.540 s 39.663 s]
                        change: [+5.5000% +8.8501% +12.522%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

riscv/src/compiler.rs
```diff
@@ -848,14 +889,53 @@ fn only_if_no_write_to_zero_vec(statements: Vec<String>, reg: Register) -> Vec<S
...
-    &[("poseidon_coprocessor", "x10 <== poseidon(x10, x11);")];
+fn poseidon_gl_call() -> String {
+    let decoding = |i| {
+        format!(
+            r#"
+        addr <=X= {} + x10;
+        P{i} <== mload();
+        addr <=X= {} + x10;
+        tmp1 <== mload();
+        P{i} <=X= P{i} + tmp1 * 2**32;
```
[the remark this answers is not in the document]
georgwiese: Actually, that's not what I mean. I though that would already be enforced somehow (because memory words are always 32 bits?).

What I mean is that `P{i} + tmp1 * 2**32` < p. Not sure what the overhead of this is, but we could just add a `assert!(x < P_GL);` for each `u64` input `x` in rust code?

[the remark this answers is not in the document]
leonardoalt: ah yes, I wanted to do that but forgot

leonardoalt:
> perhaps we'll want to include the new machines only when needed?

@georgwiese yea I think we should do this anyway. It is a bit annoying because of the way the pipeline works:

- the riscv -> powdr-asm translation happens in a String basis, it returns a single String with everything
- this String is then passed to the asm compiler crate, that parses it etc
- it's not hard to analyze which machines are used etc using the structured asm, but that happens already in a crate outside riscv
- which would still be fine, but because we inject the first call to the machines in the preamble of the program, we can't generically detect whether something is actually used or not

need to think about how to solve this. Maybe a first asm pass just to get the structured asm, and then analyze. This requires some refactoring in `riscv/src/lib.rs` though, since there "compile" means do the whole thing including witgen

leonardoalt:
updated the Rust co-processor library code to actually return the hash instead of mutating the input. The extern C function still does that because it's easier for the powdr asm code

leonardoalt:
how is this PR now passing the github tests...

chriseth:
Shouldn't we auto-detect which coprocessors are used?
If we don't I would recommend `--coprocessors abc,def,ghi`

leonardoalt:
> Shouldn't we auto-detect which coprocessors are used?

Yes but it requires either analyzing the RISCV asm instead of powdr asm which will lead to duplication for any VM, or a refactoring, and I want to get this PR merged before any of that.

>  If we don't I would recommend --coprocessors abc,def,ghi

What I dislike about that is the need for parsing/string comparison etc that's needed, but won't oppose it too much since it should be removed once auto-detect is implemented.

leonardoalt:
ok done @chriseth

riscv/src/compiler.rs
```diff
@@ -95,12 +96,163 @@ impl Architecture for RiscvArchitecture {
     }
 }
 
-pub fn machine_decls() -> Vec<&'static str> {
-    vec!["use std::binary::Binary;", "use std::shift::Shift;"]
+/// Defines which coprocessors should be used by the RISCV machine.
+/// It is important to not add unused coprocessors since they may
+/// lead to many extra columns in PIL.
+#[derive(Clone, Debug, Default)]
+pub struct CoProcessors {
```
georgwiese, line 103: Co-processors can and should be field-specific. I think this would be the right place to do it, right? Then the user would only specify the machines it needs but the `CoProcessor` type knows which implementation to chose. Right now, I think we should assert that the `PoseidonGL` and `GoldilocksGL` machines are only instantiated on the Goldilocks field, no?

One way to do this is to add a type parameter to `CoProcessors<T>`. I prepared this in #635.

georgwiese (commented):
Cool! I like the new `CoProcessors` type, I think it deserves its own module and could do more things (if not now then in the future).

georgwiese (commented):
LGTM! Please squash.

leonardoalt:
squashed

georgwiese (approved):

PR #683, opened 2023-10-09
Change expression visitor to be more flexible.

chriseth:
This makes the ExpressionVisitable trait generic over the expression type itself.

It allows us to introduce a new expression type without having to modify the more generic types like SelectedExpression etc.


georgwiese (dismissed):
👌

georgwiese (dismissed):

leonardoalt (approved):

PR #691, opened 2023-10-10
Condenser

chriseth:
The condenser is the component that turns generic expressions from the parse into the more simplified algebraic expressions in identities.

Currently it is quite simple, but in the future it will condense expression trees of the from `((x)')[2]` into single references with next set to true and the array access index set to 2.

When we are there, it will also inline functions as far as it is necessary and thus replace the macro system.


chriseth:
Depends on #686

georgwiese (commented):
Looks great, but now `pilopt::optimize_constants` can be removed again, no? As far  as I can see, the condenser now inlines constants and evaluates constant expressions.

georgwiese (approved):

PR #695, opened 2023-10-11
Split circleci jobs.

chriseth:
The biggest problem with this is that it still builds some stuff in the "test" jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,20 +1,31 @@
 version: 2.1
 
-jobs:
-  pr-tests:
+executors:
+  executor:
     machine:
       image: ubuntu-2004:current
     resource_class: large
+
+jobs:
+  build:
+    executor: executor
     steps:
+      - checkout
+
       - run:
           name: Install dependencies
           command: sudo apt update -qqy && sudo apt install libclang-dev git nodejs -qqy
 
       - run:
-          name: Install Rust
+          name: Install Rustup
           command: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 
-      - checkout
+      - run:
+          name: Install Rust
+          command: |
+            rustup toolchain install 1.72-x86_64-unknown-linux-gnu
+            rustup component add clippy --toolchain 1.72-x86_64-unknown-linux-gnu
+            rustup component add rustfmt --toolchain 1.72-x86_64-unknown-linux-gnu
 
       - run:
           name: Install pilcom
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,20 +1,31 @@
...
+  executor:
     machine:
       image: ubuntu-2004:current
     resource_class: large
+
+jobs:
+  build:
+    executor: executor
     steps:
+      - checkout
```
leonardoalt, line 13: how can it do checkout without git? or do we assume the image has git?

.circleci/config.yml
```diff
@@ -41,51 +52,121 @@ jobs:
...
-          command: rustup toolchain install nightly-2023-01-03-x86_64-unknown-linux-gnu
-
-      - run:
-          name: Install riscv target
-          command: rustup target add riscv32imac-unknown-none-elf --toolchain nightly-2023-01-03-x86_64-unknown-linux-gnu
-
-      - run:
-          name: Install stdlib
-          command: rustup component add rust-src --toolchain nightly-2023-01-03-x86_64-unknown-linux-gnu
+            - v5-cargo-pr-tests-{{ .Branch }}
```
leonardoalt, line 55: I'm not sure I agree with this. The cache from `main` can also be useful.

.circleci/config.yml
```diff
@@ -41,51 +52,121 @@ jobs:
...
+            rustup toolchain install 1.72-x86_64-unknown-linux-gnu
+            rustup toolchain install nightly-2023-01-03-x86_64-unknown-linux-gnu
+            rustup target add riscv32imac-unknown-none-elf --toolchain nightly-2023-01-03-x86_64-unknown-linux-gnu
+            rustup component add rust-src --toolchain nightly-2023-01-03-x86_64-unknown-linux-gnu
+
+      - run:
+          name: Determine which tests to run
+          command: |
+              cargo test --all --all-features --profile pr-tests  -- --list 2> /dev/null | grep ': test' | sed -e 's/: test//' | \
+              circleci tests split --split-by=timings | tee tests_to_run.txt
```
leonardoalt, line 109: is there no better way of doing this?

.circleci/config.yml
```diff
@@ -41,51 +52,121 @@ jobs:
...
           paths:
             - ~/.cargo/registry
             - ~/.cargo/git
             - target
 
+      - persist_to_workspace:
+          root: .
+          paths:
+            - target/pr-tests
+            - pilcom
```
leonardoalt, line 77: we could also keep `.rustup` and `.cargo` so we don't need to reinstall them

.circleci/config.yml
chriseth: It's a circleci-specific thing.

.circleci/config.yml
chriseth: Yes, as a fallback we should use the cache from `main`. The problem is that it should be somewhat recent. I couldn't find a good way to do this yet.

.circleci/config.yml
chriseth: It's the best way I could find.

.circleci/config.yml
chriseth: We should use a custom docker image.

.circleci/config.yml
leonardoalt: I think you can add an extra line with another cache key that is checked if the first one fails. We can actively invalidate it once in a while

leonardoalt (approved):

PR #703, opened 2023-10-16
Treat intermediate columns specially.

chriseth:
This is needed because in the future, the values of intermediate columns need to be "simple expressions".

### DIFF
--- a/ast/src/analyzed/display.rs
+++ b/ast/src/analyzed/display.rs
@@ -28,37 +28,44 @@ impl<T: Display> Display for Analyzed<T> {
         for statement in &self.source_order {
             match statement {
                 StatementIdentifier::Definition(name) => {
-                    let (symbol, definition) = &self.definitions[name];
-                    let (name, is_local) = update_namespace(name, symbol.degree, f)?;
-                    match symbol.kind {
-                        SymbolKind::Poly(poly_type) => {
-                            let kind = match &poly_type {
-                                PolynomialType::Committed => "witness ",
-                                PolynomialType::Constant => "fixed ",
-                                PolynomialType::Intermediate => "",
-                            };
-                            write!(f, "    col {kind}{name}")?;
-                            if let Some(value) = definition {
-                                writeln!(f, "{value};")?
-                            } else {
+                    if let Some((symbol, definition)) = self.definitions.get(name) {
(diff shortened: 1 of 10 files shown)


backend/src/pilstark/json_exporter/mod.rs
```diff
@@ -55,7 +55,7 @@ pub fn export<T: FieldElement>(analyzed: &Analyzed<T>) -> PIL {
                                 exporter.intermediate_poly_expression_ids[&poly.id] as usize
                             );
                         } else {
-                            panic!("Expected single value");
+                            panic!("Expected single intermediate value");
```
leonardoalt, line 58: `single value` sounds a bit misleading imo, maybe `single intermediate expression`?

ast/src/analyzed/mod.rs
```diff
@@ -107,7 +121,7 @@ impl<T> Analyzed<T> {
             // for each kind.
             self.committed_polys_in_source_order(),
             self.constant_polys_in_source_order(),
-            self.intermediate_polys_in_source_order(),
+            //self.intermediate_polys_in_source_order(),
```
leonardoalt, line 124: remove?

ast/src/analyzed/display.rs
```diff
@@ -28,37 +28,44 @@ impl<T: Display> Display for Analyzed<T> {
...
+                                write!(f, "    let {name}")?;
+                                if let Some(value) = definition {
+                                    write!(f, "{value}")?
+                                }
+                                writeln!(f, ";")?
                             }
-                            writeln!(f, ";")?
                         }
+                    } else if let Some((symbol, definition)) = self.intermediate_columns.get(name) {
+                        let (name, _) = update_namespace(name, symbol.degree, f)?;
```
leonardoalt, line 64: why does it need to "update namespace" and what does that mean?

chriseth: At this stage ("analyzed"), we eliminated the concept of namespaces. All symbols have fully qualified names.

For various reasons, we need to re-introduce namespaces again for printing, that's what "update_namespace" does: If the namespaces changed, it emits a `namespace XYZ(n):` line and updates the current namespace.

leonardoalt (approved):

PR #709, opened 2023-10-20
Implement mload with alignment correction built-in.

chriseth:
Depends on #720

### DIFF
--- a/riscv/src/compiler.rs
+++ b/riscv/src/compiler.rs
@@ -441,8 +441,6 @@ fn preamble(coprocessors: &CoProcessors) -> String {
             .collect::<Vec<_>>()
             .concat()
         + r#"
-    // TODO: addr can be removed once  mload also takes it via argument
-    reg addr;
 
     x0 = 0;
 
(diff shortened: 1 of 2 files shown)


chriseth:
This moves the keccak loop start from 22100 to 18700.
Also performance goes from 4.4 rows/sec to 5k rows/sec (because it does not need the binary machine any more for most memory accesses).
What is a bit strange is that (after optimization) the witness col count goes from 271 up to 272, the fixed from 212 to 216.

leonardoalt:
needs a rebase maybe?

riscv/src/compiler.rs
```diff
@@ -1274,8 +1274,7 @@ fn process_instruction(instr: &str, args: &[Argument], coprocessors: &CoProcesso
             vec![
                 format!("tmp1 <== wrap({rd} + {off});"),
                 "addr <== and(tmp1, 0xfffffffc);".to_string(),
```
leonardoalt: leftover `addr`?

riscv/src/compiler.rs
```diff
@@ -1257,8 +1258,7 @@ fn process_instruction(instr: &str, args: &[Argument], coprocessors: &CoProcesso
...
```
leonardoalt: leftover addr

leonardoalt (commented):
maybe remove the `addr` register?

leonardoalt:
Ah you did remove it, I was reviewing commit by commit because of the merge commits

riscv/src/compiler.rs
```diff
@@ -521,6 +519,8 @@ fn preamble(coprocessors: &CoProcessors) -> String {
         // { W, X, Z} in { shr.value, shr.amount, shr.amount}
     }
 
+    /// Stores Z at addres Y % 2**32. Y can be between 0 and 2**33.
```
leonardoalt, line 522: s/addres/address

leonardoalt:
Looks good! Just a typo left. It is weird though that there are more columns now

leonardoalt (approved):

PR #710, opened 2023-10-20
Tests for alignment constraints.

chriseth:


test_data/pil/alignment.pil
```diff
@@ -0,0 +1,34 @@
...
+    let X_b1;
+    { X_b4 } in { bytes };
+    { X_b3 } in { bytes };
+    { X_b2 } in { bytes };
+    { X_b1 } in { bytes };
+
+    // This should also work at some point:
+    // Z * (Z - 1) * (Z - 2) * (Z - 3) = 0;
+    { Z } in { bytes };
+    { Z * 2**28 } in { bytes };
```
chriseth, line 22: It would be really nice if we could deduce global constraints from this.

test_data/pil/alignment.pil
```diff
@@ -0,0 +1,38 @@
...
+    { X_b1 } in { six_bits };
+
+    Y' = (1 - first') * (Y + 1);
+    // {
+    //     X_b4 * 0x1000000 + X_b3 * 0x10000 + X_b2 * 0x100 + X_b1 * 4,
+    //     STEP,
+    //     X
+    // } is m_is_read { m_addr, m_step, m_value }
+    // // We could even do the lookup with W instead of X and then:
+    // // { W, X, Z} in { shr.value, shr.amount, shr.amount}
```
leonardoalt, line 38: Is the commented stuff supposed to work?

chriseth: No, because of the link stuff. AFAIK we cannot use lookups to machines inside pil, we can only connect machines and instructions. I think this should change.

leonardoalt:
Does this PR need to wait for https://github.com/powdr-labs/powdr/pull/719 ?

chriseth:
Replaced by #720

PR #722, opened 2023-10-25
Fix handling of ".zero" with second argument.

chriseth:


asm_utils/src/data_parser.rs
```diff
@@ -91,13 +91,15 @@ fn extract_data_value<R: Register, F: FunctionOpKind>(
     arguments: &[Argument<R, F>],
 ) -> Vec<DataValue> {
     match (directive, arguments) {
+        (".zero", [Argument::Expression(Expression::Number(n))]) => {
+            vec![DataValue::Zero(*n as usize)]
+        }
         (
             ".zero",
-            [Argument::Expression(Expression::Number(n))]
-            // TODO not clear what the second argument is
```
chriseth, line 97: Now we know!

asm_utils/src/data_parser.rs
```diff
@@ -107,47 +109,44 @@ fn extract_data_value<R: Register, F: FunctionOpKind>(
             data.push(0);
             vec![DataValue::Direct(data)]
         }
-        (".word", data) => {
```
chriseth, line 110: This is all just reformatting. weird.

leonardoalt (approved):

PR #734, opened 2023-10-31
Parse index access as operator.

chriseth:
Depends on #733 

This turns index access into a proper operator that can be applied everywhere, not just right after a symbol.

### DIFF
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -315,7 +315,7 @@ pub enum SymbolKind {
     Other(),
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum FunctionValueDefinition<T> {
     Mapping(Expression<T>),
     Array(Vec<RepeatedArray<T>>),
@@ -324,7 +324,7 @@ pub enum FunctionValueDefinition<T> {
 }
 
 /// An array of elements that might be repeated.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct RepeatedArray<T> {
     /// The pattern to be repeated
     pattern: Vec<Expression<T>>,
@@ -370,6 +370,7 @@ pub struct PublicDeclaration {
     pub source: SourceRef,
     pub name: String,
     pub polynomial: PolynomialReference,
+    pub array_index: Option<DegreeType>,
     /// The evaluation point of the polynomial, not the array index.
     pub index: DegreeType,
 }
@@ -629,7 +630,6 @@ pub struct PolynomialReference {
     /// Optional because it is filled in in a second stage of analysis.
     /// TODO make this non-optional
     pub poly_id: Option<PolyID>,
-    pub index: Option<u64>,
 }
 
 #[derive(Debug, Copy, Clone, PartialOrd, Ord, PartialEq, Eq, Hash)]
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -370,7 +370,7 @@ pub struct PublicDeclaration {
     pub source: SourceRef,
     pub name: String,
     pub polynomial: PolynomialReference,
-    pub array_index: Option<DegreeType>,
+    pub array_index: Option<usize>,
     /// The evaluation point of the polynomial, not the array index.
     pub index: DegreeType,
 }
@@ -435,7 +435,7 @@ pub struct AlgebraicReference {
     pub name: String,
     /// Identifier for a polynomial reference.
     pub poly_id: PolyID,
-    pub index: Option<u64>,
+    pub index: Option<usize>,
     pub next: bool,
 }
 
--- a/ast/src/parsed/mod.rs
+++ b/ast/src/parsed/mod.rs
@@ -23,7 +23,8 @@ pub enum PilStatement<T> {
     PublicDeclaration(
         usize,
         String,
-        NamespacedPolynomialReference<T>,
+        NamespacedPolynomialReference,
+        Option<Expression<T>>,
         Expression<T>,
     ),
     PolynomialConstantDeclaration(usize, Vec<PolynomialName<T>>),
@@ -68,7 +69,7 @@ impl<Expr> Default for SelectedExpressions<Expr> {
 }
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
-pub enum Expression<T, Ref = NamespacedPolynomialReference<T>> {
+pub enum Expression<T, Ref = NamespacedPolynomialReference> {
     Reference(Ref),
     PublicReference(String),
     Number(T),
@@ -82,7 +83,7 @@ pub enum Expression<T, Ref = NamespacedPolynomialReference<T>> {
         Box<Expression<T, Ref>>,
     ),
     UnaryOperation(UnaryOperator, Box<Expression<T, Ref>>),
-
+    IndexAccess(IndexAccess<T, Ref>),
     FunctionCall(FunctionCall<T, Ref>),
     FreeInput(Box<Expression<T, Ref>>),
     MatchExpression(Box<Expression<T, Ref>>, Vec<MatchArm<T, Ref>>),
@@ -145,8 +146,8 @@ impl<T: FieldElement, Ref> From<T> for Expression<T, Ref> {
     }
 }
 
-impl<T> From<NamespacedPolynomialReference<T>> for Expression<T> {
-    fn from(value: NamespacedPolynomialReference<T>) -> Self {
+impl<T> From<NamespacedPolynomialReference> for Expression<T> {
+    fn from(value: NamespacedPolynomialReference) -> Self {
         Self::Reference(value)
     }
 }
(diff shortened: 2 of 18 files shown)


leonardoalt:
can you use the other PR as the base of this one so that only the new commit shows?

pil_analyzer/src/condenser.rs
```diff
@@ -77,14 +74,14 @@ pub fn condense<T: FieldElement>(
 }
 
 pub struct Condenser<T> {
-    pub symbols: HashMap<String, Symbol>,
+    pub symbols: HashMap<String, (Symbol, Option<FunctionValueDefinition<T>>)>,
```
leonardoalt, line 77: Can you add a comment explaining the data structure?

backend/src/pilstark/json_exporter/mod.rs
```diff
@@ -61,7 +61,7 @@ pub fn export<T: FieldElement>(analyzed: &Analyzed<T>) -> PIL {
                 let (_, expr) = exporter.polynomial_reference_to_json(&AlgebraicReference {
                     name: pub_ref.name.clone(),
                     poly_id: pub_ref.poly_id.unwrap(),
-                    index: pub_ref.index,
+                    index: None,
```
leonardoalt, line 64: What happens to arrays here?

ast/src/parsed/mod.rs
```diff
@@ -159,129 +154,21 @@ pub struct PolynomialName<T> {
...
-
-    /// Returns an indexed polynomial using this polynomial and an index. Used for polynomial array members.
-    pub fn indexed<T>(self, index: Expression<T>) -> IndexedPolynomialReference<T> {
-        self.with_index(Some(index))
-    }
-
-    /// Returns an indexed polynomial using this polynomial and an index. Used for polynomials which are not declared in arrays.
-    pub fn single<T>(self) -> IndexedPolynomialReference<T> {
-        self.with_index(None)
-    }
```
leonardoalt, line 332: Why is none of this needed anymore?

leonardoalt (commented):
It would be nice to have some tests, both for the error cases and for the newly introduced operator

ast/src/parsed/mod.rs
chriseth: This changes `[]` to a proper operator. It allow things like `f(x)[2]` or `(f + 2)[3]`. Previously, index access was tied to a symbol, and this is the reflection of that in the rust type system.

backend/src/pilstark/json_exporter/mod.rs
chriseth: This might actually be a mistake since index on PublicDeclaration does not refer to index access but to the row number. Let me check.

chriseth:
I think AlgebraicReference might need an `index` member after all if we want to properly support arrays.

I assume that in constraints, when `x` is an array of witness columns,

`(x - 2) = 0` is valid and means that all elements of the array need to be two, while `x[3] - 2 = 0` is valid as well. Right?

chriseth:
Oh wait, that is the case....

leonardoalt:
At a first glance I don't think `(x - 2) = 0` should be valid

chriseth:
> At a first glance I don't think `(x - 2) = 0` should be valid as it's an array, not a column (polynomial)

How would you write "all elements of the array x are 2"? We could use `@x` to borrow from php :rofl:

chriseth:
It's difficult to write test since witgen does not yet support arrays. But there are tests that parse expressions that use index access, convert them to polygon-hermez-pil-json and compare this with the json output by the polygon-hermez-pil-parser (arith.pil, for example).

leonardoalt:
> How would you write "all elements of the array x are 2"? We could use `@x` to borrow from php 🤣

but if you need to do what, why do you need an array?

pil_analyzer/src/condenser.rs
leonardoalt: I still don't understand this type, also with the comment. What's a `Option<FunctionValueDefinition<T>>` ?

leonardoalt: Maybe not too important

ast/src/analyzed/mod.rs
```diff
@@ -370,6 +370,7 @@ pub struct PublicDeclaration {
     pub source: SourceRef,
     pub name: String,
     pub polynomial: PolynomialReference,
+    pub array_index: Option<DegreeType>,
```
leonardoalt, line 373: Why `DegreeType`? The array length isn't related to the size of the columns

ast/src/parsed/mod.rs
```diff
@@ -23,7 +23,8 @@ pub enum PilStatement<T> {
     PublicDeclaration(
         usize,
         String,
-        NamespacedPolynomialReference<T>,
+        NamespacedPolynomialReference,
+        Option<Expression<T>>,
         Expression<T>,
```
leonardoalt, line 28: This enum looks quite obscure now, a comment explaining what the arguments are would help.

backend/src/pilstark/json_exporter/mod.rs
leonardoalt: What's the conclusion here?

executor/src/witgen/query_processor.rs
```diff
@@ -139,30 +139,24 @@ where
...
+                    next,
+                };
+                Ok(rows
+                    .get_value(&poly_ref)
+                    .ok_or(IncompleteCause::DataNotYetAvailable)?)
+            }
+            PolynomialType::Constant => {
+                let values = self.fixed_data.fixed_cols[&poly_id].values;
+                let row = rows.current_row_index as usize + next as usize;
+                Ok(values[row % values.len()])
```
leonardoalt, line 158: I'm guessing this doesn't actually support arrays in witgen?

parser/src/lib.rs
```diff
@@ -235,6 +235,16 @@ public out = y(%last_row);"#;
             assert_eq!(input.trim(), printed.trim());
         }
 
+        #[test]
+        fn reparse_arrays() {
+            let input = "pol commit y[3];\n(y - 2) = 0;\n(y[2] - 2) = 0;\npublic out = y[1](2);";
```
leonardoalt, line 240: what does `(y - 2)` mean? Is it universally quantified?

leonardoalt (commented):
Looks good, a few questions left

ast/src/analyzed/mod.rs
chriseth: We essentially have "field element" or "degree type". Would you prefer `usize`?

ast/src/parsed/mod.rs
chriseth: I can extract it into a struct.

executor/src/witgen/query_processor.rs
chriseth: What do you mean? This is accessing a row of a fixed column. It works in witgen because this is the prover query part, it has access to anything the prover / witgen has access to.

parser/src/lib.rs
chriseth: This just parses, so yes, it should break at some later type checking phase.

pil_analyzer/src/condenser.rs
chriseth: It's an `Option` because the symbol can just be a declaration without a value. `FunctionValueDefinition` is an enum that contains all the different ways to define a column:
```
col fixed x(i) { ... };
col fixed x = [1,2,3]* + [0];
col fixed x = |i| { ... }; // or col fixed x = y;
```

ast/src/parsed/mod.rs
chriseth: Although the parsed enums all don't have structs to make it easier with the grammar. I'll add some comments...

chriseth: I mean in the end this should be an Expresison, to be honest, and then checked by the condenser that it has the right format.

chriseth: Yeah, let me change this a bit.

ast/src/analyzed/mod.rs
leonardoalt: I think that'd be clearer

executor/src/witgen/query_processor.rs
leonardoalt: the removed code below used to issue an error that "arrays are not supported", but I don't see how that's not reachable anymore

parser/src/lib.rs
leonardoalt: ok

chriseth:
Ok, I wanted to simplify PublicDeclaration, but I guess this has to wait until `FunctionCall` properly takes an Expression in the "function" part.

executor/src/witgen/query_processor.rs
chriseth: Oh right, it turns out it was easy to support them.

chriseth:
Ready for another review.

executor/src/witgen/query_processor.rs
leonardoalt: ok

leonardoalt (approved):

PR #749, opened 2023-11-06
Get degree from namespaces

Schaeff:
We currently get the degree of the table from the length of the first constant column. This creates a hard requirement of having at least one constant column. This PR removes this requirement by getting the length of the table from the degree of the namespaces: they must all match (which is already a requirement now).

I ended up simplifying code in a few places where the degree was passed around, as it is now accessible directly from `Analyzed`.

### DIFF
--- a/backend/src/pilstark/estark.rs
+++ b/backend/src/pilstark/estark.rs
@@ -1,3 +1,5 @@
+use std::iter::{once, repeat};
+
 use crate::{pilstark, BackendImpl};
 use ast::analyzed::Analyzed;
 use number::{BigInt, DegreeType, FieldElement, GoldilocksField};
@@ -58,6 +60,8 @@ impl<F: FieldElement> BackendImpl<F> for EStark {
 
         log::info!("Creating eSTARK proof.");
 
+        let degree = pil.degree();
+
         let mut pil: PIL = pilstark::json_exporter::export(pil);
 
         // TODO starky requires a fixed column with the equivalent
@@ -77,7 +81,7 @@ impl<F: FieldElement> BackendImpl<F> for EStark {
                     polType: None,
                     type_: "constP".to_string(),
                     id: fixed.len(),
-                    polDeg: fixed[0].1.len(),
+                    polDeg: degree as usize,
                     isArray: false,
                     elementType: None,
                     len: None,
@@ -85,7 +89,10 @@ impl<F: FieldElement> BackendImpl<F> for EStark {
             );
             fixed.push((
                 "main.first_step".to_string(),
-                [vec![F::one()], vec![F::zero(); fixed[0].1.len() - 1]].concat(),
+                once(F::one())
+                    .chain(repeat(F::zero()))
+                    .take(degree as usize)
+                    .collect(),
             ));
         }
 
--- a/compiler/src/lib.rs
+++ b/compiler/src/lib.rs
@@ -226,13 +226,13 @@ fn compile<T: FieldElement, Q: QueryCallback<T>>(
     log::info!("Wrote {}.", optimized_pil_file_name.to_str().unwrap());
     let start = Instant::now();
     log::info!("Evaluating fixed columns...");
-    let (constants, degree) = constant_evaluator::generate(&analyzed);
+    let constants = constant_evaluator::generate(&analyzed);
     log::info!("Took {}", start.elapsed().as_secs_f32());
 
     let witness = (analyzed.constant_count() == constants.len()).then(|| {
         log::info!("Deducing witness columns...");
         let commits =
-            executor::witgen::WitnessGenerator::new(&analyzed, degree, &constants, query_callback)
+            executor::witgen::WitnessGenerator::new(&analyzed, &constants, query_callback)
                 .with_external_witness_values(external_witness_values)
                 .generate();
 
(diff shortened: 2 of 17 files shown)


backend/src/pilstark/estark.rs
```diff
@@ -77,15 +81,18 @@ impl<F: FieldElement> BackendImpl<F> for EStark {
...
                     len: None,
                 },
             );
             fixed.push((
                 "main.first_step",
-                [vec![F::one()], vec![F::zero(); fixed[0].1.len() - 1]].concat(),
+                once(F::one())
+                    .chain(repeat(F::zero()))
+                    .take(degree as usize)
+                    .collect(),
```
Schaeff, line 95: Kinda unrelated to this PR but it seemed cleaner to me

number/src/serialize.rs
```diff
@@ -81,14 +81,24 @@ fn ceil_div(num: usize, div: usize) -> usize {
     (num + div - 1) / div
 }
 
-pub fn write_polys_file<T: FieldElement>(
-    file: &mut impl Write,
-    degree: DegreeType,
-    polys: &[(&str, Vec<T>)],
-) {
+pub fn write_polys_file<T: FieldElement>(file: &mut impl Write, polys: &[(&str, Vec<T>)]) {
```
Schaeff, line 84: This is also a bit unrelated, but all constants must have the same length anyway, so we can infer that length from the values without passing the degree.

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -361,7 +370,10 @@ impl<T: FieldElement> PILAnalyzer<T> {
             id,
             source,
             absolute_name,
-            degree: self.polynomial_degree,
+            degree: self.polynomial_degree.unwrap_or_else(|| {
+                assert!(matches!(symbol_kind, SymbolKind::Constant()));
+                0
+            }),
```
Schaeff, line 376: This is a bit weird and would require a deeper refactor: constants can be defined outside of namespaces, where the notion of degree does not exist. But we represent them as `Symbol` which requires passing a degree. 0 is already passed before this PR, I just made sure that we must be dealing a constant if this happens.

chriseth:
Looks good in general. I'm wondering if we should take this opportunity to clean up terminology (or at least not spread confusion further) and call this concept "row_count" instead of "degree". It is correct that it is the degree of a certain polynomial, but we don't even touch that polynomial.

pil_analyzer/src/pil_analyzer.rs
chriseth: Yes, degree should be removed from Symbol at some point.

leonardoalt:
> Looks good in general. I'm wondering if we should take this opportunity to clean up terminology (or at least not spread confusion further) and call this concept "row_count" instead of "degree". It is correct that it is the degree of a certain polynomial, but we don't even touch that polynomial.

yea agree if it isn't too much work

leonardoalt:
has conflicts now

compiler/src/lib.rs
```diff
@@ -238,8 +237,8 @@ fn compile<T: FieldElement, Q: QueryCallback<T>>(
             .map(|(name, c)| (name, c))
             .collect::<Vec<_>>();
 
-        write_constants_to_fs(&constants, output_dir, degree);
-        write_commits_to_fs(&witness, output_dir, degree);
```
leonardoalt, line 242: Ah, these were extracted out in my PR, so it's probably part of the conflict

executor/src/witgen/mod.rs
```diff
@@ -85,12 +81,8 @@ impl<'a, 'b, T: FieldElement, Q: QueryCallback<T>> WitnessGenerator<'a, 'b, T, Q
     /// Generates the committed polynomial values
     /// @returns the values (in source order) and the degree of the polynomials.
     pub fn generate(self) -> Vec<(&'a str, Vec<T>)> {
-        if self.degree.is_zero() {
-            panic!("Resulting degree is zero. Please ensure that there is at least one non-constant fixed column to set the degree.");
-        }
```
leonardoalt, line 90: Shouldn't we keep this?

halo2/src/prover.rs
```diff
@@ -111,7 +111,7 @@ impl Halo2Prover {
         log::info!("Starting proof aggregation...");
 
         // TODO this is hacky
```
leonardoalt, line 113: This comment can now be removed

halo2/src/prover.rs
```diff
@@ -65,7 +65,7 @@ impl Halo2Prover {
         witness: &[(&str, Vec<F>)],
     ) -> Vec<u8> {
         // TODO this is hacky
```
leonardoalt, line 67: This comment can now be removed

number/src/serialize.rs
leonardoalt: Ah, yes I did the same in my PR which was already merged, so there might be some conflicts/diffs here too

halo2/src/prover.rs
Schaeff: Isn't it still hacky to do this `leading_zero` magic?

executor/src/witgen/mod.rs
Schaeff: I don't think so: having degree zero is being used as some sort of error management, but it should not be. Having degree zero here should just return empty columns.

Schaeff:
I'd keep potential renaming to another PR. degree is used everywhere at the moment (including in user facing code, docs, etc), this PR is consistent with that and makes the future renaming easier.

halo2/src/prover.rs
leonardoalt: that we can replace by log2, but I think the hack was pointing to reading length from column

leonardoalt (approved):

PR #750, opened 2023-11-06
compiler cleanup

leonardoalt:
This PR moves *almost* all FS operations out of the `compiler` crate. PIL and ASM files are still output by that crate, but I decided to leave it to a different PR.

The main motivation to stop writing witnesses and proofs directly in `compiler` is that that's something the client should do. In fact, the client may not even want to write them to disk yet, and do some sort of continuations or aggregation. Currently the main client is the CLI crate, who does write them to disk. It's also important to keep this separated because the compiler only knew about one witness and one proof, but with continuations we will have more and that's something the compiler should not be concerned about. PIL and ASM files are fine to stay there for now because they are always a single artifact.

### DIFF
--- a/backend/src/halo2_impl.rs
+++ b/backend/src/halo2_impl.rs
@@ -14,8 +14,8 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Prover {
     fn prove(
         &self,
         pil: &Analyzed<T>,
-        fixed: &[(&str, Vec<T>)],
-        witness: &[(&str, Vec<T>)],
+        fixed: &[(String, Vec<T>)],
+        witness: &[(String, Vec<T>)],
         prev_proof: Option<Proof>,
     ) -> (Option<Proof>, Option<String>) {
         let proof = match prev_proof {
@@ -47,8 +47,8 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Mock {
     fn prove(
         &self,
         pil: &Analyzed<T>,
-        fixed: &[(&str, Vec<T>)],
-        witness: &[(&str, Vec<T>)],
+        fixed: &[(String, Vec<T>)],
+        witness: &[(String, Vec<T>)],
         prev_proof: Option<Proof>,
     ) -> (Option<Proof>, Option<String>) {
         if prev_proof.is_some() {
--- a/compiler/src/lib.rs
+++ b/compiler/src/lib.rs
@@ -4,8 +4,6 @@
 
 use std::ffi::OsStr;
 use std::fs;
-use std::io::BufWriter;
-use std::io::Write;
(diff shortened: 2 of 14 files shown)


powdr_cli/src/main.rs
```diff
@@ -605,6 +618,65 @@ fn optimize_and_output<T: FieldElement>(file: &str) {
...
+        Some(json) => {
+            let to_write = output_dir.join("constraints.json");
+            let mut file = fs::File::create(&to_write).unwrap();
+            file.write_all(json.as_bytes()).unwrap();
+            log::info!("Wrote {}.", to_write.display());
+        }
+        None => log::warn!("Constraints were not JSON serialized"),
+    }
+}
+
```
leonardoalt, line 679: The last 3 functions here were just moved from `compiler` to here.

compiler/src/lib.rs
```diff
@@ -273,66 +271,12 @@ fn compile<T: FieldElement, Q: QueryCallback<T>>(
...
-    commits: &[(&str, Vec<T>)],
-    output_dir: &Path,
-    degree: DegreeType,
-) {
-    write_polys_file(
-        &mut BufWriter::new(&mut fs::File::create(output_dir.join("commits.bin")).unwrap()),
-        degree,
-        commits,
-    );
-    log::info!("Wrote commits.bin.");
```
leonardoalt, line 335: These functions were moved to CLI.

halo2/src/mock_prover.rs
```diff
@@ -106,6 +109,10 @@ mod test {
         let witness =
             executor::witgen::WitnessGenerator::new(&analyzed, degree, &fixed, query_callback)
                 .generate();
+
+        let fixed = convert_witness(fixed);
```
chriseth, line 113: to_owned_values? weird to call a function `convert_witness` on fixed columns.

backend/src/halo2_impl.rs
```diff
@@ -14,8 +14,8 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Prover {
     fn prove(
         &self,
         pil: &Analyzed<T>,
-        fixed: &[(&str, Vec<T>)],
-        witness: &[(&str, Vec<T>)],
+        fixed: &[(String, Vec<T>)],
```
chriseth, line 17: Ah nice, the conversion to owned names is what I need for arrays as well

halo2/src/mock_prover.rs
leonardoalt: conceptually everything is witness, that's why I chose this. But will change

chriseth (dismissed):

chriseth (dismissed):

chriseth (approved):

PR #752, opened 2023-11-06
On success, RV32 instruction tests returns from __runtime_start.

lvella:
Instead of infinite looping. Leave the infinite loop for the runtime/linking.


lvella:
This looks like a mostly harmless PR, but for some reason each instruction test prints a ton binary garbage to the stdout. We have had these mysterious garbage output for sometime now, but this is next level.

leonardoalt (approved):

lvella:
As @leonardoalt pointed out, it seems the garbage on the stdout is the expected result, due to a number of `ebreak` instructions contained in the tests, which we use to implement print. We probably should use a function or some other mechanism.

PR #754, opened 2023-11-07
Instruction tests properly returning from __runtime_start.

lvella:
It seems #752 actually broke instruction tests, and this is my attempt to fix it.

### DIFF
--- a/riscv/tests/instruction_tests/generated/add.S
+++ b/riscv/tests/instruction_tests/generated/add.S
@@ -79,7 +79,7 @@
 # 12 "sources/add.S" 2
 
 
-.globl __runtime_start; __runtime_start:
+.globl __runtime_start; __runtime_start: la x10,__return_pointer; sw x1,0(x10); li x10,0
 
   #-------------------------------------------------------------
   # Arithmetic tests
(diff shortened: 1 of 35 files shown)


chriseth:
Can you explain what this does?

leonardoalt:
> Can you explain what this does?

the main change is the `.h` file.
Before it had the infinite loop we used to require, but that's not needed anymore and causes problems for the executor. The pass behavior should be to just end execution without fails

lvella:
> Can you explain what this does?

In #752 I removed the infinite loop reached at the end of a successful instruction tests. Instead, I tried to simply return from `__runtime_start`. This makes a difference for the executor, as the tests would actually terminate.

Turns out that some tests rewrites the `ra` register, breaking `ret`. In this PR I fix my initial attempt, by saving the return address in the memory location `__return_pointer` before executing the test, and loading it back before `ret`ing,

lvella:
`lb` and `lbu` are failing now. Do you know why this new entry in the data section would upset them? Shouldn't affect the alignment, right?

lvella:
I think `lrsc` is broken, but has been broken for a long time now.

lvella:
Yep just tested locally and `lrsc` was broken since before #752. I think this can be merged as is.

leonardoalt (approved):

PR #756, opened 2023-11-07
Document syntax a little.

chriseth:
Depends on https://github.com/powdr-labs/powdr/pull/773

### DIFF
--- a/book/src/pil/declarations.md
+++ b/book/src/pil/declarations.md
@@ -1,0 +1,37 @@
+# Declarations
+
+Powdr-pil allows the same syntax to declare various kinds of symbols. This includes
+constants, fixed columns, witness columns and even macros. It deduces the symbol kind
+by its type and the way the symbol is used.
+
+Symbols can be declared using ``let <name>;`` and they can be declared and defined
+using ``let <name> = <value>;``, where ``<value>`` is an expression.
+This syntax can be used for constants, fixed columns, witness columns and even (higher-order)
+functions that can transform expressions. The kind of symbol is deduced by its type and the
+way the symbol is used:
+
+- symbols without a value are witness columns,
+- symbols evaluating to a number are constants,
+- symbols defined as a function with a single parameter are fixed columns and
+- everything else is a "generic symbol" that is not a column.
+
+Examples:
+
+```rust
+// This defines a constant
+let rows = 2**16;
+// This defines a fixed column that contains the row number in each row.
+let step = |i| i;
+// Here, we have a witness column.
+let x;
+// This functions returns the square of its input (classified as a fixed column).
+let square = |x| x*x;
+// A recursive function, taking a function and an integer as parameter
+let sum = |f, i| match i {
+    0 => f(0),
+    _ => f(i) + sum(f, i - 1)
+};
+// The same function as "square" above, but employing a trick to avoid it
+// being classified as a column.
+let square_non_column = (|| |x| x*x)();
+```
--- a/book/src/pil/expressions.md
+++ b/book/src/pil/expressions.md
@@ -1,0 +1,72 @@
+# Expressions
+
+Depending on the context, powdr allows more or less features for expressions.
+
+Inside values for declarations, you can use a very flexible language which includes
+many different operators, function calls, lambda functions, tuple types, statement blocks,
+match statements and others.
+
+In statements and expressions that are required to evaluate to polynomial identities, only a much more restrictive
+language can be used. Expressions in that language are caleld Algebraic Expressions. While you can use
+the full language everywhere, in the context of a polynomial identity, the result after function evaluation
+and constant propagation has to be an algebraic expression.
+
+## Generic Expressions
+
+The expression language allows the following operators, in order of increased precedence:
+
+- lambda functions: ``|params| body``. Examples: ``|i| i`` (the identity), ``|a, b| a + b`` (sum)
+- ``||`` - logical or
+- ``&&`` - logical and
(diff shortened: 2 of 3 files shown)


book/src/pil/generic_declarations.md
```diff
@@ -0,0 +1,43 @@
+# Generic Declarations
+
+powdr-pil allows the same syntax to declare various kinds of symbols. This includes
+constants, fixed columns, witness columns and even macros. It deduces the symbol kind
+by its type nad the way the symbol is used.
```
Schaeff, line 5:
```suggestion
by its type and the way the symbol is used.
```

book/src/pil/generic_declarations.md
```diff
@@ -0,0 +1,43 @@
...
+- The operator ``**`` must have a number as exponent.
+- The operator `[i]` must have a column name on the left-hand side and the index must be a number.
+- The operator `'` must have a column or `[i]` on the left-hand-side.
+- No other operators are allowed.
+
+Of course, arbitrary parentheses are allowed.
+
+### Generic Expressions
+
+TODO
```
Schaeff, line 43: Is this WIP for this PR?

book/src/pil/generic_declarations.md
```diff
@@ -0,0 +1,43 @@
...
+After inlining function calls and constant propagation, the resulting expression has to
+be an "algebraic expression". These are restricted in the following way:
+
+- You can freely use the operators  ``+``, ``-``,``*``.
+- The operator ``**`` must have a number as exponent.
+- The operator `[i]` must have a column name on the left-hand side and the index must be a number.
+- The operator `'` must have a column or `[i]` on the left-hand-side.
+- No other operators are allowed.
+
+Of course, arbitrary parentheses are allowed.
```
Schaeff, line 39:
```suggestion
Arbitrary parentheses are allowed.
```

book/src/pil/generic_declarations.md
```diff
@@ -0,0 +1,43 @@
...
+
```
Schaeff, line 40: A code block with examples and counter examples would be nice here as well

book/src/pil/generic_declarations.md
chriseth: no, just wanted to get something going for now.

chriseth:
Updated this a little.

book/src/pil/declarations.md
```diff
@@ -0,0 +1,27 @@
...
+let step = |i| i;
+// Here, we have a witness column.
+let x;
+// This functions returns the square of its input.
+let square = |x| x*x;
+// A recursive function, taking a function and an integer as parameter
+let sum = |f, i| match i {
+    0 => f(0),
+    _ => f(i) + sum(f, i - 1)
+};
```
leonardoalt, line 26: Wait, can we actually evaluate this?!

book/src/pil/declarations.md
```diff
@@ -0,0 +1,27 @@
+# Declarations
+
+Powdr-pil allows the same syntax to declare various kinds of symbols. This includes
+constants, fixed columns, witness columns and even macros. It deduces the symbol kind
+by its type and the way the symbol is used.
+
```
leonardoalt, line 6: Maybe worth mentioning here that the default type is the given field element?

book/src/pil/expressions.md
```diff
@@ -0,0 +1,66 @@
...
+- ``'`` - "next row" operator (suffix)
+- ``[]``, ``()`` - array index access and function calls
+
+Elementary expressions are
+- number literals (integers)
+- string literals, written in double quotes, e.g. ``"hello"``
+- array literals written in square brackets, e.g. ``[1, 2, 3]``
+- tuples, having at least two elements, e.g. `(1, "abc")`
+- match expressions (see below).
+
```
leonardoalt, line 39: What about the fixed array operators?

leonardoalt:
build/test fails

book/src/pil/declarations.md
chriseth: Actually the way it is implemented now is more like python. Maybe better to not mention anything until we have a proper type checker?

book/src/pil/declarations.md
chriseth: Sure, why not? I mean when all the pending PRs are merged.

book/src/pil/expressions.md
leonardoalt: they're not mentioned here

leonardoalt: I mean like `[1, 0, 0] + [0, 1]*`

chriseth: Oh these are not part of the expression language. I mean with allowing `+` for array concatenation, they kind of are, but iteration is not.

leonardoalt: I don't understand why not. If we mention array literals here, aren't these just a way to represent array literals?

chriseth: `*` is not implemented for array literals. What you are referring to is a special syntax to declare fixed columns. I don't see how to especially implement the `*` in the generic expression syntax, since the result is not really an array type (you cannot determine the length without context). It would be better to evaluate those to functions instead of arrays (in the end, fixed columns are also functions and not arrays).

```
x + y => |i| if i < x.len { x(i) } else { y(i - x.len) }
x* => |i| x(i % x.len)
```

Note that both operators need a `.len` property on the input, so they actually should also return such a property, but for `*` it does not exist.

leonardoalt: ok

leonardoalt (approved):

leonardoalt (approved):

PR #758, opened 2023-11-07
Arbitrary expressions at statement level

chriseth:


ast/src/parsed/mod.rs
```diff
@@ -54,7 +54,7 @@ pub enum PilStatement<T> {
         Vec<PilStatement<T>>,
         Option<Expression<T>>,
     ),
-    FunctionCall(usize, String, Vec<Expression<T>>),
+    Expression(usize, Box<Expression<T>>),
```
Schaeff, line 57: I don't think it needs to be boxed, see the `Namespace` variant of this enum.

parser/src/powdr.lalrpop
```diff
@@ -257,8 +257,7 @@ InstructionBodyElement: PilStatement<T> = {
     PolynomialIdentity,
     PlookupIdentity,
     PermutationIdentity,
-    // We could use FunctionCallStatement here, but it makes lalrpop fail to build
-    <@L> <Identifier> "(" <ExpressionList> ")" => PilStatement::FunctionCall(<>)
+    <@L> <BoxedExpression> => PilStatement::Expression(<>)
```
Schaeff, line 260:
```suggestion
    ExpressionStatement,
```

ast/src/parsed/mod.rs
chriseth: Ah right!

chriseth:
Updated.

analysis/src/macro_expansion.rs
```diff
@@ -23,7 +20,6 @@ pub struct MacroExpander<T> {
     macros: HashMap<String, MacroDefinition<T>>,
     arguments: Vec<Expression<T>>,
     parameter_names: HashMap<String, usize>,
-    shadowing_locals: HashSet<String>,
```
chriseth, line 26: This was unused. Removing it did not break any tests, and we are replacing macros by functions anyway soon.

chriseth:
Can squash after review

georgwiese (approved):
Hmm, looks good to me, although I must say I'm not very deep in the compiler stack yet and not sure I fully understand the implications of this change :/

ast/src/parsed/mod.rs
```diff
@@ -54,7 +54,7 @@ pub enum PilStatement<T> {
         Vec<PilStatement<T>>,
         Option<Expression<T>>,
     ),
-    FunctionCall(usize, String, Vec<Expression<T>>),
+    Expression(usize, Expression<T>),
```
leonardoalt, line 57: How is this still used to represent the same thing?

leonardoalt: ah I see now

leonardoalt (approved):

Issue #367, opened 2023-07-06
Parser cannot parse namespaced names


chriseth:
The parser is not able to parse namespaced names inside expressions (inside function bodies for example).

Example:
```
namespace Assembly(2);
    col fixed A = [0]*;
    col fixed C(i) { Assembly.A(i) };
    col witness w;
    
    w = C;
```

This fails to parse `Assembly.A` properly.

emilyaherbert:
@chriseth this isn't just a limitation of parsing but also the AST:
https://github.com/powdr-org/powdr/blob/5d14f71e5389c699403c4e2c5625ac6c29cfd377/ast/src/parsed/asm.rs#L94

Do we want to allow for the use of `String` and any `PolynomialReference`? like:
https://github.com/powdr-org/powdr/blob/5d14f71e5389c699403c4e2c5625ac6c29cfd377/ast/src/parsed/mod.rs#L71

(`Assembly.A` in this example is parsed to `PolynomialReference` here):
https://github.com/powdr-org/powdr/blob/5d14f71e5389c699403c4e2c5625ac6c29cfd377/parser/src/powdr.lalrpop#L367

chriseth:
Yes, I think the proper way would be to replace the whole `PolynomialReference` construct by a combination of Path, IndexAccess and a Next unary operator.

The Analyzed AST can then combine it again into a single construct.

PR #759, opened 2023-11-07
Allow arbitrary expressions to be function-called.

chriseth:
Before this PR, the function part of a function call (`<function>(<arguments>)`) only allowed an explicit string. This change allows terms, i.e. you can do things like `x[0](7)` or `(|i| i + 1)(8)` and all the other fun stuff you could expect from a language.

Fixes https://github.com/powdr-labs/powdr/issues/367

Depends on #776

### DIFF
--- a/analysis/src/macro_expansion.rs
+++ b/analysis/src/macro_expansion.rs
@@ -7,7 +7,7 @@ use ast::parsed::{
     asm::{ASMProgram, Instruction, InstructionBody, Machine, MachineStatement},
     folder::Folder,
     visitor::ExpressionVisitable,
-    Expression, FunctionCall, FunctionDefinition, PilStatement,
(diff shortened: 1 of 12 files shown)


Schaeff:
nice pun

chriseth:
Just realized that in order for this to work, I should probably first extend the evaluator to return values of arbitrary types, because at least in order to evaluate `f(x)` for arbitrary `f`, the evaluator needs to be able to evaluate arbitrary function-typed `f`.

leonardoalt:
> This change allows terms, i.e. you can do things like x[0](7) or (|i| i + 1)(8) and all the other fun stuff you could expect from a language.

Solidity function types are calling you

leonardoalt (approved):

Issue #673, opened 2023-10-05
Constant evaluator refactoring


chriseth:
The constant evaluator is quite lonely in the executor crate, and it actually does not really fit there. It might be better to create a new crate that is just about evaluating fully-fledged PIL, and that crate will be used both by the pil analyzer (to evaluate compile time constants used in e.g. array lengths) and the constant evaluator.

Then we can also simplify some things in the witness generator: The expressions the witness generator uses are much much simpler than the fully-fledged PIL language, it only has some binary operations, only works on field elements, has no functions, etc.
So it would make sense to create a new expression type for the witness generator.

PR #761, opened 2023-11-09
Evaluate complex expressions including recursion and lambda functions

chriseth:
Fixes https://github.com/powdr-labs/powdr/issues/673

Depends on #767

### DIFF
--- a/compiler/src/lib.rs
+++ b/compiler/src/lib.rs
@@ -335,6 +335,12 @@ pub fn inputs_to_query_callback<T: FieldElement>(inputs: Vec<T>) -> impl QueryCa
         // TODO In the future, when match statements need to be exhaustive,
         // This function probably gets an Option as argument and it should
         // answer None by Ok(None).
+
+        // We are expecting a tuple
+        let query = query
+            .strip_prefix('(')
+            .and_then(|q| q.strip_suffix(')'))
+            .ok_or_else(|| "Prover query has to be a tuple".to_string())?;
         let items = query.split(',').map(|s| s.trim()).collect::<Vec<_>>();
         match &items[..] {
             ["\"input\"", index] => {
--- a/executor/src/constant_evaluator/mod.rs
+++ b/executor/src/constant_evaluator/mod.rs
@@ -1,9 +1,9 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, fmt::Display, rc::Rc};
 
 use ast::analyzed::{Analyzed, FunctionValueDefinition};
 use itertools::Itertools;
 use number::{DegreeType, FieldElement};
-use pil_analyzer::evaluator::Evaluator;
+use pil_analyzer::evaluator::{self, Closure, Custom, EvalError, SymbolLookup, Value};
 use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 
 /// Generates the constant polynomial values for all constant polynomials
@@ -28,34 +28,36 @@ fn generate_values<T: FieldElement>(
     analyzed: &Analyzed<T>,
     degree: DegreeType,
     body: &FunctionValueDefinition<T>,
-    other_constants: &HashMap<&str, Vec<T>>,
+    computed_columns: &HashMap<&str, Vec<T>>,
 ) -> Vec<T> {
+    let symbols = Symbols {
+        analyzed,
+        computed_columns,
+    };
+    // TODO we should maybe pre-compute some symbols here.
     match body {
         FunctionValueDefinition::Mapping(body) => (0..degree)
             .into_par_iter()
             .map(|i| {
-                Evaluator {
-                    definitions: &analyzed.definitions,
-                    variables: &[i.into()],
-                    function_cache: other_constants,
-                }
-                .evaluate(body)
-                .unwrap()
+                evaluator::evaluate_function_call(body, vec![i.into()], &symbols)
+                    .unwrap()
+                    .try_to_number()
+                    .unwrap()
             })
             .collect(),
         FunctionValueDefinition::Array(values) => {
-            let evaluator = Evaluator {
-                definitions: &analyzed.definitions,
-                variables: &[],
-                function_cache: other_constants,
-            };
             let values: Vec<_> = values
                 .iter()
                 .flat_map(|elements| {
                     let items = elements
                         .pattern()
                         .iter()
-                        .map(|v| evaluator.evaluate(v).unwrap())
+                        .map(|v| {
+                            evaluator::evaluate(v, &symbols)
+                                .unwrap()
+                                .try_to_number()
+                                .unwrap()
+                        })
                         .collect::<Vec<_>>();
 
                     items
@@ -72,6 +74,77 @@ fn generate_values<T: FieldElement>(
         FunctionValueDefinition::Expression(_) => {
             panic!("Expression used for fixed column, only expected for intermediate polynomials")
         }
+    }
+}
+
+struct Symbols<'a, T> {
+    pub analyzed: &'a Analyzed<T>,
+    pub computed_columns: &'a HashMap<&'a str, Vec<T>>,
+}
+
+impl<'a, T: FieldElement> SymbolLookup<'a, T, FixedColumnRef<'a>> for Symbols<'a, T> {
+    fn lookup(&self, name: &str) -> Result<Value<'a, T, FixedColumnRef<'a>>, EvalError> {
+        Ok(
+            if let Some((name, _)) = self.computed_columns.get_key_value(name) {
+                Value::Custom(FixedColumnRef { name })
+            } else if let Some((_, value)) = self.analyzed.definitions.get(&name.to_string()) {
+                match value {
+                    Some(FunctionValueDefinition::Expression(value)) => {
+                        evaluator::evaluate(value, self)?
+                    }
+                    Some(FunctionValueDefinition::Mapping(body)) => (Closure {
+                        parameter_count: 1,
+                        body,
+                        environment: vec![],
+                    })
+                    .into(),
+                    Some(_) => Err(EvalError::Unsupported(
+                        "Cannot evaluate arrays and queries.".to_string(),
+                    ))?,
+                    None => Err(EvalError::Unsupported(
+                        "Cannot evaluate witness columns.".to_string(),
+                    ))?,
+                }
+            } else {
+                Err(EvalError::SymbolNotFound(format!(
+                    "Symbol {name} not found."
+                )))?
+            },
+        )
+    }
+
+    fn eval_function_application(
+        &self,
+        function: FixedColumnRef<'a>,
+        arguments: &[Rc<Value<'a, T, FixedColumnRef<'a>>>],
+    ) -> Result<Value<'a, T, FixedColumnRef<'a>>, EvalError> {
+        if arguments.len() != 1 {
+            Err(EvalError::TypeError(format!(
+                "Expected one argument, but got {}",
+                arguments.len()
+            )))?
+        };
+        let Value::Number(row) = arguments[0].as_ref() else {
+            return Err(EvalError::TypeError(format!(
+                "Expected number but got {}",
+                arguments[0]
+            )));
+        };
+        let data = &self.computed_columns[function.name];
+        Ok(Value::Number(data[row.to_degree() as usize % data.len()]))
+    }
+}
+
+#[derive(Clone, PartialEq)]
+pub struct FixedColumnRef<'a> {
+    pub name: &'a str,
+}
+
+impl<'a> Custom for FixedColumnRef<'a> {}
+
+impl<'a> Display for FixedColumnRef<'a> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.name)
     }
 }
 
@@ -343,5 +416,61 @@ mod test {
             constants[12],
             ("F.greater_eq", convert([0, 0, 0, 1, 1, 1].to_vec()))
         );
+    }
+
+    #[test]
+    #[should_panic = "Cannot evaluate witness columns"]
+    pub fn calling_witness() {
+        let src = r#"
+            constant %N = 10;
+            namespace F(%N);
+            let w;
+            let x = |i| w(i) + 1;
+        "#;
+        let analyzed = analyze_string::<GoldilocksField>(src);
+        assert_eq!(analyzed.degree(), 10);
+        generate(&analyzed);
+    }
+
+    #[test]
+    #[should_panic = "Symbol F.w not found"]
+    pub fn symbol_not_found() {
+        let src = r#"
+            constant %N = 10;
+            namespace F(%N);
+            let x = |i| w(i) + 1;
+        "#;
+        let analyzed = analyze_string::<GoldilocksField>(src);
+        assert_eq!(analyzed.degree(), 10);
+        generate(&analyzed);
+    }
+
+    #[test]
+    #[should_panic = "Cannot evaluate arrays"]
+    pub fn forward_reference_to_array() {
+        let src = r#"
+            constant %N = 10;
+            namespace F(%N);
+            let x = |i| y(i) + 1;
+            col fixed y = [1, 2, 3]*;
+        "#;
+        let analyzed = analyze_string::<GoldilocksField>(src);
+        assert_eq!(analyzed.degree(), 10);
+        generate(&analyzed);
+    }
+
+    #[test]
+    pub fn forward_reference_to_function() {
+        let src = r#"
+            constant %N = 4;
+            namespace F(%N);
+            let x = |i| y(i) + 1;
+            let y = |i| i + 20;
+        "#;
+        let analyzed = analyze_string::<GoldilocksField>(src);
+        assert_eq!(analyzed.degree(), 4);
+        let constants = generate(&analyzed);
+        assert_eq!(constants[0], ("F.x", convert([21, 22, 23, 24].to_vec())));
+        assert_eq!(constants[1], ("F.y", convert([20, 21, 22, 23].to_vec())));
     }
 }
--- a/executor/src/witgen/query_processor.rs
+++ b/executor/src/witgen/query_processor.rs
@@ -1,14 +1,11 @@
-use ast::{
-    analyzed::{AlgebraicReference, Expression, PolyID, PolynomialType, Reference},
-    evaluate_binary_operation, evaluate_unary_operation,
-    parsed::{FunctionCall, MatchArm, MatchPattern},
-};
+use std::{fmt::Display, rc::Rc};
 
-use number::{DegreeType, FieldElement};
+use ast::analyzed::{AlgebraicReference, Expression, PolyID, PolynomialType};
+use num_traits::ToPrimitive;
+use number::FieldElement;
+use pil_analyzer::evaluator::{self, Custom, EvalError, SymbolLookup, Value};
 
-use super::{
-    rows::RowPair, Constraint, EvalError, EvalResult, EvalValue, FixedData, IncompleteCause,
-};
+use super::{rows::RowPair, Constraint, EvalResult, EvalValue, FixedData, IncompleteCause};
 
 /// Computes value updates that result from a query.
 pub struct QueryProcessor<'a, 'b, T: FieldElement, QueryCallback: Send + Sync> {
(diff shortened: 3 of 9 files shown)


pil_analyzer/src/evaluator.rs
```diff
@@ -40,65 +44,253 @@ pub fn compute_constants<T: FieldElement>(
...
+}
+
+impl<T: Eq> PartialEq for Value<T> {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Number(l), Self::Number(r)) => l == r,
+            (Self::String(l), Self::String(r)) => l == r,
+            (Self::Tuple(l), Self::Tuple(r)) => l == r,
+            (Self::Array(l), Self::Array(r)) => l == r,
+            (Self::Closure(_, _), Self::Closure(_, _)) => panic!("Tried to compare closures."),
```
Schaeff, line 65: It seems like introducing a `Closure` type and putting the panic in its `PartialEq` impl would allow to derive `PartialEq` for `Value`, reducing boilerplate

chriseth: True! Also Closure is a rather complicated type.

chriseth:
On error, the evaluator currently creates strings, but some of those error cases ("no match arm found" (fixable) and "data not yet available") regularly occur during prover query string evaluation. We should avoid that.

asm_to_pil/src/vm_to_constrained.rs
```diff
@@ -576,6 +576,7 @@ impl<T: FieldElement> ASMPILConverter<T> {
             Expression::ArrayLiteral(_) => panic!(),
             Expression::MatchExpression(_, _) => panic!(),
             Expression::FreeInput(expr) => {
+                // TODO transform references
```
chriseth, line 579:
```suggestion
```

compiler/src/lib.rs
```diff
@@ -285,11 +285,16 @@ fn compile<T: FieldElement, Q: QueryCallback<T>>(
 #[allow(clippy::print_stdout)]
 pub fn inputs_to_query_callback<T: FieldElement>(inputs: Vec<T>) -> impl Fn(&str) -> Option<T> {
     move |query: &str| -> Option<T> {
+        // We are expecting a tuple
+        let query = query.strip_prefix('(').and_then(|q| q.strip_suffix(')'))?;
         let items = query.split(',').map(|s| s.trim()).collect::<Vec<_>>();
+        println!("it: {}", items[0]);
```
chriseth, line 291:
```suggestion
```

compiler/src/lib.rs
```diff
@@ -285,11 +285,16 @@ fn compile<T: FieldElement, Q: QueryCallback<T>>(
...
         match items[0] {
             "\"input\"" => {
+                println!("INPUT");
```
chriseth, line 294:
```suggestion
```

compiler/src/lib.rs
```diff
@@ -285,11 +285,16 @@ fn compile<T: FieldElement, Q: QueryCallback<T>>(
...
                 assert_eq!(items.len(), 2);
                 let index = items[1].parse::<usize>().unwrap();
+                println!("Index: {index}");
```
chriseth, line 297:
```suggestion
```

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,383 @@
...
+                        environment,
+                    }) => {
+                        assert_eq!(parameter_count, arguments.len());
+
+                        let local_vars =
+                            arguments.into_iter().chain(environment).collect::<Vec<_>>();
+
+                        evaluate(body, &local_vars, symbols)?
+                    }
+                    Value::Custom(values) => values.eval_function_application(&arguments)?,
```
chriseth, line 288: It might make sense to pass `symbols` here as a function argument. In implementations it turned out that we needed to store a reference inside the custom value. We could avoid that by just passing `symbols` here.

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,383 @@
...
+    String(String),
+    Tuple(Vec<Self>),
+    Array(Vec<Self>),
+    Closure(Closure<'a, T, C>),
+    Custom(C),
+}
+
+// TODO PartialEq is only needed because don't have proper patterns in match expressions.
+pub trait Custom<V>: Display + Clone + PartialEq {
+    fn eval_function_application(self, arguments: &[Rc<V>]) -> Result<V, String>;
```
chriseth, line 79: Maybe it would be better to move this into the Symbols trait and keep the Custom trait fuction-less (i.e. reduce it to the template parameter in Value).

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,362 @@
...
+                .into()
             }
-            Expression::LambdaExpression(_) => {
-                Err("Cannot evaluate lambda expression.".to_string())
+            Expression::IndexAccess(index_access) => {
+                match evaluate(&index_access.array, locals, symbols)? {
+                    Value::Array(elements) => {
+                        let index = evaluate_to_number(&index_access.index, locals, symbols)?;
+                        // TODO avoid conversion to bigint
+                        if index.to_arbitrary_integer() > elements.len().into() {
```
chriseth, line 237:
```suggestion
                        if index.to_arbitrary_integer() >= elements.len().into() {
```

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,395 @@
...
+    }
+
+    #[test]
+    pub fn capturing() {
+        let src = r#"namespace Main(16);
+            let f = |n, g| match n { 99 => |i| n, 1 => g(3) };
+            let result = f(1, f(99, |x| x + 3000));
+        "#;
+        // If the lambda function that f(99, ...) returns does not properly capture
+        // the value of n in a closure, f(1, ...) would return 1.
```
leonardoalt, line 392: This comment sounds a bit off

chriseth: Yeah, its grammar is a bit convoluted. What about:
```
If the lambda function returned by the expression f(99, ...) does not properly capture
the value of n in a closure, then f(1, ...) would return 1.
```

leonardoalt: sounds good!

chriseth:
Updated. New commits should be squsahed.

executor/src/witgen/query_processor.rs
```diff
@@ -65,105 +70,95 @@ where
...
+struct Symbols<'a, T: FieldElement> {
+    // TODO we should also provide access to non-column symbols.
+    fixed_data: &'a FixedData<'a, T>,
+    rows: &'a RowPair<'a, 'a, T>,
+}
 
-    fn evaluate_poly(
+impl<'a, T: FieldElement> SymbolLookup<'a, T, Reference> for Symbols<'a, T> {
+    fn lookup(&self, name: &str) -> Result<Value<'a, T, Reference>, EvalError> {
+        let poly_id = self.fixed_data.column_by_name(name);
```
georgwiese, line 96: This panics if the column doesn't exist, I guess instead we want to return a `EvalError::SymbolNotFound`?

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,394 @@
...
+                    parameter_count: lambda.params.len(),
+                    body: lambda.body.as_ref(),
+                    environment: locals.to_vec(),
+                })
+                .into()
             }
-            Expression::LambdaExpression(_) => {
-                Err("Cannot evaluate lambda expression.".to_string())
+            Expression::IndexAccess(index_access) => {
+                match evaluate(&index_access.array, locals, symbols)? {
```
georgwiese, line 257: Is it a performance issue that we evaluate the entire array each time we access a single element? I'm not that deep into how arrays are actually used in practice, so maybe it's fine.

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,394 @@
...
+        let src = r#"namespace Main(16);
+            let words = ["the", "quick", "brown", "fox"];
+            let translate = |w| match w {
+                "the" => "franz",
+                "quick" => "jagt",
+                "brown" => "mit",
+                "fox" => "dem",
+                _ => "?",
+            };
+            let extract = |arr, i| arr[i];
```
georgwiese, line 363: This is unused. is there a point in having it in the test?

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,394 @@
...
+            let translated = map_array(words, translate);
+        "#;
+        let result = parse_and_evaluate_symbol(src, "Main.translated");
+        assert_eq!(result, r#"["franz", "jagt", "mit", "dem"]"#);
+    }
+
+    #[test]
+    pub fn fibonacci() {
+        let src = r#"namespace Main(16);
+            let fib = |i| match i {
```
georgwiese, line 374: 🤯

georgwiese (commented):
Wow, this stuff is pretty wild, nice!

compiler/src/lib.rs
```diff
@@ -285,6 +285,8 @@ fn compile<T: FieldElement, Q: QueryCallback<T>>(
 #[allow(clippy::print_stdout)]
 pub fn inputs_to_query_callback<T: FieldElement>(inputs: Vec<T>) -> impl Fn(&str) -> Option<T> {
     move |query: &str| -> Option<T> {
+        // We are expecting a tuple
+        let query = query.strip_prefix('(').and_then(|q| q.strip_suffix(')'))?;
```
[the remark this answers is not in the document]
chriseth: Yes, good idea!

executor/src/constant_evaluator/mod.rs
```diff
@@ -75,6 +69,72 @@ fn generate_values<T: FieldElement>(
...
+                        body,
+                        environment: vec![],
+                    })
+                    .into(),
+                    _ => Err(EvalError::Unsupported(
+                        "Cannot evaluate arrays and queries.".to_string(),
+                    ))?,
+                }
+            } else {
+                panic!("Symbol {name} not found.")
```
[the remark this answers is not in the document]
chriseth: Right, I had more panics in the beginning and forgot this one.

executor/src/constant_evaluator/mod.rs
```diff
@@ -75,6 +69,72 @@ fn generate_values<T: FieldElement>(
...
+                    Some(FunctionValueDefinition::Expression(value)) => {
+                        evaluator::evaluate(value, self)?
+                    }
+                    Some(FunctionValueDefinition::Mapping(body)) => (Closure {
+                        parameter_count: 1,
+                        body,
+                        environment: vec![],
+                    })
+                    .into(),
+                    _ => Err(EvalError::Unsupported(
```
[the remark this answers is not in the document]
chriseth: True. So this would happen for `x(i)` where `x` is a witness column. I will add a test for that.

executor/src/constant_evaluator/mod.rs
```diff
@@ -75,6 +69,72 @@ fn generate_values<T: FieldElement>(
...
+                arguments[0]
+            )));
+        };
+        let data = &self.computed_columns[function.name];
+        Ok(Value::Number(data[row.to_degree() as usize % data.len()]))
+    }
+}
+
+#[derive(Clone, PartialEq)]
+pub struct FixedColumnRef<'a> {
```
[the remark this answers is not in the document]
chriseth: The distinction is that `computed_columns` only contain the columns that we already fully evaluated. So this is another point that should be tested: forward references. I mean with the evaluator this is not such a big deal anymore because it can just continue evaluating the other column instead of re-using the already computed values.

executor/src/witgen/query_processor.rs
```diff
@@ -49,7 +48,13 @@ where
     ) -> EvalValue<&'a AlgebraicReference, T> {
         let query_str = match self.interpolate_query(query, rows) {
             Ok(query) => query,
-            Err(incomplete) => return EvalValue::incomplete(incomplete),
+            Err(e) => match e {
+                EvalError::DataNotAvailable => {
+                    return EvalValue::incomplete(IncompleteCause::DataNotYetAvailable)
+                }
+                // All other errors are non-recoverable, so we should not run the query again.
+                _ => return EvalValue::complete([]),
```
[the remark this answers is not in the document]
chriseth: What is failing hard in that case? Would I need to change the signature of process_query to return a Result?

executor/src/witgen/query_processor.rs
```diff
@@ -65,105 +70,95 @@ where
...
+                let next =
+                    self.rows
+                        .row_number_next(row.to_degree())
+                        .ok_or(EvalError::OutOfBounds(format!(
+                            "Referenced row outside of window: {row}"
+                        )))?;
                 let poly_ref = AlgebraicReference {
-                    name: poly.to_string(),
-                    poly_id,
+                    name: String::new(),
```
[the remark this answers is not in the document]
chriseth: The lifetime of this object is just for the value lookup inside `self.rows`. I didn't want to store references inside the value returned by `lookup` because that's a mess. We could store a String, but it's really not necessary. Another way would be to provide an interface in `self.rows` that works by `poly_id`. But I can also just store an AlgebraicReference. Yes, it's only a matter of performance.

executor/src/witgen/query_processor.rs
```diff
@@ -65,105 +70,95 @@ where
...
+                self.rows
                     .get_value(&poly_ref)
-                    .ok_or(IncompleteCause::DataNotYetAvailable)?)
+                    .ok_or(EvalError::DataNotAvailable)?
             }
             PolynomialType::Constant => {
-                let values = self.fixed_data.fixed_cols[&poly_id].values;
-                Ok(values[row as usize % values.len()])
+                let values = self.fixed_data.fixed_cols[&function.poly_id].values;
+                // TODO can we avoid bigint?
```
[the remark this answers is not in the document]
chriseth: We could implement modulo on FieldElements, but I guess it is fine like this.

executor/src/witgen/query_processor.rs
```diff
@@ -65,105 +70,95 @@ where
...
+}
+
+impl PartialEq for Reference {
+    fn eq(&self, other: &Self) -> bool {
+        self.poly_id == other.poly_id
     }
 }
+
+// TODO not sure if we really need "Display"
+// If we don't then there is no need to actually implement a new type.
```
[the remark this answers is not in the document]
chriseth: I can try, but I'll stop before I burn in the borrow-checker hell.

pil_analyzer/src/condenser.rs
```diff
@@ -239,3 +239,24 @@ impl<T: FieldElement> Condenser<T> {
...
 }
+
+/// Returns a HashMap of all symbols that have a constant single value.
+fn compute_constants<T: FieldElement>(
+    definitions: &HashMap<String, (Symbol, Option<FunctionValueDefinition<T>>)>,
+) -> HashMap<String, T> {
+    definitions
+        .iter()
+        .filter_map(|(name, (symbol, value))| {
+            // TODO we could try to compute anything that evaluates to a "value" here.
```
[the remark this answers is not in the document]
chriseth: I think this was more difficult when I still had more panics in the code. I'll try!

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,394 @@
...
+}
+
+#[derive(Clone, PartialEq)]
+pub enum Value<'a, T, C> {
+    Number(T),
+    String(String),
+    Tuple(Vec<Self>),
+    Array(Vec<Self>),
+    Closure(Closure<'a, T, C>),
+    Custom(C),
```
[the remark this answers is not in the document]
chriseth: No, it can be anything. The idea is that this file implement the basic evaluator on a minimalistic functional language. The `Custom` and `SymbolLookup` mechanism allows the language to be extended. The extension is already different in the query_processor and in the constant_evaluator. Another extension will be added in https://github.com/powdr-labs/powdr/pull/773 where the condenser uses this enum alternative to either store an AlgebraicExpression or a Constraint.

So essentially all the other types can be seen as "built-in" core language and this one is the "user-defined" part.

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,394 @@
...
+        panic!("Tried to compare closures.");
+    }
+}
+
+impl<'a, T: Display, C> Display for Closure<'a, T, C> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "|{}| {}",
+            repeat('_').take(self.parameter_count).format(", "),
```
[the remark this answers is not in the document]
chriseth: I'll try to fix this in https://github.com/powdr-labs/powdr/pull/776

pil_analyzer/src/evaluator.rs
chriseth: I wondered about this as well, but I think this is also what you would expect from a programming language (except if you are used to lazy evaluation as default). If you call a function that returns an array, you expect this array to be "created" and are aware of the performance implications. Instead, if you already know that you just need a single element, you would not use an array but a function that implements index access.

pil_analyzer/src/evaluator.rs
chriseth: ah, I think I wanted to use it in the next line.

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,383 @@
...
-                };
-                (
-                    name.to_owned(),
-                    evaluate_expression(definitions, value).unwrap(),
-                )
-            })
+#[derive(Clone)]
+pub struct Closure<'a, T, C> {
+    // TODO we could also store the names of the parameters (for printing)
+    // In order to do this, we would need to add a parameter name to Mapping, which might be a good idea anyway.
```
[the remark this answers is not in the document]
chriseth: Yep, it's nicely fixed there: https://github.com/powdr-labs/powdr/pull/776/files#diff-6120a69f31b7a89a6b837da84213bce669bba81735efa11cd949e266f07042feR119

executor/src/constant_evaluator/mod.rs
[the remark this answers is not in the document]
chriseth: It's as I thought. Added two tests.

executor/src/witgen/query_processor.rs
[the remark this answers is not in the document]
chriseth: Got thought it unscathed.

pil_analyzer/src/condenser.rs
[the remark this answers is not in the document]
chriseth: Hm, I think it does too much.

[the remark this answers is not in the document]
chriseth: Yeah, it inlines columns that are treated as intermediate columns. I think I would prefer to leave it like this for now and then just replace this by a proper type checker analysis.

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,394 @@
...
-    definitions: &HashMap<String, (Symbol, Option<FunctionValueDefinition<T>>)>,
     expression: &Expression<T>,
-) -> Result<T, String> {
-    Evaluator {
-        definitions,
-        function_cache: &Default::default(),
-        variables: &[],
+    definitions: &HashMap<String, (Symbol, Option<FunctionValueDefinition<T>>)>,
+) -> Result<T, EvalError> {
+    internal::evaluate_to_number::<T, NoCustom>(expression, &[], &Definitions(definitions))
```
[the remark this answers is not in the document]
chriseth: I don't understand this change. Now I'm wondering if I should provide an `expect_number` utility function instead of having all these combinations.

executor/src/witgen/query_processor.rs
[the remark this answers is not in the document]
georgwiese: Yeah I guess it would be that or a panic...

chriseth:
Now depends on #786

pil_analyzer/src/evaluator.rs
```diff
@@ -1,104 +1,382 @@
...
+}
+
+pub fn evaluate<'a, T: FieldElement, C: Custom>(
+    expr: &'a Expression<T>,
+    symbols: &impl SymbolLookup<'a, T, C>,
+) -> Result<Value<'a, T, C>, EvalError> {
+    internal::evaluate(expr, &[], symbols)
+}
+
+// TODO this function should be removed in the future, or at least it should
```
chriseth, line 26: This is now fixed in https://github.com/powdr-labs/powdr/pull/776 as well.

georgwiese (approved):

PR #762, opened 2023-11-09
Regression: fixing instruction sc.w

lvella:
Execution result report was being skipped in case there was no reservation.


chriseth (approved):
Sorry, that was me.

leonardoalt (approved):

lvella:
There was an
```addr <=X= ... ```
being skipped, too. It was removed when address became an argument of `mstore`, but the jump count was not adjusted.

PR #764, opened 2023-11-13
Removing `tail` instruction.

lvella:
In our label based implementation, `tail` should actually be identical to jump.

### DIFF
--- a/riscv/src/compiler.rs
+++ b/riscv/src/compiler.rs
@@ -550,8 +550,6 @@ fn preamble(degree: u64, coprocessors: &CoProcessors) -> String {
     instr jump_dyn X { pc' = X }
     instr jump_and_link_dyn X { pc' = X, x1' = pc + 1 }
     instr call l: label { pc' = l, x1' = pc + 1 }
-    // TODO x6 actually stores some relative address, but only part of it.
-    instr tail l: label { pc' = l, x6' = l }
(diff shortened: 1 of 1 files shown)


riscv/src/compiler.rs
```diff
@@ -1184,10 +1182,14 @@ fn process_instruction(instr: &str, args: &[Argument], coprocessors: &CoProcesso
                 _ => None,
             };
             match (replacement, instr) {
+                (None, instr) => {
+                    let instr = if instr == "tail" { "jump" } else { instr };
```
chriseth, line 1186: Can you add a comment saying something like "tail is allowed to use x6 for computation, but does not have to" - is that correct?

lvella: I am not sure this is true. Strictly speaking, both `tail` and `call` should use it in a very specific way, although no sane program will rely on this behavior.

The thing is, in our current implementation, `call` doesn't do it, and `tail` does it wrong (only the highest bits of the address must be stored in `x6`, not the whole thing).

lvella: I've added a comment explaining it.

chriseth:
Looks good, can you squash?

lvella:
squashed

chriseth (approved):

PR #766, opened 2023-11-13
More context for logging.

chriseth:


executor/src/witgen/vm_processor.rs
```diff
@@ -607,13 +619,19 @@ impl<'a, T: FieldElement> VmProcessor<'a, T> {
...
-                .render("Current Row", true, &self.witnesses)
+            self.row(row_index).render(
+                &format!("Current row ({row_index})"),
+                true,
+                &self.witnesses
+            )
         );
         log::debug!(
             "{}",
-            self.row(row_index)
```
chriseth, line 615: The missing `+1` was a bug.

leonardoalt (approved):

PR #769, opened 2023-11-14
exp: remove anything non pil related

Maddiaa0:
In the long run we are thinking of just using the pil compiler part of powdr, and not using witgen, this is a little experimental pr so see how much of it we can remove to maintain functionality. 

A more favourable approach may be to create an additional binary that just does barretenberg pilgen, such that we do not diverge from powdr main too much, this I am still waying up; a pr implementing this will likely follow


Maddiaa0:
Whoops, meant to make this on my fork

PR #773, opened 2023-11-16
Allow function calls in condenser

chriseth:
Depends on #759 and #776

### DIFF
--- a/pil_analyzer/src/pil_analyzer.rs
+++ b/pil_analyzer/src/pil_analyzer.rs
@@ -834,7 +834,7 @@ namespace N(65536);
     }
 
     #[test]
-    #[should_panic = "Arrays cannot be used as a whole in this context"]
+    #[should_panic = "Operator - not supported on types"]
     fn no_direct_array_references() {
         let input = r#"namespace N(16);
     col witness y[3];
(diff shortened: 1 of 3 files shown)


chriseth:
Check if we actually still need the constants.

chriseth:
Removed "compute_constants".

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -865,4 +865,74 @@ namespace N(65536);
...
+
+    #[test]
+    fn fixed_concrete_and_symbolic() {
+        let input = r#"namespace N(16);
+    let last_row = 15;
+    let ISLAST = |i| match i { last_row => 1, _ => 0, };
+    let x;
+    let y;
+    y - ISLAST(3) = 0;
+    x - ISLAST = 0;
```
leonardoalt, line 925: It makes sense, but I find it confusing that ISLAST(3) evaluates at compile-time and ISLAST compiles to a runtime column, even though both look kind of the same syntactically.

leonardoalt (approved):

pil_analyzer/src/pil_analyzer.rs
chriseth: Once we have a distinction by types, we could change it so that you can never evaluate fixed columns in an algebraic expression context, but that could also create more problems than it solves.

Another option would be to require explicit syntax to turn a symbol into an actual column reference. But then it would look quite ugly:

```
ref(y) - ISLAST(3) = 0
ref(x) - ref(ISLAST) = 0
```

I think that in the end, people will just not notice this difference between ISLAST and ISLAST(3) and find it quite natural.

leonardoalt: yea maybe

PR #774, opened 2023-11-17
Allow functions to process constraints

chriseth:


leonardoalt (changes requested):
Whats the base field of the equations here? Or is it bigint arith?

chriseth:
This is the 256 bit arithmetic on goldilocks.

test_data/pil/arith_improved.pil
```diff
@@ -0,0 +1,1529 @@
...
+
+let N = 2**16;
+
+let BYTE2 = |i| i & 0xffff;
+
+namespace Arith(N);
+    let fold = |length, f, initial, folder| match length {
+        0 => initial,
+        _ => folder(fold(length - 1, f, initial, folder), f(length - 1))
+    };
```
leonardoalt, line 21: we're this close to Prolog

leonardoalt:
```prolog
fold(0, _, Initial, _, Initial).
fold(Length, Func, Initial, Folder, Result) :-
    Length > 0,
    NewLength is Length - 1,
    call(Func, Length, Value),
    call(Folder, Value, Initial, Folded),
    fold(NewLength, Func, Folded, Folder, Result).
```

leonardoalt:
build/test fails

chriseth:
The test actually fails because the fixed columns need something like 2**21 rows or so...

pil_analyzer/src/evaluator.rs
```diff
@@ -248,13 +248,17 @@ mod internal {
...
-                match (&left, &right) {
-                    (Value::Custom(_), _) | (_, Value::Custom(_)) => {
+                let mut left = evaluate(left, locals, symbols)?;
+                let mut right = evaluate(right, locals, symbols)?;
+                match (&mut left, op, &mut right) {
+                    (Value::Custom(_), _, _) | (_, _, Value::Custom(_)) => {
                         symbols.eval_binary_operation(left, *op, right)?
                     }
-                    (Value::Number(l), Value::Number(r)) => {
+                    (Value::Array(l), BinaryOperator::Add, Value::Array(r)) => {
```
chriseth, line 257: Any reason not to allow `+` to implement concatenation on arrays?

leonardoalt: no

PR #775, opened 2023-11-17
Remove degree as property of symbol.

chriseth:
The degree is provided with the namespaces and stored in the Analyzed struct.


lvella (approved):
👌

PR #780, opened 2023-11-21
Refactor out compiler function calls to compile() to reduce callstack.

inventednight:
This PR replaces some of the callstack to the `compile()` function directly to `compile()`. This might lead some of these functions to presently be redundant, so their usage in the CLI/tests might be worth removing and replacing.

Previously, for example:
compile_pil_or_asm -> compile_asm -> compile_asm_string -> compile_pil_ast -> compile

Is now:
compile_pil_or_asm -> compile

Each of the functions in between also now call compile() directly, so any usage of the intermediate functions have also had their callstack reduced.


compiler/src/lib.rs
```diff
@@ -144,6 +153,28 @@ pub fn compile_asm_string<T: FieldElement>(
...
+            output_dir,
+            inputs_to_query_callback(inputs),
+            prove_with,
+            external_witness_values,
+        )),
+    ))
+}
+
+// Returns the relative pil file name and PILFile, for passing to compile()
+fn get_pil_file<T: FieldElement>(
```
leonardoalt, line 172:
```suggestion
fn compile_asm_to_pil<T: FieldElement>(
```

leonardoalt:
Looks good! Just one small suggestion.

I think eventually (different PR) we should also split/rename things there and/or make a new crate with more granular interfaces. Currently when we call "compile" it not only compiles, but also runs the executor/witgen.

The single steps we should have are probably:

- compile asm to PIL
- compile PIL to Analyzed
- execute Analyzed

Then we could have end to end functions that compose these.

leonardoalt:
This has conflicts now after we merged the executor PR

leonardoalt:
Thanks! We ended up going with a different refactoring in https://github.com/powdr-labs/powdr/pull/826 , so I think we should close this one.

PR #782, opened 2023-11-22
Updating to latest riscv-executor branch.

lvella:


leonardoalt:
maybe both branches should rebase on main before this? looks like a massive diff

lvella:
Eventually, when either is about to merge, yes.

A quarta, 22/11/2023, 16:55, Leo ***@***.***> escreveu:

leonardoalt:
But what is this diff then? It has a lot of stuff from main. If both braches rebased on main this would be much smaller

lvella:
This PR is a merge from main + riscv-executor into Georg's continuation
branch, that is why it is big.

I prefer to rebase/squash/whatever just when we are about to merge into
main, just because it is much easier to handle merges than rebases.

A quarta, 22/11/2023, 22:56, Leo ***@***.***> escreveu:

PR #784, opened 2023-11-23
Implement if expression.

chriseth:
The `else` part is currently required. It does not make sense to skip the else part because we don't have statement lists as part of expressions yet.

### DIFF
--- a/pil_analyzer/src/pil_analyzer.rs
+++ b/pil_analyzer/src/pil_analyzer.rs
@@ -631,6 +631,17 @@ namespace N(65536);
     }
 
     #[test]
+    fn if_expr() {
+        let input = r#"namespace Assembly(2);
+    col fixed A = [0]*;
+    col fixed C(i) { if (i < 3) { Assembly.A(i) } else { (i + 9) } };
+    col fixed D(i) { if Assembly.C(i) { 3 } else { 2 } };
+"#;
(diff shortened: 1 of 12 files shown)


pil_analyzer/src/pil_analyzer.rs
```diff
@@ -890,6 +900,17 @@ namespace N(65536);
...
+        let formatted = process_pil_file_contents::<GoldilocksField>(input).to_string();
+        assert_eq!(formatted, input);
+    }
+
+    #[test]
+    fn if_expr() {
+        let input = r#"namespace Assembly(2);
+    col fixed A = [0]*;
+    col fixed C(i) { if (i < 3) { Assembly.A(i) } else { (i + 9) } };
+    col fixed D(i) { if Assembly.C(i) { 3 } else { 2 } };
```
leonardoalt, line 913: How does this work if C isn't boolean?

chriseth: We don't have a type checker yet, so it is defined in the same way as it is defined in C.

leonardoalt:
need to update the executor

leonardoalt:
> need to update the executor

still missing

chriseth:
Updated.

leonardoalt (dismissed):

chriseth:
Still depends on #756 to be merged.

PR #790, opened 2023-11-24
Support array elements in operations.

chriseth:


std/hash/poseidon_gl.asm
```diff
@@ -57,26 +51,28 @@ machine PoseidonGL(LASTBLOCK, operation_id) {
...
+    pol a2 = inp[2] + C_2;
+    pol a3 = inp[3] + C_3;
+    pol a4 = inp[4] + C_4;
+    pol a5 = inp[5] + C_5;
+    pol a6 = inp[6] + C_6;
+    pol a7 = inp[7] + C_7;
+    pol a8 = cap[0] + C_8;
+    pol a9 = cap[1] + C_9;
+    pol a10 = cap[2] + C_10;
+    pol a11 = cap[3] + C_11;
```
leonardoalt, line 75: I guess all these blocks here could also be transformed into folds?

leonardoalt (approved):

std/hash/poseidon_gl.asm
chriseth: The ones below, yes, but I'm not sure about intermediate polys.

PR #796, opened 2023-11-27
Regs to memory

chriseth:


chriseth:
This seems to result in 82 witness columns.

leonardoalt:
> This seems to result in 82 witness columns instead of 130.

in what example?

Currently this is what I get when running powdr-riscv on revm:

`This seems to result in 82 witness columns instead of 130.`

leonardoalt:
@chriseth how far was this from ready?

leonardoalt:
Closing in favor of the new approach.

PR #800, opened 2023-11-28
Use sccache.

chriseth:


chriseth:
If this works well, we could ditch the "attach workspace" step. It takes really long.

leonardoalt:
do we close this then?

PR #801, opened 2023-11-28
riscv mulh

leonardoalt:


riscv_executor/src/lib.rs
```diff
@@ -554,6 +554,13 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
                 vec![r.into()]
             }
             "fail" => {
+                let x3 = self.proc.get_reg("x3").0;
```
chriseth, line 557: Is this leftover?

leonardoalt: ah yea was debugging

chriseth (approved):

PR #802, opened 2023-11-28
More linking table entries

leonardoalt:
These were needed by `revme` which runs the Ethereum tests on revm.

### DIFF
--- a/riscv/src/compiler.rs
+++ b/riscv/src/compiler.rs
@@ -686,55 +686,36 @@ fn preamble(degree: u64, coprocessors: &CoProcessors) -> String {
 }
 
 fn runtime(coprocessors: &CoProcessors) -> String {
-    r#"
-.globl __udivdi3@plt
-.globl __udivdi3
-.set __udivdi3@plt, __udivdi3
-
-.globl __umoddi3@plt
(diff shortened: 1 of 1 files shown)


leonardoalt:
@chriseth @lvella please check

riscv/src/compiler.rs
```diff
@@ -690,10 +690,46 @@ fn runtime(coprocessors: &CoProcessors) -> String {
...
 
+.globl __udivti3@plt
+.globl __udivti3
+.set __udivti3@plt, __udivti3
+
+.globl __divdf3@plt
+.globl __divdf3
+.set __divdf3@plt, __divdf3
+
+.globl __muldf3@plt
```
chriseth, line 701: Can you use something like
`["udivdi3", "__udivti3", "__divdf3", ...].map(|n| format!(".globl {n}@plt\n.globl {n}\n.set {n}@plt, {n}\n")).join("\n\n")`?

chriseth:
These seem to be software emulation of floating point operations...

leonardoalt:
> These seem to be software emulation of floating point operations...

yea, it's good to know these are supported

lvella (approved):

PR #803, opened 2023-11-28
Remove macros.

chriseth:


parser/src/lib.rs
```diff
@@ -201,22 +169,14 @@ mod test {
...
+pol commit arr[8];
 pol commit x, y;
-macro constrain_equal_expr(A, B) { (A - B) };
-macro force_equal_on_last_row(poly, value) { (ISLAST * constrain_equal_expr(poly, value)) = 0; };
-force_equal_on_last_row(x', 1);
-force_equal_on_last_row(y', 1);
-macro on_regular_row(cond) { ((1 - ISLAST) * cond) = 0; };
-on_regular_row(constrain_equal_expr(x', y));
-on_regular_row(constrain_equal_expr(y', (x + y)));
+{ (x + 2), y' } in { ISLAST, 7 };
```
chriseth, line 177: I realized that this test is for printing parsed PIL, but it actually never printed a lookup or even constraint, so I added some.

leonardoalt (approved):

PR #811, opened 2023-11-29
Panic on parse errors instead of turning them into illegible escaped …

chriseth:
…strings.

### DIFF
--- a/importer/src/powdr_std.rs
+++ b/importer/src/powdr_std.rs
@@ -40,8 +40,12 @@ fn load_std<T: FieldElement>() -> ASMModule<T> {
         }
         Ok(std_source) => {
             let std_content =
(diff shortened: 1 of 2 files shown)


importer/src/powdr_std.rs
```diff
@@ -40,8 +40,12 @@ fn load_std<T: FieldElement>() -> ASMModule<T> {
         }
         Ok(std_source) => {
             let std_content =
-                parse_asm::<T>(Some(std_path.as_path().to_str().unwrap()), &std_source).unwrap();
-
+                parse_asm::<T>(Some(std_path.as_path().to_str().unwrap()), &std_source)
```
chriseth, line 43: I have no idea why this calls `parse_asm` directly instead of calling the module loader.

leonardoalt (approved):

PR #815, opened 2023-11-30
New data input coprocessors

leonardoalt:
This PR is the beginning of the data passing refactoring.

Being able to pass only a word at a time using `get_prover_input` is not ideal. This PR adds two new coprocessors, `get_data_len` and `get_data` that make it easier to get full slices of data from the prover.

This is also not super elegant and not ideal, but it already allows for better UX in some cases.

### DIFF
--- a/powdr_cli/src/main.rs
+++ b/powdr_cli/src/main.rs
@@ -426,6 +426,7 @@ fn run_command(command: Commands) {
             continuations,
         } => match (just_execute, continuations) {
             (true, true) => {
+                assert!(matches!(field, FieldArgument::Gl));
                 let contents = fs::read_to_string(&file).unwrap();
                 let inputs = split_inputs::<GoldilocksField>(&inputs);
                 rust_continuations(file.as_str(), contents.as_str(), inputs);
@@ -433,6 +434,10 @@ fn run_command(command: Commands) {
             (true, false) => {
                 let contents = fs::read_to_string(&file).unwrap();
                 let inputs = split_inputs::<GoldilocksField>(&inputs);
+                let inputs: HashMap<GoldilocksField, Vec<GoldilocksField>> =
+                    vec![(GoldilocksField::from(0), inputs)]
+                        .into_iter()
+                        .collect();
                 riscv_executor::execute::<GoldilocksField>(&contents, &inputs, &default_input());
             }
             (false, true) => {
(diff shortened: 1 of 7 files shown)


powdr_cli/src/main.rs
```diff
@@ -431,7 +431,14 @@ fn run_command(command: Commands) {
             (true, false) => {
                 let contents = fs::read_to_string(&file).unwrap();
                 let inputs = split_inputs::<GoldilocksField>(&inputs);
-                riscv_executor::execute::<GoldilocksField>(&contents, &inputs, &default_input());
+                let mut inputs_hash: HashMap<GoldilocksField, Vec<GoldilocksField>> =
+                    HashMap::default();
+                inputs_hash.insert(0u32.into(), inputs);
```
chriseth, line 436: I think it would be clearer to use `mapping` or something instead of `hash`, or just `inputs`.

powdr_cli/src/main.rs
```diff
@@ -624,10 +633,18 @@ fn rust_continuations<F: FieldElement>(file_name: &str, contents: &str, inputs:
     let program =
         compiler::compile_asm_string_to_analyzed_ast::<F>(file_name, contents, None).unwrap();
 
+    let mut inputs_hash: HashMap<F, Vec<F>> = HashMap::default();
+    inputs_hash.insert(0u32.into(), inputs);
```
chriseth, line 637:
```suggestion
    let inputs_hash: HashMap<F, Vec<F>> = [(0, inputs)].into();
```

riscv/runtime/src/coprocessors.rs
```diff
@@ -5,6 +5,19 @@
 // during the reachability analysis.
 extern "C" {
     fn poseidon_gl_coprocessor(data: *mut [u64; 12]);
+    fn input_coprocessor(index: u32, what: u32) -> u32;
```
chriseth, line 8: Please update the comment above.

riscv/src/coprocessors.rs
```diff
@@ -8,68 +8,85 @@ type RuntimeFunctionImpl = (&'static str, fn() -> String);
 #[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq)]
 struct CoProcessor {
     name: &'static str,
-    ty: &'static str,
-    import: &'static str,
-    instructions: &'static str,
+    ty: Option<&'static str>,
```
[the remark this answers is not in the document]
chriseth: Shouldn't this rather be an enum splitting into runtime function impl and external machine?

riscv/tests/riscv_data/evm/src/lib.rs
```diff
@@ -21,6 +21,11 @@ fn main() {
 
     let bytecode_len = get_prover_input(0);
     let bytecode: Vec<_> = (1..(bytecode_len + 1)).map(|idx| get_prover_input(idx) as u8).collect();
+    /*
+    let bytecode_len = get_data_len(0);
```
chriseth, line 25: todo

chriseth (approved):

PR #816, opened 2023-11-30
Remove some unneeded (and misleading) derived implementations.

chriseth:
I think just blindly deriving Clone, Eq and Ord is very misleading. IMO it is totally OK for basic data types but the more complicated the data gets, it gets less and less clear what it acutally means. If I have two local variables with the same ID and different names, are they the same or not? Are they even the same if they have the same local ID and local name, but used in different contexts?

Also providing Clone for a gigantic data structure like a whole ASM machine is not really useful. If there is a clone, then it should be reasonably expensive, or at least needed, I think.

### DIFF
--- a/ast/src/asm_analysis/mod.rs
+++ b/ast/src/asm_analysis/mod.rs
@@ -23,7 +23,7 @@ use crate::parsed::{
 
 pub use crate::parsed::Expression;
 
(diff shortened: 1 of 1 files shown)


inventednight:
LGTM, but is there a way for us to turn on warnings for unused derived impls?

leonardoalt (approved):

PR #821, opened 2023-11-30
Pil in asm at global level

chriseth:
This now is able to handle:
```
let x = 12;

mod M {
    use super::x as r;
    
    let y = r + 2;
}

mod R {
    use super::x;
    use super::M::y;

    machine FullConstant {
        degree 2;

        let C = |i| match i % 2 {
            0 => x,
            1 => y,
        };
        col commit w;
        w = C;
    }
}
```

and translates this through all the stages of the assembly handling, object, pilgraph, linker, whatever ... to:
```
namespace M(2);
let y = (x + 2);
namespace Global(2);
let x = 12;
namespace main(2);
let C = |i| match (i % 2) { 0 => x, 1 => M::y, };
pol commit w;
w = C;
```
which is then properly parsed and executed!

TODO:
 - [x] clean up forward references resolving
 - [x] write documentation
 - [x] implement let statements as local variables (maybe in another PR)

### DIFF
--- a/ast/src/parsed/asm.rs
+++ b/ast/src/parsed/asm.rs
@@ -1,6 +1,6 @@
 use std::{
     fmt::{Display, Formatter, Result},
-    iter::{once, repeat},
+    iter::{empty, once, repeat},
 };
 
 use itertools::Itertools;
@@ -346,6 +346,22 @@ impl Display for Part {
 pub struct Machine<T> {
     pub arguments: MachineArguments,
     pub statements: Vec<MachineStatement<T>>,
+}
+
+impl<T: Clone> Machine<T> {
+    /// Returns a vector of all local variables / names defined in the machine.
+    pub fn local_names(&self) -> Box<dyn Iterator<Item = &String> + '_> {
+        Box::new(self.statements.iter().flat_map(|s| match s {
+            MachineStatement::RegisterDeclaration(_, name, _) => Box::new(once(name)),
+            MachineStatement::Pil(_, statement) => statement.symbol_definition_names(),
+            MachineStatement::Degree(_, _)
+            | MachineStatement::Submachine(_, _, _)
+            | MachineStatement::InstructionDeclaration(_, _, _)
+            | MachineStatement::LinkDeclaration(_)
+            | MachineStatement::FunctionDeclaration(_, _, _, _)
+            | MachineStatement::OperationDeclaration(_, _, _, _) => Box::new(empty()),
+        }))
+    }
 }
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Default, Clone)]
--- a/ast/src/parsed/asm.rs
+++ b/ast/src/parsed/asm.rs
@@ -47,6 +47,8 @@ pub enum SymbolValue<T> {
     Import(Import),
     /// A module definition
     Module(Module<T>),
+    /// A generic symbol / function.
+    Expression(Expression<T>),
 }
 
 impl<T> SymbolValue<T> {
@@ -55,6 +57,7 @@ impl<T> SymbolValue<T> {
             SymbolValue::Machine(machine) => SymbolValueRef::Machine(machine),
             SymbolValue::Import(i) => SymbolValueRef::Import(i),
             SymbolValue::Module(m) => SymbolValueRef::Module(m.as_ref()),
+            SymbolValue::Expression(e) => SymbolValueRef::Expression(e),
         }
     }
 }
@@ -67,6 +70,8 @@ pub enum SymbolValueRef<'a, T> {
     Import(&'a Import),
     /// A module definition
     Module(ModuleRef<'a, T>),
+    /// A generic symbol / function.
+    Expression(&'a Expression<T>),
 }
 
 #[derive(Debug, PartialEq, Eq, From)]
--- a/ast/src/parsed/folder.rs
+++ b/ast/src/parsed/folder.rs
@@ -25,6 +25,11 @@ pub trait Folder<T> {
                     SymbolValue::Machine(machine) => self.fold_machine(machine).map(From::from),
                     SymbolValue::Import(import) => self.fold_import(import).map(From::from),
                     SymbolValue::Module(module) => self.fold_module(module).map(From::from),
+                    SymbolValue::Expression(e) => {
+                        // Not folding expressions by default because the ExpressionFolder
+                        // is a different trait.
+                        Ok(SymbolValue::Expression(e))
+                    }
                 }
                 .map(|value| ModuleStatement::SymbolDefinition(SymbolDefinition { value, ..d })),
             })
--- a/ast/src/parsed/mod.rs
+++ b/ast/src/parsed/mod.rs
@@ -5,7 +5,10 @@ pub mod folder;
 pub mod utils;
 pub mod visitor;
 
-use std::ops;
+use std::{
+    iter::{empty, once},
+    ops,
+};
 
 use number::{DegreeType, FieldElement};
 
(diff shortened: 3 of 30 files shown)


chriseth:
And yes, we should have `mod{...}` in pil, but I'm wondering if we shouldn't just use the asm module parser for both.

importer/src/path_canonicalizer.rs
```diff
@@ -205,6 +245,8 @@ fn check_path_internal<'a, T>(
                     SymbolValueRef::Import(p) => {
                         location.pop().unwrap();
 
+                        // TODO we do a recursive call here, but shouldn't this actually
```
chriseth, line 248: To be handled at a later point. This concerns existing code.

test_data/asm/pil_at_module_level.asm
```diff
@@ -0,0 +1,43 @@
...
+    machine FullConstant {
+        degree 2;
+
+        let C = |i| match i % 2 {
+            0 => x,
+            1 => y,
+        };
+        col commit w[2];
+
+        // This and the next line is the same.
```
leonardoalt, line 35:
```suggestion
        // This and the next line are the same.
```

test_data/asm/pil_at_module_level.asm
```diff
@@ -0,0 +1,43 @@
...
+        super::utils::sum(2, |i| w[i]) == 8;
+        sum(2, |i| w[i]) == 8;
+
+        // Try to see if we only clear local variables
+        // if they not already exist in the outer scope.
```
leonardoalt, line 40:
```suggestion
        // if they don't already exist in the outer scope.
```

leonardoalt: wdym by "clear" here?

ast/src/parsed/asm.rs
```diff
@@ -343,6 +348,22 @@ pub struct Machine<T> {
     pub statements: Vec<MachineStatement<T>>,
 }
 
+impl<T: Clone> Machine<T> {
+    /// Returns a vector of all local variables / names defined in the machine.
+    pub fn local_names(&self) -> Box<dyn Iterator<Item = &String> + '_> {
```
leonardoalt, line 353: why not just `Vec<String>`?

ast/src/parsed/folder.rs
```diff
@@ -25,6 +25,10 @@ pub trait Folder<T> {
                     SymbolValue::Machine(machine) => self.fold_machine(machine).map(From::from),
                     SymbolValue::Import(import) => self.fold_import(import).map(From::from),
                     SymbolValue::Module(module) => self.fold_module(module).map(From::from),
+                    SymbolValue::Expression(e) => {
+                        // Not folding expressions by default
```
leonardoalt, line 29: why?

book/src/asm/declarations.md
```diff
@@ -0,0 +1,13 @@
+# Declarations
+
+Symbols can be defined via ``let <name> = <value>``. The value is a generic [PIL-expression](../pil/expressions.md)
```
leonardoalt, line 3:
```suggestion
Symbols can be defined via ``let <name> = <value>;``. The value is a generic [PIL-expression](../pil/expressions.md)
```

book/src/asm/declarations.md
```diff
@@ -0,0 +1,13 @@
...
+and its type is unconstrained (it can be a value, a function or even a higher-order function).
+
+Other symbols available in the current module can be accessed by name, but it is also possible to specify
+full relative paths in the form of e.g. ``super::super::module_name::symbol``.
```
leonardoalt, line 7: double super?

pil_analyzer/src/condenser.rs
```diff
@@ -103,6 +103,7 @@ impl<T: FieldElement> Condenser<T> {
         let (poly, _) = self
             .symbols
             .get(&reference.name)
+            .or_else(|| self.symbols.get(&format!("Global.{}", reference.name)))
```
leonardoalt, line 106: Why is `Global.` hardcoded here?

ast/src/parsed/mod.rs
```diff
@@ -52,6 +55,89 @@ pub enum PilStatement<T> {
     Expression(usize, Expression<T>),
 }
 
+impl<T> PilStatement<T> {
+    /// If the statement is a symbol definition, returns all (local) names of defined symbols.
+    pub fn symbol_definition_names(&self) -> Box<dyn Iterator<Item = &String> + '_> {
```
leonardoalt, line 60: Why not `Vec<String>`?

ast/src/parsed/mod.rs
```diff
@@ -52,6 +55,89 @@ pub enum PilStatement<T> {
...
+            | PilStatement::PolynomialIdentity(_, _)
+            | PilStatement::PlookupIdentity(_, _, _)
+            | PilStatement::PermutationIdentity(_, _, _)
+            | PilStatement::ConnectIdentity(_, _, _)
+            | PilStatement::Expression(_, _) => Box::new(empty()),
+        }
+    }
+
+    /// Returns an iterator over all (top-level) expressions in this statement.
+    pub fn expressions(&self) -> Box<dyn Iterator<Item = &Expression<T>> + '_> {
```
leonardoalt, line 83: It feels like this is gonna bite as back at some point, but I don't have a suggestion of how to simplify this.

ast/src/parsed/mod.rs
```diff
@@ -52,6 +55,89 @@ pub enum PilStatement<T> {
...
+            | PilStatement::PolynomialCommitDeclaration(_, _, Some(fundef)) => fundef.expressions(),
+            PilStatement::PolynomialCommitDeclaration(_, _, None)
+            | PilStatement::Include(_, _)
+            | PilStatement::PolynomialConstantDeclaration(_, _)
+            | PilStatement::LetStatement(_, _, None) => Box::new(empty()),
+        }
+    }
+
+    /// Returns an iterator over all (top-level) expressions in this statement.
+    pub fn expressions_mut(&mut self) -> Box<dyn Iterator<Item = &mut Expression<T>> + '_> {
```
leonardoalt, line 111: Same here as above.

ast/src/parsed/mod.rs
```diff
@@ -276,6 +374,24 @@ pub enum FunctionDefinition<T> {
     Expression(Expression<T>),
 }
 
+impl<T> FunctionDefinition<T> {
+    /// Returns an iterator over all (top-level) expressions.
+    pub fn expressions(&self) -> Box<dyn Iterator<Item = &Expression<T>> + '_> {
```
leonardoalt, line 379: Same here and below.

importer/src/path_canonicalizer.rs
leonardoalt: Please open an issue

importer/src/path_canonicalizer.rs
```diff
@@ -302,20 +342,113 @@ fn check_machine<T: Clone>(
...
+/// # Errors
+///
+/// This function will return an error if any of the paths does not resolve to anything
+fn check_expression<T: Clone>(
+    location: &AbsoluteSymbolPath,
+    e: &Expression<T>,
+    state: &mut State<'_, T>,
+    local_variables: &HashSet<String>,
+) -> Result<(), String> {
+    // We cannot use the visitor here because we need to change the local variables
```
leonardoalt, line 381: why doesn't the mut visitor work for that?

linker/src/lib.rs
```diff
@@ -27,113 +29,156 @@ pub fn link<T: FieldElement>(graph: PILGraph<T>) -> Result<PILFile<T>, Vec<Strin
...
-                }
+        .sorted_by_cached_key(|(namespace, _)| {
+            let mut namespace = namespace.clone();
+            let name = namespace.pop();
+            // This is needed to group into namespaces first.
+            (namespace, name)
+        })
+        .flat_map(|(mut namespace, e)| {
+            let name = namespace.pop().unwrap();
+            let def = PilStatement::LetStatement(0, name.to_string(), Some(e));
```
leonardoalt, line 45: what's this 0?

linker/src/lib.rs
```diff
@@ -27,113 +29,156 @@ pub fn link<T: FieldElement>(graph: PILGraph<T>) -> Result<PILFile<T>, Vec<Strin
...
+
+            if current_namespace != Some(namespace.clone())
+                && !(current_namespace.is_none() && namespace.is_empty())
+            {
+                current_namespace = Some(namespace.clone());
+                vec![
+                    PilStatement::Namespace(
+                        0,
```
leonardoalt, line 53: what's this 0?

linker/src/lib.rs
```diff
@@ -27,113 +29,156 @@ pub fn link<T: FieldElement>(graph: PILGraph<T>) -> Result<PILFile<T>, Vec<Strin
...
+                        namespace.relative_to(&AbsoluteSymbolPath::default()),
+                        Expression::Number(T::from(main_degree)),
+                    ),
+                    def,
+                ]
+            } else {
+                vec![def]
+            }
+        })
```
leonardoalt, line 62: This whole block is a bit confusing, can you elaborate more?

ast/src/parsed/asm.rs
chriseth: I had `Vec<&String>` before. Mainly wanted to do it this way, because it is more rusty and also wanted to check if it will be complicated. Also it saves heap allocations, but not sure if this is a strong argument.

ast/src/parsed/folder.rs
chriseth: Because the expression folder trait is a different trait. So if you want to change the behaviour on expressions, you have to implement the other trait and then just call that other trait from here.

ast/src/parsed/mod.rs
chriseth: Same as above.

ast/src/parsed/mod.rs
chriseth: I actually think this will save us.

chriseth: Oh and by the way: if we implement lookups as operators in the expression language, most of this will go away.

book/src/asm/declarations.md
chriseth: On purpose, but I can also change it.

importer/src/path_canonicalizer.rs
chriseth: Actually I think this is indeed fine.

importer/src/path_canonicalizer.rs
chriseth: Because we need a push/pop behavior:

When entering something that has local variables, we need to change them and then change them back when leaving that object. The mut iterator only calls once for each element, but does not care about how it is nested.

linker/src/lib.rs
chriseth: source location. At some point we really need to fix up our error messages and have to pull through all the source information properly. It is zero here because none of the asm stuff has source locations.

pil_analyzer/src/condenser.rs
chriseth: Good point! I think this can be removed.

test_data/asm/pil_at_module_level.asm
chriseth: `i` needs to be added to the list of local variables when entering the body of `|i| ...`. It needs to be cleared from that list again when we exit. There is a special situation where `i` is also a local variable of the outer function. In that case we of course do not clear.

linker/src/lib.rs
chriseth: Actually I was able to simplify the code.

test_data/asm/secondary_block_machine_add2.asm
```diff
@@ -18,8 +18,9 @@ machine Main {
     // The input column needs to be constant for the entire block
     col witness add_two_input;
 
+    col fixed _first_step = [1] + [0]*;
     // Because constraints are not cyclic, we need to explicitly constrain the first state
-    first_step * (add_two_state - add_two_input) = 0;
```
chriseth, line 22: The column `first_step` is introduced by the vm-to-constrained transformation. We should find a way that it does not clash with existing columns (and also combine equal columns with an optimizer).

chriseth: Another way would be to provide `first_step` column in the std library.

book/src/asm/declarations.md
leonardoalt: no that's fine, was just wondering

ast/src/parsed/asm.rs
leonardoalt: ok

importer/src/path_canonicalizer.rs
leonardoalt: right

linker/src/lib.rs
leonardoalt: yea

leonardoalt (approved):

leonardoalt (approved):

PR #851, opened 2023-12-18
fix evm test inputs

leonardoalt:
Fixing the EVM test inputs and trying it out on our CI again. If CI fails because of OOM I'll comment the function out but will leave the updated inputs and `verify_riscv_crate` adjustment.

### DIFF
--- a/riscv/tests/riscv.rs
+++ b/riscv/tests/riscv.rs
@@ -127,6 +127,7 @@ return(0, 32)
 static BYTECODE: &str = "61029a60005260206000f3";
 
 #[cfg(feature = "complex-tests")]
+#[ignore = "Too slow"]
 #[test]
 fn test_evm() {
     let case = "evm";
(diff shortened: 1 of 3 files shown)


riscv/tests/riscv.rs
```diff
@@ -124,21 +124,21 @@ fn test_password() {
 mstore(0, 666)
 return(0, 32)
 */
-/*
 static BYTECODE: &str = "61029a60005260206000f3";
```
chriseth, line 127: This could use PUSH0 ;)

chriseth (dismissed):

leonardoalt:
updated this PR to run the evm test with the slow tests, let's see

leonardoalt:
yea this is still too much for our CI...

leonardoalt:
I commented out the test again because of CI, but at least the test is correct now if you uncomment and run it

leonardoalt:
the test is at least compiled now, hidden behind a feature.

Ready for review

chriseth:
looks good, but tests are failing

lvella (approved):

PR #852, opened 2023-12-18
Use spaces for indentation.

chriseth:
And do not indent empty lines.


ast/src/lib.rs
```diff
@@ -80,7 +80,10 @@ pub fn evaluate_unary_operation<T: FieldElement>(op: UnaryOperator, v: T) -> T {
 fn indent<S: ToString>(s: S, indentation: usize) -> String {
     s.to_string()
         .split('\n')
-        .map(|line| format!("{}{line}", "\t".repeat(indentation)))
+        .map(|line| match line {
+            "" => "".to_string(),
```
chriseth, line 84: Without this, we add spaces to the end of inputs that end in `\n`, which is usually not what you want.

leonardoalt (approved):

PR #853, opened 2023-12-18
Run importer and resolver in tests.

chriseth:


georgwiese (approved):
LGTM.

Although I do wonder if we should just use the `Pipeline` everywhere. It seems like we still have big callstacks in the tests (e.g. `analysis::vm::test_utils::asm_to_pil_str` -> `batch_str` -> `infer_str` -> `typecheck_str` -> `resolve_str`-> `parse_asm` -> `parse_module`). Can we just use the pipeline instead?

I see two disadvantages:
- Crates would have to depend on `compiler` in dev dependencies (but I think that's fine?)
- We might have to break up the pipeline in more steps. E.g., it currently calls `convert_analyzed_to_pil_constraints` which calls a bunch of pipeline steps. So that's a problem if you only want to run half of them in the test to test a specific step.

But maybe it's worth it?

chriseth:
I don't think that crates _can_ depend on compiler. Wouldn't this be a circular dependency? Maybe it's fine for the tests.

PR #854, opened 2023-12-19
Fix public parsing and labels.

chriseth:


analysis/README.md
```diff
@@ -234,7 +234,7 @@ The diff for our example program is as follows:
 -       instr nothing = sub.nothing;
 -
 -       function main {
--               start::
+-               start:
 -               A <=Y= one();
 -               // END BATCH Unimplemented
```
chriseth, line 239: The `-` and the red color here is from the syntax hilighter of the file, it's not part of the commit's diff.

PR #858, opened 2023-12-22
Proper name resolution in PIL.

chriseth:
Changes references in PIL to use SymbolPath instead of NamspacedPolynomialReference, i.e. we can use `super` and arbitrarily nested namespaces. It also changes `namespace` to allow a `SymbolPath`. Because namespaces currently cannot be nested, you need to use

```
namespace A::B::C(N);
```

if you want nested namespaces.

It also changes `PILAnalyzer` to work in three stages instead of doing it all at once:
1. import resolution and parsing dependent files
2. register all declarations / names
3. resolve names.

And finally, the name resolution in PILAnalyzer does not blindly add the current namespace, but instead, it tries to find the reference name relative to the current namespace but if that fails, it descends step by step to the root to discover it.

This is preparation for combining PIL and asm (modules and imports).

### DIFF
--- a/asm_to_pil/src/vm_to_constrained.rs
+++ b/asm_to_pil/src/vm_to_constrained.rs
@@ -12,7 +12,7 @@ use ast::{
         LinkDefinitionStatement, Machine, RegisterDeclarationStatement, RegisterTy, Rom,
     },
     parsed::{
-        asm::InstructionBody,
+        asm::{InstructionBody, Part},
         build::{direct_reference, next_reference},
         folder::ExpressionFolder,
         visitor::ExpressionVisitable,
@@ -381,8 +381,11 @@ impl<T: FieldElement> ASMPILConverter<T> {
                 body.iter_mut().for_each(|s| {
                     s.post_visit_expressions_mut(&mut |e| {
                         if let Expression::Reference(r) = e {
-                            if let Some(sub) = substitutions.get(&r.name) {
-                                r.name = sub.to_string();
+                            if let Some(name) = r.try_to_identifier() {
+                                if let Some(sub) = substitutions.get(name) {
+                                    *r.path.parts.last_mut().unwrap() =
+                                        Part::Named(sub.to_string());
+                                }
                             }
                         }
                     });
@@ -466,16 +469,13 @@ impl<T: FieldElement> ASMPILConverter<T> {
         function: Expression<T>,
         mut args: Vec<Expression<T>>,
     ) -> CodeLine<T> {
-        let Expression::Reference(NamespacedPolynomialReference {
-            namespace: _,
-            name: instr_name,
-        }) = function
-        else {
+        let Expression::Reference(reference) = function else {
             panic!("Expected instruction name");
         };
+        let instr_name = reference.try_to_identifier().unwrap();
         let instr = &self
             .instructions
-            .get(&instr_name)
+            .get(instr_name)
             .unwrap_or_else(|| panic!("Instruction not found: {instr_name}"));
         let output = instr.outputs.clone();
 
@@ -487,7 +487,7 @@ impl<T: FieldElement> ASMPILConverter<T> {
         }
 
         args.extend(lhs_with_regs.iter().map(|(lhs, _)| direct_reference(lhs)));
-        self.handle_instruction(instr_name, args)
+        self.handle_instruction(instr_name.clone(), args)
     }
 
     fn handle_instruction(&mut self, instr_name: String, args: Vec<Expression<T>>) -> CodeLine<T> {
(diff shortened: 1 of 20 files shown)


ast/src/parsed/mod.rs
```diff
@@ -158,11 +160,31 @@ pub struct PolynomialName<T> {
...
 #[derive(Debug, PartialEq, Eq, Default, Clone, PartialOrd, Ord)]
 /// A polynomial with an optional namespace
+/// This is different from SymbolPath mainly due to different formatting.
 pub struct NamespacedPolynomialReference {
-    /// The optional namespace, if `None` then this polynomial inherits the next enclosing namespace, if any
-    pub namespace: Option<String>,
-    /// The name of the polynomial / symbol.
-    pub name: String,
+    pub path: SymbolPath,
+}
```
leonardoalt, line 166: Would it make sense to destruct the path into whatever this struct needs?

ast/src/parsed/asm.rs
```diff
@@ -109,6 +110,13 @@ impl SymbolPath {
         self.parts.extend(other.into().parts);
         self
     }
+
+    /// Formats the path and uses `.` as separator if
+    /// there are at most two components.
+    pub fn to_dotted_string(&self) -> String {
+        let separator = if self.parts.len() <= 2 { "." } else { "::" };
```
leonardoalt, line 117: This feels a bit confusing. Maybe we should keep `::` for namespacing and static access, and `.` for instance access? (if any of that makes sense)

ast/src/parsed/asm.rs
```diff
@@ -203,6 +224,13 @@ impl AbsoluteSymbolPath {
         parts.push(part.to_string());
         Self { parts }
     }
+
+    /// Formats the path without leading `::` and uses `.` as separator if
+    /// there are at most two components.
+    pub fn to_dotted_string(&self) -> String {
```
leonardoalt, line 230: This looks the same as the other `to_dotted_string` impl above?

asm_to_pil/src/vm_to_constrained.rs
```diff
@@ -585,10 +585,8 @@ impl<T: FieldElement> ASMPILConverter<T> {
             Expression::FunctionCall(_) => panic!(),
             Expression::Reference(reference) => {
                 // TODO check it actually is a register
-                vec![(
-                    1.into(),
-                    AffineExpressionComponent::Register(reference.name),
-                )]
+                let name = reference.try_to_identifier().unwrap();
+                vec![(1.into(), AffineExpressionComponent::Register(name.clone()))]
```
leonardoalt, line 589: I know it's unrelated to this PR, but do you know what this hardcoded 1 is?

pil_analyzer/src/statement_processor.rs
```diff
@@ -90,6 +90,31 @@ where
         }
     }
 
+    /// Returns all names of the symbols defined inside the given statemnet.
```
leonardoalt, line 93:
```suggestion
    /// Returns all names of the symbols defined inside the given statement.
```

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -181,31 +243,34 @@ struct Driver<'a, T>(&'a PILAnalyzer<T>);
 
 impl<'a, T: FieldElement> AnalysisDriver<T> for Driver<'a, T> {
     fn resolve_decl(&self, name: &str) -> String {
-        if name.starts_with('%') {
+        (if name.starts_with('%') {
```
leonardoalt, line 246: 😭😭😭😭

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -40,60 +50,97 @@ struct PILAnalyzer<T> {
     /// appear in the source.
     source_order: Vec<StatementIdentifier>,
     symbol_counters: Option<Counters>,
-    included_files: HashSet<PathBuf>,
-    line_starts: Vec<usize>,
     current_file: PathBuf,
+    current_line_starts: Vec<usize>,
```
leonardoalt, line 54: what's "line starts"?

chriseth: This is a lookup data structure to help convert file positions to line-column-pairs.

asm_to_pil/src/vm_to_constrained.rs
chriseth: Yes, it's the factor one. The return type is to be interpreted roughly as `v.map(|(f, x)| f * x).sum()`. Maybe we could create a custom data structure with operators at some point.

ast/src/parsed/asm.rs
chriseth: I think currently we don't have any instances in pil. When we add structs, then we should use `.` for field access and `::` for namespaces, yes.

ast/src/parsed/asm.rs
chriseth: It is, but this is for AbsoluteSymbolPath and the other is for SymbolPath.

ast/src/parsed/mod.rs
chriseth: We could just replace all occurrences of NamespacedPolynomialReference by SymbolPath, but that would have created so many changes that I found it easier like this. Also it's a bit of a conceptual difference as well.

leonardoalt:
tests are failing

chriseth:
I think I fixed it...

leonardoalt (approved):

PR #859, opened 2023-12-24
[DO NOT MERGE] New powdr crate

leonardoalt:
This PR

- renames the `compiler` crate to `pipeline`
- creates a new crate `powdr` which re-exports some internal crates. Not sure if it's better to do it how it's done now, just `pub use <crate>;` or to re-export specific types/traits. For now I only re-exported the internal crates needed to run the tests in https://github.com/powdr-labs/powdr_revm.

Open question:

Should we move the tests from `pipeline` into `powdr`? It kinda feels like it makes sense.


leonardoalt:
I'll try to get the Ethereum tests working end-to-end in https://github.com/powdr-labs/powdr_revm, so we know which specific items need to be re-exported in the powdr crate, so we don't have to re-export every internal crate.

leonardoalt:
doing this in a more organized way in a series of PRs

PR #862, opened 2023-12-28
[WIP] Upgrade backend estark

eigmax:
Upgrade backend estark from eigen-zkvm for:
* Some parallelizations for some big loops by Rayon.
* AVX2 and AVX512 Poseidon hash and this brings 50% time saving for MerkleHash: RUSTFLAGS="-C target-feature=+avx2" cargo test --release. Notice that the avx512 can not be enabled due to it imports features `stdsimd`, which is not on the stable release channel. 
* RAM saving: drop constant and commit polynomials in time


eigmax:
Regarding the eStark backend, here we have something to discuss:

1. `StarkSetup`:  for any circuits, the `StarkSetup` need do once only;
2.  Proof recursion: generate the Circom circuit and its inputs of StarkVerifier, then dump them into files. eStark can then load the dumped circuit and input file to generate the recursive proof and aggregate the recursive proof.  I have s sample of this at https://github.com/eigmax/zkml-rust/blob/main/powdr.diff#L20. 

For above features,  @leonardoalt can you leave us some suggestions about how to implement it in powdr?

eigmax:
@leonardoalt We have some breaking changes for the current `BackendImpleWithSetup`:

1.  write_setup needs pil and fixed due to the `setup` is circuit related, so it may look like below. 
```
    fn write_setup(&self, pil: &Analyzed<F>, fixed: &[(String, Vec<F>)], output_dir: &str) -> Result<(), io::Error>;
```
We'll write the `StarkInfo`, `Program`, and `const_tree` into output_dir, in which the `const_tree` may be large. The code is at [here](https://github.com/0xEigenLabs/eigen-zkvm/pull/176/files#diff-2e6e02d9cd6035d2709a31fe59837d0f2cb1812b32c8ac222a9e39b5130bf203R24). 

Hence, for `setup`, we already need the PIL, I have no idea how to get the PIL at `setup` stage.

2.   For the `prove`, I will produce a `circom`  and  a `input`(starkproof), and will consider them as proof, and write them into 2 files. This can be done by your tip 2.  But this PR will not generate the final snark proof, cause the following steps will involve more deps from eigen-zkvm, we may discuss in the next PR.

leonardoalt:
1. I think instead of passing these to the setup function you could change the prover struct instead to already contain these, like in https://github.com/powdr-labs/powdr/blob/main/halo2/src/prover.rs#L40 . But maybe we need to re-think the whole thing to use the new Pipeline (https://github.com/powdr-labs/powdr/blob/main/pipeline/src/pipeline.rs#L128)  actually

2. You mean as a serialized single proof object?

eigmax:
For point 2, I mean we can output 2 proofs(one is the stark verifier circuit, written by Circom) and starkproof,  and then I can use eigen-zkvm to produce the recursive proof. 

For point 1, the `StarkSetup` is to build a circuit related constant Merkle tree.  From your suggestion, I should first create a PILProver to carry on the PIL, hence the PILProver can do `StarkSetup` and `prove`.  Am I making sense?

leonardoalt:
1. Yea I think that makes sense!

2. You could pack both proofs in a single new object and serialize it to match the current proof structure, which is just a Vec<u8> I think

lvella:
I am not following what is going on here, I am just writing to tell you yo please undraft this PR when it is ready for review and merging.

PR #873, opened 2024-01-04
Use LambdaExpression for Query already in parser AST.

chriseth:
The analyzed AST uses LambdaExpression for a Query, but the parsed AST still had the names of the parameters as an extra field. This changes makes it easier to analyze the code because local variables are now only defined via parameters in LambdaExpressions.


leonardoalt (dismissed):

leonardoalt:
tests are failing

chriseth:
should be fixed now

leonardoalt (approved):

PR #874, opened 2024-01-05
Introduced get_pc()

lvella:
`get_pc()` saves a hash table lookup on every instruction that reads PC

### DIFF
--- a/riscv_executor/src/lib.rs
+++ b/riscv_executor/src/lib.rs
@@ -246,6 +246,11 @@ mod builder {
             }
         }
 
+        /// get current value of PC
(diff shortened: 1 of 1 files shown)


leonardoalt (commented):
Why did you remove len? It's useful to know it in fast mode, I'm using it in powdr_revm

lvella:
> Why did you remove len? It's useful to know it in fast mode, I'm using it in powdr_revm

Sorry, I didn't find any user inside powdr crate, so I assumed it had no users. Let me revert it.

leonardoalt (approved):

PR #876, opened 2024-01-07
Fix display of lambda functions.

chriseth:
Without the additional parentheses, displaying and re-parsing `(|x| x)(2)` would result in `|x| x(2)`.

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -169,8 +169,8 @@ mod test {
 constant %N = 16;
 namespace Fibonacci(%N);
 constant %last_row = (%N - 1);
-let bool = [|X| (X * (1 - X))][0];
-let one_hot = |i, which| match i { which => 1, _ => 0, };
+let bool = [(|X| (X * (1 - X)))][0];
+let one_hot = (|i, which| match i { which => 1, _ => 0, });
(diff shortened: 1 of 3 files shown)


leonardoalt:
tests are failing

parser/src/lib.rs
```diff
@@ -169,8 +169,8 @@ mod test {
 constant %N = 16;
 namespace Fibonacci(%N);
 constant %last_row = (%N - 1);
-let bool = [|X| (X * (1 - X))][0];
-let one_hot = |i, which| match i { which => 1, _ => 0, };
+let bool = [(|X| (X * (1 - X)))][0];
+let one_hot = (|i, which| match i { which => 1, _ => 0, });
```
leonardoalt, line 173: this looks uglier now :D

why are the parens needed? Shouldn't that be done by the parent of that node if needed?

chriseth: It's the same as requiring parens for `a + b`, we add them because the Display trait has no sense of context.
We should fix that at some point and build a prettier expression printer.

chriseth: If you put the responsibility on the parent, you will get `a + b` being formatted as `(1) + (b)`.

leonardoalt (approved):

PR #888, opened 2024-01-09
new powdr crate

leonardoalt:
this is done in a very simple way right now just to get it started. Follow up PRs should

- use this crate in `powdr_cli`
- re-export relevant symbols directly

### DIFF
--- a/powdr/Cargo.toml
+++ b/powdr/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "powdr"
+version = "0.1.0"
+edition = "2021"
+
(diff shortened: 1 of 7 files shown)


powdr/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "powdr"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
lvella, line 6: Maybe remove this? I am ok either way.

lvella (dismissed):

PR #891, opened 2024-01-09
change the data type that continuations takes

leonardoalt:
This is also just a hot fix needed for powdr_revm, we should find a proper solution using a new ser/de data format for both the Pipeline and this


georgwiese (approved):
LGTM

georgwiese (approved):

PR #911, opened 2024-01-15
Much smaller and faster trace to build.

lvella:
This saves a tiny bit of time when running continuations (from 18.7s to 18.1s on `many_chunks.rs`), and uses quite less memory when transposing the trace.

### DIFF
--- a/riscv_executor/src/lib.rs
+++ b/riscv_executor/src/lib.rs
@@ -105,36 +105,83 @@ pub enum MemOperationKind {
 
 #[derive(Debug)]
 pub struct MemOperation {
-    /// Line of the register trace the memory operation happened.
-    pub idx: usize,
+    /// The row of the execution trace the memory operation happened.
+    pub row: usize,
     pub kind: MemOperationKind,
     pub address: u32,
 }
 
+pub struct RegWrite {
+    /// The row of the execution trace this write will result into. Multiple
+    /// writes at the same row are valid: the last write to a given reg_idx will
+    /// define the final value of the register in that row.
+    row: usize,
+    /// Index of the register in the register bank.
+    reg_idx: u16,
+    val: Elem,
+}
+
 pub struct ExecutionTrace {
-    pub reg_map: HashMap<String, usize>,
+    pub reg_map: HashMap<String, u16>,
 
     /// Values of the registers in the execution trace.
     ///
     /// Each N elements is a row with all registers, where N is the number of
     /// registers.
-    pub regs: Vec<Elem>,
+    pub reg_writes: Vec<RegWrite>,
 
     /// Writes and reads to memory.
-    pub mem: Vec<MemOperation>,
+    pub mem_ops: Vec<MemOperation>,
 
-    /// The length of the trace.
+    /// The length of the trace, after applying the reg_writes.
     pub len: usize,
 }
 
 impl ExecutionTrace {
-    /// Split the values of the registers' trace into rows.
-    pub fn regs_rows(&self) -> impl Iterator<Item = &[Elem]> {
-        self.regs.chunks_exact(self.reg_map.len())
+    /// Replay the execution and get the register values per trace row.
+    pub fn replay(&self) -> TraceReplay {
+        TraceReplay {
+            trace: self,
+            regs: vec![0.into(); self.reg_map.len()],
+            pc_idx: self.reg_map["pc"] as usize,
+            next_write: 0,
+            next_r: 0,
+        }
     }
+}
 
-    pub fn row(&self, idx: usize) -> &[Elem] {
-        &self.regs[(idx * self.reg_map.len())..((idx + 1) * self.reg_map.len())]
+pub struct TraceReplay<'a> {
+    trace: &'a ExecutionTrace,
+    regs: Vec<Elem>,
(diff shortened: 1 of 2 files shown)


riscv_executor/src/lib.rs
```diff
@@ -105,36 +105,83 @@ pub enum MemOperationKind {
...
     pub kind: MemOperationKind,
     pub address: u32,
 }
 
+pub struct RegWrite {
+    /// The row of the execution trace this write will effect. Multiple writes
+    /// at the same row are valid: the last write to a given reg_idx will define
+    /// the final value of the register in that row.
+    row: usize,
+    /// Index of the register in the register bank written.
```
leonardoalt, line 119:
```suggestion
    /// Index of the register in the register bank.
```

riscv_executor/src/lib.rs
```diff
@@ -105,36 +105,83 @@ pub enum MemOperationKind {
...
 
 impl ExecutionTrace {
-    /// Split the values of the registers' trace into rows.
-    pub fn regs_rows(&self) -> impl Iterator<Item = &[Elem]> {
-        self.regs.chunks_exact(self.reg_map.len())
+    /// Replay the execution and get the register values per trace row.
+    pub fn replay(&self) -> TraceReplay {
+        TraceReplay {
+            trace: self,
+            regs: vec![0.into(); self.reg_map.len()],
```
leonardoalt, line 145: Why 0? Wasn't this supposed to replay something?

riscv_executor/src/lib.rs
```diff
@@ -276,7 +327,7 @@ mod builder {
 
         /// set next value of register, accounting to x0 writes
         ///
-        /// to set the PC, use s_pc() instead of this
+        /// to set the PC, use set_pc() instead of this
```
leonardoalt, line 330: maybe assert this isn't a pc write?

leonardoalt (commented):
some minor comments

leonardoalt:
test failing

riscv_executor/src/lib.rs
lvella: Zero is the initial state of all registers. This vector only holds one row at a time.

leonardoalt: what is this function supposed to do? Just create the object?

riscv_executor/src/lib.rs
```diff
@@ -105,36 +105,83 @@ pub enum MemOperationKind {
...
-    /// Line of the register trace the memory operation happened.
-    pub idx: usize,
+    /// The row of the execution trace the memory operation happened.
+    pub row: usize,
     pub kind: MemOperationKind,
     pub address: u32,
 }
 
+pub struct RegWrite {
+    /// The row of the execution trace this write will effect. Multiple writes
```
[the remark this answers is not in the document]
lvella: The weird wording was on purpose, because the result of the write will end up on that row. But to use "effect" implies that a row is the consequence of a single write, which is not necessarily true.

Let me think of a better way of writing that.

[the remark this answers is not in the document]
leonardoalt: If you want to use the verb, it should be `affect`. The noun, `effect`. That's all I meant.

riscv_executor/src/lib.rs
lvella: Yes, this object was supposed to be an iterator over the trace rows, but I couldn't implement `Iterator` on it because `next()` would have to return a slice borrowed from `self`, and the iterator interface doesn't allow for that, as the result of a proper `Iterator` should be able to outlive the iterator itself.

So now it is a plain struct with a method called `next_row()` instead of `next()`, because clippy complains that it might be confused with `next()` from trait `Iterator` (and suggest me to implement `Iterator` for it instead, which is impossible).

leonardoalt: if this function only creates the object, why not call it `new`?

riscv_executor/src/lib.rs
lvella: Ok, I've included an assert.

riscv_executor/src/lib.rs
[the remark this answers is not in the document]
lvella: "to effect" is a verb too, and is similar, but a little different from "to affect"

Which is irrelevant now because I changed to neither affect nor effect.

riscv_executor/src/lib.rs
lvella: Because it creates an object of another type, not `Self`.

The other object is an iterator-like, and the convention is the thing they are iterating on to create them. E.g., it is `vec.drain()`, not `Drain::new(&mut vec)`.

leonardoalt: ok

leonardoalt (approved):
maybe squash the commits

lvella:
> maybe squash the commits

This is so tiresome. We should find a way to squash automatically.

leonardoalt:
it takes 5 seconds to squash..

leonardoalt:
> without the need to rerun the tests if the files are identical.

I think GH keeps the approval status if that's the case, no?

leonardoalt (approved):

PR #914, opened 2024-01-17
Better errors for fixed col evaluation.

chriseth:


executor/src/constant_evaluator/mod.rs
```diff
@@ -35,45 +42,46 @@ fn generate_values<T: FieldElement>(
...
+            .map(|values| {
+                let values: Vec<T> = values.into_iter().flatten().collect();
+                assert_eq!(values.len(), degree as usize);
+                values
+            }),
         FunctionValueDefinition::Query(_) => panic!("Query used for fixed column."),
+    };
+    match result {
+        Err(err) => {
+            eprintln!("Error evaluating fixed polynomial {name} = {body}:\n{err}");
```
chriseth, line 81: Is it fine to print to stderr here or should we propagate the error further upwards?
The problem with propagating is that we have to create yet another error class that contains the name and the body - or just turn it into a string here.

pacheco: Is there a possibility for the error to be handled somewhere else later (e.g., nicer reporting maybe), so that the temporary unwraps elsewhere are worth it? If so maybe `Result<_, String>` is enough?

leonardoalt: I think eventually we should really refactor all the error stuff to properly propagate errors upwards. I think the `backend` crate already does some of that.

leonardoalt (approved):

PR #918, opened 2024-01-18
Generate references to the source while parsing

pacheco:
This PR at first was only intended to add a way for analyzing a `PILFile` directly without having to go back to source form.
In doing that, I ran into some logic in the analyzer related to matching parsed statements to the originating source.
This PR encapsulates this logic into the parser, so parsed PIL and ASM statements carry a `SourceRef` which points to the file, line and column where it originated from.
For now, generated statements (e.g., linker step) use a SourceRef::unknown(), which just points to no file.

### DIFF
--- a/backend/src/pilstark/json_exporter/mod.rs
+++ b/backend/src/pilstark/json_exporter/mod.rs
@@ -1,6 +1,6 @@
 use number::FieldElement;
-use std::cmp;
 use std::collections::HashMap;
+use std::{cmp, path::PathBuf};
 
 use ast::analyzed::{
     AlgebraicBinaryOperator, AlgebraicExpression as Expression, AlgebraicUnaryOperator, Analyzed,
@@ -77,7 +77,18 @@ pub fn export<T: FieldElement>(analyzed: &Analyzed<T>) -> PIL {
             }
             StatementIdentifier::Identity(id) => {
                 let identity = &analyzed.identities[*id];
-                let file_name = identity.source.file.clone();
+                // PILCOM strips the path from filenames, we do the same here for compatibility
+                let file_name = identity
+                    .source
+                    .file
+                    .as_deref()
+                    .and_then(|s| {
+                        PathBuf::from(s)
+                            .file_name()
+                            .and_then(|n| n.to_str())
+                            .map(String::from)
+                    })
+                    .unwrap_or_default();
                 let line = identity.source.line;
                 let selector_degree = if identity.kind == IdentityKind::Polynomial {
                     2
@@ -444,6 +455,8 @@ mod test {
 
     fn compare_export_file(file: &str) {
         let (json_out, pilcom_parsed) = generate_json_pair(file);
+        // TODO: test fails because PILCOM "fileName" fields output the filename
+        // stripped of its path
         assert_eq!(json_out, pilcom_parsed);
     }
 
--- a/parser/src/powdr.lalrpop
+++ b/parser/src/powdr.lalrpop
@@ -2,8 +2,9 @@ use std::str::FromStr;
 use ast::parsed::{*, asm::*};
 use number::{AbstractNumberType, FieldElement};
 use num_traits::Num;
+use crate::ParserContext;
 
-grammar<T> where T: FieldElement;
+grammar<T>(ctx: &ParserContext) where T: FieldElement;
 
 match {
     r"\s*" => { },
@@ -14,7 +15,6 @@ match {
 
 pub PILFile: PILFile<T> = {
     (<PilStatement> ";")* => PILFile(<>)
-
 };
 
 pub ASMModule: ASMModule<T> = {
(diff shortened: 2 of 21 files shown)


backend/src/pilstark/json_exporter/mod.rs
```diff
@@ -77,7 +77,12 @@ pub fn export<T: FieldElement>(analyzed: &Analyzed<T>) -> PIL {
             }
             StatementIdentifier::Identity(id) => {
                 let identity = &analyzed.identities[*id];
-                let file_name = identity.source.file.clone();
+                let file_name: String = identity
+                    .source
+                    .file
+                    .as_deref()
+                    .map(String::from)
+                    .unwrap_or_default();
```
chriseth, line 85: I think you could just extract the file name from the path here to make the tests work.

parser/src/powdr.lalrpop
```diff
@@ -80,39 +80,39 @@ pub PilStatement = {
 };
 
 Include: PilStatement<T> = {
-    <start:@L> "include" <file:StringLiteral> => PilStatement::Include(<>)
+    <start:@L> "include" <file:StringLiteral> => PilStatement::Include(ctx.source_ref(start), file)
 };
 
 Namespace: PilStatement<T> = {
-    <start:@L> "namespace" <name:SymbolPath> "(" <pol_degree:Expression> ")" => PilStatement::Namespace(<>)
+    <start:@L> "namespace" <name:SymbolPath> "(" <pol_degree:Expression> ")" => PilStatement::Namespace(ctx.source_ref(start), name, pol_degree)
```
chriseth, line 87: I'm wondering if we could use a macro to make these less ugly...

chriseth: On the other hand, the macro would probably also look ugly ;)

parser_util/src/lines.rs
```diff
@@ -4,11 +4,12 @@ pub fn compute_line_starts(source: &str) -> Vec<usize> {
         .collect::<Vec<_>>()
 }
 
-pub fn offset_to_line(offset: usize, line_starts: &[usize]) -> usize {
-    match line_starts.binary_search(&offset) {
+pub fn offset_to_line_col(offset: usize, line_starts: &[usize]) -> (usize, usize) {
```
chriseth, line 7: Could you document if this is 0- or 1-based?

pil_analyzer/src/lib.rs
```diff
@@ -27,7 +31,5 @@ pub trait AnalysisDriver<T>: Clone + Copy {
     fn resolve_decl(&self, name: &str) -> String;
     /// Turns a reference to a name with an optional namespace into an absolute name.
     fn resolve_ref(&self, path: &SymbolPath) -> String;
-    /// Translates a file-local source position into a proper SourceRef.
-    fn source_position_to_source_ref(&self, pos: usize) -> SourceRef;
```
chriseth, line 31: Oh nice! This was a really annoying part of the trait.

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -26,15 +27,23 @@ pub fn process_pil_file<T: FieldElement>(path: &Path) -> Analyzed<T> {
...
 pub fn process_pil_file_contents<T: FieldElement>(contents: &str) -> Analyzed<T> {
     let file_name = Path::new("input");
     let mut analyzer = PILAnalyzer::new();
 
-    analyzer.process(
-        [(file_name.to_owned(), ParsedFile::parse(file_name, contents))]
-            .into_iter()
-            .collect(),
-    );
+    let ast = parser::parse(Some("input"), contents).unwrap_or_else(|err| {
```
chriseth, line 40: I'm wondering - maybe "stdin" or "unspecified input" could be a separate enum value for file name? But we could also do this another time.

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -26,15 +27,23 @@ pub fn process_pil_file<T: FieldElement>(path: &Path) -> Analyzed<T> {
...
+        eprintln!("Error parsing .pil file:");
+        err.output_to_stderr();
+        panic!();
+    });
+
+    analyzer.process([(file_name.to_owned(), ast)].into_iter().collect());
```
chriseth, line 46: You could call `process_pil_ast` here.

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -50,38 +59,47 @@ struct PILAnalyzer<T> {
...
-        let line_starts = parser_util::lines::compute_line_starts(contents);
-        let ast = parser::parse(Some(path.to_str().unwrap()), contents).unwrap_or_else(|err| {
-            eprintln!("Error parsing .pil file:");
-            err.output_to_stderr();
-            panic!();
-        });
-        ParsedFile { line_starts, ast }
+// when starting from an already parsed PILFile there's no reference to source code.
+// Make it so each statement is considred to be in its own line.
+fn pilfile_set_one_statement_per_line<T>(pil: &mut PILFile<T>) {
```
chriseth, line 66: Can this be deleted now?

parser/src/powdr.lalrpop
pacheco: Something like `pil_statement!(Namespace, <>)`?
I think its not too bad as is, since the ugliness is contained to the lalrpop file and makes the need for the `ParserContext` more obvious

pil_analyzer/src/pil_analyzer.rs
pacheco: I was using `Option<SourceRef>` at first, so that `None` would represent "generated" statements... but I got the feeling I was overcomplicating for the current usage (was having to decide when to propagate the Option or unwrap).
Using an enum could be nice for a future PR.

pil_analyzer/src/pil_analyzer.rs
pacheco: nice catch!

pil_analyzer/src/pil_analyzer.rs
pacheco: the idea here was to have lines matching the .pil file generated in the previous pipeline step... but even if this would be useful, the analyzer is now the wrong place to do it. Will remove.

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -127,22 +144,33 @@ impl<T: FieldElement> PILAnalyzer<T> {
...
-            self.current_line_starts = parsed.line_starts;
-            for statement in parsed.ast.0 {
+        for (_name, ast) in files {
+            for statement in ast.0 {
                 self.handle_statement(statement);
             }
         }
     }
 
+    pub fn process_ast(&mut self, mut pil_file: PILFile<T>) {
```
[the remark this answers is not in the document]
pacheco: yep, seems there's no need to pass filenames around anymore

pil_analyzer/src/pil_analyzer.rs
```diff
@@ -110,27 +110,14 @@ impl<T: FieldElement> PILAnalyzer<T> {
...
 
-    pub fn process(&mut self, files: Vec<(PathBuf, PILFile<T>)>) {
+    pub fn process(&mut self, files: Vec<PILFile<T>>) {
         self.current_namespace = Default::default();
-        for statement in files.iter().flat_map(|(_, f)| f.0.iter()) {
+        for statement in files.iter().flat_map(|f| f.0.iter()) {
             self.collect_names(statement);
         }
 
         self.current_namespace = Default::default();
```
chriseth, line 119: I think resetting the namespace at the beginning of each file instead of at the beginning of processing is the right thing to do. It is a fix and not a refactoring, but I think we should do it in this PR nevertheless.

chriseth: (same for the collect_names loop above)

chriseth (approved):

PR #920, opened 2024-01-19
Builtin functions.

chriseth:


std/check.asm
```diff
@@ -0,0 +1,9 @@
+/// This is a built-in function taking a string argument and terminating
+/// evaluation unsuccessfully with this argument as explanation.
+/// This symbol is not an empty array, the actual semantics are overridden.
+let panic = [];
```
leonardoalt, line 4: This is quite confusing

test_data/asm/pil_at_module_level.asm
```diff
@@ -38,6 +38,6 @@ mod R {
 
         // Try to see if we only clear local variables
         // if they not already exist in the outer scope.
-        make_array(2, |i| ((|i| w)(2))[i] == 4);
+        array::new2, |i| ((|i| w)(2))[i] == 4);
```
leonardoalt, line 41: that's a weird name

test_data/pil/book/generic_to_algebraic.pil
```diff
@@ -31,4 +31,4 @@ namespace Main(16);
     let make_array = |length, f| fold(length, f, [], |acc, e| acc + [e]);
     // If an expression evaluates to an array of constraints, all the
     // constraints in the array are added to the system.
-    make_array(15, |i| wit[i] == 1);
\ No newline at end of file
+    array::new15, |i| wit[i] == 1);
```
leonardoalt, line 34: the poseidon code does `new(len, ...)`, why is it in the function name here?

test_data/pil/arith_improved.pil
```diff
@@ -140,16 +140,16 @@ namespace Arith(N);
...
+    array::new16, |i| fixed_inside_32_block(x1[i]));
+    array::new16, |i| fixed_inside_32_block(y1[i]));
+    array::new16, |i| fixed_inside_32_block(x2[i]));
+    array::new16, |i| fixed_inside_32_block(y2[i]));
+    array::new16, |i| fixed_inside_32_block(x3[i]));
+    array::new16, |i| fixed_inside_32_block(y3[i]));
+    array::new16, |i| fixed_inside_32_block(s[i]));
+    array::new16, |i| fixed_inside_32_block(q0[i]));
+    array::new16, |i| fixed_inside_32_block(q1[i]));
+    array::new16, |i| fixed_inside_32_block(q2[i]));
```
leonardoalt, line 152: is this a wrongly applied replace?

std/array.asm
```diff
@@ -0,0 +1,16 @@
+/// This is a built-in function taking an array argument and returning
+/// the length of the array.
+/// This symbol is not an empty array, the actual semantics are overridden.
+let len = [];
```
leonardoalt, line 4: This is also confusing. Why is it done like this?

std/check.asm
chriseth: The main other way I could think of is something like:
```
std::core:
let builtin = |name| (); // The only weird thing
std::check:
let panic = std::core::builtin("panic");
```

And then `builtin` would be the only real builtin function.

chriseth: although this would not pass strict type checking

chriseth: ok, we could use something like `let panic = std::core::builtin::<"panic">();`

leonardoalt: if it's already built-in, why not make it fully built-in?

test_data/pil/arith_improved.pil
chriseth: yes

std/array.asm
chriseth: If I just use `let len;`, then it will be a witness column. I have to assign something.
Although thinking about it, maybe it could work?
But in any case, since we don't have any annotations or stuff like this, it has to be weird in some way.

leonardoalt:
if you rebase circle vanishes :D

std/check.asm
leonardoalt: so we keep it like this for now?

chriseth:
rebased

pil_analyzer/src/evaluator.rs
```diff
@@ -130,12 +137,27 @@ impl<'a, T: FieldElement, C: Custom> Value<'a, T, C> {
                 format!("[{}]", elements.iter().map(|e| e.type_name()).format(", "))
             }
             Value::Closure(c) => c.type_name(),
+            Value::BuiltinFunction(b) => format!("builtin{b:?}"),
```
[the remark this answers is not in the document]
chriseth: Actually this should usually be the type name of a closure. Let me make it proper.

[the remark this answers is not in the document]
chriseth: ok, turns out we don't even know the types in a closure.

What about `builtin_`? I don't particularly like `::` because it is not a namespace.

As soon as we add proper types, this will be replaced anyway.

leonardoalt (approved):

leonardoalt (approved):

PR #921, opened 2024-01-20
Add powdr- prefix to crate names

leonardoalt:
I want to start preparing the crates for release and publishing on crates.io . The first convention is to name all crates `powdr-*`.

For now I only renamed the crates in Cargo.toml files, so the code doesn't compile. If we decide that this is fine and decide on some conventions, I'll go ahead and change the code as well.


asm_to_pil/Cargo.toml
```diff
@@ -1,15 +1,16 @@
 [package]
-name = "asm_to_pil"
+name = "powdr-asm_to_pil"
```
leonardoalt, line 2: I think the main thing we need to decide is the name convention of crates with multiple words. Do we use `powdr-asm_to_pil` or `powdr-asm-to-pil`?

pacheco: IMO it's better to use the same separator

leonardoalt:
I think I'd prefer to use the `powdr-` prefix everywhere to make it obvious what's from powdr

chriseth (approved):

chriseth (dismissed):

pacheco (approved):

PR #923, opened 2024-01-22
Remove reordering of labels in data segment, as it is no longer needed.

lvella:


riscv/src/compiler.rs
```diff
@@ -137,18 +137,6 @@ pub fn compile(
     // for compilation, and will not be called.
     statements = replace_coprocessor_stubs(statements, coprocessors).collect::<Vec<_>>();
 
-    // Sort the objects according to the order of the names in object_order.
```
chriseth, line 140: Could you add an assertion checking that the sum of the sizes is less than 2**31 or something?

lvella:
We still have a gigantic heap (1GB), but it is fine now because, according to @leonardoalt, quote:
> one problem is that the difference between two consecutive addresses in the memory machine could not be > degree
> but we fixed that
> now the only constraint left is the address space of 32 bits itself, which is inherent in RISCV32

I guess this is correct because almost all test are passing now. I found one is broken, I am testing now if that one is also broken in main.

riscv/src/compiler.rs
chriseth: Ah sorry, this is not needed anymore.

chriseth:
Yep, I see it now.

lvella:
`test_print_rv32_executor` is broken in main

chriseth (approved):

PR #927, opened 2024-01-22
remove old files

leonardoalt:


chriseth:
I was actually just yesterday reading through those and got some good new ideas...
I think we should at least archive them somewhere

leonardoalt:
> I was actually just yesterday reading through those and got some good new ideas... I think we should at least archive them somewhere

ok, can put them on hackmd

leonardoalt:
https://hackmd.io/@leoalt/IDEAS
https://hackmd.io/@leoalt/notes_asm

posted here for now

PR #928, opened 2024-01-22
remove runtime crate

leonardoalt:
For the tests that use crates it's possible to use the same powdr-riscv-rt crate from the checkout. For file tests we need to use git. I kept them separate for now but I can make everything git if that's preferred.


riscv/src/lib.rs
```diff
@@ -152,7 +152,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-runtime = {{ path = "./runtime" }}
+powdr_riscv_rt = {{ git = "https://github.com/powdr-labs/powdr", branch = "main" }}
```
chriseth, line 155: Doesn't this also need the `runtime` subdirectory?

riscv/tests/riscv_data/vec_median/Cargo.toml
```diff
@@ -4,6 +4,6 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-runtime = { path = "../../../runtime" }
+powdr-riscv-rt = { path = "../../../../powdr_riscv_rt" }
```
chriseth, line 7: I would opt for actually spelling `runtime` out. `rt` can mean so many things.

riscv/src/lib.rs
leonardoalt: What for?

riscv/tests/riscv_data/vec_median/Cargo.toml
leonardoalt: I think this is rather common. We use another crate called riscv_rt for qemu, so i thought powdr-riscv-rt fits well

leonardoalt: I mean, common also if you include the context you get from the "riscv" prefix to "rt"

chriseth: Even if it's common, I really don't think it's a good idea. We also don't abbreviate executor or other stuff.

chriseth: I mean isn't even runtime already an abbreviation for runtime environment?

leonardoalt: Ok. I still have a somewhat strong opinion about this. `powdr-riscv-runtime` is just really long and kinda clumsy. If someone else says they prefer it spelled out I'll give in.

chriseth: What does it matter if it's long and clumsy? Do you need to type it that often?

leonardoalt: just aesthetics

chriseth (approved):

PR #933, opened 2024-01-23
github actions: only store cache from main

pacheco:
Cache entry from `main` keeps getting evicted. Prevent other branches from writing cache entries (they can only fetch from main).

### DIFF
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -1,8 +1,12 @@
 name: PR tests
 
 on:
+  workflow_dispatch:
   pull_request:
   merge_group:
+  push:
+    branches:
+      - main
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


chriseth (dismissed):

.github/workflows/pr-tests.yml
```diff
@@ -49,8 +56,19 @@ jobs:
...
+      if: github.ref == 'refs/heads/main'
+      uses: actions/cache/save@v3
+      with:
+        path: |
+          ~/.cargo/registry
+          ~/.cargo/git
+          target
+        key: ${{ runner.os }}-cargo-pr-tests-${{ hashFiles('**/Cargo.toml') }}
     - name: Run default tests
+      if: github.ref != 'refs/heads/main'
```
chriseth, line 69: I'll merge this to see the effect now, but why did you add this restriction here?

chriseth (approved):

.github/workflows/pr-tests.yml
chriseth: Yes, I think it's better to run them again. It's might also be easier to track problems if we ever get unstable tests or so.

PR #934, opened 2024-01-23
Print builtin

chriseth:


parser/src/lib.rs
```diff
@@ -67,6 +67,29 @@ pub fn parse_module<'a, T: FieldElement>(
         .map_err(|err| handle_parse_error(err, file_name, input))
 }
 
+/// Parse an escaped string - used in the grammar.
+pub fn unescape_string(s: &str) -> String {
```
leonardoalt, line 71: is there no utility for this?

leonardoalt (approved):

parser/src/lib.rs
chriseth: No. we could use an external crate for that, though.

leonardoalt: yea better not

PR #944, opened 2024-01-24
Split tests.

chriseth:


.github/workflows/pr-tests.yml
```diff
@@ -65,9 +64,71 @@ jobs:
...
+  test_slow:
+    needs: build
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
+    - name: Download build artifacts
+      uses: actions/download-artifact@v2
```
pacheco, line 111: These artifacts last only for the workflow duration?

.github/workflows/pr-tests.yml
```diff
@@ -65,9 +64,71 @@ jobs:
           ~/.cargo/git
           target
         key: ${{ runner.os }}-cargo-pr-tests-${{ hashFiles('**/Cargo.toml') }}
+
+  test_quick:
+    needs: build
```
pacheco, line 69: Should `test_quick` and `test_slow` also run after the merge to main?
If not, we could add `if: github.ref != 'refs/heads/main'` here and on the `test_slow` section.

.github/workflows/pr-tests.yml
chriseth: Yes - is there a better way in this case?

pacheco (approved):

PR #950, opened 2024-01-24
More conforming handling of asm data sections

lvella:


asm_utils/src/data_parser.rs
```diff
@@ -14,76 +18,131 @@ pub enum DataValue {
...
     /// Returns the size of the value in bytes.
-    pub fn size(&self) -> usize {
+    ///
+    /// The address is necessary because the size of the alignment padding
+    /// depending on what address it is defined on.
+    pub fn size(&self, from_addr: usize) -> usize {
         match self {
             DataValue::Direct(data) => data.len(),
             DataValue::Zero(length) => *length,
+            DataValue::Alignment(bytes) => alignment_size(from_addr, *bytes),
```
chriseth, line 28: so the alignment directive essentially creates a padding, right?

asm_utils/src/utils.rs
```diff
@@ -1,7 +1,15 @@
...
-pub fn next_multiple_of_four(x: usize) -> usize {
-    ((x + 3) / 4) * 4
+pub fn next_aligned(val: usize, alignment: usize) -> usize {
+    // Alignment will probably always be a power of two, which can be aligned in
+    // a much faster bitwise operation. But then we would have to assert!() it,
+    // so it is just better to use the generic version.
+    ((val + (alignment - 1)) / alignment) * alignment
+}
+
+pub fn alignment_size(from: usize, alignment: usize) -> usize {
```
chriseth, line 10: Can you document this? Can the be described as "padding to next alignment boundary"?

asm_utils/src/data_parser.rs
```diff
@@ -14,76 +18,131 @@ pub enum DataValue {
...
 }
 
+struct DataObjects<'a> {
+    objects: BTreeMap<String, Vec<DataValue>>,
+    object_order: Vec<String>,
+    current_label: Option<&'a str>,
+}
+
+impl<'a> DataObjects<'a> {
+    fn new() -> Self {
```
chriseth, line 42: Default::default?

chriseth: This should even be `#[derive(Default)]`

asm_utils/src/data_parser.rs
```diff
@@ -14,76 +18,131 @@ pub enum DataValue {
...
+        Self {
+            objects: BTreeMap::new(),
+            object_order: Vec::new(),
+            current_label: None,
+        }
+    }
+
+    fn curr_entry(&mut self) -> &mut Vec<DataValue> {
```
chriseth, line 50:
```suggestion
    fn current_entry(&mut self) -> &mut Vec<DataValue> {
```

asm_utils/src/data_parser.rs
```diff
@@ -14,76 +18,131 @@ pub enum DataValue {
...
+
+fn is_data_section<R: Register, F: FunctionOpKind>(arg: &Argument<R, F>) -> bool {
+    let full_name = match arg {
+        Argument::StringLiteral(name) => name.as_slice(),
+        Argument::Expression(Expression::Symbol(name)) => name.as_bytes(),
+        _ => return false,
+    };
+
+    // split the initial '.' out
+    let name = &full_name[(1 + full_name.iter().position(|c| *c == b'.').unwrap())..];
```
chriseth, line 137: can you use `split_prefix` here?

asm_utils/src/data_parser.rs
lvella: Not necessarily. Only if needed.

asm_utils/src/data_parser.rs
lvella: Shouldn't we implement traits as needed? Anyway, this current version, using this struct, is broken.

asm_utils/src/data_parser.rs
lvella: ok

asm_utils/src/data_parser.rs
lvella: What `split_prefix` ?

asm_utils/src/utils.rs
lvella: Done.

leonardoalt:
Is this ready for review?

lvella:
> Is this ready for review?

Considering the conflicts, I'd say no.
Also, I don't know if I should try to restore the original instruction tests in this PR or another.

asm-utils/src/data_parser.rs
```diff
@@ -14,76 +16,134 @@ pub enum DataValue {
 
 impl DataValue {
     /// Returns the size of the value in bytes.
-    pub fn size(&self) -> usize {
+    ///
+    /// The address is necessary because the size of the alignment padding
+    /// depending on what address it is defined on.
```
leonardoalt, line 21:
```suggestion
    /// depends on what address it is defined on.
```

asm-utils/src/data_parser.rs
```diff
@@ -14,76 +16,134 @@ pub enum DataValue {
...
+            DataValue::Alignment(bytes) => alignment_size(from_addr, *bytes),
             DataValue::Reference(_) => 4,
             DataValue::Offset(..) => 4,
         }
     }
 }
 
+#[derive(Default)]
+struct DataSections {
+    sections: Vec<Vec<(Option<String>, Vec<DataValue>)>>,
```
leonardoalt, line 35: Can you add a comment explaining what this represents? Is it a "sequence of sectors, where a sector is a maybe named sequence of values"?

asm-utils/src/data_parser.rs
```diff
@@ -14,76 +16,134 @@ pub enum DataValue {
...
+}
+
+impl DataSections {
+    fn new() -> Self {
+        Default::default()
+    }
+
+    fn curr_entry(&mut self) -> &mut Vec<DataValue> {
```
leonardoalt, line 43:
```suggestion
    fn current_entry(&mut self) -> &mut Vec<DataValue> {
```

This wasn't applied from the previous review

asm-utils/src/data_parser.rs
```diff
@@ -14,76 +16,134 @@ pub enum DataValue {
...
+                }
+                (".p2align", [Argument::Expression(Expression::Number(pow_of_2))]) => {
+                    if is_in_data_section {
+                        data.curr_entry()
+                            .push(DataValue::Alignment((1 << pow_of_2) as usize));
+                    }
+                }
+                (".balign" | ".p2align", _) => {
+                    // TODO: implement the optional arguments of .balign and .p2align
+                    unimplemented!()
```
leonardoalt, line 120: What about `.type` and `.size` directives?

asm-utils/src/data_parser.rs
```diff
@@ -14,76 +16,134 @@ pub enum DataValue {
...
+
+fn is_data_section<R: Register, F: FunctionOpKind>(arg: &Argument<R, F>) -> bool {
+    let full_name = match arg {
+        Argument::StringLiteral(name) => name.as_slice(),
+        Argument::Expression(Expression::Symbol(name)) => name.as_bytes(),
+        _ => return false,
+    };
+
+    // split the initial '.' out
+    let name = &full_name[(1 + full_name.iter().position(|c| *c == b'.').unwrap())..];
```
leonardoalt, line 138: Is this really the easiest way to do this?

asm-utils/src/data_parser.rs
```diff
@@ -14,76 +16,134 @@ pub enum DataValue {
...
+
+    // isolate name until next '.'
+    let name = &name[..(name.iter().position(|e| *e == b'.').unwrap_or(name.len()))];
```
leonardoalt, line 141: and here

asm-utils/src/data_parser.rs
```diff
@@ -14,76 +16,134 @@ pub enum DataValue {
...
+
```
leonardoalt, line 142: gpt says

```rust
let parts: Vec<&str> = full_name.split('.').collect();
let name = parts.get(1).unwrap_or(&"");
```

asm-utils/src/data_parser.rs
```diff
@@ -109,6 +169,24 @@ fn extract_data_value<R: Register, F: FunctionOpKind>(
             data.push(0);
             vec![DataValue::Direct(data)]
         }
+        (".dword" | ".half" | ".hword" | ".short", data) => {
+            let len = match directive {
+                ".dword" => 8,
+                _ => 2,
+            };
```
leonardoalt, line 176: Since this is already handling different sizes, can't we also support `.byte` here directly? The rest of the code looks similar

asm-utils/src/reachability.rs
```diff
@@ -10,24 +10,34 @@ use crate::ast::{Argument, Expression, FunctionOpKind, Register, Statement};
...
-            .into_iter()
-            .map(|s| s.to_owned())
-            .collect::<HashSet<_>>();
+    let (referenced_labels, referenced_data_sections) =
+        find_reachable_labels::<R, F, A>(label, statements, data_sections, &replacement_refs);
+
+    {
+        let mut idx = 0usize;
+        data_sections.retain(|_| {
+            let must_retain = referenced_data_sections.contains(&idx);
```
leonardoalt, line 29: What's this index? Why is it growing?

asm-utils/src/data_storage.rs
```diff
@@ -2,76 +2,157 @@
...
+    data_parser::DataValue,
+    utils::{alignment_size, next_aligned},
+};
 
 pub enum SingleDataValue<'a> {
     Value(u32),
-    LabelReference(&'a String),
-    Offset(&'a String, &'a String),
+    LabelReference(&'a str),
+    Offset(&'a str, &'a str),
```
leonardoalt, line 13: Why str, just to complicate lifetimes? :D

asm-utils/src/data_storage.rs
```diff
@@ -2,76 +2,157 @@
...
+    latest_label: Option<&'b str>,
+}
+
+impl<'a, 'b> WordWriter<'a, 'b> {
+    fn new(
+        starting_pos: u32,
+        data_writer: &'a mut dyn FnMut(u32, SingleDataValue) -> Vec<String>,
+    ) -> Self {
+        // sanitary alignment to 8 bytes
+        let current_pos = next_aligned(starting_pos as usize, 8) as u32;
```
leonardoalt, line 31: where's this 8 coming from?

asm-utils/src/data_storage.rs
```diff
@@ -2,76 +2,157 @@
...
+            SingleDataValue::LabelReference(label),
+        ));
+
+        assert_eq!(self.partial, 0);
+        self.current_pos += 4;
+    }
+
+    fn finish(mut self) -> Vec<String> {
+        // ensure the latest partial word is written
+        self.advance(4);
```
leonardoalt, line 100: why 4?

leonardoalt (commented):
The new writer looks nice!

asm-utils/src/data_parser.rs
lvella: Both are flags stored in the ELF and not validated by the assembler.

We used `.type` to identify the label type, i.e., if it is data or text. Now that role is performed by the `.section`.

`.size` was used to assert the size of the data pointed by label, if it happened to be the last label seen. But now `size()` needs the address, and we don't have it at this point.

I could still do the assert, but I would have to make the following assumptions that might not be true for every assembly code:
- there is actually exactly that many bytes between the label definition and the `.size`;
- there are no `.align` and `.p2align` between the label and the `.size`;
- the `.size` is actually inside the section of the label it refers too (not sure GNU `as` would accept differently).

asm-utils/src/data_parser.rs
lvella: That is the best Google could find.

asm-utils/src/data_parser.rs
leonardoalt: that's fine, just wanted to know what happened to them

asm-utils/src/data_parser.rs
leonardoalt: gpt

asm-utils/src/data_parser.rs
lvella: That is not an `&str`. I used `&[u8]` because that is the common denominator between `String` and `Vec<u8>` of the two different enum alternatives that could reach this code.

Besides, this code sucks. Searches the entire string and allocates a new `Vec`, and is not even shorter.

lvella: I can do this:

```rs
let next_dot = name.iter().position(|e| *e == b'.');
let name = &name[..next_dot.unwrap_or(name.len())];
```

leonardoalt: sure, just kinda still surprised there's nothing better

asm-utils/src/data_parser.rs
lvella: Indeed.

asm-utils/src/data_storage.rs
lvella: I didn't touch the lifetimes. It was already a lifetime annotated reference. I just changed to `str` because it is more general.

asm-utils/src/data_storage.rs
lvella: Because that is the size of a word.

asm-utils/src/data_storage.rs
leonardoalt: @chriseth do you know why this is aligning to 8 bytes?

asm-utils/src/reachability.rs
lvella: `data_sections` is a Vec. This is the Vec index of the current iteration of the `retain()` loop.

asm-utils/src/data_storage.rs
leonardoalt: that's for RISCV-32 though, maybe this library should be parametric over the native word size

asm-utils/src/data_parser.rs
lvella: not an `&str`

asm-utils/src/data_storage.rs
lvella: I think this is future proof. The calling code pass a hardcoded `starting_pos` value that is already aligned to 8 bytes.

asm-utils/src/data_storage.rs
lvella: This is more about powdr's word than RV32. And I don't see powdr's memory word being larger that 32 bits, otherwise it would break Goldilocks.

That said, making this struct generic is easy enough, but probably would depend on `num_traits` crate. There was another thing that I did not made generic over native ints because `asm-utils` did not on `num_traits`.

leonardoalt: That's fine, the only thing I think doesn't fit is that this is supposed to be "asm-utils" which would also work for WASM for example. But we can also go for WASM32 and everything still works. Fine to keep.

asm-utils/src/data_parser.rs
lvella: fixed

asm-utils/src/data_parser.rs
lvella: fixed

asm-utils/src/data_parser.rs
lvella: fixed

leonardoalt (approved):

asm-utils/src/data_storage.rs
lvella: Even for 64-bit processors, the memory word in the zkvm can't be bigger than the smallest field we support, so we would still use 4 byte words.

PR #980, opened 2024-01-29
In off(reg) syntax, makes register optinal, defaulting to x0.

lvella:


chriseth (dismissed):
Looks good, but can you add a one-liner comment?

lvella:
Like now?

chriseth (approved):

PR #996, opened 2024-01-31
Include Halo2 backend in install instructions

conor10:
At the moment the docs for newcomers are misleading, if they follow the Installation page and then navigate to the Hello World page, the SNARK generation for the minimal VM fails as the Halo2 backend is not built by default in the CLI module.

I've proposed that the feature flag is enabled for the Halo2 backend in the Installation page to prevent others from having this issue.

A note could alternatively be added to the Hello World page, but I'm thinking in terms of which place will ensure people aren't going to hit snags when they first try out powdr.

The simplest solution would just be to enable the Halo2 backend by default, but no doubt you've disabled it for a reason.

### DIFF
--- a/backend/src/pilstark/json_exporter/mod.rs
+++ b/backend/src/pilstark/json_exporter/mod.rs
@@ -372,7 +372,7 @@ impl<'a, T: FieldElement> Exporter<'a, T> {
 
 #[cfg(test)]
 mod test {
-    use powdr_pil_analyzer::analyze;
+    use powdr_pil_analyzer::analyze_file;
     use pretty_assertions::assert_eq;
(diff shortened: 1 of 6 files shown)


leonardoalt (approved):
Thanks! We've disabled it because its dependency tree is rather large and slows down compiler development and CI. I think it makes sense to add it to the docs.

leonardoalt:
Can you please rebase your branch on top of `origin/main`? We had a bug fix that's needed for CI.

leonardoalt:
Hi @conor10 a small correction: please rebase instead of merging our main into yours, otherwise we get this merge commit as part of the commits. Ideally there will be only one commit in this PR, the original one.

PR #1001, opened 2024-02-01
require semicolon in `instr and Y, Z -> X = binary.and;` declarations

pacheco:
Require semicolon on instr `=` declarations.
Added test that parses, writes and reparses all `pil|asm` files under `test_data`.
This revealed some issues with `Display` implementations that this PR also fixes

### DIFF
--- a/ast/src/parsed/display.rs
+++ b/ast/src/parsed/display.rs
@@ -37,9 +37,12 @@ impl<T: Display> Display for ModuleStatement<T> {
                     write!(f, "machine {name} {m}")
                 }
                 SymbolValue::Import(i) => {
-                    write!(f, "{i} as {name}")
+                    write!(f, "{i} as {name};")
                 }
-                SymbolValue::Module(m) => {
+                SymbolValue::Module(m @ Module::External(_)) => {
+                    write!(f, "mod {m}")
+                }
+                SymbolValue::Module(m @ Module::Local(_)) => {
                     write!(f, "mod {name} {m}")
                 }
                 SymbolValue::Expression(e) => {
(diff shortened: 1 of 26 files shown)


ast/src/lib.rs
```diff
@@ -77,32 +77,40 @@ where
...
+    by: usize,
+    line_end: Option<&str>,
+) -> Result
 where
     S: Display,
     I: IntoIterator<Item = S>,
     W: Write,
 {
     for item in items.into_iter() {
         write_indented_by(f, item, by)?;
```
chriseth, line 110: I never really liked this indentation mechanism. It forces every multi-line Display implementation to use the same mechanism and I have the impression that requiring this special line_end parameter is a symptom of that.
Also it does not work for nested structures.

Could you check if you could use the indentation mechanism used by Expression (i.e. just format to a string and then call `indent` on that string)?

chriseth: ah wait, this uses that mechanism under the hood. Ok, I think I need to take a closer look.

ast/src/parsed/display.rs
```diff
@@ -13,7 +13,7 @@ use super::{asm::*, *};
 
 impl<T: Display> Display for PILFile<T> {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
-        write_items(f, &self.0)
+        write_items(f, &self.0, Some(";"))
```
chriseth, line 16: I think we should add the `;` to the statement's fmt function

chriseth: Ok, I'm completely lost - why do we need this here?

ast/src/parsed/display.rs
```diff
@@ -119,7 +144,7 @@ impl<T: Display> Display for MachineStatement<T> {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
         match self {
             MachineStatement::Degree(_, degree) => write!(f, "degree {};", degree),
-            MachineStatement::Pil(_, statement) => write!(f, "{statement};"),
+            MachineStatement::Pil(_, statement) => write!(f, "{statement}"),
```
chriseth, line 147: Did this produce a duplicate `;` before?

parser/src/lib.rs
```diff
@@ -191,49 +192,60 @@ mod test {
...
-        parse_file("polygon-hermez/rom.pil");
-        parse_file("polygon-hermez/storage.pil");
+    /// Test that (source -> AST -> source -> AST) works properly
+    fn parse_write_reparse_asm() {
+        let crate_dir = env!("CARGO_MANIFEST_DIR");
+        let basedir = std::path::PathBuf::from(format!("{crate_dir}/../test_data/"));
+        let asm_files = find_files_with_ext(basedir, "asm".into());
+        for (file, orig_string) in asm_files {
+            let orig_asm =
+                parse_asm::<Bn254Field>(Some(&file), &orig_string).unwrap_err_to_stderr();
```
chriseth, line 220: Why not goldilocks?

parser/src/lib.rs
```diff
@@ -191,49 +192,60 @@ mod test {
...
+            let orig_asm_to_string = format!("{}", orig_asm);
+            let reparsed_asm =
+                parse_asm::<Bn254Field>(Some((file + " reparsed").as_ref()), &orig_asm_to_string)
+                    .unwrap_err_to_stderr();
+            // TODO: we can't directly assert ASTs because SourceRefs will differ
```
chriseth, line 225: I think we should still do the comparison, otherwise if we juts "lose" something during the ast to string conversion, this won't be detected.

ast/src/parsed/display.rs
pacheco: yep

parser/src/lib.rs
pacheco: some files have numbers that are too big for Goldilocks (e.g., `test_data/std/split_bn254_test.asm`)

parser/src/powdr.lalrpop
```diff
@@ -566,7 +566,6 @@ SpecialIdentifier: &'input str = {
 
 ConstantIdentifier: String = {
     // TODO it seems the lexer splits the token after %
-    "%N" => <>.to_string(),
```
chriseth, line 569: If this works, can you also remove the TODO?

Issue #890, opened 2024-01-09
Expose proof verification in CLI


leonardoalt:
Currently CLI does not allow for proof verification which needs to be done by using the Rust library directly. CLI should expose verifier generation and direct verification.

Differences in the backends:

- Halo2 verification requires the proof, verification key and params
- eStark requires the proof and verification key (`StarkSetup`). It also depends on `StarkStruct`, but that's quickly created from the degree

PR #1007, opened 2024-02-02
Export verification key and use it for proof

leonardoalt:
Fixes https://github.com/powdr-labs/powdr/issues/890

### DIFF
--- a/backend/src/halo2_impl.rs
+++ b/backend/src/halo2_impl.rs
@@ -1,6 +1,6 @@
 use std::io::{self};
 
-use crate::{BackendImpl, BackendImplWithSetup, Proof};
+use crate::{BackendImpl, BackendImplWithSetup, Error, Proof};
 use powdr_ast::analyzed::Analyzed;
 use powdr_halo2::Halo2Prover;
 use powdr_number::{DegreeType, FieldElement};
@@ -11,19 +11,49 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Prover {
         Halo2Prover::new(degree)
     }
 
+    fn add_verification_key(
+        &mut self,
+        pil: &Analyzed<T>,
+        fixed: &[(String, Vec<T>)],
+        vkey: Vec<u8>,
+    ) {
+        self.add_verification_key(pil, fixed, vkey)
+    }
+
+    fn verification_key(
+        &self,
+        pil: &Analyzed<T>,
+        fixed: &[(String, Vec<T>)],
+    ) -> Result<Vec<u8>, Error> {
+        match self.verification_key(pil, fixed) {
+            Ok(vkey) => Ok(vkey),
+            Err(e) => Err(Error::VerificationKeyFailed(e)),
+        }
+    }
+
+    fn verify(&self, proof: &Proof, instances: &[Vec<T>]) -> Result<(), Error> {
+        match self.verify(proof, instances) {
+            Ok(_) => Ok(()),
+            Err(e) => Err(Error::VerificationFailed(e)),
+        }
+    }
+
     fn prove(
         &self,
         pil: &Analyzed<T>,
         fixed: &[(String, Vec<T>)],
         witness: &[(String, Vec<T>)],
         prev_proof: Option<Proof>,
-    ) -> (Option<Proof>, Option<String>) {
+    ) -> Result<(Proof, Option<String>), Error> {
         let proof = match prev_proof {
             Some(proof) => self.prove_aggr(pil, fixed, witness, proof),
             None => self.prove_ast(pil, fixed, witness),
         };
 
-        (Some(proof), None)
+        match proof {
+            Ok(proof) => Ok((proof, None)),
+            Err(e) => Err(Error::ProofFailed(e.to_string())),
+        }
     }
 }
 
@@ -44,19 +74,40 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Mock {
         Self
     }
 
+    fn add_verification_key(
+        &mut self,
+        _pil: &Analyzed<T>,
+        _fixed: &[(String, Vec<T>)],
+        _vkey: Vec<u8>,
+    ) {
+        unimplemented!("Halo2Mock backend does not require verification key");
+    }
+
+    fn verification_key(
+        &self,
+        _pil: &Analyzed<T>,
+        _fixed: &[(String, Vec<T>)],
+    ) -> Result<Vec<u8>, Error> {
+        unimplemented!("Halo2Mock backend does not require verification key");
+    }
+
+    fn verify(&self, _proof: &Proof, _instances: &[Vec<T>]) -> Result<(), Error> {
+        unimplemented!("Halo2Mock backend does not yet support separate proof verification");
+    }
+
     fn prove(
         &self,
         pil: &Analyzed<T>,
         fixed: &[(String, Vec<T>)],
         witness: &[(String, Vec<T>)],
         prev_proof: Option<Proof>,
-    ) -> (Option<Proof>, Option<String>) {
+    ) -> Result<(Proof, Option<String>), Error> {
         if prev_proof.is_some() {
             unimplemented!("Halo2Mock backend does not support aggregation");
         }
 
         powdr_halo2::mock_prove(pil, fixed, witness);
 
-        (None, None)
+        Ok((vec![], None))
     }
 }
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -66,13 +66,34 @@ struct ConcreteBackendWithoutSetup<B>(B);
 
 /// Concrete implementation for backends with setup.
 impl<F: FieldElement, B: BackendImpl<F>> Backend<F> for ConcreteBackendWithoutSetup<B> {
+    fn add_verification_key(
+        &mut self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+        vkey: Vec<u8>,
+    ) {
+        self.0.add_verification_key(pil, fixed, vkey);
+    }
+
+    fn verification_key(
+        &self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+    ) -> Result<Vec<u8>, Error> {
+        self.0.verification_key(pil, fixed)
+    }
+
+    fn verify(&self, proof: &Proof, instances: &[Vec<F>]) -> Result<(), Error> {
+        self.0.verify(proof, instances)
+    }
+
     fn prove(
         &self,
         pil: &Analyzed<F>,
         fixed: &[(String, Vec<F>)],
         witness: &[(String, Vec<F>)],
         prev_proof: Option<Proof>,
-    ) -> (Option<Proof>, Option<String>) {
+    ) -> Result<(Proof, Option<String>), Error> {
         self.0.prove(pil, fixed, witness, prev_proof)
     }
 
@@ -104,13 +125,34 @@ struct ConcreteBackendWithSetup<B>(B);
 
 /// Concrete implementation for backends with setup.
 impl<F: FieldElement, B: BackendImplWithSetup<F>> Backend<F> for ConcreteBackendWithSetup<B> {
+    fn add_verification_key(
+        &mut self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+        vkey: Vec<u8>,
+    ) {
+        self.0.add_verification_key(pil, fixed, vkey);
+    }
+
+    fn verification_key(
+        &self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+    ) -> Result<Vec<u8>, Error> {
+        self.0.verification_key(pil, fixed)
+    }
+
+    fn verify(&self, proof: &Proof, instances: &[Vec<F>]) -> Result<(), Error> {
+        self.0.verify(proof, instances)
+    }
+
     fn prove(
         &self,
         pil: &Analyzed<F>,
         fixed: &[(String, Vec<F>)],
         witness: &[(String, Vec<F>)],
         prev_proof: Option<Proof>,
-    ) -> (Option<Proof>, Option<String>) {
+    ) -> Result<(Proof, Option<String>), Error> {
         self.0.prove(pil, fixed, witness, prev_proof)
     }
 
@@ -123,8 +165,16 @@ impl<F: FieldElement, B: BackendImplWithSetup<F>> Backend<F> for ConcreteBackend
 pub enum Error {
     #[error("input/output error")]
     IO(#[from] std::io::Error),
-    #[error("the backend has not setup operations")]
+    #[error("the witness is empty")]
+    EmptyWitness,
+    #[error("the backend has no setup operations")]
     NoSetupAvailable,
+    #[error("proof generation failed")]
+    ProofFailed(String),
+    #[error("proof verification failed")]
+    VerificationFailed(String),
+    #[error("verification key generation failed")]
+    VerificationKeyFailed(String),
 }
 
 pub type Proof = Vec<u8>;
@@ -136,6 +186,21 @@ pub type Proof = Vec<u8>;
 
 /// Dynamic interface for a backend.
 pub trait Backend<F: FieldElement> {
+    fn add_verification_key(
+        &mut self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+        vkey: Vec<u8>,
+    );
+
+    fn verification_key(
+        &self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+    ) -> Result<Vec<u8>, Error>;
+
+    fn verify(&self, proof: &Proof, instances: &[Vec<F>]) -> Result<(), Error>;
+
     /// Perform the proving.
     ///
     /// If prev_proof is provided, proof aggregation is performed.
@@ -148,7 +213,7 @@ pub trait Backend<F: FieldElement> {
         fixed: &[(String, Vec<F>)],
         witness: &[(String, Vec<F>)],
         prev_proof: Option<Proof>,
-    ) -> (Option<Proof>, Option<String>);
+    ) -> Result<(Proof, Option<String>), Error>;
 
     /// Write the prover setup to a file, so that it can be loaded later.
     fn write_setup(&self, output: &mut dyn io::Write) -> Result<(), Error>;
@@ -171,13 +236,28 @@ pub trait BackendFactory<F: FieldElement> {
 trait BackendImpl<F: FieldElement> {
     fn new(degree: DegreeType) -> Self;
 
+    fn add_verification_key(
+        &mut self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+        vkey: Vec<u8>,
+    );
+
+    fn verification_key(
+        &self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+    ) -> Result<Vec<u8>, Error>;
+
+    fn verify(&self, proof: &Proof, instances: &[Vec<F>]) -> Result<(), Error>;
+
     fn prove(
         &self,
         pil: &Analyzed<F>,
         fixed: &[(String, Vec<F>)],
         witness: &[(String, Vec<F>)],
         prev_proof: Option<Proof>,
-    ) -> (Option<Proof>, Option<String>);
+    ) -> Result<(Proof, Option<String>), Error>;
 }
 
 /// Trait implemented by backends that have a setup phase that must be saved to
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -12,7 +12,7 @@ use powdr_number::{Bn254Field, FieldElement, GoldilocksField};
 use powdr_pipeline::{Pipeline, Stage};
 use powdr_riscv::continuations::{rust_continuations, rust_continuations_dry_run};
 use powdr_riscv::{compile_riscv_asm, compile_rust};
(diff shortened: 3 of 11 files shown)


backend/src/lib.rs
```diff
@@ -136,6 +180,22 @@ pub type Proof = Vec<u8>;
 
 /// Dynamic interface for a backend.
 pub trait Backend<F: FieldElement> {
+    fn add_verification_key(
+        &mut self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+        vkey: Vec<u8>,
+    );
+
```
leonardoalt, line 189: Ideally I wanted to add this function as `with_verification_key(self, ...) -> Self` but that won't compile because of the dynamic dispatcher.

backend/src/halo2_impl.rs
```diff
@@ -11,6 +11,28 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Prover {
...
+
+    fn verify(
+        &self,
+        proof: &Proof,
+        verification_key: &[u8],
+        instances: &[Vec<T>],
+    ) -> Result<bool, String> {
+        self.verify(proof, verification_key, instances)
+    }
+
```
leonardoalt, line 35: We should really rethink how our backend is implemented. There's so much repetition. These functions are implemented 4 times in a row just calling each other. `Backend -> BackendImpl -> ConcreteBackend -> Halo2Impl -> Halo2`.

halo2/src/prover.rs
```diff
@@ -38,20 +42,38 @@ pub use halo2_proofs::poly::kzg::commitment::ParamsKZG;
...
+        &mut self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+        vkey: Vec<u8>,
+    ) {
+        let plaf = analyzed_to_plaf(pil, fixed);
+        let vkey: VerifyingKey<G1Affine> = VerifyingKey::<G1Affine>::from_bytes::<PlafH2Circuit>(
+            &vkey,
+            SerdeFormat::Processed,
+            plaf,
```
leonardoalt, line 73: @georgwiese `Default::default` actually doesn't work here, otherwise it complains that the circuit is empty (no constraints). So we need to pass the same circuit that we pass for the proof.
This causes this function to also require `pil` and `fixed`, because we need those to compute the `plaf` circuit.

halo2/src/prover.rs
```diff
@@ -74,13 +96,16 @@ impl Halo2Prover {
...
+        let (circuit, publics) =
+            analyzed_to_circuit_with_witness(pil, plaf_circuit.plaf, circuit_data, witness);
 
         log::debug!("{}", PlafDisplayBaseTOML(&circuit.plaf));
 
         log::info!("Generating VK and PK for snark...");
-        let vk = keygen_vk(&params, &circuit).unwrap();
-        let pk = keygen_pk(&params, vk, &circuit).unwrap();
+        //let vk = keygen_vk(&params, &circuit).unwrap();
+        let pk = keygen_pk(&params, self.vkey.clone().unwrap(), &circuit).unwrap();
```
leonardoalt, line 107: This is one of the main hacks that I have no idea how to solve.
- we want to pass the verification key to the prover integration so that it can use a previously generated one and not need to compute on the fly
- Halo2 wants the `vk` by value in this function
- so we need to either `clone` or `take` it, not sure which is better

pipeline/src/pipeline.rs
```diff
@@ -390,6 +397,33 @@ impl<T: FieldElement> Pipeline<T> {
         }
     }
 
+    /// Reads previously generated constants from the provided directory and
+    /// advances the pipeline to the `PilWithEvaluatedFixedCols` stage.
+    pub fn read_constants(mut self, directory: &Path) -> Self {
```
leonardoalt, line 402: This is similar to `read_witness...` below.

halo2/src/circuit_builder.rs
```diff
@@ -17,12 +19,10 @@ use powdr_number::{BigInt, FieldElement};
 
 use super::circuit_data::CircuitData;
 
-/// Converts an analyzed PIL, fixed and witness columns to a PlafH2Circuit and publics for the halo2 backend.
-pub(crate) fn analyzed_to_circuit<T: FieldElement>(
+pub(crate) fn analyzed_to_plaf<T: FieldElement>(
```
leonardoalt, line 22: This file is quite hacky right now. Before, there was only one version of this function, `analyzed_to_circuit`, which took `pil`, `fixed` and `witness` and returned the `PlafH2Circuit` needed for a Halo2 proof, and the publics. This is definitely not needed for the vkey.

Right now there are 3 versions of this function here:

- `analyzed_to_plaf`: Takes (`pil`, `fixed`) and returns `Plaf`, needed to deserialize a vkey
- `analyzed_to_circuit`: Takes (`pil`, `fixed`) and returns `PlafH2Circuit`, needed to compute a vkey
- `analyzed_to_circuit_with_witness`: Takes (`pil`, `fixed`, `witness`) and returns (`PlafH2Circuit`, `publics`), needed for the proof

I still need to de-dup and cleanup a lot of stuff here.

backend/src/lib.rs
lvella: Did you mean `with_verification_key(...) -> Self` ? There is a factory class to create backends, and it has `create(&self, degree: DegreeType)` and `create_from_setup(&self, input: &mut dyn io::Read)`.

We could add `fn create_with_verification_key(...)` and `fn create_with_setup_and_verification_key(...)`, but that is too combinatoric for my taste.

I propose instead just having one `fn create(...)`, taking as `Option<...>` the setup and the verification key.

backend/src/halo2_impl.rs
leonardoalt: To be honest I still don't even understand what's what in `lib.rs` because there are so many traits and impls. I don't think `BackendImplWithVerification` is needed because all proof systems have a verification key and verification, so it's just like `prove`.

backend/src/lib.rs
leonardoalt: > but that is too combinatoric for my taste.

yes that's also what I thought

> Did you mean with_verification_key(...) -> Self

No I meant an object that's already created (with `create` or `creat_from_setup` and you can just append the verification key

backend/src/lib.rs
```diff
@@ -171,6 +231,22 @@ pub trait BackendFactory<F: FieldElement> {
...
+    fn verification_key(&self, pil: &Analyzed<F>, fixed: &[(String, Vec<F>)]) -> Vec<u8>;
+
+    fn verify(
+        &self,
+        proof: &Proof,
+        verification_key: &[u8],
+        instances: &[Vec<F>],
+    ) -> Result<bool, String>;
+
```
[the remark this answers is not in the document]
leonardoalt: All backends have verification. I think it's fine to assume that prove <=> verify and keep it all in the same trait.

halo2/src/circuit_builder.rs
leonardoalt: Ok this is a bit cleaner now.

backend/src/halo2_impl.rs
```diff
@@ -11,6 +11,28 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Prover {
         Halo2Prover::new(degree)
     }
 
+    fn add_verification_key(
+        &mut self,
+        pil: &Analyzed<T>,
+        fixed: &[(String, Vec<T>)],
+        vkey: Vec<u8>,
```
lvella, line 18: This argument seems read from a file. Passing it as a `Vec<u8>` seems inconsistent with the setup argument, that receives a `&mut dyn io::Read`. Not sure if it is comparable or should be uniformized.

leonardoalt: I tried it before as `&mut dyn io::Read` but it wouldn't compile, because Halo2 only reads from `impl io::Read` not dyn

backend/src/halo2_impl.rs
```diff
@@ -11,6 +11,28 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Prover {
         Halo2Prover::new(degree)
     }
 
+    fn add_verification_key(
+        &mut self,
+        pil: &Analyzed<T>,
+        fixed: &[(String, Vec<T>)],
```
lvella, line 17: These two arguments probably should be given in the object's creation.

leonardoalt: I'm not sure. What if you still want to generate the verification key? For that you need more than just easy data like the size, field and backend type. You also need the optimized pil and fixed columns. The easiest way to get these is via Pipeline.

lvella: I mean that they are used both in proving and verifying. Passing them on create, you can create the Halo2 `Plaf` once and store in the object, then it will be used for proving and verifying.

halo2/src/circuit_builder.rs
```diff
@@ -171,6 +176,115 @@ pub(crate) fn analyzed_to_circuit<T: FieldElement>(
...
+        .iter()
+        .map(|(_, row)| {
+            row.iter()
+                .map(|value| Some(value.to_arbitrary_integer()))
+                .collect()
+        })
+        .collect();
+
+    let mut copies = vec![];
+    let mut publics = vec![];
```
georgwiese, line 201: This is is unused.

Also, this block seems to be copied, maybe we can extract a function?

halo2/src/circuit_builder.rs
```diff
@@ -251,31 +339,24 @@ pub(crate) fn analyzed_to_circuit<T: FieldElement>(
...
     let plaf = Plaf {
-        info,
-        columns,
-        polys,
-        metadata: Default::default(),
-        lookups,
-        shuffles,
         copys: copies,
-        fixed,
+        ..plaf
```
georgwiese, line 359: I think this doesn't change anything, the `plaf` already has the copy constraints, no? That's good, then we can get rid of the code duplication :) We just only still need to compute the publics...

backend/src/lib.rs
```diff
@@ -136,6 +180,22 @@ pub type Proof = Vec<u8>;
...
+    );
+
+    fn verification_key(&self, pil: &Analyzed<F>, fixed: &[(String, Vec<F>)]) -> Vec<u8>;
+
+    fn verify(
+        &self,
+        proof: &Proof,
+        verification_key: &[u8],
+        instances: &[Vec<F>],
+    ) -> Result<bool, String>;
```
lvella, line 197: Use the `Error` type, not a string to return error.

halo2/src/prover.rs
georgwiese: Oh OK. I think that makes sense. The circuit params I think are needed to be able to specify the constraints at runtime, and `Plaf` contains all information needed to do that,

halo2/src/prover.rs
georgwiese: I think cloning should be fine? It's supposed to be small, right?

halo2/src/prover.rs
```diff
@@ -179,6 +207,83 @@ impl Halo2Prover {
...
+    pub fn verification_key<F: FieldElement>(
+        &self,
+        pil: &Analyzed<F>,
+        fixed: &[(String, Vec<F>)],
+    ) -> Vec<u8> {
+        let vk = self.verification_key_inner(pil, fixed);
+        vk.to_bytes(SerdeFormat::Processed)
+    }
+
+    pub fn verify_inner<
```
georgwiese, line 230: I think I would just inline these `_inner` functions, but might be a taste thing. At least it shouldn't be public, right?

georgwiese (commented):
Cool! That all makes sense to me, just a few comments.

The CLI seems to be buggy still though. For example, this fails now:
`cargo run --features halo2 -r pil test_data/pil/fibonacci.pil -o output -f --field bn254 --prove-with halo2`

I haven't yet managed to create a proof and verify it.

halo2/src/prover.rs
leonardoalt: You mean Rust `inline` or replace the code? The reason why it's a function is that it's used both at proof generation and verification times. But yes, shouldn't be public

halo2/src/prover.rs
leonardoalt: that's true, I was thinking in the order of magnitude of the setup

cli/src/main.rs
```diff
@@ -281,7 +281,61 @@ enum Commands {
...
+
+    Verify {
+        /// Input PIL file
+        file: String,
+
+        /// The field to use
+        #[arg(long)]
+        #[arg(default_value_t = FieldArgument::Gl)]
+        #[arg(value_parser = clap_enum_variants!(FieldArgument))]
+        field: FieldArgument,
```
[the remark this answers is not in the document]
leonardoalt: yea agre...

cli/src/main.rs
```diff
@@ -281,7 +281,61 @@ enum Commands {
         #[arg(long)]
         proof: Option<String>,
 
+        /// File containing previously generated verification key.
```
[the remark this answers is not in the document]
leonardoalt: It is... because halo2... it requires the vk to create the pk

cli/src/main.rs
```diff
@@ -699,18 +811,53 @@ fn read_and_prove<T: FieldElement>(
...
     Ok(())
 }
 
+fn read_and_verify<T: FieldElement>(
+    _file: &Path,
+    _backend_type: &BackendType,
+    _proof_path: &Path,
+    _verification_key_path: &Path,
+) -> Result<(), Vec<String>> {
+    /*
```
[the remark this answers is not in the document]
leonardoalt: I updated this PR with that too

halo2/src/circuit_builder.rs
leonardoalt: will check

pipeline/src/pipeline.rs
```diff
@@ -817,4 +857,24 @@ impl<T: FieldElement> Pipeline<T> {
     pub fn data_callback(&self) -> Option<&dyn QueryCallback<T>> {
         self.arguments.query_callback.as_deref()
     }
+
+    pub fn verification_key(&self) -> Vec<u8> {
+        match self.artifact.as_ref().unwrap() {
```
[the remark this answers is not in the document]
leonardoalt: > Hmm, I would think computing the verification key should be a pipeline step, no?

I thought about doing that too but didn't want it to be too invasive in Pipeline. Will check your suggestion.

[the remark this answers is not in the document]
leonardoalt: Yea I see. Let's keep this PR as is then, and think about how to modify the pipeline later.

cli/src/main.rs
```diff
@@ -281,7 +281,61 @@ enum Commands {
...
+        /// File containing the proof.
+        #[arg(long)]
+        proof: String,
+
+        /// File containing the verification ley.
+        #[arg(long)]
+        verification_key: String,
+    },
+
+    VerificationKey {
```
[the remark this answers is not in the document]
leonardoalt: Not sure I agree with this one. Setup is also only called setup, and this would cause the subcommand to be `create-verification-key`

halo2/src/circuit_builder.rs
leonardoalt: After simplifying this the `analyzed_to_*` API also looks much better!

leonardoalt:
This is ready for review again. I think I handled all comments for now.

backend/src/lib.rs
lvella: In the end, do you want the object to be created with the verification key or not? Why not just pass it with `verify()`?

halo2/src/circuit_data.rs
```diff
@@ -78,7 +71,11 @@ impl<'a, T: FieldElement> CircuitData<'a, T> {
...
         let values = values.into_iter().collect::<Vec<_>>();
-        assert_eq!(values.len(), self.len());
+
+        if !self.fixed.is_empty() {
+            assert_eq!(values.len(), self.len());
+        }
+
         self.fixed.push((name.to_string(), values));
         let column = Column {
             kind: ColumnKind::Fixed,
```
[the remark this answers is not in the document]
georgwiese: AIs nitpicking our PR is how AI Armageddon starts

backend/src/lib.rs
leonardoalt: I don't have a strong opinion about it. I'd be fine without it being created with it. The reason not to pass it with `verify` is that we also need `params` in verify. So I found it easier to already have everything in the object.

backend/src/halo2_impl.rs
```diff
@@ -11,19 +11,49 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Prover {
...
+    }
+
     fn prove(
         &self,
         pil: &Analyzed<T>,
         fixed: &[(String, Vec<T>)],
         witness: &[(String, Vec<T>)],
         prev_proof: Option<Proof>,
-    ) -> (Option<Proof>, Option<String>) {
+    ) -> Result<(Option<Proof>, Option<String>), Error> {
```
lvella, line 47: I don't think we need the first `Option` anymore. Either we have a proof or there was an error. There is no scenario there is no error and no proof.

backend/src/halo2_impl.rs
```diff
@@ -44,19 +74,40 @@ impl<T: FieldElement> BackendImpl<T> for Halo2Mock {
...
         if prev_proof.is_some() {
             unimplemented!("Halo2Mock backend does not support aggregation");
         }
 
         powdr_halo2::mock_prove(pil, fixed, witness);
 
-        (None, None)
+        Ok((None, None))
```
lvella, line 111: Ok, there is this case, but shouldn't be. Just return an empty proof.

backend/src/halo2_impl.rs
leonardoalt: Done

cli/src/main.rs
```diff
@@ -699,18 +817,55 @@ fn read_and_prove<T: FieldElement>(
...
+    let mut pipeline = Pipeline::<T>::default()
+        .from_file(file.to_path_buf())
+        .read_constants(dir)
+        .with_setup_file(params.map(PathBuf::from))
+        .with_vkey_file(Some(vkey))
+        .with_backend(*backend_type);
+
+    // TODO add support for publics
+    pipeline.verify(proof, &[vec![]])?;
+    println!("Proof is valid!");
```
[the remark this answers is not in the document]
leonardoalt: The reason was to always print it. INFO log is quite verbose already, and I thought the user would always want to know whether the proof is valid or not, also without INFO log.

lvella:
On the backend side, look as good as this PR will ever be. I'll leave to @georgwiese to give the final approval.

halo2/src/circuit_builder.rs
```diff
@@ -75,11 +84,7 @@ pub(crate) fn analyzed_to_circuit<T: FieldElement>(
             .iter()
             .map(|(name, _)| ColumnFixed::new(name.to_string()))
             .collect(),
-        witness: cd
-            .witness
-            .iter()
-            .map(|(name, _)| ColumnWitness::new(name.to_string(), 0))
-            .collect(),
+        witness: wit_columns.clone(),
```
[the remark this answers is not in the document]
leonardoalt: Surprised clippy didn't catch these

halo2/src/circuit_builder.rs
```diff
@@ -251,36 +325,25 @@ pub(crate) fn analyzed_to_circuit<T: FieldElement>(
         })
         .collect();
 
-    let witness_cols = cd
-        .witness
+    let witness_cols = witness
         .iter()
         .enumerate()
         .map(|(n, (name, _))| (name.to_string(), (ColumnKind::Fixed, n)));
```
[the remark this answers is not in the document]
leonardoalt: ah yea, this is leftover from when I was splitting the original `analyzed_to_circuit` function into 4 different functions.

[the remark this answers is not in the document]
leonardoalt: ah no @georgwiese , this was weird but fine. It existed only to collect the names of witness columns, and it was re-used in different places. I'm gonna fix it here.

pipeline/src/pipeline.rs
```diff
@@ -264,6 +266,7 @@ impl<T: FieldElement> Pipeline<T> {
     pub fn with_tmp_output(self, tmp_dir: &mktemp::Temp) -> Self {
         Pipeline {
             output_dir: Some(tmp_dir.to_path_buf()),
+            force_overwrite: true,
```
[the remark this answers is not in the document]
leonardoalt: yea but in the new test I'm running the pipeline twice from constant pols, which causes the witness to be generated twice. Without this the second one fails. I can also generate a new entry point just for `force_overwrite` since that doesn't exist now, but this felt easier.

pipeline/src/test_util.rs
```diff
@@ -71,13 +73,50 @@ pub fn gen_estark_proof(file_name: &str, inputs: Vec<GoldilocksField>) {
...
     let tmp_dir = mktemp::Temp::new_dir().unwrap();
-    Pipeline::default()
+    let mut pipeline = Pipeline::default()
         .with_tmp_output(&tmp_dir)
         .from_file(PathBuf::from(file_name))
         .with_prover_inputs(inputs)
-        .with_backend(powdr_backend::BackendType::Halo2)
-        .proof()
-        .unwrap();
+        .with_backend(powdr_backend::BackendType::Halo2);
```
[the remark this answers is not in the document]
leonardoalt: This is to test that the proof is valid in a separate step. Before we tested that the proof is generated, and setup and vkey were created during proof time. Now we test 3 paths:

- same as before, collect the proof
- verify the proof using the setup and vkey that were generated separately for the same circuit
- same as first, but using the generated setup and vkey, instead of generating them on the fly

georgwiese (approved):
Let's go 🚀

PR #1010, opened 2024-02-02
Type inference

chriseth:
TODO:

 - [x] documentation
 - [x] add inferred types to declarations.


executor/src/constant_evaluator/mod.rs
```diff
@@ -432,76 +453,68 @@ mod test {
...
-            col fixed b = [0, 0, 1, 1, 0, 5];
-            col fixed or(i) { if (a(i) != std::convert::fe(0)) || (b(i) != std::convert::fe(0)) { 1 } else { 0 } };
-            col fixed and(i) { if (a(i) != std::convert::fe(0)) && (b(i) != std::convert::fe(0)) { 1 } else { 0 } };
-            col fixed not(i) { if !(a(i) != std::convert::fe(0)) { 1 } else { 0 } };
-            col fixed less(i) { if std::convert::int(id(i)) < std::convert::int(inv(i)) { 1 } else { 0 } };
-            col fixed less_eq(i) { if std::convert::int(id(i)) <= std::convert::int(inv(i)) { 1 } else { 0 } };
+            namespace F(N);
+            let id = |i| i;
+            let inv = |i| N - i;
+            let a: int -> int = |i| [0, 1, 0, 1, 2, 1][i];
```
pacheco, line 463: just curious: what happens if the array is less than the degree here? when is it caught?

executor/src/constant_evaluator/mod.rs
```diff
@@ -432,76 +453,68 @@ mod test {
...
+            col fixed and(i) { if (a(i) != 0) && (b(i) != 0) { 1 } else { 0 } };
+            col fixed not(i) { if !(a(i) != 0) { 1 } else { 0 } };
+            col fixed less(i) { if id(i) < inv(i) { 1 } else { 0 } };
+            col fixed less_eq(i) { if id(i) <= inv(i) { 1 } else { 0 } };
             col fixed eq(i) { if id(i) == inv(i) { 1 } else { 0 } };
             col fixed not_eq(i) { if id(i) != inv(i) { 1 } else { 0 } };
-            col fixed greater(i) { if std::convert::int(id(i)) > std::convert::int(inv(i)) { 1 } else { 0 } };
-            col fixed greater_eq(i) { if std::convert::int(id(i)) >= std::convert::int(inv(i)) { 1 } else { 0 } };
+            col fixed greater(i) { if id(i) > inv(i) { 1 } else { 0 } };
+            col fixed greater_eq(i) { if id(i) >= inv(i) { 1 } else { 0 } };
```
pacheco, line 473: nice!

executor/src/constant_evaluator/mod.rs
```diff
@@ -555,19 +568,21 @@ mod test {
         let src = r#"
             constant %N = 4;
             namespace F(%N);
-            let x: col = |i| y(i) + 1;
-            let y: col = |i| i + 20;
+            let x = |i| y(i) + 1;
+            let y = |i| i + 20;
+            let X: col = x;
```
pacheco, line 573: so this is building a col from a generic fn that becomes concrete on `fe`?

executor/src/constant_evaluator/mod.rs
chriseth: You mean what happens if someone evaluates `a(7)`? Then we get an "array out of bounds" runtime error in the evaluator.

executor/src/constant_evaluator/mod.rs
chriseth: No, symbols are only generic if they are explicitly generic, i.e. `let<T> x: T -> T = ...`. If you do not specify a type, then a unique concrete type must be inferred from the way the symbol is used and from its definition. If the type is not unique (mostly meaning it is generic), the type checker fails.

What happens here is that the type of the value of a column can either be `int -> fe` or `int -> int` ('int -> fe' is preferred, if that fails, the type checker tries `int -> int`). Since the type of `y` is at least `T -> T`, the type of `x` cannot be `int -> fe` and thus the types of both `x` and `y` are inferred to be `int -> int`.

A reference to a column is `expr`. I think these simple rules allow a quite intuitive use without the need for implicit conversions everywhere. In the end, there is no real implicit conversion happening anyway.

executor/src/constant_evaluator/mod.rs
```diff
@@ -140,16 +143,30 @@ struct Symbols<'a, T> {
...
                     Some(FunctionValueDefinition::Expression(TypedExpression {
                         e,
-                        type_scheme: _,
-                    })) => evaluator::evaluate(e, self)?,
+                        type_scheme,
+                    })) => {
+                        let generic_args = type_scheme
+                            .as_ref()
+                            .map(|type_scheme| {
+                                generic_arg_mapping(type_scheme, generic_args.iter())
```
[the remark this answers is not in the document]
chriseth: I mean the problem is that we have an `Option<TypeSheme>` here. But I can change the `type_scheme` parameter of `generic_arg_mapping` to be an `Option`.

chriseth:
Here is the issue that contains items for each TODO in this PR: https://github.com/powdr-labs/powdr/issues/1106

Will write the documentation next.

chriseth:
Ok it seems we have to implement some of the optimization ideas here before we can merge this :)

pil-analyzer/src/call_graph.rs
```diff
@@ -0,0 +1,64 @@
...
+    graph: &'b HashMap<&'a str, HashSet<String>>,
+    visited: &'b mut HashSet<&'a str>,
+    result: &'b mut Vec<String>,
+) {
+    if !visited.insert(name) {
+        return;
+    }
+    if let Some(called) = graph.get(name) {
+        for c in called {
+            let n = graph.get_key_value(c.as_str()).unwrap().0;
```
pacheco, line 39: can't `c.as_str()` be used directly in place of `n`?

chriseth: No, this is a lifetime transfer thing....

chriseth:
I got tremendous speed up by avoiding creating a type variable for each literal number. This keeps the substitution hashmap small and I think this is sufficient for now.

pil-analyzer/src/evaluator.rs
```diff
@@ -515,18 +568,41 @@ mod internal {
...
+                )))?,
+            }
+        } else {
+            // TODO Default is to convert literals to integers.
+            // We need to change the parser here to parse integers, not field elements,
+            // so that we can process larger numbers.
+            ty.as_ref().cloned().unwrap_or_else(|| TypeName::Int)
+        };
+        Ok(match ty {
+            TypeName::Fe => Value::FieldElement(n.to_arbitrary_integer().into()),
```
pacheco, line 589: isn't `n` already a field element?

pil-analyzer/src/evaluator.rs
```diff
@@ -695,12 +776,12 @@ mod test {
...
     pub fn fibonacci() {
         let src = r#"namespace Main(16);
-            let fib = |i| match i {
+            let fib: int -> int = |i| match i {
                 0 => 0,
                 1 => 1,
                 _ => fib(i - 1) + fib(i - 2),
             };
-            let result = fib(20);
+            let result: int = fib(20);
```
pacheco, line 784: can't this `int` be inferred from `fib`s type?

executor/src/constant_evaluator/mod.rs
chriseth: Yes

executor/src/constant_evaluator/mod.rs
```diff
@@ -616,8 +626,10 @@ mod test {
     pub fn arrays_of_fixed() {
         let src = r#"
             namespace F(4);
-                let x: int -> col = |k| |i| i + k;
+                let x: fe -> (int -> fe) = |k| |i| std::convert::fe(i) + k;
                 let y: col[2] = [x(0), x(1)];
+            namespace std::convert(4);
+                let fe = || fe();
```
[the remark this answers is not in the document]
chriseth: `std::convert::fe` is a built-in symbol. We still have to declare it (because I don't think it's a good idea to also clutter name resolution with builtins), but the actual semantics is replaced during type-checking and execution.
We can maybe find a better solution at some point.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -126,6 +130,82 @@ impl<T: FieldElement> PILAnalyzer<T> {
...
+    pub fn type_check(&mut self) {
+        let mut expressions = vec![];
+        // Collect all definitions with their types and expressions.
+        // For Arrays, we also collect the inner expressions and expect them to be field elements.
+        let definitions = self
+            .definitions
+            .iter_mut()
+            .map(|(name, (symbol, value))| {
+                (
+                    name.clone(),
```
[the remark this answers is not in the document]
chriseth: Changed it a bit.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -594,4 +680,20 @@ namespace N(16);
         let formatted = analyze_string::<GoldilocksField>(input).to_string();
         assert_eq!(formatted, input);
     }
+
+    #[test]
+    #[should_panic = "Set of declared and used type variables are not the same"]
```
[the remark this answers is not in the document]
chriseth: Extended it a little and added another test.

book/src/pil/declarations.md
```diff
@@ -7,18 +7,21 @@ from the type of the symbol and the way the symbol is used.
...
 
-- Symbols without a value are witness columns or arrays of witness columns. Their type can be omitted. If it is given, it must be ``col`` or ``col[]``.
-- Symbols evaluating to a number or with type ``fe`` are constants.
-- Symbols without type but with a value that is a function with a single parameter are fixed columns.
-- Symbols defined with a value and type ``col`` (or ``col[]``) are fixed columns (or arrays of fixed columns).
+- Symbols without a value are witness columns or arrays of witness columns. Their type can be omitted. If it is given, it must be ``col`` or ``col[k]``.
+- Symbols with type ``fe`` are constants.
+- Symbols defined with a value and type ``col`` (or ``col[k]``) are fixed columns (or arrays of fixed columns).
+- Symbols defined with a value and type ``expr`` (or ``expr[k]``) are intermediate columns (or arrays of intermediate columns).
 - Everything else is a "generic symbol" that is not a column or constant.
```
pacheco, line 21: I don't get the distinction between a generic symbol and a constant. Isn't every definition constant?

book/src/pil/declarations.md
```diff
@@ -7,18 +7,21 @@ from the type of the symbol and the way the symbol is used.
...
 
-> Note that the type ``col`` is the same as ``int -> fe``, so ``let w: int -> fe`` also declares a witness column.
+> Note that a symbol with declared type ``int -> fe`` is not considered a column.
```
pacheco, line 23: I think stating this is unnecessary, and maybe even slightly confusing, without the context that this used to be the case.

book/src/pil/declarations.md
```diff
@@ -7,18 +7,21 @@ from the type of the symbol and the way the symbol is used.
...
+> Its type has to be ``col`` (or ``col[k]``). See [Types](./types.md) for details.
```
pacheco, line 24: and this line needs to be removed/reworded without the line above it

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
 - tuple
 - array
 - function type
-- expression
-- constraint
+- `expr` (expression)
+- `constr` (constraint)
+- `!` ("bottom" or "unreachable" type)
+
+> The `col` type is special in that it is not a proper type, but can be used for declaring symbols.
```
pacheco, line 16: I'm a bit confused: what's a proper type? Isn't `col` just a type with special rules for how values can be declared (from `int->fe` and `int->fe`)?

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
...
+> See [Types of Symbols](#types-of-symbols) for details.
+
+Powdr-pil does perform a Hindley-Milner-kind type checking. After type checking, for all symbols
```
pacheco, line 19: "Powdr-pil has type inference (Hindley-Milner)."

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
...
+without an explicit type, their type has to be uniquely determined. If you want a symbol to have
+a generic type, that generic type has to be explicitly specified.
+
+Declarations of generic symbols can constraint type variables to satisfy certain bounds.
```
pacheco, line 23: Maybe clarify `generic symbol`/`generic type` a bit? A `generic symbol` refers to a symbol with a generic type right?

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
...
+Currently, only built-in bounds or traits are supported (see the next section).
```
pacheco, line 24: is bound something other than a trait here?

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
...
+## Types of Symbols
+
+A symbol declared to have type `col` (or `col[k]`) is a bit special:
+
+If you assign it a value, that value is expected to have type `int -> fe` or `int -> int` (or an array thereof).
+This allows the simple declaration of a column `let byte = |i| i & 0xff;` without complicated conversions.
+The integer value is converted to a field element during evaluation, but it has to be non-negative and less than
+the field modulus.
+
+If you reference such a symbol, the type of the reference is `expr`, which means it is only symbolically referenced.
```
pacheco, line 49: what does symbolically referenced mean?

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
...
+The following snippet defines a function that takes a value of a generic type and returns the value incremented by one.
+The type bounds on the generic type are `FromLiteral` and `Add`. The type checker will complain if we do not specify
+the type bounds. The bound `Add` is required because we use the `+` operator in the function and `FromLiteral` is needed
+because we use the literal `1` as a value of that type.
+
+```rust
+let<T: FromLiteral + Add> add_one: T -> T = |i| i + 1;
+```
+
+## Types of Symbols
```
pacheco, line 40: Maybe call this section `Declaring and referencing columns` or sth in those lines?

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
...
+
+A symbol declared to have type `col` (or `col[k]`) is a bit special:
+
+If you assign it a value, that value is expected to have type `int -> fe` or `int -> int` (or an array thereof).
+This allows the simple declaration of a column `let byte = |i| i & 0xff;` without complicated conversions.
```
pacheco, line 45: don't you need the explicit `: col` for columns?

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
...
+The integer value is converted to a field element during evaluation, but it has to be non-negative and less than
+the field modulus.
+
+If you reference such a symbol, the type of the reference is `expr`, which means it is only symbolically referenced.
+A byte constraint is as easy as `{ X } in { byte }` now, since the expected types in plookup columns is `expr`.
```
pacheco, line 50: remove `now`?

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
...
+The downside is that you cannot evaluate columns as functions. If you want to do that, you have to assign
+a copy to an `int -> int` symbol: `let byte_f = |i| i & 0xff; let byte = byte_f;`.
```
pacheco, line 52: shouldn't `byte_f` and `byte` be explicitly typed?

test_data/pil/sum_via_witness_query.pil
```diff
@@ -1,10 +1,10 @@
...
-namespace Sum(%N);
-    constant %last_row = %N - 1;
+namespace Sum(N);
+    let last_row = N - 1;
 
-    pol fixed ISLAST(i) { if i == %last_row { 1 } else { 0 } };
-    pol fixed ISALMOSTLAST(i) { if i == %last_row - 1 { 1 } else { 0 } };
+    pol fixed ISLAST(i) { if i == last_row { 1 } else { 0 } };
+    pol fixed ISALMOSTLAST(i) { if i == last_row - 1 { 1 } else { 0 } };
     pol fixed ISFIRST = [ 1, 0 ] + [0]*;
```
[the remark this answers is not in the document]
chriseth: I think so, I don't know why it was done like that.

test_data/asm/book/declarations.asm
```diff
@@ -1,12 +1,17 @@
...
+        0 => 0,
+        _ => arr[len - 1] + sum(len - 1, arr)
+    };
+    // A simple function that returns the input incremented by one,
+    // as an expression.
+    let incremented: expr -> expr = |x| x + 1;
     // This is a function that takes an expression as input and returns
     // a constraint enforcing this expression increments by a certain value
     // between rows.
+    // The type will be inferred here because `'` is only valid on `expr`.
```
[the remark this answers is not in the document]
chriseth: In principle yes, the `next` operator is "linear" in that `(5*x + y + 1)' = 5*(x') + y' + 1`, but I think the condenser would have to do that transformation and it currently does not do it (although I think it would be simple), but I might be wrong.

std/hash/poseidon_bn254.asm
```diff
@@ -50,7 +50,12 @@ machine PoseidonBN254(FIRSTBLOCK, operation_id) {
 
     // Utility method needed until the parser can parse large numbers outside the current field.
     // Takes a list of 8 u32 numbers and returns a u256 number.
-    let bn = |a1, a2, a3, a4, a5, a6, a7, a8| (a1 << 224) | (a2 << 192) | (a3 << 160) | (a4 << 128) | (a5 << 96) | (a6 << 64) | (a7 << 32) | a8;
+    // It is generic because it also needs to work for expr.
+    let<T: FromLiteral + Mul + Add> bn: T, T, T, T, T, T, T, T -> T = |a1, a2, a3, a4, a5, a6, a7, a8| array::fold(
+        [a1, a2, a3, a4, a5, a6, a7, a8],
+        0,
+        |acc, el| acc * 0x100000000 + el
```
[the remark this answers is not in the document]
chriseth: It would be, but `<<` is only implemented on integers, here we also have to support `expr`.

book/src/pil/types.md
chriseth: I can extend it. What I meant was that `col` can only be used at that location and no expression will ever have the type `col` (because it is not a proper type).

book/src/pil/types.md
chriseth: Yes, `byte` should be, `byte_f` does not need to be, but I can add types for clarity.

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+                        )
+                    ))?;
+                }
+            }
+        }
+
+        // We have to check type schemes last, because only at this point do we know
+        // that other types that should be concrete do not occur as type variables in the
+        // inferred type scheme any more.
+        let type_var_mapping = self.inferred_types.iter().map(|(name, inferred_type)| {
```
pacheco, line 242: It seems `inferred_types` doesn't need to be a member of `Self`, as it's only used inside this function. As a variable, you could `into_iter` here

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+                    .cloned()
+                    .zip(declared_type_vars.into_iter().map(|tv| Type::TypeVar(tv.clone())))
+                    .collect(),
+            ))
+        }).collect::<Result<_, String>>()?;
+
+        Ok(type_var_mapping)
+    }
+
+    /// Updates generic arguments and literal annotations with the proper resolved types.
```
pacheco, line 269: what's a literal annotation?

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+        Ok(())
+    }
+
+    /// Type-checks the isolated expressions.
+    fn check_expressions<T: FieldElement>(
+        &mut self,
+        expressions: &mut [(&mut Expression<T>, ExpectedType)],
+    ) -> Result<(), String> {
+        for (e, expected_type) in expressions {
+            if expected_type.allow_array {
```
pacheco, line 381: can you clarify a bit why `.allow_array` is needed and where it comes from?

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+            Expression::IfExpression(if_expr) => {
+                self.expect_type(&Type::Bool, &mut if_expr.condition)?;
+                let result = self.process_expression(&mut if_expr.body)?;
+                self.expect_type(&result, &mut if_expr.else_body)?;
+                result
+            }
+        })
+    }
+
+    fn process_function_call<'b, T: FieldElement>(
```
pacheco, line 543: `infer_function_call_type`?

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+                        )
+                    })?;
+            } else {
+                self.expect_type(&expected_type.ty, e)?;
+            }
+        }
+        Ok(())
+    }
+
+    fn process_expression<T: FieldElement>(
```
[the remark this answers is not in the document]
pacheco: same for `process_function_call`

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+        format!("T{}", self.last_type_var)
+    }
+
+    fn new_type_var(&mut self) -> Type {
+        Type::TypeVar(self.new_type_var_name())
+    }
+
+    /// Creates a type scheme out of a type by making all unsubstituted
+    /// type variables generic.
+    /// TODO this is wrong for mutually recursive generic functions.
```
pacheco, line 649: does this mean that for now those would fail type check?

pil-analyzer/tests/types.rs
```diff
@@ -41,3 +58,206 @@ fn use_fun_in_expr_context() {
...
+
+#[test]
+#[should_panic = "Unable to derive concrete type for reference to generic symbol std::array::len"]
+fn non_concrete_inner_type_arr() {
+    let input = "let x: int = std::array::len([]); namespace std::array(2); let len = 99;";
+    type_check(input, &[("x", "", "int")]);
+}
+
+#[test]
+fn type_check_arrays() {
```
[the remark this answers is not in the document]
chriseth: it verifies that type checking is performed inside RepeatedArrays

ast/src/analyzed/mod.rs
```diff
@@ -826,6 +871,9 @@ pub struct PolynomialReference {
     /// Optional because it is filled in in a second stage of analysis.
     /// TODO make this non-optional
     pub poly_id: Option<PolyID>,
+    /// The type arguments if the symbol is generic.
+    /// Might be empty before type checking is completed.
```
leonardoalt, line 875: This sounds a bit sketchy. Shouldn't it use a different type then? Or be a different type completely, pre and post type checking?

executor/src/constant_evaluator/mod.rs
leonardoalt: I don't think declaring a variable twice should be allowed. Or rather: we should decide if the language is supposed to remain declarative or not

pipeline/src/test_util.rs
```diff
@@ -163,7 +163,7 @@ pub fn evaluate_function<'a, T: FieldElement>(
     arguments: Vec<Rc<evaluator::Value<'a, T, evaluator::NoCustom>>>,
 ) -> evaluator::Value<'a, T, evaluator::NoCustom> {
     let symbols = evaluator::Definitions(&analyzed.definitions);
-    let function = symbols.lookup(function).unwrap();
+    let function = symbols.lookup(function, &[]).unwrap();
```
leonardoalt, line 166: Not about this PR but: why is this test in Pipeline actually?

std/array.asm
```diff
@@ -1,21 +1,22 @@
 /// This is a built-in function taking an array argument and returning
 /// the length of the array.
 /// This symbol is not an empty array, the actual semantics are overridden.
-let len = [];
+let<T> len: T[] -> int = [];
 
 /// Evaluates to the array [f(0), f(1), ..., f(length - 1)].
-let new = |length, f| std::utils::fold(length, f, [], |acc, e| (acc + [e]));
+let<T> new: int, (int -> T) -> T[] = |length, f| std::utils::fold(length, f, [], |acc, e| (acc + [e]));
```
leonardoalt, line 7: Nice!

std/array.asm
```diff
@@ -1,21 +1,22 @@
...
 /// Computes folder(...folder(folder(initial, arr[0]), arr[1]) ..., arr[len(arr) - 1])
-let fold = |arr, initial, folder| std::utils::fold(len(arr), |i| arr[i], initial, folder);
+let<T1, T2> fold: T1[], T2, (T2, T1 -> T2) -> T2 = |arr, initial, folder| std::utils::fold(len(arr), |i| arr[i], initial, folder);
 
 /// Returns the sum of the array elements.
-/// This actually also works on field elements, so the type is currently too restrictive.
-let sum: int[] -> int = |arr| fold(arr, 0, |a, b| a + b);
+// TODO: Should make use of the Default or Zero trait instead of FromLiteral (then we can also
+// use this function to flatten an array of arrays.
+let<T: Add + FromLiteral> sum: T[] -> T = |arr| fold(arr, 0, |a, b| a + b);
```
leonardoalt, line 18: Fancy

test_data/asm/failing_assertion.asm
```diff
@@ -1,9 +1,10 @@
 use std::check;
 
 machine Empty {
-    let line: col = |i| i - 7;
+    let line = |i| i - 7;
+    let line_col: col = line;
```
leonardoalt, line 5: So the previous case is not valid anymore?

test_data/asm/secondary_block_machine_add2.asm
```diff
@@ -22,12 +21,12 @@ machine Main {
     // Because constraints are not cyclic, we need to explicitly constrain the first state
     _first_step * (add_two_state - add_two_input) = 0;
 
-    // Add %offset in a single step of computation
-    constant %offset = 1;
+    // Add offset in a single step of computation
+    let offset = 1;
```
leonardoalt, line 25: Nice

leonardoalt (commented):
Review for all crates except `pil-analyzer` and `book`.

book/src/pil/declarations.md
```diff
@@ -7,18 +7,21 @@ from the type of the symbol and the way the symbol is used.
...
 This syntax can be used for constants, fixed columns, witness columns and even (higher-order)
 functions that can transform expressions. The kind of symbol is deduced by its type and the
 way the symbol is used:
 
-- Symbols without a value are witness columns or arrays of witness columns. Their type can be omitted. If it is given, it must be ``col`` or ``col[]``.
-- Symbols evaluating to a number or with type ``fe`` are constants.
-- Symbols without type but with a value that is a function with a single parameter are fixed columns.
-- Symbols defined with a value and type ``col`` (or ``col[]``) are fixed columns (or arrays of fixed columns).
+- Symbols without a value are witness columns or arrays of witness columns. Their type can be omitted. If it is given, it must be ``col`` or ``col[k]``.
+- Symbols with type ``fe`` are constants.
```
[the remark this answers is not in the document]
leonardoalt: yea agree

book/src/pil/types.md
leonardoalt: I mentioned this in another comment, maybe it's worth a chat discussion: it's almost like the `col` and `wit` business is a lower level language inside pil that's arithmetization specific.

book/src/pil/types.md
leonardoalt: agree

book/src/pil/types.md
leonardoalt: still needs update

book/src/pil/types.md
```diff
@@ -9,13 +9,75 @@ The powdr-pil language has the following types:
...
+
+All other symbols use their proper type both for their value and for references to these symbols.
```
leonardoalt, line 54:
```suggestion
All other symbols use their declared type both for their value and for references to these symbols.
```

book/src/pil/types.md
```diff
@@ -170,27 +233,23 @@ Expressions are eagerly evaluated from left to right.
...
+The function `f` adds the two columns `x` and `y` symbolically - it essentially returns the expression `x + y`.
 The last line is at statement level and it is expected that it evaluates to a constraint, in this case, a polynomial identity.
-Because of that, `f()` is expected to evaluate to an expression, which means that the references
-to `x` and `y` that appear when evaluating `f` are already turned into expressions.
-The same is true about the `7` inside `g`.
-
-If `f` is called in a different context (for example inside a function that defines a constant column),
-then the column references `x` and `y` are interpreted to have the type `int -> fe` and thus
-adding them results in a type error.
+Because of that, `g` is inferred to have type `-> expr`, which is compatible with the literal `7`.
```
leonardoalt, line 252: Nice

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+        // symbol is referenced). Unifying the declared type with the inferred
+        // type is done at the end.
+        for name in names {
+            // Ignore builtins (removed from definitions) and definitions without value.
+            let Some((_, Some(value))) = definitions.get_mut(&name) else {
+                continue;
+            };
+
+            let declared_type = self.declared_types[&name].clone();
+            (if declared_type.vars.is_empty() {
```
leonardoalt, line 142: where's this `(` closing? Can you rearrange this code a bit? It's kinda hard to read atm

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+                                        self.substitute_to(t)
+                                    )
+                                }),
+                        }
+                    }
+                    Type::Array(ArrayType {
+                        base,
+                        length: Some(_),
+                    }) if base.as_ref() == &Type::Expr => {
+                        // An array of intermediate columns with fixed length. We ignore the length.
```
leonardoalt, line 196: How do you know here that this is an array of intermediate columns?

executor/src/constant_evaluator/mod.rs
chriseth: It is currently not allowed to declare two symbols of the same name in the same scope. The example above has `x` and `X`.

pil-analyzer/src/type_unifier.rs
```diff
@@ -0,0 +1,134 @@
...
+        self.type_var_bounds
+            .get(type_var)
+            .cloned()
+            .unwrap_or_default()
+    }
+
+    pub fn ensure_bound(&mut self, ty: &Type, bound: String) -> Result<(), String> {
+        if let Type::TypeVar(n) = ty {
+            if let Some(sub) = self.substitutions.get(n) {
+                self.ensure_bound(&sub.clone(), bound)
```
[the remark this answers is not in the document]
chriseth: Yes, because it is a ref to self and `ensure_bound` needs a `&mut` to self. Changed it to avoid the clone nevertheless.

pil-analyzer/src/type_unifier.rs
```diff
@@ -0,0 +1,134 @@
...
+            if bounds.contains(&bound.as_str()) {
+                Ok(())
+            } else {
+                Err(format!("Type {ty} does not satisfy trait {bound}."))
+            }
+        }
+    }
+
+    pub fn unify_types(&mut self, mut inner: Type, mut expected: Type) -> Result<(), String> {
+        if let (Type::TypeVar(n1), Type::TypeVar(n2)) = (&inner, &expected) {
```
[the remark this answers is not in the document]
chriseth: I wanted to avoid the more complicated recursive comparison. I think I can also just remove the whole thing.

pil-analyzer/src/type_unifier.rs
```diff
@@ -0,0 +1,134 @@
...
+        inner.substitute_type_vars(&self.substitutions);
+        expected.substitute_type_vars(&self.substitutions);
+
+        if inner == expected {
+            return Ok(());
+        }
+        match (inner, expected) {
+            (Type::Bottom, _) | (_, Type::Bottom) => Ok(()),
+            (Type::TypeVar(n1), Type::TypeVar(n2)) if n1 == n2 => Ok(()),
+            (Type::TypeVar(name), ty) | (ty, Type::TypeVar(name)) => {
```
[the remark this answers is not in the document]
chriseth: Yes, both types are fully substituted. Also `add_substitution` has a check that the to-be-substituted variable does not appear in the substitution.

pil-analyzer/src/type_unifier.rs
```diff
@@ -0,0 +1,134 @@
...
+use powdr_ast::analyzed::types::Type;
+
+use crate::type_builtins::elementary_type_bounds;
+
+// TODO Optimization ideas:
+// the substitutions are applied a lot.
+// 1) One idea would be to extract a separate substitution map
+// that contains only the newly added substitutions during a unification run. Then we only have to use
+// that smaller map during the recursive substitutions.
+// 2) We could have a wrapper around type that stores which type variables are contained in it.
```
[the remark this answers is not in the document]
chriseth: I actually think that currently, it is fast enough.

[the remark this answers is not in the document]
chriseth: I started a preliminary implementation, but it did not provide a very big speedup.

pil-analyzer/src/type_unifier.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn add_type_var_bound(&mut self, type_var: String, bound: String) {
+        self.type_var_bounds
+            .entry(type_var)
+            .or_default()
+            .insert(bound);
+    }
+
+    fn add_substitution(&mut self, type_var: String, ty: Type) -> Result<(), String> {
+        if ty.contains_type_var(&type_var) {
+            Err(format!(
```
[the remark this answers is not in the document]
chriseth: Yes, there is even a test.

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+    local_var_types: Vec<Type>,
+    /// Declared types for symbols. Type scheme for polymorphic symbols
+    /// and unquantified type variables for symbols without type.
+    declared_types: HashMap<String, TypeScheme>,
+    /// These are the inferred types for symbols that are declared
+    /// as type schemes. They are compared to the declared types
+    /// at the very end.
+    inferred_types: HashMap<String, Type>,
+    unifier: Unifier,
+    last_type_var: usize,
```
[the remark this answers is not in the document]
chriseth: Actually no. Added a comment.

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+                .iter()
+                .map(|(n, (_, v))| (n.as_str(), v.as_deref())),
+        );
+
+        // Remove builtins from definitions and check they types are correct.
+        for (name, ty) in builtin_schemes() {
+            if let Some((_, (Some(defined_ty), _))) = definitions.remove_entry(name) {
+                assert!(
+                    ty == &defined_ty,
+                    "Invalid type for built-in scheme {name}: {}",
```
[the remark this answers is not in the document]
chriseth: It serves as documentation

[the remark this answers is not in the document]
chriseth: But actually we only check that the type is correct if it is specified (`Some(defined_ty)`)

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+            .map_err(|err| {
+                format!(
+                    "Error checking sub-expression {expr}:\nExpected type: {}\nInferred type: {}\n{err}",
+                    self.substitute_to(expected_type.clone()),
+                    self.substitute_to(inferred_type)
+                )
+            })
+    }
+
+    fn substitute_to(&self, mut ty: Type) -> Type {
```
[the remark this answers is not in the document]
chriseth: Hm, the target is a type, so I think `into_substitute` would be confusing.

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+        }
+
+        self.check_expressions(expressions)?;
+
+        // From this point on, the substitutions are fixed.
+
+        // Now we check for all symbols that are not declared as a type scheme that they
+        // can resolve to a concrete type.
+        for (name, declared_type) in &self.declared_types {
+            if declared_type.vars.is_empty() {
```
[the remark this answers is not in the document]
chriseth: It means that the type is not a type scheme. For declarations without types, we assign a fresh type variable, but do not add it to the type variables of the type scheme. I.e. the type is `T1` and not `<T1> T1`.

So `vars` of a type scheme is not the type variables that are contained in the type, it is the type variables that are newly introduced by the type scheme.

pil-analyzer/src/type_inference.rs
chriseth: Oh wow, great observation! Will change!

pil-analyzer/src/type_inference.rs
chriseth: The second element of Expression::Number that specifies which type the literal has.

pil-analyzer/src/type_inference.rs
chriseth: Should be documented at the type: It says if arrays of the expected type are allowed as well. This is mainly used at statement context where we expect either `constr` or `constr[]`.

pil-analyzer/src/type_inference.rs
chriseth: I actually haven't tried what happens...
Maybe it works because we specify the type schemes explicitly.

ast/src/analyzed/mod.rs
chriseth: We had that before and it was a big mess especially with the assembly that is still not properly integrated. Would prefer to keep it an option for now.

test_data/asm/failing_assertion.asm
chriseth: No, unfortunately not.

pil-analyzer/tests/types.rs
```diff
@@ -41,3 +58,206 @@ fn use_fun_in_expr_context() {
...
+#[test]
+fn single_literal() {
+    let input = "let<T: FromLiteral> x: T[] = [1, 2];";
+    type_check(input, &[("x", "T: FromLiteral", "T[]")]);
+}
+
+#[test]
+fn assignment() {
+    // This should derive a concrete type for x due to how it is used by y.
+    let input = "let x = [|i| i]; let y: int[] = [x[0](2)];";
```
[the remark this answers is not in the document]
chriseth: It's not needed any more. Maybe it was when I wrote the test. But regardless, it makes the test a bit more complicated so probably still good.

pil-analyzer/tests/types.rs
```diff
@@ -41,3 +58,206 @@ fn use_fun_in_expr_context() {
...
+    type_check(input, &[]);
+}
+
+#[test]
+fn if_statement() {
+    let input = "
+        let g = || g();
+        let x = |a, b| if g() { a } else { b + 2 };
+        let c: int = 2;
+        let y = [|i| x(c, i)];
```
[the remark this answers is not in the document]
chriseth: Nothing

pil-analyzer/src/expression_processor.rs
```diff
@@ -179,6 +179,10 @@ impl<T, D: AnalysisDriver<T>> ExpressionProcessor<T, D> {
         PolynomialReference {
             name: self.driver.resolve_ref(path),
             poly_id: None,
+            // These will be filled by the type checker.
```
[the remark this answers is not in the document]
chriseth: Yep, will make it an option.

pil-analyzer/src/type_builtins.rs
```diff
@@ -0,0 +1,161 @@
...
+pub fn unary_operator_scheme(op: UnaryOperator) -> TypeScheme {
+    UNARY_OPERATOR_SCHEMES[&op].clone()
+}
+
+pub fn elementary_type_bounds(ty: &Type) -> &'static [&'static str] {
+    match ty {
+        Type::Bottom => &[],
+        Type::Bool => &[],
+        Type::Int => &[
+            "FromLiteral",
```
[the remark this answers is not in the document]
chriseth: Yes, I want to allow user-defined traits.

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+
+use crate::{
+    call_graph::sort_called_first,
+    type_builtins::{
+        binary_operator_scheme, builtin_schemes, type_for_reference, unary_operator_scheme,
+    },
+    type_unifier::Unifier,
+};
+
+// TODO we should check at the end that the literals are not too large for the inferred types.
```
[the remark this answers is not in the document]
chriseth: later, it's on the list: https://github.com/powdr-labs/powdr/issues/1106

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,675 @@
...
+        }
+    }
+}
+
+#[derive(Default)]
+struct TypeChecker {
+    /// Types for local variables, might contain type variables.
+    local_var_types: Vec<Type>,
+    /// Declared types for symbols. Type scheme for polymorphic symbols
+    /// and unquantified type variables for symbols without type.
```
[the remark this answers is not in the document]
chriseth: In `let<T> x: (T, K) = (1,2)`, `T` is quantified, `K` is unquantified. Of course, in the source code, unquantified type variables are illegal but they appear during type checking until they are substituted by something else that does hopefully not contain any unquantified type variables any more.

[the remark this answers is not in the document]
chriseth: I can also write "new unused" if that is easier to understand?

book/src/pil/types.md
```diff
@@ -9,13 +9,77 @@ The powdr-pil language has the following types:
...
 - array
 - function type
-- expression
-- constraint
+- `expr` (expression)
+- `constr` (constraint)
+- `!` ("bottom" or "unreachable" type)
+
+> The `col` type is special in that it is only used for declaring columns, but cannot appear as the type of an expression.
+> See [Types of Symbols](#declaring-and-referencing-columns) for details.
```
pacheco, line 17: name of the section has changed

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,684 @@
...
+                        )
+                    })?;
+            } else {
+                self.expect_type(&expected_type.ty, e)?;
+            }
+        }
+        Ok(())
+    }
+
+    /// Process and expresison and return the type of the expression.
```
pacheco, line 419: `expression`

pil-analyzer/src/type_inference.rs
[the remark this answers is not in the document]
pacheco: The comment in the `TypeScheme` struct definition, it mentions that `ty: Type` uses the type variables from `vars`... can this be clarified in the struct definition?

pil-analyzer/src/type_inference.rs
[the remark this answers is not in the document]
pacheco: maybe "undeclared"?

book/src/pil/types.md
chriseth: Haha, I updated the link but forget the text :)

pil-analyzer/src/type_inference.rs
```diff
@@ -0,0 +1,717 @@
...
+                        format_type_scheme_around_name(
+                            name,
+                            &Some(inferred_scheme),
+                        )
+                    ));
+                }
+            }
+        }
+
+        // We check type schemes last, because only at this point do we know
```
pacheco, line 162: maybe this last part could be a `check_type_schemes` method that takes in the `inferred_types`?

ast/src/analyzed/mod.rs
```diff
@@ -874,8 +874,8 @@ pub struct PolynomialReference {
     /// TODO make this non-optional
     pub poly_id: Option<PolyID>,
     /// The type arguments if the symbol is generic.
-    /// Might be empty before type checking is completed.
-    pub generic_args: Vec<Type>,
+    /// Guarenteed to be Some(_) after type checking is completed.
```
leonardoalt, line 877: If it's guaranteed maybe it should be a GuaranteedGenericArgs type :D

pil-analyzer/src/call_graph.rs
```diff
@@ -7,7 +7,7 @@ use powdr_ast::{
 use powdr_number::FieldElement;
 
 /// Returns a sorted list of symbols such that called symbols appear before the symbols that reference them.
-/// Circular dependencies appear in an arbitray order.
+/// Circular dependencies appear in an arbitrary order.
```
leonardoalt, line 10:
```suggestion
/// Circular dependencies appear in an arbitrary order.
```

pacheco (approved):
LGTM!

leonardoalt (approved):

PR #1019, opened 2024-02-06
Support expression for latch.

chriseth:


leonardoalt:
is this ready for review?

chriseth:
No, there is a problem with the namespace. During machine instantiation, we need to change the namespaces to the instantiation point (plus a suffix for the machine). I have the feeling that this could be one in a better way with functions returning constraints and columns.

PR #1024, opened 2024-02-06
Remove load_dynamic instruction

lvella:
It is redundant now that `li` can do the same thing.

### DIFF
--- a/riscv/src/compiler.rs
+++ b/riscv/src/compiler.rs
@@ -239,13 +239,13 @@ fn replace_dynamic_label_references(statements: &mut Vec<Statement>, data_labels
     lui	a0, %hi(LABEL)
     addi	s10, a0, %lo(LABEL)
     -
-    turn this into the pseudo-riscv-instruction
-    load_dynamic s10, LABEL
+    turn this into the pseudoinstruction
+    li s10, LABEL
     which is then turned into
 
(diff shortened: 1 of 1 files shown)


riscv/tests/riscv_data/function_pointer/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+#![no_std]
+
+use powdr_riscv_runtime::get_prover_input;
+
+// Never inline to make sure the function is not optimized away, and public to
+// make sure op argument is not removed, forcing a function pointer to be loaded
+// and passed.
+#[inline(never)]
+pub fn apply_op(op: fn(u32, u32) -> u32, a: u32, b: u32) -> u32 {
```
chriseth, line 9: Wouldn't it be better to force dynamic dispatch?

chriseth:
Ah that's what you meant!

riscv/tests/riscv_data/function_pointer/src/lib.rs
lvella: Anyway, this commit was already merged in another PR. I made this branch on top of another branch and forgot to set the target appropriately. Sorry.

chriseth (approved):

PR #1035, opened 2024-02-08
Refactor of backend API

lvella:
Would like feedback first before changing eStark.

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -6,7 +6,7 @@ mod pilstark;
 
 use powdr_ast::analyzed::Analyzed;
 use powdr_number::{DegreeType, FieldElement};
-use std::{io, marker::PhantomData};
+use std::{io, path::Path};
 use strum::{Display, EnumString, EnumVariantNames};
 
 #[derive(Clone, EnumString, EnumVariantNames, Display, Copy)]
@@ -26,138 +26,20 @@ pub enum BackendType {
 impl BackendType {
     pub fn factory<T: FieldElement>(&self) -> &'static dyn BackendFactory<T> {
         #[cfg(feature = "halo2")]
-        const HALO2_FACTORY: WithSetupFactory<powdr_halo2::Halo2Prover> =
-            WithSetupFactory(PhantomData);
+        const HALO2_FACTORY: halo2_impl::Halo2ProverFactory = halo2_impl::Halo2ProverFactory;
         #[cfg(feature = "halo2")]
-        const HALO2_MOCK_FACTORY: WithoutSetupFactory<halo2_impl::Halo2Mock> =
-            WithoutSetupFactory(PhantomData);
-        const ESTARK_FACTORY: WithoutSetupFactory<pilstark::estark::EStark> =
-            WithoutSetupFactory(PhantomData);
-        const PIL_STARK_CLI_FACTORY: WithoutSetupFactory<pilstark::PilStarkCli> =
-            WithoutSetupFactory(PhantomData);
+        const HALO2_MOCK_FACTORY: halo2_impl::Halo2MockFactory = halo2_impl::Halo2MockFactory;
+        const ESTARK_FACTORY: pilstark::estark::EStarkFactory = pilstark::estark::EStarkFactory;
+        const PIL_STARK_CLI_FACTORY: pilstark::PilStarkCliFactory = pilstark::PilStarkCliFactory;
 
         match self {
             #[cfg(feature = "halo2")]
             BackendType::Halo2 => &HALO2_FACTORY,
             #[cfg(feature = "halo2")]
             BackendType::Halo2Mock => &HALO2_MOCK_FACTORY,
-            BackendType::PilStarkCli => &PIL_STARK_CLI_FACTORY,
             BackendType::EStark => &ESTARK_FACTORY,
+            BackendType::PilStarkCli => &PIL_STARK_CLI_FACTORY,
         }
-    }
-}
-
-/// Factory for backends without setup.
-struct WithoutSetupFactory<B>(PhantomData<B>);
-
-/// Factory implementation for backends without setup.
-impl<F: FieldElement, B: BackendImpl<F> + 'static> BackendFactory<F> for WithoutSetupFactory<B> {
-    fn create(&self, degree: DegreeType) -> Box<dyn Backend<F>> {
-        Box::new(ConcreteBackendWithoutSetup(B::new(degree)))
-    }
-
-    fn create_from_setup(&self, _input: &mut dyn io::Read) -> Result<Box<dyn Backend<F>>, Error> {
-        Err(Error::NoSetupAvailable)
-    }
-}
-
-/// Concrete dynamic dispatch Backend object, for backends without setup.
-struct ConcreteBackendWithoutSetup<B>(B);
-
-/// Concrete implementation for backends with setup.
-impl<F: FieldElement, B: BackendImpl<F>> Backend<F> for ConcreteBackendWithoutSetup<B> {
-    fn add_verification_key(
-        &mut self,
-        pil: &Analyzed<F>,
-        fixed: &[(String, Vec<F>)],
-        vkey: Vec<u8>,
-    ) {
-        self.0.add_verification_key(pil, fixed, vkey);
-    }
-
-    fn verification_key(
-        &self,
-        pil: &Analyzed<F>,
-        fixed: &[(String, Vec<F>)],
-    ) -> Result<Vec<u8>, Error> {
-        self.0.verification_key(pil, fixed)
-    }
-
-    fn verify(&self, proof: &Proof, instances: &[Vec<F>]) -> Result<(), Error> {
-        self.0.verify(proof, instances)
-    }
-
-    fn prove(
-        &self,
-        pil: &Analyzed<F>,
-        fixed: &[(String, Vec<F>)],
-        witness: &[(String, Vec<F>)],
-        prev_proof: Option<Proof>,
-    ) -> Result<(Proof, Option<String>), Error> {
-        self.0.prove(pil, fixed, witness, prev_proof)
-    }
-
-    fn write_setup(&self, _output: &mut dyn io::Write) -> Result<(), Error> {
-        Err(Error::NoSetupAvailable)
-    }
-}
-
-/// Factory for backends with setup.
-struct WithSetupFactory<B>(PhantomData<B>);
-
-/// Factory implementation for backends with setup.
-impl<F: FieldElement, B: BackendImplWithSetup<F> + 'static> BackendFactory<F>
-    for WithSetupFactory<B>
-{
-    fn create(&self, degree: DegreeType) -> Box<dyn Backend<F>> {
-        Box::new(ConcreteBackendWithSetup(B::new(degree)))
-    }
-
-    fn create_from_setup(&self, input: &mut dyn io::Read) -> Result<Box<dyn Backend<F>>, Error> {
-        Ok(Box::new(ConcreteBackendWithSetup(B::new_from_setup(
-            input,
-        )?)))
-    }
-}
-
-/// Concrete dynamic dispatch Backend object, for backends with setup.
-struct ConcreteBackendWithSetup<B>(B);
-
-/// Concrete implementation for backends with setup.
-impl<F: FieldElement, B: BackendImplWithSetup<F>> Backend<F> for ConcreteBackendWithSetup<B> {
-    fn add_verification_key(
-        &mut self,
-        pil: &Analyzed<F>,
-        fixed: &[(String, Vec<F>)],
-        vkey: Vec<u8>,
-    ) {
-        self.0.add_verification_key(pil, fixed, vkey);
-    }
-
-    fn verification_key(
-        &self,
-        pil: &Analyzed<F>,
-        fixed: &[(String, Vec<F>)],
-    ) -> Result<Vec<u8>, Error> {
-        self.0.verification_key(pil, fixed)
-    }
-
-    fn verify(&self, proof: &Proof, instances: &[Vec<F>]) -> Result<(), Error> {
-        self.0.verify(proof, instances)
-    }
-
-    fn prove(
-        &self,
-        pil: &Analyzed<F>,
-        fixed: &[(String, Vec<F>)],
-        witness: &[(String, Vec<F>)],
-        prev_proof: Option<Proof>,
-    ) -> Result<(Proof, Option<String>), Error> {
-        self.0.prove(pil, fixed, witness, prev_proof)
-    }
-
-    fn write_setup(&self, output: &mut dyn io::Write) -> Result<(), Error> {
-        Ok(self.0.write_setup(output)?)
     }
 }
 
@@ -169,12 +51,18 @@ pub enum Error {
     EmptyWitness,
     #[error("the backend has no setup operations")]
     NoSetupAvailable,
-    #[error("proof generation failed")]
-    ProofFailed(String),
-    #[error("proof verification failed")]
-    VerificationFailed(String),
-    #[error("verification key generation failed")]
-    VerificationKeyFailed(String),
+    #[error("the backend does not implement proof verification")]
+    NoVerificationAvailable,
+    #[error("the backend does not support proof aggregation")]
+    NoAggregationAvailable,
+    #[error("internal backend error")]
+    BackendError(String),
+}
+
+impl From<String> for Error {
+    fn from(s: String) -> Self {
+        Error::BackendError(s)
+    }
 }
 
 pub type Proof = Vec<u8>;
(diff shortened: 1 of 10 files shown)


pipeline/src/pipeline.rs
```diff
@@ -724,15 +724,16 @@ impl<T: FieldElement> Pipeline<T> {
     }
 
     fn maybe_write_proof(&self, proof_result: &ProofResult<T>) -> Result<(), Vec<String>> {
-        if let Some(constraints_serialization) = &proof_result.constraints_serialization {
+        // TODO: this must go inside the backend...
```
leonardoalt, line 727: I disagree with this. IMO the user needs to access the artifact and do whatever it needs with it

cli/src/main.rs
```diff
@@ -624,32 +624,23 @@ fn verification_key<T: FieldElement>(
         .with_setup_file(params.map(PathBuf::from))
         .with_backend(*backend_type);
 
-    let vkey = pipeline.verification_key()?;
-
-    write_verification_key_to_fs(vkey, dir);
+    let mut vkey_file = BufWriter::new(fs::File::create(dir.join("vkey.bin")).unwrap());
+    pipeline.write_verification_key(&mut vkey_file)?;
+    vkey_file.flush().unwrap();
+    log::info!("Wrote vkey.bin.");
```
leonardoalt, line 630: I find it odd that here the pipeline writes to FS, and in the setup case it returns the data and the user writes to FS.

pipeline/src/pipeline.rs
```diff
@@ -865,7 +866,10 @@ impl<T: FieldElement> Pipeline<T> {
         self.arguments.query_callback.as_deref()
     }
 
-    pub fn verification_key(&mut self) -> Result<Vec<u8>, Vec<String>> {
+    pub fn write_verification_key<W: io::Write>(
```
leonardoalt, line 869: Why does Pipeline write it to FS instead of just returning the key?

backend/src/lib.rs
```diff
@@ -169,12 +58,14 @@ pub enum Error {
...
     #[error("the backend has no setup operations")]
     NoSetupAvailable,
-    #[error("proof generation failed")]
-    ProofFailed(String),
-    #[error("proof verification failed")]
-    VerificationFailed(String),
-    #[error("verification key generation failed")]
-    VerificationKeyFailed(String),
+    #[error("backend specific error")]
+    BackendError(String),
```
leonardoalt, line 62: I think we could keep the old errors since they are generic enough for the things we want to do.

backend/src/lib.rs
```diff
@@ -184,91 +75,40 @@ pub type Proof = Vec<u8>;
...
-        fixed: &[(String, Vec<F>)],
-    ) -> Result<Vec<u8>, Error>;
+        pil: &'a Analyzed<F>,
+        fixed: &'a [(String, Vec<F>)],
+        setup: Option<&mut dyn io::Read>,
+        verification_key: Option<&mut dyn io::Read>,
+    ) -> Result<Box<dyn Backend<'a, F> + 'a>, Error>;
+
+    /// Generate a new setup.
+    fn generate_setup(&self, size: DegreeType, output: &mut dyn io::Write) -> Result<(), Error>;
```
leonardoalt, line 90: Why is setup the only function that goes into this trait?

leonardoalt: shouldn't it also go into the Backend trait like the other functionalities?

lvella: This is because how we use it in `powdr setup` (I think), where we can call it without needing the analyzed pil or the fixed columns. Verifying and proving, on the other hand, do require the full object with the common stuff.

pipeline/src/pipeline.rs
lvella: Why is the pipeline returning a key in the first place?

I am very uncomfortable with uninterpreted `Vec<u8>` being passed around contained backend specific serialized data. We are bypassing the type system. This is no better than casting to `void *` or using `unsafe` to reinterpret cast. Serializations should go directly to disk.

backend/src/lib.rs
lvella: I changed it because this way it is simpler and less error prone to write the backend implementation: you can  just use `?` to turn return from a `Result<_, String>` and it becomes a backend error. Whether it is a proving or a verification error, it is obvious from the function that was called.

cli/src/main.rs
leonardoalt: Because at the moment of writing that code the pipeline was the easiest way to collect all required artifacts at the same time. Try collecting the verification key in a different way and you'll see. You'll need to repeatedly call the Pipeline to get different artifacts, or call the specific crates directly.

leonardoalt: That said, I'm fine with this not being in Pipeline and if it's easy to remove it and do it like Setup is done, sounds good to me.

pipeline/src/pipeline.rs
leonardoalt: I disagree it should go directly to disk. I think the user should decide what to do with it. Vec<u8> is a universal serialized data type.

backend/src/lib.rs
leonardoalt: Sure, not too strongly opinionated here.

pipeline/src/pipeline.rs
lvella: The proposed API accommodates both `File` and `Vec<u8>`. Both implement `io::Write`.

backend/src/lib.rs
```diff
@@ -184,91 +75,40 @@ pub type Proof = Vec<u8>;
...
-where
-    Self: Sized + BackendImpl<F>,
-{
-    /// Create a backend object from a setup loaded from a file.
-    fn new_from_setup(input: &mut dyn io::Read) -> Result<Self, io::Error>;
 
-    /// Write the setup to a file.
-    fn write_setup(&self, output: &mut dyn io::Write) -> Result<(), io::Error>;
+    /// Write the verification key.
+    fn write_verification_key(&self, output: &mut dyn io::Write) -> Result<(), Error>;
```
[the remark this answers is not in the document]
leonardoalt: Vec<u8> is universal enough. But sure, fine with `serialize`

backend/src/lib.rs
leonardoalt: I see. Ok... I still find it ugly that a backend that has nothing to do with setup needs to actively implement this function and refuse it, but well.. that's the trade off of having the dynamic interfaces

lvella: Oh, yeah. I forgot about it, but I was planning to provide default implementation to everything that is optional for a backend to have.

backend/src/lib.rs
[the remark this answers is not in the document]
leonardoalt: Actually, fine with "write"

backend/src/lib.rs
lvella: Much better now.

backend/src/lib.rs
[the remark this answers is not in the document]
lvella: Now it is called "export_".

pipeline/src/verify.rs
```diff
@@ -11,7 +11,7 @@ pub fn verify(temp_dir: &Path, name: &str, constants_name: Option<&str>) {
         temp_dir.to_str().unwrap()
     );
     let commits_file = format!("{}/{name}_commits.bin", temp_dir.to_str().unwrap());
-    let constraints_file = format!("{}/{name}_constraints.json", temp_dir.to_str().unwrap());
+    let constraints_file = format!("{}/constraints.json", temp_dir.to_str().unwrap());
```
leonardoalt, line 14: why?

pipeline/src/pipeline.rs
```diff
@@ -681,10 +682,9 @@ impl<T: FieldElement> Pipeline<T> {
 
     fn maybe_write_constants(&self, constants: &[(String, Vec<T>)]) -> Result<(), Vec<String>> {
         if let Some(path) = self.path_if_should_write(|name| format!("{name}_constants.bin"))? {
-            write_polys_file(
-                &mut BufWriter::new(&mut fs::File::create(path).unwrap()),
-                constants,
-            );
+            let mut file = BufWriter::new(fs::File::create(path).unwrap());
+            write_polys_file(&mut file, constants);
+            file.flush().unwrap();
```
leonardoalt, line 687: Why not keep it as a separate function? These 3 lines are duplicated in other places

pipeline/src/pipeline.rs
```diff
@@ -874,16 +864,28 @@ impl<T: FieldElement> Pipeline<T> {
...
+                let mut setup_file = self
+                    .arguments
+                    .setup_file
+                    .as_ref()
+                    .map(|path| BufReader::new(fs::File::open(path).unwrap()));
+
+                let backend = factory
+                    .create(
+                        pil.borrow(),
+                        &fixed_cols[..],
```
leonardoalt, line 877: In general I find it annoying to pass the fixed_cols around like this (which is not this PR's fault). We should have some type "low level program" that has the PIL constraints AND the fixed columns.

pipeline/src/pipeline.rs
```diff
@@ -51,8 +51,6 @@ pub struct ProofResult<T: FieldElement> {
     pub witness: Option<Vec<(String, Vec<T>)>>,
     /// Proof, potentially None (if success is false)
     pub proof: Option<Proof>,
-    /// Serialized low level constraints, potentially None (if success is false)
-    pub constraints_serialization: Option<String>,
```
leonardoalt, line 55: why?

leonardoalt: don't we need it for the tests?

cli/src/main.rs
```diff
@@ -624,32 +623,23 @@ fn verification_key<T: FieldElement>(
         .with_setup_file(params.map(PathBuf::from))
         .with_backend(*backend_type);
 
-    let vkey = pipeline.verification_key()?;
-
-    write_verification_key_to_fs(vkey, dir);
+    let mut vkey_file = BufWriter::new(fs::File::create(dir.join("vkey.bin")).unwrap());
+    pipeline.export_verification_key(&mut vkey_file)?;
+    vkey_file.flush().unwrap();
```
leonardoalt, line 628: a bit ugly to always have to do this flush

backend/src/lib.rs
```diff
@@ -184,91 +72,51 @@ pub type Proof = Vec<u8>;
...
 
-/// Trait implemented by backends that have a setup phase that must be saved to
-/// a file.
-trait BackendImplWithSetup<F: FieldElement>
-where
-    Self: Sized + BackendImpl<F>,
-{
-    /// Create a backend object from a setup loaded from a file.
-    fn new_from_setup(input: &mut dyn io::Read) -> Result<Self, io::Error>;
+    /// Exports the setup in a backend specific format. Can be used to create
```
leonardoalt, line 111: wdym "can be used to create a new instance of the same backend"?

halo2/src/prover.rs
```diff
@@ -36,90 +35,89 @@ use std::{
...
+pub use halo2_proofs::poly::commitment::Params;
 pub use halo2_proofs::poly::kzg::commitment::ParamsKZG;
-
-/// Create a halo2 proof for a given PIL, fixed column values and witness column values
-/// We use KZG ([GWC variant](https://eprint.iacr.org/2019/953)) and Keccak256
-pub struct Halo2Prover {
+pub use halo2_proofs::SerdeFormat;
+
+/// Create a halo2 proof for a given PIL, fixed column values and witness column
+/// values We use KZG ([GWC variant](https://eprint.iacr.org/2019/953)) and
```
leonardoalt, line 43:
```suggestion
/// values. We use KZG ([GWC variant](https://eprint.iacr.org/2019/953)) and
```

halo2/src/prover.rs
```diff
@@ -132,27 +130,16 @@ impl Halo2Prover {
...
     ) -> Result<Vec<u8>, String> {
         log::info!("Starting proof aggregation...");
 
-        // TODO this is hacky
-        let degree = usize::BITS - pil.degree().leading_zeros() + 1;
-        let params_app = {
-            let mut params = self.params.clone();
-            params.downsize(degree);
-            params
-        };
```
leonardoalt, line 150: Why was this removed? This is important because the universal setup may be a common large one, and here we want to downsize it to the circuit's degree

halo2/src/prover.rs
```diff
@@ -219,32 +206,19 @@ impl Halo2Prover {
...
-        pil: &Analyzed<F>,
-        fixed: &[(String, Vec<F>)],
-    ) -> Result<VerifyingKey<G1Affine>, String> {
-        let plaf_circuit = analyzed_to_plaf(pil, fixed);
-        let circuit = analyzed_to_circuit_with_zeroed_witness(plaf_circuit, pil);
-        match keygen_vk(&self.params, &circuit) {
-            Ok(vkey) => Ok(vkey),
-            Err(e) => Err(e.to_string()),
-        }
+    pub fn add_verification_key(&mut self, mut vkey: &mut dyn io::Read) {
```
leonardoalt, line 209: I guess we don't need this function anymore?

halo2/src/prover.rs
```diff
@@ -219,32 +206,19 @@ impl Halo2Prover {
...
     }
 
-    pub fn verification_key<F: FieldElement>(
-        &self,
-        pil: &Analyzed<F>,
-        fixed: &[(String, Vec<F>)],
-    ) -> Result<Vec<u8>, String> {
-        let vk = self.verification_key_inner(pil, fixed)?;
-        Ok(vk.to_bytes(SerdeFormat::Processed))
+    pub fn verification_key(&self) -> Result<VerifyingKey<G1Affine>, String> {
```
leonardoalt, line 219: Maybe this could already check if we have one and simply clone that?

leonardoalt (commented):
Nice! Some small comments:

pipeline/src/verify.rs
lvella: Would be another argument into the backend. Maybe the backend should take instead a reference to the pipeline.

pipeline/src/pipeline.rs
lvella: I hope not. I remember running all the tests last friday. This was just so we export the json to the prover-cpp.

cli/src/main.rs
lvella: This is just so write errors aren't ignored. I thought of two solutions for this:

1. a new BufWriter-like struct that panics if there is an error flushing inside `drop()` (instead of ignoring it);
2. a plain function that encapsulates this repeated code.

backend/src/lib.rs
lvella: "can be used to create a new instance of the same backend type" makes it clearer? Means that I can pass the exported setup to `create()` function.

halo2/src/prover.rs
lvella: As I said previously, I think `powdr_halo2` belongs as a mod inside backend crate, but I didn't want to do any any deeper change to the `Halo2Prover` in this PR.

leonardoalt:
test failing

halo2/src/prover.rs
lvella: I have placed this `downsize()` in `create()`. Do you think it is correct there? Still fails the test.

leonardoalt: which test?

backend/src/lib.rs
lvella: reworded

halo2/src/prover.rs
lvella: The one that was failing. But it is fixed now.

pipeline/src/pipeline.rs
lvella: Placed it in a different function.

cli/src/main.rs
lvella: Wrote a new function for these calls.

halo2/src/prover.rs
lvella: This mimics the behavior of what is in main. `verification_key()` always creates a new key, and if this key is not manually added to the backend, it panics when verifying.

Doesn't look good. Maybe if a key is needed, either by this function or by `verify()`, a new one should be created and set. Or either, create the key if none is supplied in `create()`.

Should I change this in this PR?

pipeline/src/pipeline.rs
leonardoalt: Yea it's tricky. I agree Pipeline shouldn't really need to know about this, but at the same time I also don't think a backend should be writing specific things to the disk.

halo2/src/prover.rs
leonardoalt: The key is deterministic so I don't think we need to create a new one. But either is fine (here or another PR)

leonardoalt:
please squash then I'll approve

halo2/src/prover.rs
```diff
@@ -36,90 +35,96 @@ use std::{
...
-    ) {
+        setup: Option<&mut dyn io::Read>,
+    ) -> Result<Self, io::Error> {
+        Self::assert_field_is_bn254();
+
+        let params = setup
+            .map(|mut setup| ParamsKZG::<Bn256>::read(&mut setup))
+            .transpose()?
+            .map(|mut params| {
+                params.downsize(degree_bits(pil.degree()));
```
leonardoalt, line 76: I'm also not sure it should be done here. Question: if I want to compute two separate proofs, then aggregate them, do you see that as 3 backend instances, or would I use the same? If the same, this here isn't ideal because I might want to keep a large `setup` that works for all circuits, but when making specific proofs I downsize the setup params to that specific circuit.

lvella: If the PIL and fixed cols do change in between the separated proofs, this API is not very good. The degree is taken from the PIL, so it would require one instance per proof.

leonardoalt: they don't change for separate proofs, that's the whole point.

lvella: Then I don't understand. Either PIL and degree needs to change in between proofs, or they don't. The way it is, if the setup is loaded, it is downsized. If it is generated from the factory, it is not.

leonardoalt (approved):

PR #1043, opened 2024-02-09
add halo2 to powdr crate by default

leonardoalt:


lvella:
I'll let @chriseth review this one, because he was the one to complain about halo2.

leonardoalt:
He complained about it because it slows down our dev env. But this is about how the powdr crate is shipped externally, and we don't use this crate internally nor for development. I think it should be shipped with full features.

PR #1044, opened 2024-02-09
`instr` declaration parameter mapping

pacheco:
Allow mapping register usage on the RHS of `instr` declarations.
Included file `vm_instr_param_mapping.asm` shows possible usages.

### DIFF
--- a/airgen/src/lib.rs
+++ b/airgen/src/lib.rs
@@ -174,12 +174,26 @@ impl<'a, T: FieldElement> ASMPILConverter<'a, T> {
         LinkDefinitionStatement {
             source: _,
             flag,
-            params,
-            to: CallableRef { instance, callable },
+            params: lhs_params,
+            to:
+                CallableRef {
+                    instance,
+                    callable,
+                    params: rhs_params,
+                },
         }: LinkDefinitionStatement<T>,
     ) -> Link<T> {
+        // The actual mapping of parameters used for the link constraint comes
+        // from the RHS of the instr/link declaration (as it allows for extra
+        // registers/columns being used and parameter reordering).
+        // We allow declarations with an empty RHS as syntactic sugar for when RHS = LHS.
+        let params = if rhs_params.is_empty() {
+            lhs_params
+        } else {
+            rhs_params
+        };
         let from = LinkFrom {
-            params: params.clone(),
+            params,
             flag: flag.clone(),
         };
 
--- a/analysis/src/vm/inference.rs
+++ b/analysis/src/vm/inference.rs
@@ -50,12 +50,11 @@ fn infer_machine<T: FieldElement>(mut machine: Machine<T>) -> Result<Machine<T>,
                             .instructions
                             .iter()
                             .find(|i| i.name == *instr_name)
-                            .unwrap();
+                            .unwrap_or_else(|| panic!("invalid instruction: {}", instr_name));
 
-                        let outputs = def.instruction.params.outputs.clone().unwrap_or_default();
-
-                        outputs
+                        def.instruction
                             .params
+                            .outputs
                             .iter()
                             .map(|o| {
                                 assert!(o.ty.is_none());
--- a/asm-to-pil/src/romgen.rs
+++ b/asm-to-pil/src/romgen.rs
@@ -9,7 +9,7 @@ use powdr_ast::asm_analysis::{
 use powdr_ast::parsed::visitor::ExpressionVisitable;
 use powdr_ast::parsed::NamespacedPolynomialReference;
 use powdr_ast::parsed::{
-    asm::{OperationId, Param, ParamList, Params},
+    asm::{OperationId, Param, Params},
     Expression,
 };
 use powdr_ast::SourceRef;
@@ -105,18 +105,12 @@ pub fn generate_machine_rom<T: FieldElement>(
         // the number of inputs is the max of the number of inputs needed in each function
         let input_count = machine
             .functions()
-            .map(|f| f.params.inputs.params.len())
+            .map(|f| f.params.inputs.len())
             .max()
             .unwrap_or(0);
         let output_count = machine
             .functions()
-            .map(|f| {
-                f.params
-                    .outputs
-                    .as_ref()
-                    .map(|o| o.params.len())
-                    .unwrap_or(0)
-            })
+            .map(|f| f.params.outputs.len())
             .max()
             .unwrap_or(0);
         // create one read-only register for each input
@@ -146,38 +140,33 @@ pub fn generate_machine_rom<T: FieldElement>(
             let input_substitution = function
                 .params
                 .inputs
-                .params
                 .iter()
                 .enumerate()
                 .map(|(index, param)| (param.name.clone(), input_at(index)))
                 .collect();
 
-            let operation_inputs = ParamList {
-                params: function
-                    .params
-                    .inputs
-                    .params
-                    .iter()
-                    .enumerate()
-                    .map(|(i, _)| Param {
-                        name: input_at(i),
-                        index: None,
-                        ty: None,
-                    })
-                    .collect(),
-            };
-            let operation_outputs = function.params.outputs.clone().map(|outputs| ParamList {
-                params: outputs
-                    .params
-                    .into_iter()
-                    .enumerate()
-                    .map(|(i, _)| Param {
-                        name: output_at(i),
-                        index: None,
-                        ty: None,
-                    })
-                    .collect(),
-            });
+            let operation_inputs = function
+                .params
+                .inputs
+                .iter()
+                .enumerate()
+                .map(|(i, _)| Param {
+                    name: input_at(i),
+                    index: None,
+                    ty: None,
+                })
+                .collect();
+            let operation_outputs = function
+                .params
+                .outputs
+                .iter()
+                .enumerate()
+                .map(|(i, _)| Param {
+                    name: output_at(i),
+                    index: None,
+                    ty: None,
+                })
+                .collect();
 
             // substitute the names in the operation body and extend the return arguments
             for s in function.body.statements.iter_mut() {
--- a/asm-to-pil/src/vm_to_constrained.rs
+++ b/asm-to-pil/src/vm_to_constrained.rs
@@ -12,7 +12,7 @@ use powdr_ast::{
         LinkDefinitionStatement, Machine, RegisterDeclarationStatement, RegisterTy, Rom,
     },
     parsed::{
-        asm::InstructionBody,
+        asm::{CallableRef, InstructionBody, Params},
         build::{self, direct_reference, next_reference},
         folder::ExpressionFolder,
         visitor::ExpressionVisitable,
@@ -30,13 +30,7 @@ use crate::common::{instruction_flag, return_instruction, RETURN_NAME};
 pub fn convert_machine<T: FieldElement>(machine: Machine<T>, rom: Option<Rom<T>>) -> Machine<T> {
     let output_count = machine
         .operations()
-        .map(|f| {
-            f.params
-                .outputs
-                .as_ref()
-                .map(|list| list.params.len())
-                .unwrap_or(0)
-        })
+        .map(|f| f.params.outputs.len())
         .max()
         .unwrap_or_default();
     ASMPILConverter::with_output_count(output_count).convert_machine(machine, rom)
@@ -88,22 +82,18 @@ impl<T: FieldElement> ASMPILConverter<T> {
         }
 
         // turn internal instructions into constraints and external ones into links
-        input.links.extend(
-            input
-                .instructions
-                .drain(..)
-                .filter_map(|instr| self.handle_instruction_def(instr)),
-        );
+        for instr in std::mem::take(&mut input.instructions) {
+            self.handle_instruction_def(&mut input, instr);
+        }
 
         // introduce `return` instruction
-        assert!(
-            self.handle_instruction_def(InstructionDefinitionStatement {
+        self.handle_instruction_def(
+            &mut input,
+            InstructionDefinitionStatement {
                 source: SourceRef::unknown(),
                 name: RETURN_NAME.into(),
-                instruction: self.return_instruction()
-            })
-            .is_none(),
-            "return instruction cannot link to an external function"
+                instruction: self.return_instruction(),
+            },
         );
 
         let assignment_registers = self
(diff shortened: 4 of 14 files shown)


test_data/asm/vm_instr_param_mapping.asm
```diff
@@ -0,0 +1,88 @@
...
+
+        addAB_to_A;
+        assert_eq A, 77;
+
+        sub_to_C 6, 3;
+        assert_eq C, 3;
+
+        return;
+    }
+}
```
leonardoalt, line 88: Nice!

linker/src/lib.rs
```diff
@@ -93,31 +93,32 @@ pub fn link<T: FieldElement>(graph: PILGraph<T>) -> Result<PILFile<T>, Vec<Strin
...
+                Vec::with_capacity(1 + from.params.inputs.len() + from.params.outputs.len());
+            //flag
+            expressions.extend(to.operation.id.iter().cloned().map(Expression::Number));
+            // inputs
+            expressions.extend(
+                from.params
+                    .inputs
+                    .into_iter()
+                    .map(|i| {
+                        assert!(i.ty.is_none() || i.ty == Some("write".to_string()));
```
leonardoalt, line 106: isn't there something in the type's type that tell us this, so we don't need to use strings?

linker/src/lib.rs
```diff
@@ -93,31 +93,32 @@ pub fn link<T: FieldElement>(graph: PILGraph<T>) -> Result<PILFile<T>, Vec<Strin
...
+                        (i.name, i.index)
+                    })
+                    .map(|(name, index)| index_access(direct_reference(name), index)),
+            );
+            // outputs
+            expressions.extend(from.params.outputs.into_iter().map(|p| match p.ty {
+                None => index_access(direct_reference(p.name), p.index),
+                // write register as output is mapped as a next ref in the plookup
+                Some(s) if s == "write" => index_access(next_reference(p.name), p.index),
```
leonardoalt, line 115: same here re register's type, I think we shouldn't use strings. There must be something in the type

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -454,6 +456,86 @@ impl<T: FieldElement> ASMPILConverter<T> {
...
+
+            // if rhs is not empty, check it's valid
+            if !rhs.is_empty() {
+                // rhs params must either be assignment registers declared in the lhs or write registers
+                rhs
+                .inputs
+                .iter_mut()
+                .chain(rhs.outputs.iter_mut())
+                .for_each(|p| {
+                    assert!(p.ty.is_none());
```
leonardoalt, line 480: why is this true?

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -454,6 +456,86 @@ impl<T: FieldElement> ASMPILConverter<T> {
...
+                    match reg.ty {
+                        RegisterTy::Assignment => {
+                            let mut lhs_params = lhs.inputs.iter().chain(lhs.outputs.iter());
+                            assert!(lhs_params.any(|p_lhs| p_lhs.name == p.name),
+                                    "assignment register on rhs must be present on lhs params");
+                        }
+                        RegisterTy::Write => {
+                            // the linker needs this information to build the
+                            // plookup when a write register is used as output.
+                            // TODO: less hackish way of passing this info? proper type for param.ty?
```
leonardoalt, line 491: yea agree

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -454,6 +456,86 @@ impl<T: FieldElement> ASMPILConverter<T> {
...
+                            p.ty = Some("write".to_string());
+                        },
+                        _ => panic!("rhs of instruction/link declaration must only use write or assignemtn registers"),
```
leonardoalt, line 494:
```suggestion
                        _ => panic!("rhs of instruction/link declaration must only use write or assignment registers"),
```

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -454,6 +456,86 @@ impl<T: FieldElement> ASMPILConverter<T> {
...
+                    .get_mut(&p.name)
+                    .expect("rhs params must be registers");
+                if reg.ty.is_write() {
+                    let Expression::Reference(ref poly) = l.flag else {
+                        panic!();
+                    };
+                    let flag = poly.try_to_identifier().unwrap();
+                    let reference = direct_reference(flag);
+                    let value = next_reference(&p.name);
+                    reg.conditioned_updates.push((reference, value));
```
leonardoalt, line 533: what does this do?

leonardoalt (commented):
Looks good overall! Left a few small comments.

Bigger comments:

- Can you add some "should panic" tests that cover some of the forbidden cases?
- Can you add a PIL test case where we can see the new constraints that are generated/modified?
- Can you add/modify docs in `book.asm.instructions` and `book.asm.links`?

linker/src/lib.rs
```diff
@@ -93,31 +93,32 @@ pub fn link<T: FieldElement>(graph: PILGraph<T>) -> Result<PILFile<T>, Vec<Strin
             let to = link.to;
 
             // the lhs is `instr_flag { inputs, outputs }`
+            let mut expressions =
+                Vec::with_capacity(1 + from.params.inputs.len() + from.params.outputs.len());
```
chriseth, line 97: I don't think it's worth it reserving the capacity.

ast/src/parsed/asm.rs
```diff
@@ -464,19 +448,20 @@ pub enum MachineStatement<T> {
...
-    pub to: CallableRef,
+    pub to: CallableRef<T>,
 }
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
-pub struct CallableRef {
+pub struct CallableRef<T> {
     pub instance: String,
     pub callable: String,
+    pub params: Params<T>,
```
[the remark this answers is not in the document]
pacheco: yes, this is the result of parsing the RHS of the declaration

linker/src/lib.rs
pacheco: I can remove it, but it doesn't hurt either... I've added it because I broke the single iterator chaining call into a few extends to make it more readable.

asm-to-pil/src/vm_to_constrained.rs
pacheco: This `Param.ty` here comes from parsing text of the form: `arg1: type1, arg2: type2 -> arg3: type3`.
We're currently using this same `Param` type to parse every kind of parameter declaration (instructions and functions etc), even when the `:type` part is not allowed, as is the case here, thus the assertion.
For example, you could try to do `instr a = somevm.func a: label`, which currently gets parsed, but is invalid.

asm-to-pil/src/vm_to_constrained.rs
pacheco: when the instruction is active, `C' = C'`, when it is not active `C'` behaves as usual (depends on everything else)

asm-to-pil/src/vm_to_constrained.rs
leonardoalt: ah I see

asm-to-pil/src/vm_to_constrained.rs
pacheco: I'll do this change on a follow-up PR as it touches some other unrelated stuff.

leonardoalt: can you make an issue so we don't forget

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -454,6 +456,86 @@ impl<T: FieldElement> ASMPILConverter<T> {
         res
     }
```
chriseth, line 458: Could you please document this function? It's quite large and takes `&mut self`

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -454,6 +456,86 @@ impl<T: FieldElement> ASMPILConverter<T> {
...
 
+    fn handle_instr_declaration_params(&mut self, m: &mut Machine<T>) {
+        for l in &mut m.links {
```
chriseth, line 460: I'm a bit lost here. Why is this function called `handle_instr_declaration_params` and then iterates over the links?
Also I think the code would be easier to understand when the iteration is pulled out of the function call, assuming the different links are not interacting (and by pulling that out of the call, you can more easily see that).

airgen/src/lib.rs
```diff
@@ -174,12 +174,24 @@ impl<'a, T: FieldElement> ASMPILConverter<'a, T> {
...
+            params: lhs_params,
+            to:
+                CallableRef {
+                    instance,
+                    callable,
+                    params: rhs_params,
+                },
         }: LinkDefinitionStatement<T>,
     ) -> Link<T> {
+        // Here, we care about actual mapping of columns to target op input/output.
```
chriseth, line 186: Sorry, I still don't understand. Can you explain why we overwrite the lhs params?

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -454,6 +456,86 @@ impl<T: FieldElement> ASMPILConverter<T> {
...
+            lhs.inputs.iter().chain(lhs.outputs.iter()).for_each(|p| {
+                assert!(p.ty.is_none());
+                assert!(self
+                    .registers
+                    .get(&p.name)
+                    .is_some_and(|r| r.ty.is_assignment()));
+            });
+
+            // if rhs is not empty, check it's valid
+            if !rhs.is_empty() {
```
chriseth, line 473: Isn't this redundant if you are iterating over empty lists otherwise?

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -454,6 +456,86 @@ impl<T: FieldElement> ASMPILConverter<T> {
...
+                rhs
+                .inputs
+                .iter_mut()
+                .chain(rhs.outputs.iter_mut())
+                .for_each(|p| {
+                    let reg = self.registers.get(&p.name).expect("rhs params must be registers");
+                    assert!(p.ty.is_none());
+                    match reg.ty {
+                        RegisterTy::Assignment => {
+                            let mut lhs_params = lhs.inputs.iter().chain(lhs.outputs.iter());
```
chriseth, line 484: Maybe it would be useful to have a function that returns an iterator over inputs and outputs?

book/src/asm/instructions.md
```diff
@@ -20,10 +20,50 @@ Instructions feature:
...
+    reg C;
+    ...
+    instr add_to_A X, Y = addvm.add X, Y -> A;          // - output written to a regular register
+    instr addAB -> X = addvm.add A, B -> X;             // - input from the value of regular registers
+    instr addAB_to_C = addvm.add A, B -> C;             // - regular registers used for both inputs and output
+    instr addAB_to_A = addvm.add A, B -> A;             // - read from and then write to the same regular register
+    instr sub_from_add X, Y -> Z = addvm.add Y, Z -> X; // - find input for a given output (doesn't always work!)
+    ...
+}
 ```
```
chriseth, line 67: I think we should also mention somewhere that assignment registers have to be "repeated" on the lhs while non-assignment registers can only appear on the rhs. Not sure how to say that in an understandable way, though :)

asm-to-pil/src/vm_to_constrained.rs
pacheco: This also checks that all left side (assignment) registers are used on the right side.
If right side is empty (syntactic sugar for `lhs = rhs`), this check is not made.

asm-to-pil/src/vm_to_constrained.rs
pacheco: Yeah this was not great. 
I've restructured the code handling instruction definitions (this is now part of handling external instructions), should be clearer now what's going on.

airgen/src/lib.rs
pacheco: I've improved the comment here.
Essentially, what matters for the lookup is the RHS, as it may have extra (write) registers, reorderings etc.
The RHS being empty is just syntactic sugar for RHS = LHS.

asm-to-pil/src/vm_to_constrained.rs
pacheco: Good idea

asm-to-pil/src/vm_to_constrained.rs
pacheco: should be better now after restructuring the code

airgen/src/lib.rs
```diff
@@ -174,12 +174,26 @@ impl<'a, T: FieldElement> ASMPILConverter<'a, T> {
...
     ) -> Link<T> {
+        // The actual mapping of parameters used for the link constraint comes
+        // from the RHS of the instr/link declaration (as it allows for extra
+        // registers/columns being used and parameter reordering).
+        // We allow declarations with an empty RHS as syntactic sugar for when RHS = LHS.
+        let params = if !rhs_params.is_empty() {
+            rhs_params
+        } else {
+            lhs_params
+        };
```
chriseth, line 194: This might be nitpicking, but if there is no compelling reason to do otherwise, I often try to avoid negations and instead switch the both cases, because statements without negations are usually easier to understand:
```suggestion
        let params = if rhs_params.is_empty() {
            lhs_params
        } else {
            rhs_params
        };
```

analysis/src/vm/inference.rs
```diff
@@ -50,13 +50,13 @@ fn infer_machine<T: FieldElement>(mut machine: Machine<T>) -> Result<Machine<T>,
...
+                            .unwrap_or_else(|| panic!("invalid instruction: {}", instr_name));
 
-                        let outputs = def.instruction.params.outputs.clone().unwrap_or_default();
-
-                        outputs
+                        def.instruction
                             .params
+                            .outputs
                             .iter()
+                            .cloned()
```
chriseth, line 59: Do we really need the `cloned()` here?

asm-to-pil/src/romgen.rs
```diff
@@ -146,38 +140,33 @@ pub fn generate_machine_rom<T: FieldElement>(
             let input_substitution = function
```
chriseth, line 140: `.params.inputs.params.iter()` or similar call chains appear very often in this code. Could we have a functions `input_params` / `output_params` that return an iterators or something? (could be done in another pr)

book/src/asm/instructions.md
```diff
@@ -20,10 +20,54 @@ Instructions feature:
...
+    reg Y[<=];
+    reg Z[<=];
+
+    instr add X, Y -> Z = submachine.add; // will call submachine.add(X, Y) and return through Z
+    // this is equivalent to:
+    // instr add X, Y -> Z = submachine.add X, Y -> Z;
+    ...
+}
+```
+The left-hand side of the definition declares the instruction parameters, that is, how the instruction is called in the local machine.
```
chriseth, line 48:
```suggestion
The left-hand side of the definition states which local registers to use for calling the instruction.
```
(in my understanding, "declare" introduces a new symbol)

book/src/asm/instructions.md
```diff
@@ -20,10 +20,54 @@ Instructions feature:
...
+The right-hand side of the definition declares the external operation being called, and how the instruction parameters are used in that call.
```
chriseth, line 49:
```suggestion
The right-hand side of the definition specifies  the external operation being called, and how the instruction parameters are used in that call.
```

chriseth: Hm, I fear this still does not really catch the essence...
But I also don't have a better suggestion right now.

chriseth: Maybe something along the lines of "left hand side defines the name of the local instruction and and its parameters, who all need to be assignment registers. The right hand side specifies the name of the sub-machine instruction and how to call it. Here, in addition to the assignment registers used on the left hand side, non-assignment registers can be used. Those can be seen as implicit parameters."

book/src/asm/instructions.md
```diff
@@ -20,10 +20,54 @@ Instructions feature:
...
+    reg A;
+    reg B;
+    reg C;
+    ...
+    instr add_to_A X, Y = addvm.add X, Y -> A;          // - output written to a regular register
+    instr addAB -> X = addvm.add A, B -> X;             // - input from the value of regular registers
+    instr addAB_to_C = addvm.add A, B -> C;             // - regular registers used for both inputs and output
+    instr addAB_to_A = addvm.add A, B -> A;             // - read from and then write to the same regular register
+    instr sub_from_add X, Y -> Z = addvm.add Y, Z -> X; // - find input for a given output (doesn't always work!)
+    ...
```
chriseth, line 69: Maybe also add examples how the instruction are used.

book/src/asm/instructions.md
pacheco: reworked the text and expanded the example

analysis/src/vm/inference.rs
pacheco: removed

PR #1049, opened 2024-02-12
Serialize optimized PIL and load it for the prove phase

Champii:
This PR adds serde for  `Analyzed<T>` and every included data structures (as well as the `FieldElement` trait).

This aims to avoid a second analyze/optimize phase when calling `powdr prove`.
In order to do so, `powdr prove` can now accept the serialized `Analyzed<T>` as `{name}_opt.cbor` and start from there.

Note: This adds a new constraint, namely to use the same field for both phases. There should be a way to check that at runtime and print a nice error message, because actually it just fails to deserialize with an obscure serde message like `failed to fill the whole buffer`. Any input on that would be greatly appreciated :)

### DIFF
--- a/ast/src/parsed/mod.rs
+++ b/ast/src/parsed/mod.rs
@@ -11,6 +11,7 @@ use std::{
 };
 
 use powdr_number::{DegreeType, FieldElement};
+use serde::{Deserialize, Serialize};
 
 use self::asm::{Part, SymbolPath};
 use crate::SourceRef;
@@ -151,7 +152,7 @@ impl<T> PilStatement<T> {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
 pub struct SelectedExpressions<Expr> {
     pub selector: Option<Expr>,
     pub expressions: Vec<Expr>,
@@ -178,7 +179,7 @@ impl<Expr> SelectedExpressions<Expr> {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
 pub enum Expression<T, Ref = NamespacedPolynomialReference> {
     Reference(Ref),
     PublicReference(String),
@@ -292,18 +293,18 @@ impl NamespacedPolynomialReference {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct LambdaExpression<T, Ref = NamespacedPolynomialReference> {
     pub params: Vec<String>,
     pub body: Box<Expression<T, Ref>>,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct ArrayLiteral<T, Ref = NamespacedPolynomialReference> {
     pub items: Vec<Expression<T, Ref>>,
 }
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Serialize, Deserialize)]
 pub enum UnaryOperator {
     Minus,
     LogicalNot,
@@ -320,7 +321,7 @@ impl UnaryOperator {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Serialize, Deserialize)]
 pub enum BinaryOperator {
     Add,
     Sub,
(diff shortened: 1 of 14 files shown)


number/src/traits.rs
```diff
@@ -1,6 +1,7 @@
 use std::{fmt, hash::Hash, ops::*, str::FromStr};
 
 use num_traits::{One, Zero};
+use serde::{de::DeserializeOwned, Serialize};
```
leonardoalt, line 4: Is there a problem that the other structs use `Deserialize` and this one uses `DeserializeOwned`?

pipeline/src/pipeline.rs
```diff
@@ -455,6 +475,14 @@ impl<T: FieldElement> Pipeline<T> {
         path.file_stem().unwrap().to_str().unwrap().to_string()
     }
 
+    // This is used for parsing file names than ends with '_{suffix}'
+    fn name_from_path_with_suffix(path: &Path) -> String {
+        let file_name = Self::name_from_path(path);
+        let mut splitted = file_name.split("_").collect::<Vec<_>>();
```
leonardoalt, line 481:
```suggestion
        let mut split = file_name.split("_").collect::<Vec<_>>();
```

ast/src/parsed/mod.rs
```diff
@@ -151,7 +152,7 @@ impl<T> PilStatement<T> {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
```
leonardoalt, line 155: Are you sure the structs in this file also need serde? I'd imagine only `analyzed/*` would need it for `Analyzed<T>`, but also not sure right now

leonardoalt (commented):
Really nice, thanks! Left some small comments

number/src/traits.rs
Champii: To use `Deserialize` on this trait would introduce a lifetime `'de` that would propagate on every use of the `FieldElement<'de>`, because here we add a new bound to the trait, but we don't actually implement `Deserialize` for it

pipeline/src/pipeline.rs
Champii: Good catch, thanks :)

number/src/traits.rs
leonardoalt: Yea I've had to use this before too, just not sure if it conflicts/works well with the others

ast/src/parsed/mod.rs
Champii: Because the `Analyzed<T>` uses some of those directly or indirectly, we definitely have to make them serializable yes :)
For example, the `definitions` field of that struct contains a `FunctionValueDefinition<T>` that itself contains a `Expression<T>` in the `parsed/*` folder (and thus also all of its children)

leonardoalt:
Can you also add a test (maybe in Pipeline), that de(ser(pil)) == pil ?

number/src/traits.rs
Champii: It shouldn't, it just indicates that every implementors of `FieldElement` should have owned data (If I recall correctly)

pipeline/src/pipeline.rs
```diff
@@ -398,6 +398,26 @@ impl<T: FieldElement> Pipeline<T> {
         }
     }
 
+    pub fn from_maybe_optimized_file(self, file: PathBuf) -> Self {
+        if file.extension().unwrap() == "cbor" {
```
mratsim, line 402: I think we should use a non-generic file extension, and one that allows being grepped also by `*.pil*`
so either `.pilo` (PIL optimized or PIL object) or `.pilb` (PIL binary).

pipeline/src/pipeline.rs
```diff
@@ -677,6 +706,18 @@ impl<T: FieldElement> Pipeline<T> {
         Ok(())
     }
 
+    fn maybe_write_serialized_pil(
+        &self,
+        pil: &Analyzed<T>,
+        suffix: &str,
+    ) -> Result<(), Vec<String>> {
+        if let Some(path) = self.path_if_should_write(|name| format!("{name}{suffix}.cbor"))? {
```
mratsim, line 714: Beyond the change of file extension,
we might want to add a magic header and a version if we ever want forward compatibility or different PIL compilers like Polygon to add a loader for our own PIL.

pipeline/src/pipeline.rs
mratsim: Or if temporaries that people aren't supposed to use `.pilo` (PIL object), `.pila` (PIL artifact), `.pilt` (PIL temporary), `.pili` (PIL intermediate), ...

mratsim (approved):
LGTM, apart from the extension

This raises the following design question once Powdr is used in the wild:
- Do we expect people to save the CBOR serialized code or is it used internally only.

In the second case, we're fine as-is.

In the first case we likely want at least a magic header and a version number.
And to ease on the versioning burden, we might want to use something that can generate a serde schema alongside of the file in CI. And we check in CI that if the schema generated is different from the previous commit/PR, we need to bump the version. For example something like https://github.com/tailhook/probor (it's unmaintained but I would expect someone wrote something similar since 6 years ago).

pipeline/src/pipeline.rs
leonardoalt: Makes sense, `.pilo` or `.pil.cbor` sounds best to me, but the others are fine too

pipeline/src/pipeline.rs
leonardoalt: Agree, though I think we can do it in a different PR.

leonardoalt:
> Do we expect people to save the CBOR serialized code or is it used internally only?

I think it makes sense for people to keep the serialized code, for continuations, verification, etc.
Agree with the magic number and version in the binary. I think we can do that in a follow-up PR.

Champii:
I was thinking about a way to indicate the version of the binary blob in case of breaking changes yes.
I will do some research on existing schema generators and craft something in a new PR.
In the meantime, can we agree on the extension ? I like `.pilo` and it seems like you both do too :)

leonardoalt:
Sounds good!

pipeline/src/pipeline.rs
```diff
@@ -398,6 +398,26 @@ impl<T: FieldElement> Pipeline<T> {
...
 
+    pub fn from_maybe_optimized_file(self, file: PathBuf) -> Self {
+        if file.extension().unwrap() == "pilo" {
+            self.from_optimized_pil_file(file)
+        } else {
+            self.from_file(file)
+        }
+    }
+
+    pub fn from_optimized_pil_file(self, pil_file: PathBuf) -> Self {
```
leonardoalt, line 409: Sorry for being pedantic, but since we're changing names...

```suggestion
    pub fn from_pil_object(self, pil_file: PathBuf) -> Self {
```

pipeline/src/pipeline.rs
```diff
@@ -677,6 +706,18 @@ impl<T: FieldElement> Pipeline<T> {
         Ok(())
     }
 
+    fn maybe_write_serialized_pil(
```
leonardoalt, line 709:
```suggestion
    fn maybe_write_pil_object(
```

pipeline/src/pipeline.rs
Champii: Haha no problem, I'm on it

Champii: I should maybe also change the `from_maybe_optimized_file` to `from_maybe_pil_object` ?

Champii:
All done !

leonardoalt (approved):

PR #1052, opened 2024-02-12
daily rust cache for PR-tests action

pacheco:


lvella (commented):
Why not do it in the already existing daily job? What are the cons?

lvella:
Or maybe make the daily tests depend on the daily cache build?

pacheco:
The daily tests build with `release`, we have a specific profile for PR tests.
Also, this way its possible to rebuild the cache without also running release tests.

lvella (approved):

PR #1057, opened 2024-02-14
Pilo versionning and checks

Champii:
This PR introduces schemas for `Analyzed<T>` and some checks on the validity of the binary format.

It uses `schemars` to produce a `JsonSchema` of the structure, and adds an executable (`powdr-schema`) that performs a diff between the current code structure and the one previously computed and then bumps the version number. This executable is meant to be called from a CI process automatically.

To perform the checks, this PR introduces a wrapper around `Analyzed<T>` that keeps track of a magic number, the version of the code structure and the type of the `FieldElement` used. 

Possible improvements:

 - We could also store a hash of the generated `JsonSchema` instead of the whole structure to save a few bytes (12kb at the time of writing)
 - Add some tests for the schema generation and the version bumping


leonardoalt:
Just a quick comment before I check the whole thing: I think this shouldn't be in the ast crate, but rather in some more externally facing crate. The AST is ofc used for the binary but it's rather internal. I think it should be in one of these crates:

- powdr
- pipeline
- its own crate? This is actually my fav option rn I think

number/src/macros.rs
```diff
@@ -23,12 +23,14 @@ macro_rules! powdr_field {
...
             Deserialize,
+            JsonSchema,
         )]
         pub struct $name {
             #[serde(
                 serialize_with = "crate::serialize::ark_se",
                 deserialize_with = "crate::serialize::ark_de"
             )]
+            #[schemars(skip)]
             value: $ark_type,
```
Champii, line 34: I cannot find any way to include this field in the `JsonSchema`, I'm wondering if this would be safe to skip it altogether considering how deep it is tight to the types in `powdr`. It means that when upgrading to another major `ark` version that breaks backward compatibility, one would have to bump the `analyzed_type.version` file manually.
It might be an edge case that is not worth pursuing but I'm not really sure

ast/src/analyzed/serialized.rs
```diff
@@ -0,0 +1,84 @@
+use std::path::PathBuf;
+
+use powdr_number::{FieldElement, KnownField};
+use serde::{Deserialize, Serialize};
+
+use super::Analyzed;
+
+const MAGIC: u32 = 0x4c_41_50_42;
```
Champii, line 8: This MAGIC number might be useless in the end, as the deserialization will fail anyway if the file is not in the good format.

mratsim: But if Powdr becomes so important that file explorer wants to display something about the file, it will help ;).

number/src/macros.rs
mratsim: I think it's OK.

There are limited ways to store BigInt / FieldElement / Elliptic curves and the serialization they use should be stable due to the size of their ecosystem.

They just dump bigint in little-endian: https://github.com/arkworks-rs/algebra/blob/cc2ad8c368c323ee9d9aacaabf8e353de95682e6/serialize/src/impls.rs#L170-L181

mratsim: And coincidentally Halo2 also uses little-endian.

mratsim (approved):
LGTM, awesome.

number/src/macros.rs
leonardoalt: Yea little-endian is rather common across the board

ast/src/analyzed/serialized.rs
leonardoalt: how did you pick the bytes

Champii: Randomly, and checked if they existed in a repository of magic files. 
Any other series of bytes will do if you have a specific idea :)

ast/src/analyzed/mod.rs
```diff
@@ -19,7 +21,7 @@ use crate::SourceRef;
 
 use self::types::TypedExpression;
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
```
leonardoalt, line 24: I'm a bit wary of all these changes in the ast crate, that are specific to the versioning schema. Can't we just use the serialization that is already available for Analyzed, and only apply the schema stuff in the powdr-schema crate?

ast/src/analyzed/serialized.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct SerializedAnalyzed {
+    magic: u32,
+    version: u32,
+    field: KnownField,
+    analyzed: Vec<u8>,
+}
+
```
leonardoalt, line 17: Similarly to my other comment, ideally we don't need to touch the AST crate

ast/src/analyzed/mod.rs
Champii: It would be theoretically possible, but it means using [serde-reflection](https://github.com/zefchain/serde-reflection) that requires to do the reflection at runtime with a given value, and because of the nature of enums and how they are serialized, it means to keep a piece of code somewhere with every enums present in the serialized data. 
This defeat the purpose of automatically-generated schema, as one has to keep this collection of enums up to date like [here](https://github.com/diem/diem/blob/main/testsuite/generate-format/src/diem.rs#L47) for example

ast/src/analyzed/serialized.rs
Champii: For futur references, the UNIX style magic format for this particular file can be defined as such:
```
8       belong          0x4c415042      Powdr PIL binary object
```

ast/src/analyzed/mod.rs
Champii: Note that `schemars` also have a similar runtime-based reflection feature, but it suffers from the same drawbacks
[https://graham.cool/schemars/examples/9-from_value/](https://graham.cool/schemars/examples/9-from_value/)

leonardoalt: Right, ok. I think our usage of `Analyzed` as the "type of the circuit constraints" isn't great in the first place, so let's keep it like this for now and maybe one day we can refactor it into a better circuit data structure.

ast/src/analyzed/serialized.rs
leonardoalt: we could use 0x706f776472 for `powdr` :)

leonardoalt: wait that's too many bytes

leonardoalt: yea the original sequence is fine!

Champii: > we could use 0x706f776472 for `powdr` :)

Haha that was my first idea also, we can definitely use it tho, because we can define magic files based on strings. Your call :p

Champii: Ok I got it working, pushing right now

leonardoalt (approved):

PR #1058, opened 2024-02-14
Some more uses of cross_product

chriseth:
I also changed cross_product in the way it iterates through the rows. Now the first array element always updates and the last array elements updates when all the others have completed a cycle.

### DIFF
--- a/std/shift.asm
+++ b/std/shift.asm
@@ -1,4 +1,5 @@
 use std::utils::unchanged_until;
+use std::utils::cross_product;
 use std::convert::int;
 
 machine Shift(latch, operation_id) {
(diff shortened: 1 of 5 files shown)


std/shift.asm
```diff
@@ -14,14 +15,19 @@ machine Shift(latch, operation_id) {
...
-    col fixed P_A(i) { i % 256 };
-    col fixed P_B(i) { (i / 256) % 32 };
-    col fixed P_ROW(i) { (i / (256 * 32)) % 4 };
-    col fixed P_operation(i) { (i / (256 * 32 * 4)) % 2 };
+    let inputs = cross_product([256, 32, 4, 2]);
+    let a: int -> int = inputs[0];
+    let b: int -> int = inputs[1];
+    let row: int -> int = inputs[2];
+    let op: int -> int = inputs[3];
+    let P_A: col = a;
```
leonardoalt, line 23: how does this conversion work, if `P_A` needs to return a `fe`?

chriseth: That's one of the very few leftover implicit conversion rules: `col` is not really a type, it is more like a marker. We could also say that we go back to `col P_A;` to declare witness columns and `col P_A = ...` for fixed columns.

Having said that:

If you declare a symbol of type `col`, then the value of that declaration can either be `int -> int` or `int -> fe`.

If you reference a symbol of type `col`, the type of that reference will be `expr`.

leonardoalt: I think `let ... col` is better

leonardoalt (approved):

PR #1059, opened 2024-02-14
balign and p2align pad value

pacheco:


asm-utils/src/data_storage.rs
```diff
@@ -65,10 +65,11 @@ impl<'a, 'b> WordWriter<'a, 'b> {
         self.current_pos = next_pos;
     }
 
-    fn align(&mut self, alignment: u32) {
+    fn align(&mut self, alignment: u32, pad_value: u8) {
         let padding_size = alignment_size(self.current_pos as usize, alignment as usize);
         if padding_size != 0 {
-            self.advance(padding_size as u32);
+            let bytes = vec![pad_value; padding_size];
+            self.write_bytes(&bytes);
```
lvella, line 72: My ideal solution: implement a `write_repeated(value, number_of_times)` and used it instead of allocating a `Vec` for this.

Less than ideal, but IMO better: check the common case of `pad_value == 0` and use `advance()` in this case. At should avoid the `Vec` most of the time.

lvella: Maybe `write_bytes()` should take an `IntoIter` instead of `&[u8]`, so that it would work for this case, too. Then you also would need to check for the special case of `pad_value == 0` and `advance()` in this case.

pacheco: went for the iterator route

PR #1072, opened 2024-02-19
Registers in the riscv-executor can hold fields.

lvella:


leonardoalt:
tests failing

riscv-executor/src/lib.rs
```diff
@@ -30,62 +34,105 @@ pub mod poseidon_gl;
...
+impl<F: FieldElement> Elem<F> {
+    /// Interprets the value of a field as a binary, if it can be represented either as a
+    /// u32 or a i32.
+    ///
+    /// Panics otherwise.
+    pub fn new_from_fe_as_bin(value: &F) -> Self {
+        if let Some(v) = value.to_integer().try_into_u32() {
+            Self::Binary(v as i64)
+        } else if let Some(v) = value.try_into_i32() {
+            Self::Binary(v as i64)
```
leonardoalt, line 54: if the input is F is fits into i32, why wouldn't it have fit u32 above?

riscv-executor/src/lib.rs
```diff
@@ -30,62 +34,105 @@ pub mod poseidon_gl;
...
     }
 }
 
-impl From<u64> for Elem {
+/*impl<F: FieldElement> From<u64> for Elem<F> {
     fn from(value: u64) -> Self {
         Self(value as i64)
     }
-}
+}*/
```
leonardoalt, line 108: Remove?

riscv-executor/src/lib.rs
```diff
@@ -30,62 +34,105 @@ pub mod poseidon_gl;
...
 }
 
-impl From<i64> for Elem {
+/*
+impl<F: FieldElement> From<i64> for Elem<F> {
     fn from(value: i64) -> Self {
         Self(value)
     }
 }
+*/
```
leonardoalt, line 122: Remove?

riscv-executor/src/lib.rs
```diff
@@ -807,10 +854,10 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
 pub fn execute_ast<T: FieldElement>(
     program: &AnalysisASMFile<T>,
     inputs: &Callback<T>,
-    bootloader_inputs: &[T],
+    bootloader_inputs: &[Elem<T>],
```
leonardoalt, line 857: Since `Elem` is only used here, wouldn't it be easier to take in `&[T]` and build the `Elem`s here?

riscv-executor/src/lib.rs
lvella: If it is negative, it doesn't match `u32`, and if it is >= 2**31, it doesn't match `i32`.

riscv-executor/src/lib.rs
lvella: I did like this to preserve the type across chunk executions. I.e. imagine an execution was split between a `poseidon_gl` and a `split_gl`. The "fieldness" of the register containing the poseidon result must be preserved.

riscv-executor/src/lib.rs
leonardoalt: but if the input is F, how is it negative?

riscv/src/coprocessors.rs
```diff
@@ -271,19 +271,22 @@ pub fn call_every_submachine(coprocessors: &CoProcessors) -> Vec<String> {
     if coprocessors.has(SHIFT_COPROCESSOR.name) {
         calls.push("x10 <== shl(x10, x10);".to_string());
     }
-    if coprocessors.has(SPLIT_GL_COPROCESSOR.name) {
-        calls.push("x10, x11 <== split_gl(x10);".to_string());
-    }
     if coprocessors.has(POSEIDON_GL_COPROCESSOR.name) {
         calls.extend(vec![
-            "P0, P1, P2, P3 <== poseidon_gl(P0, P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11);"
+            "P0, P1, P2, x10 <== poseidon_gl(P0, P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11);"
```
leonardoalt, line 276: why?

riscv/src/coprocessors.rs
```diff
@@ -271,19 +271,22 @@ pub fn call_every_submachine(coprocessors: &CoProcessors) -> Vec<String> {
...
             "P0 <=X= 0;".to_string(),
             "P1 <=X= 0;".to_string(),
             "P2 <=X= 0;".to_string(),
-            "P3 <=X= 0;".to_string(),
         ]);
     }
+    if coprocessors.has(SPLIT_GL_COPROCESSOR.name) {
+        // SplitGL only makes sense when used with PoseidonGL, because there is
+        // no other way to get a value in the full range of Goldilocks into a
+        // register of the RISC-V zkVM.
```
leonardoalt, line 286: ah because of this

leonardoalt: I don't think this really matters since this piece of code is used only for the dummy block and isn't really testing anything

leonardoalt (approved):

lvella:
I thought you would tell me to squash before merging.

PR #1080, opened 2024-02-21
Add type schemes.

chriseth:


ast/src/analyzed/display.rs
```diff
@@ -138,16 +159,20 @@ impl<T: Display> Display for FunctionValueDefinition<T> {
...
+            }) => format_outer_function(e, None, f),
+            FunctionValueDefinition::Expression(TypedExpression {
+                e,
+                type_scheme: Some(ty),
+            }) if *ty == Type::Col.into() => format_outer_function(e, None, f),
+            FunctionValueDefinition::Expression(TypedExpression {
+                e,
+                type_scheme: Some(ts),
+            }) => {
+                // TODO format type vars
```
chriseth, line 174: todo

executor/src/constant_evaluator/mod.rs
```diff
@@ -138,15 +141,18 @@ struct Symbols<'a, T> {
 }
 
 impl<'a, T: FieldElement> SymbolLookup<'a, T, FixedColumnRef<'a>> for Symbols<'a, T> {
-    fn lookup(&self, name: &str) -> Result<Value<'a, T, FixedColumnRef<'a>>, EvalError> {
+    fn lookup<'b>(&self, name: &str) -> Result<Value<'a, T, FixedColumnRef<'a>>, EvalError> {
         Ok(
             if let Some((name, _)) = self.computed_columns.get_key_value(name) {
+                // TODO it is actually not possible anymore to reference another fixed column,
```
chriseth, line 147: TODO check if we can move this into the next PR

test_data/pil/book/declarations.pil
```diff
@@ -3,15 +3,13 @@ namespace Main(16);
     // This defines a constant
     let rows = 2**16;
     // This defines a fixed column that contains the row number in each row.
-    let step = |i| i;
-    // This defines a copy of the column, also a fixed column because the type
-    // is explicitly specified.
-    let also_step: col = step;
+    // Only symbols whole type is "col" are considered fixed columns.
```
leonardoalt, line 6:
```suggestion
    // Only symbols whose type is "col" are considered fixed columns.
```

ast/src/parsed/display.rs
```diff
@@ -649,6 +658,43 @@ impl<E: Display> Display for FunctionTypeName<E> {
...
 }
 
+pub fn format_type_scheme_around_name<E: Display>(
+    name: &str,
+    type_scheme: &Option<TypeScheme<E>>,
+) -> String {
+    if let Some(type_scheme) = type_scheme {
+        format!(
+            "{} {name}: {}",
+            type_scheme.type_vars_to_string(),
```
leonardoalt, line 668: is this the generics?

ast/src/parsed/display.rs
```diff
@@ -649,6 +658,43 @@ impl<E: Display> Display for FunctionTypeName<E> {
...
+            self.0
+                .iter()
+                .map(|(v, b)| {
+                    format!(
+                        "{v}{}",
+                        if b.is_empty() {
+                            String::new()
+                        } else {
+                            format!(": {}", b.iter().join(" + "))
+                        }
```
leonardoalt, line 690: beautiful nested madness

ast/src/asm_analysis/display.rs
```diff
@@ -47,13 +47,17 @@ impl<T: Display> Display for AnalysisASMFile<T> {
                 Item::Machine(machine) => {
                     write_indented_by(f, format!("machine {name}{machine}"), current_path.len())?;
                 }
-                Item::Expression(ExpressionWithTypeName { e, type_name }) => write_indented_by(
+                Item::Expression(ExpressionWithTypeScheme { e, type_scheme }) => write_indented_by(
                     f,
                     format!(
-                        "let {name}{} = {e};\n",
-                        type_name
+                        "let{} {name}{} = {e};\n",
```
leonardoalt, line 53: This is done >=3 times

ast/src/analyzed/types.rs
```diff
@@ -116,25 +198,86 @@ pub struct FunctionType {
...
+            0 => return self,
+            1 => {
+                let var = self.vars.vars().next().unwrap();
+                [(var.clone(), "T".to_string())].into()
+            }
+            _ => self
+                .vars
+                .vars()
+                .enumerate()
+                .map(|(i, v)| ((*v).clone(), format!("T{}", i + 1)))
```
leonardoalt, line 232: why do we need these concrete names?

ast/src/analyzed/types.rs
```diff
@@ -116,25 +198,86 @@ pub struct FunctionType {
...
+            ty,
+        }
+    }
+
+    /// Formats the type variables and bounds part.
+    pub fn type_vars_to_string(&self) -> String {
+        if self.vars.is_empty() {
+            String::new()
+        } else {
+            format!("<{}>", self.vars)
```
leonardoalt, line 258: does the parser support this already?

parser/src/lib.rs
```diff
@@ -72,6 +75,22 @@ pub fn parse_module<'a, T: FieldElement>(
...
+    let ctx = ParserContext::new(None, input);
+    TYPE_NAME_PARSER
+        .parse(&ctx, input)
+        .map_err(|err| handle_parse_error(err, None, input))
+}
+
+pub fn parse_type_var_bounds(input: &str) -> Result<TypeBounds, ParseError<'_>> {
+    let ctx = ParserContext::new(None, input);
+    TYPE_VAR_BOUNDS_PARSER
+        .parse::<GoldilocksField>(&ctx, input)
```
leonardoalt, line 90: why Goldilocks?

pil-analyzer/src/statement_processor.rs
```diff
@@ -437,20 +449,34 @@ where
     fn resolve_type_name(&self, mut n: TypeName<parsed::Expression<T>>) -> Result<Type, EvalError> {
         // Replace all expressions by number literals.
         for e in n.expressions_mut() {
-            let v = self.evaluate_expression(e.clone())?;
-            *e = parsed::Expression::Number(v);
+            let v = self.evaluate_expression_to_int(e.clone())?;
+            let v_u64 = v.to_u64().ok_or(EvalError::TypeError(format!(
```
leonardoalt, line 453: why u64?

leonardoalt (commented):
A few questions and small comments.

If fixed columns have to be explicitly declared as `col`, what happens in a case like this?

```
let x = |i| i;
col witness x = i + 2;
```

ast/src/analyzed/types.rs
chriseth: This is used for error reporting or if two type schemes are the same up to renaming of type variables.

ast/src/analyzed/types.rs
chriseth: Yes, the grammar symbol is "GenericTypedName"

ast/src/asm_analysis/display.rs
chriseth: Fixed in https://github.com/powdr-labs/powdr/pull/1083

ast/src/parsed/display.rs
chriseth: Yes

parser/src/lib.rs
chriseth: True, this should be generic. We really need to change the grammar to use integers instead...

pil-analyzer/src/statement_processor.rs
chriseth: This is a number literal inside a type scheme, so it must be an array length. Will add a comment.

ast/src/parsed/display.rs
leonardoalt: Nah not too important here

parser/src/lib.rs
leonardoalt: But what happens with Bn then?

leonardoalt (approved):

PR #1086, opened 2024-02-22
Small fixes for witgen performance

pacheco:
This mainly removes useless allocations.
Improvement in benchmarks is `kekkac` by ~9% and `many_chunks` by around ~15%


executor/src/witgen/machines/fixed_lookup_machine.rs
```diff
@@ -207,11 +208,12 @@ impl<T: FieldElement> FixedLookup<T> {
...
         // get the values of the fixed columns
-        let right = right
+        let mut right = right
             .expressions
             .iter()
-            .map(try_to_simple_poly_ref)
-            .collect::<Option<Vec<_>>>()?;
+            .filter_map(try_to_simple_poly_ref)
+            .peekable();
+        right.peek()?;
```
chriseth, line 216: Can you explain what the purpose of this peek call is?

ast/src/analyzed/mod.rs
```diff
@@ -841,6 +842,13 @@ impl From<&Symbol> for PolyID {
     }
 }
 
+impl Hash for PolyID {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        // Only use `id` for hashing
```
chriseth, line 847: If we do this, the hash gets incompatible with Eq, and we should take extra care never to put any polynomials of different kinds in the same hash set/map. Can we guarantee that?

executor/src/witgen/machines/fixed_lookup_machine.rs
pacheco: The previous behaviour was:
- collect an iterator of Option<T> into an Option<Vec<T>>, then early return if None
Here, we use peekable to test that the iterator is non-empty without consuming.
Peek is also needed in the internal function, which relied on Vec::first() for a few tests.

chriseth: Can you add a short comment in front of that line?

ast/src/analyzed/mod.rs
chriseth: Oh right! Can you add a quick comment about that?

pacheco: ok, so actually did it a bit differently to keep the old behavior and save the extra hash call

chriseth (approved):

PR #1098, opened 2024-02-26
use `Cow` for returning `nonzero_coefficients`

pacheco:
- with `Cow` we can return a reference to coefficients directly when the affine is `clean`
- also removed the `nonzero_variables` method which as used in a single place and required another cloning


georgwiese (approved):
Nice!

PR #1104, opened 2024-02-28
AffineExpression enum (improve executor benchmarks ~13/23%)

pacheco:
Internally represent as an enum of Constant | One | Many, to avoid allocations in the common case of a single variable.
Also rewrote `ExpressionEvaluator::evaluate_binary_operation`, allowing it to shortcut earlier on Err when possible.

### DIFF
--- a/executor/src/witgen/affine_expression.rs
+++ b/executor/src/witgen/affine_expression.rs
@@ -1,4 +1,3 @@
-use std::borrow::Cow;
 use std::fmt::Display;
 
 use itertools::{Either, Itertools};
@@ -13,12 +12,11 @@ use super::{EvalError::*, EvalResult, EvalValue, IncompleteCause};
 
 /// An expression affine in the committed polynomials (or symbolic variables in general).
 #[derive(Debug, Clone)]
-pub struct AffineExpression<K, T> {
-    pub coefficients: Vec<(K, T)>,
-    pub offset: T,
-    /// If true, all coefficients have nonzero values and there are no duplicate keys
-    /// (a constant affine expression is always clean).
-    clean: bool,
+pub enum AffineExpression<K, T> {
+    Constant(T),
+    OneVar((K, T), T),
+    /// Many contains 1 or more variables, in variable order, without duplicates or zero coefficients
+    ManyVars(Vec<(K, T)>, T),
 }
 
 pub type AffineResult<K, T> = Result<AffineExpression<K, T>, IncompleteCause<K>>;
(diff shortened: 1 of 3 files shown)


executor/src/witgen/affine_expression.rs
```diff
@@ -444,28 +421,110 @@ where
...
-            self.clean = !rhs.coefficients.iter().any(|(k, _)| k == &lk);
-            self.coefficients = rhs.coefficients;
-            self.coefficients.push((lk, lv));
-        } else if let [(rk, rv)] = rhs.coefficients[..] {
-            self.clean = !self.coefficients.iter().any(|(k, _)| k == &rk);
-            self.coefficients.push((rk, rv));
-        } else {
-            self.coefficients.append(&mut rhs.coefficients);
-            self.clean = false;
+    fn add(self, rhs: Self) -> Self::Output {
```
pacheco, line 424: I hate the auto formatting here... is there some annotation to prevent it?

executor/src/witgen/affine_expression.rs
```diff
@@ -13,128 +12,116 @@ use super::{EvalError::*, EvalResult, EvalValue, IncompleteCause};
...
 #[derive(Debug, Clone)]
-pub struct AffineExpression<K, T> {
-    pub coefficients: Vec<(K, T)>,
-    pub offset: T,
-    /// If true, all coefficients have nonzero values and there are no duplicate keys
-    /// (a constant affine expression is always clean).
-    clean: bool,
+pub enum AffineExpression<K, T> {
+    Constant(T),
+    One((K, T), T),
```
chriseth, line 17: Maybe `OneVar`?

chriseth: Did you try using `Option<T>` for the second `T` (optimizing the case where it is zero)?

executor/src/witgen/affine_expression.rs
```diff
@@ -13,128 +12,116 @@ use super::{EvalError::*, EvalResult, EvalValue, IncompleteCause};
...
+    /// @returns the nonzero coefficients and their variable IDs (but not the offset).
+    /// The coefficients are ordered by variable
+    pub fn nonzero_coefficients(&self) -> impl Iterator<Item = &(K, T)> {
+        match self {
+            AffineExpression::Many {
+                coefficients,
+                offset: _,
+            } => {
+                debug_assert!(Self::is_proper_coefficients(coefficients));
+                Either::Right(Either::Left(coefficients.iter()))
```
chriseth, line 88: Hm, maybe I'm not too familiar with `either`, but what does this do? Is the idea to unify the returned type? Is there not a more obvious way?

executor/src/witgen/affine_expression.rs
```diff
@@ -266,31 +247,36 @@ where
...
-        if coeffs.next().is_some() {
-            return None;
-        }
+        let (first, second, offset) = match self {
+            AffineExpression::Many {
+                coefficients,
+                offset,
+            } if coefficients.len() == 2 => {
+                let mut iter = coefficients.iter().map(|(v, c)| (v, c));
+                (iter.next().unwrap(), iter.next().unwrap(), offset)
```
chriseth, line 256: What about something like
```suggestion
```
let [first, second] = &coefficients[..] else { unreachable!() };
(first, second, offset)
```

File: executor/src/witgen/affine_expression.rs (hunk 2 above)
Comment (pacheco) [reply executor/src/witgen/affine_expression.rs:17]: you mean wrapping the offset in Option<>? If so, I think it wouldn't change much and complicate all the `match`es in the code

File: executor/src/witgen/affine_expression.rs [hunk 5]
```diff
@@ -426,14 +407,10 @@ where
...
-        if self.offset != other.offset {
+        if self.offset() != other.offset() {
             return false;
         };
-        let mut self_coeff = self.nonzero_coefficients().to_vec();
-        self_coeff.sort_unstable();
-        let mut other_coeff = other.nonzero_coefficients().to_vec();
-        other_coeff.sort_unstable();
-        self_coeff == other_coeff
+        self.nonzero_coefficients().eq(other.nonzero_coefficients())
```
chriseth, line 413: The main problem here is that the One and the Many cases overlap, right? Does it make sense to force Many to have at least two variables?

executor/src/witgen/affine_expression.rs
```diff
@@ -444,28 +421,110 @@ where
...
+        use std::cmp::Ordering;
+        use AffineExpression::*;
+        match (self, rhs) {
+            // constant + constant
+            (Constant(o1), Constant(o2)) => Constant(o1 + o2),
+            // constant + one
+            (One(coeff, o1), Constant(o2)) | (Constant(o2), One(coeff, o1)) => One(coeff, o1 + o2),
+            // constant + many
+            (
+                Many {
```
chriseth, line 434: it's weird - this is not too long to fit a single line, right? Does it always break patterns like this?

chriseth: Is there a way to use something like `many @ Many` and then use `many.coefficients` below?

chriseth: I mean another "solution" would be to not use named parameters for Many, as you do already for One.

executor/src/witgen/affine_expression.rs
pacheco: Idea is: I want to return an `impl Trait`, but there are 3 concrete implementations being returned here.
General solution in these situations is to return Box<dyn Trait>, but that is one allocation and uses dynamic dispatch.
`Either` implements `Iterator<Item=T>` when both sides implement `Iterator<T>` type.
And this works recursively for each side.
This way we can return a concrete `impl Trait`, avoiding allocations or dynamic dispatch!

executor/src/witgen/affine_expression.rs
```diff
@@ -444,28 +421,110 @@ where
...
+            ) => {
+                debug_assert!(Self::is_proper_coefficients(&coefficients));
+                match coefficients.binary_search_by_key(&&v, |(k, _)| k) {
+                    Ok(idx) => coefficients[idx].1 += c,
+                    Err(idx) => coefficients.insert(idx, (v, c)),
+                }
+                coefficients.retain(|(_, c)| !c.is_zero());
+                if coefficients.is_empty() {
+                    Constant(offset + o)
+                } else {
```
chriseth, line 492: So this is the case where you could reduce to One but don't right?

executor/src/witgen/affine_expression.rs
pacheco: `nonzero_coefficients()` unifies both cases. I did do the conversion down to One, but it was a lil bit slower and the code was not simpler

executor/src/witgen/affine_expression.rs
pacheco: yes, there are two places where `coefficients.is_empty()` where we could also test for `len() == 1`

executor/src/witgen/affine_expression.rs
pacheco: the thing is that `One` avoids allocating a vec![] for a single variable case (basically every constructed expr starts as a single variable), but if we already have the vec![] it's not worth it to get rid of it

executor/src/witgen/affine_expression.rs
```diff
@@ -444,28 +421,110 @@ where
...
+                    coefficients: mut c1,
+                    offset: o1,
+                },
+                Many {
+                    coefficients: c2,
+                    offset: o2,
+                },
+            ) => {
+                debug_assert!(Self::is_proper_coefficients(&c1));
+                for (k2, v2) in c2 {
```
chriseth, line 511: This could be faster if you used some kind of merge sort here. Also you could swap `c1` and `c2` if `c2` is larger.

executor/src/witgen/expression_evaluator.rs
```diff
@@ -55,47 +55,62 @@ where
...
-                    } else if let Some(f) = right.constant_value() {
-                        Ok(left * f)
-                    } else {
-                        Err(IncompleteCause::QuadraticTerm)
+            AlgebraicBinaryOperator::Sub => Ok(self.evaluate(left)? - self.evaluate(right)?),
+            AlgebraicBinaryOperator::Mul => {
+                // don't short circuit on err as rhs might still be 0
+                let left_res = self.evaluate(left);
+                match left_res {
+                    Ok(left_expr) if left_expr.is_zero() => Ok(left_expr),
```
chriseth, line 75: This returns `Ok` if the rhs has an error, right? I think we have to double-check that this is OK.

executor/src/witgen/affine_expression.rs
pacheco: > Is there a way to use something like `many @ Many` and then use `many.coefficients` below?
I think not, but non-named parameters could be better indeed.

executor/src/witgen/expression_evaluator.rs
chriseth: Oh, it already did that, sorry!

executor/src/witgen/expression_evaluator.rs
```diff
@@ -55,47 +55,62 @@ where
...
+                    Ok(left_expr) if left_expr.is_one() => self.evaluate(right),
+                    Ok(left_expr) => {
+                        if let Some(n) = left_expr.constant_value() {
+                            return Ok(self.evaluate(right)? * n);
+                        }
+                        // lhs not a constant
+                        let right_expr = self.evaluate(right)?;
```
chriseth, line 82: This can be moved to the start of the match body.

executor/src/witgen/expression_evaluator.rs
```diff
@@ -55,47 +55,62 @@ where
...
+                        match right_expr.constant_value() {
+                            Some(n) if n.is_zero() => Ok(right_expr),
```
chriseth, line 84: Can you rename `n` to `r`?

executor/src/witgen/expression_evaluator.rs
pacheco: This part is replicating the behaviour we had before, for multiplication:
- either side being zero allows for the other arm to Err and still returns Ok
In fact, without this behavior, witgen was failing in some tests if I remember correctly.

executor/src/witgen/expression_evaluator.rs
```diff
@@ -55,47 +55,62 @@ where
...
+                            Some(n) if n.is_one() => Ok(left_expr),
```
chriseth, line 85: Does it make sense to pull the `is_one()` and `is_zero()` checks into the multiplication by constant function?

executor/src/witgen/expression_evaluator.rs
pacheco: you can't because the `lhs.is_zero()` arm avoids evaluating the right side

executor/src/witgen/expression_evaluator.rs
pacheco: AFAIR i tested it and was a bit slower, can check again

executor/src/witgen/expression_evaluator.rs
```diff
@@ -55,47 +55,62 @@ where
...
-                    if let (Some(l), Some(r)) = (left.constant_value(), right.constant_value()) {
-                        Ok(l.pow(r.to_integer()).into())
-                    } else {
-                        Err(IncompleteCause::ExponentiationTerm)
-                    }
+            }
+            AlgebraicBinaryOperator::Pow => {
+                if let (Some(l), Some(r)) = (
+                    self.evaluate(left)?.constant_value(),
+                    self.evaluate(right)?.constant_value(),
```
chriseth, line 107: It might be that `right` in this case could even always be a literal number - please check!

executor/src/witgen/affine_expression.rs
chriseth: Can you add a short comment?

executor/src/witgen/expression_evaluator.rs
chriseth, line 82: This is what I meant:
```suggestion
                        let right_expr = self.evaluate(right)?;
                        if let Some(n) = left_expr.constant_value() {
                            return Ok(right_expr * n);
                        }
                        // lhs not a constant
```

executor/src/witgen/expression_evaluator.rs
pacheco: added the `expect`

executor/src/witgen/expression_evaluator.rs
pacheco: checked again, its a tiny bit faster doing the checks here

pacheco: yep, checking it here is a bit faster

executor/src/witgen/affine_expression.rs
pacheco: changed it to a parallel iteration + merge and it is indeed a bit faster (~3%) for the many chunks benchmark!

executor/src/witgen/affine_expression.rs
```diff
@@ -453,41 +453,53 @@ where
...
+                let mut c1 = c1.into_iter();
                 let mut c2 = c2.into_iter();
-                while let Some((k2, v2)) = c2.next() {
-                    let range = &mut c1[range_start..];
-                    match range.binary_search_by_key(&&k2, |(k, _)| k) {
-                        Ok(idx) => {
-                            range[idx].1 += v2;
-                            range_start = idx + 1;
+                // loop over the two expressions in parallel, merging their variables in order
+                let mut ov1 = c1.next();
```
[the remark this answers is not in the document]
pacheco: cool, didn't know about those! let me see

chriseth (approved):

PR #1107, opened 2024-02-28
Uniformizing indentation all over RISC-V zkVM .asm file.

lvella:
@georgwiese complained about indentation in the output .asm file, so this PR is his fault.

### DIFF
--- a/riscv/src/compiler.rs
+++ b/riscv/src/compiler.rs
@@ -149,7 +149,7 @@ pub fn compile(
     let data_positions =
         store_data_objects(data_sections, data_start, &mut |label, addr, value| {
             if let Some(label) = label {
-                let comment = format!(" // data {label}");
+                let comment = format!("// data {label}");
                 if with_bootloader && !matches!(value, SingleDataValue::LabelReference(_)) {
                     &mut initial_mem
                 } else {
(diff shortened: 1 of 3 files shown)


riscv/src/compiler.rs
```diff
@@ -200,44 +200,50 @@ pub fn compile(
...
+        program.push((2, "x1 <== jump(__data_init);".to_string()));
     }
     program.extend([
-        format!("// Set stack pointer\nx2 <=X= {stack_start};"),
-        "x1 <== jump(__runtime_start);".to_string(),
-        "return;".to_string(), // This is not "riscv ret", but "return from powdr asm function".
+        (2, "// Set stack pointer".to_string()),
+        (2, format!("x2 <=X= {stack_start};")),
+        (2, "x1 <== jump(__runtime_start);".to_string()),
+        (2, "return;".to_string()), // This is not "riscv ret", but "return from powdr asm function".
```
leonardoalt, line 231: I'm not sure all this weirdness (2, ...) everywhere is worth it

leonardoalt:
Instead of outputting formatted code, wouldn't it make more sense to output whatever and have a separate formatter? Don't we already have one for PIL?

riscv/src/compiler.rs
lvella: I thought about it halfway through. Anyway, this 2 is the indentation level.

leonardoalt: I got that, but I don't think it's a good solution.

lvella:
We close it.

PR #1119, opened 2024-03-04
Some fixes.

chriseth:
Somehow the type annotation is needed for my ide...

### DIFF
--- a/pipeline/src/test_util.rs
+++ b/pipeline/src/test_util.rs
@@ -132,7 +132,7 @@ pub fn gen_halo2_proof(file_name: &str, inputs: Vec<Bn254Field>) {
 
     // Create the proof before adding the setup and vkey to the backend,
     // so that they're generated during the proof
-    let proof = pipeline.compute_proof().unwrap().clone();
+    let proof: Vec<u8> = pipeline.compute_proof().unwrap().clone();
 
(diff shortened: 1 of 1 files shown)


pipeline/src/test_util.rs
```diff
@@ -132,7 +132,7 @@ pub fn gen_halo2_proof(file_name: &str, inputs: Vec<Bn254Field>) {
 
     // Create the proof before adding the setup and vkey to the backend,
     // so that they're generated during the proof
-    let proof = pipeline.compute_proof().unwrap().clone();
+    let proof: Vec<u8> = pipeline.compute_proof().unwrap().clone();
```
lvella, line 135: What happens if you don't annotate this?

leonardoalt:
https://github.com/rust-lang/rust-analyzer/issues/16609#issuecomment-1970199689

Since this is a rust-analyzer bug I don't think we should change our code

chriseth:
> It is in our unwritten coding guidelines that we shall never name a type that can be inferred. But it is not really a guideline, we just started to believe it.

I would say the guidelines say "write readable code, do not clutter it"

leonardoalt:
unnecessary explicit types are clutter

chriseth:
comments are clutter

PR #1120, opened 2024-03-04
Remove constant check.

chriseth:
This check is now better done by the type checker.

### DIFF
--- a/pil-analyzer/src/statement_processor.rs
+++ b/pil-analyzer/src/statement_processor.rs
@@ -144,19 +144,13 @@ where
                     Some(definition),
                 )
             }
-            PilStatement::ConstantDefinition(source, name, value) => {
-                // Check it is a constant.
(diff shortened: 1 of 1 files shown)


pil-analyzer/src/statement_processor.rs
```diff
@@ -144,19 +144,13 @@ where
                     Some(definition),
                 )
             }
-            PilStatement::ConstantDefinition(source, name, value) => {
-                // Check it is a constant.
-                if let Err(err) = self.evaluate_expression_to_fe(value.clone()) {
```
pacheco, line 149: why is this not needed anymore?

chriseth: This is used to check if an expression evaluates to a single number that is in the range of the field. The actual number is ignored, though. The evaluation at this point, though, is type-blind: It assumes all numbers are integers. It is much better to just let the type checker check that the expression properly evaluates to `fe`.

pacheco (approved):

PR #1122, opened 2024-03-05
Unify bigints

chriseth:
This allows to swap out the bigint implementation more easily.


leonardoalt (commented):
What's the difference between BigInt and LargeInt?

chriseth:
LargeInt is the underlying data structure for the field element. LargeInt is fixed-sized and BigInt is unlimited size.

chriseth:
Essentially you probably never have to deal with LargeInt, unless you want to convert a field element to a u64 or something, they you maybe don't want to go through BigInt.

PR #1123, opened 2024-03-05
Use ibig instead of num_bigint

chriseth:


chriseth:
I think this does indeed provide a 20 to 50% speedup.

chriseth:
ibig on top of the evaulation cache does not improve the fixed column generation much, but it does improve our evaluation benchmark, so I think it is still good to have.

number/Cargo.toml
```diff
@@ -19,6 +19,7 @@ csv = "1.3"
 serde = { version = "1.0", default-features = false, features = ["alloc", "derive", "rc"] }
 serde_with = "3.6.1"
 schemars = { version = "0.8.16", features = ["preserve_order"]}
+ibig = "0.3.6"
```
leonardoalt, line 22: is `num-bigint` above still needed?

number/src/macros.rs
```diff
@@ -70,25 +70,25 @@ macro_rules! powdr_field {
             }
         }
 
-        impl Shl<u64> for LargeIntImpl {
+        impl Shl<usize> for LargeIntImpl {
             type Output = Self;
 
-            fn shl(self, other: u64) -> Self {
+            fn shl(self, other: usize) -> Self {
                 // TODO: avoid using BigUint
```
leonardoalt, line 77: Remove this comment (and same comment below) and import?

number/Cargo.toml
chriseth: Yes, because the large integer type of ark can only convert from num_bigint.

leonardoalt (approved):
please squash

PR #1124, opened 2024-03-05
`source_order` points to identities vec index, not poly id

pacheco:


chriseth (commented):
Good catch!

Can you add a test that shows the bug if it's easy to do?

chriseth (approved):

chriseth:
Maybe  we should remove by ID, especially when we introduce stages...

PR #1130, opened 2024-03-06
Use evaluation cache for symbols.

chriseth:
Improves constant evaluation for keccak from 11 to 4.4 seconds.

This only caches the symbols themselves, i.e. for a column, it does not store the actual cell values, but only the function closure.

The Arc commit reduces it to 4 seconds.

### DIFF
--- a/executor/src/constant_evaluator/mod.rs
+++ b/executor/src/constant_evaluator/mod.rs
@@ -1,15 +1,18 @@
-use std::{collections::HashMap, rc::Rc};
+use std::{
+    collections::HashMap,
+    sync::{Arc, RwLock},
+};
 
 use itertools::Itertools;
 use powdr_ast::{
     analyzed::{
         types::{ArrayType, Type, TypedExpression},
-        Analyzed, Expression, FunctionValueDefinition,
+        Analyzed, Expression, FunctionValueDefinition, Symbol,
     },
     parsed::IndexAccess,
 };
 use powdr_number::{BigInt, DegreeType, FieldElement};
-use powdr_pil_analyzer::evaluator::{self, Definitions, Value};
+use powdr_pil_analyzer::evaluator::{self, Definitions, SymbolLookup, Value};
 use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 
 /// Generates the fixed column values for all fixed columns that are defined
@@ -45,7 +48,10 @@ fn generate_values<T: FieldElement>(
     body: &FunctionValueDefinition<T>,
     index: Option<u64>,
 ) -> Vec<T> {
-    let symbols = Definitions(&analyzed.definitions);
+    let symbols = CachedSymbols {
+        symbols: &analyzed.definitions,
+        cache: Arc::new(RwLock::new(HashMap::new())),
+    };
     let result = match body {
         FunctionValueDefinition::Expression(TypedExpression { e, type_scheme }) => {
             if let Some(type_scheme) = type_scheme {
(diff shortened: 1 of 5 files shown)


executor/src/constant_evaluator/mod.rs
```diff
@@ -45,7 +48,10 @@ fn generate_values<T: FieldElement>(
     body: &FunctionValueDefinition<T>,
     index: Option<u64>,
 ) -> Vec<T> {
-    let symbols = Definitions(&analyzed.definitions);
+    let symbols = CachedSymbols {
+        symbols: &analyzed.definitions,
+        cache: Arc::new(RwLock::new(HashMap::new())),
```
lvella, line 53: Are you using `Arc` and `RwLock` instead of `Rc` and `RefCell` to be prepared for multithreading?

pil-analyzer/src/evaluator.rs
```diff
@@ -325,10 +326,10 @@ pub struct Definitions<'a, T>(
 
 impl<'a, T: FieldElement> SymbolLookup<'a, T> for Definitions<'a, T> {
     fn lookup<'b>(
-        &self,
+        &mut self,
```
lvella, line 329: I am having a hard time tracing the origin of the need for `mut` here...

lvella: In fact, I removed locally and it still compiles. I think clippy didn't complain because the `mut` is part of a trait that all implementations must follow.

lvella: I removed it locally and it still compiled. Doesn't seem to be necessary. See PR #1133.

pil-analyzer/src/evaluator.rs
```diff
@@ -535,7 +541,11 @@ mod internal {
                             if p == v {
                                 Some(value)
                             } else {
-                                match (p.try_to_integer(), v.clone().try_to_integer()) {
+                                // TODO I don't think this part is needed now that we have the type checker.
+                                match (
+                                    p.as_ref().clone().try_to_integer(),
+                                    v.as_ref().clone().try_to_integer(),
```
lvella, line 547: Maybe `try_to_integer()` should take `&self` instead of `self`, so no cloning would be needed here.

pil-analyzer/src/evaluator.rs
```diff
@@ -841,7 +858,7 @@ mod test {
         else {
             panic!()
         };
-        evaluate(symbol, &Definitions(&analyzed.definitions))
+        evaluate::<_>(symbol, &mut Definitions(&analyzed.definitions))
```
lvella, line 861: `::<_>` ?

This adds nothing... Does it help with rust-analyzer problems?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -296,6 +296,8 @@ impl<T: FieldElement> PILAnalyzer<T> {
         let namespace_degree: u64 = u64::try_from(
             evaluator::evaluate_expression(&degree, &self.definitions)
                 .unwrap()
+                .as_ref()
+                .clone()
```
lvella, line 300: Not needed if `try_to_integer()` takes `&self` instead of `self`.

pil-analyzer/src/statement_processor.rs
```diff
@@ -468,6 +468,8 @@ where
             &ExpressionProcessor::new(self.driver).process_expression(expr),
             self.driver.definitions(),
         )?
+        .as_ref()
+        .clone()
```
lvella, line 472: Same thing.

lvella (changes requested):
Left some comments above...

executor/src/constant_evaluator/mod.rs
chriseth: It already is multithreaded, see line 74 / 80: `.into_par_iter()`

pil-analyzer/src/evaluator.rs
chriseth: Thought about that already as well...
Will check how ofter `try_to_integer` would needlessly clone in that case, but I guess almost never.

chriseth:
Addressed all your comments.

lvella (approved):
Your craftsmanship is evident in this PR!

Please squash before merging.

lvella (dismissed):

lvella (approved):
Please squash.

PR #1133, opened 2024-03-06
Removing mut from symbols, as it doesn't seem to be needed.

lvella:


chriseth:
Interesting! I was not aware that `RwLock` has interior mutability!

PR #1137, opened 2024-03-07
Reverse local variable numbering.

chriseth:
I initially thought it would be good to re-start local variable numbering at zero in each closure, so that these closures could potentially be moved without harming the numbering, but then we would check for captured variables and so on. So in the end, I think it is better not to re-start at zero. Also it is easier for `let` statements to just introduce a new variable at the end instead of at the beginning.


georgwiese (approved):
LGTM

PR #1138, opened 2024-03-07
Let statements in expressions.

### LINKED ISSUE #959: PIL: Implement `let` statement as expression
At expression level, `let`-statements should be allowed, together with blocks of statements.

This means, the following should be added to the alternatives for "expression":
`"{"` (LetStatement ";")* Expression "}"`

(Also the `IfStatement` and others should be extended so that we don't have to use double `{`.)

For now, the let statement requires a value (but see #960 for an extension).

During symbol resolution and evaluation, the defined variable is added to the `locals` just as we do for closures.

chriseth:
Allows braced blocks everywhere where expressions are expected.
The statements in those blocks can be `let x` or `let x = ...`.
The former declares a new witness column, the latter just binds a local variable.

fixes #959


chriseth:
I'll do the `constr` part in a different pr.

ast/src/parsed/mod.rs
```diff
@@ -658,6 +671,22 @@ impl<R> Children<Expression<R>> for IfExpression<R> {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct LetStatementInsideBlock<Ref = NamespacedPolynomialReference> {
```
leonardoalt, line 675: What's the difference between this and a normal let statement?

pil-analyzer/src/condenser.rs
```diff
@@ -241,4 +293,39 @@ impl<'a, T: FieldElement> SymbolLookup<'a, T> for CondenserSymbols<'a> {
...
         Definitions(self.symbols).lookup_public_reference(name)
     }
+
+    fn new_witness_column(
+        &mut self,
+        name: &str,
+    ) -> Result<Arc<Value<'a, T>>, evaluator::EvalError> {
+        self.next_witness_id += 1;
+        let id = self.next_witness_id;
+        // Find an unused name.
```
leonardoalt, line 303: Maybe extract this?

pil-analyzer/tests/condenser.rs
```diff
@@ -0,0 +1,49 @@
...
+    N.x_1 = N.y;
+"#;
+    let formatted = analyze_string::<GoldilocksField>(input).to_string();
+    assert_eq!(formatted, expected);
+}
+
+#[test]
+fn new_witness_column_name_clash() {
+    let input = r#"namespace N(16);
+    let new_wit = || { let x; x };
```
leonardoalt, line 34: lol

pil-analyzer/tests/condenser.rs
```diff
@@ -0,0 +1,49 @@
+use powdr_number::GoldilocksField;
+use powdr_pil_analyzer::analyze_string;
+use test_log::test;
+
+use pretty_assertions::assert_eq;
+
+#[test]
+fn new_witness_column() {
+    let input = r#"namespace N(16);
+    let new_wit = || { let x; x };
```
leonardoalt, line 10: Can you also add some tests where this function adds constraints over the new column?

ast/src/parsed/mod.rs
chriseth: The other let statement is just an enum variant. The other difference is that here, we do not support types yet. Once add types, we can unify the two.

chriseth: Actually I could unify them now and just not parse the type scheme.

chriseth: ok, this gets a bit too complicated and I would like to defer that to later. The reason is that at parsing time, we have `Type<Expression<NamespacedPolynomialReference>>` and `Expression<NamespacedPolynomialReference>`, while after analysis we have `Type<u64>` and `Expression<Reference>`. Probably better to fix that by splitting the Expression type into multiple types.

pil-analyzer/tests/condenser.rs
chriseth: Ah nice, that actually found a bug.

ast/src/parsed/mod.rs
leonardoalt: ok

pil-analyzer/tests/condenser.rs
leonardoalt: I meant a case that adds constraints to the new column inside the function itself, like
```
let new_wit = || { let x; x = 2; x }
```

Or is that not possible atm?

chriseth: constraints can only be added in #1187

leonardoalt (approved):
please squash

leonardoalt (dismissed):

leonardoalt (approved):

PR #1140, opened 2024-03-07
fix `remove_polynomials` to work with zero sized arrays

pacheco:
This fixes a bug where zero sized arrays could cause a crash when removing polynomials during optimization.
The issue is that `array_elements()` won't return anything for a zero sized array, but the array definition itself still has an `id` which is referenced by the code and needs to be updated.

### DIFF
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -181,6 +181,19 @@ impl<T> Analyzed<T> {
                                     },
                                 );
                             }
+                            // we still need a replacement for zero sized array,
+                            // but array_elements() won't return any items in this case
+                            // TODO: handle this differently? https://github.com/powdr-labs/powdr/issues/1142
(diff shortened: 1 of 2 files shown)


ast/src/analyzed/mod.rs
```diff
@@ -181,6 +181,18 @@ impl<T> Analyzed<T> {
                                     },
                                 );
                             }
+                            // we still need a replacement for zero sized array,
```
chriseth, line 184: I think we should not allocate IDs for zero-sized arrays. Maybe we could make it an error in `statement_processor` and/or in the condenser?

chriseth (approved):

ast/src/analyzed/mod.rs
pacheco: it's already not allocated, but the Symbol is assigned the id where element id's would start from

PR #1141, opened 2024-03-08
Fix recursive lookups.

chriseth:
When implementing the SymbolLookup trait and delegating to the Definitions implementation, that one could do recursive evaluations that would ignore the first implementation of SymbolLookup. This PR provides a mechanism where you can re-use parts of the Definitions implementation while recursive evaluations would still go through your own implementation.


leonardoalt (commented):
were there any tests that broke without this? Should we add some?

leonardoalt (approved):

PR #1143, opened 2024-03-08
Allow assignment registers to be declared after write registers

### LINKED ISSUE #1100: asm parser nonsense
```asm
machine Test {                                                                                                                                                                                                                               
    reg pc[@pc];                                                                                                                                                                                                                             
    reg X[<=];                                                                                                                                                                                                                               
    reg A;                                                                                                                                                                                                                                   
    reg B;                                                                                                                                                                                                                                   
    reg Y[<=];                                                                                                                                                                                                                               
                                                                                                                                                                                                                                             
    instr inc {                                                                                                                                                                                                                              
        A' = A + 1                                                                                                                                                                                                                           
    }                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                             
    instr dec {                                                                                                                                                                                                                              
        A' = A - 1                                                                                                                                                                                                                               }                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                             
    instr assert_eq X, Y {                                                                                                                                                                                                                   
        X = Y                                                                                                                                                                                                                                
    }                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                             
    function main {                                                                                                                                                                                                                          
        inc;                                                                                                                                                                                                                                 
        inc;                                                                                                                                                                                                                                 
        B <=X= A;                                                                                                                                                                                                                            
        A <=Y= B;                                                                                                                                                                                                                            
        inc;                                                                                                                                                                                                                                 
        dec;                                                                                                                                                                                                                                 
        assert_eq A, 2;                                                                                                                                                                                                                          }                                                                                                                                                                                                                                        
}
```

Schaeff:
Closes #1100 

Unsure:
- should this be implemented within `handle_register_declaration` and simply make sure we start with assignment registers?
- should the hello world be kept as is as it's more idiomatic and test on another file?


chriseth:
I think it's good to allow forward references, of things that are not declared using `let` (and maybe even some things that are, if e disallow shadowing by subsequent let statements on the same level) so nothing should really depend on the order.

Schaeff:
I agree! That's already what powdr is supposed to do, simply this piece of code did rely on assignment registers being first.

test_data/asm/vm_to_block_array.asm
```diff
@@ -7,8 +7,8 @@ machine Main {
     reg pc[@pc];
     reg X[<=];
     reg Y[<=];
-    reg Z[<=];
     reg A;
+    reg Z[<=];
```
lvella, line 11: Maybe add a comment here saying this weirdness is on purpose?

lvella (changes requested):

lvella (approved):
Please squash before merging.

If the contents are identical, GitHub should not dismiss the approval.

PR #1145, opened 2024-03-08
add test that uses serde in riscv

leonardoalt:


pil-analyzer/src/pil_analyzer.rs
```diff
@@ -162,6 +159,9 @@ impl<T: FieldElement> PILAnalyzer<T> {
                                         .map(|e| (e, Type::Fe.into())),
                                 );
                             }
+                            // TODO add support for prover queries properly.
+                            // They should accept a tuple of any size.
```
chriseth, line 163:
```suggestion
                            // They should accept a certain enum.
```

chriseth: Then we can also ditch the weird string identifiers.

chriseth (approved):

chriseth (approved):

chriseth (approved):

Issue #1155, opened 2024-03-12
read_polys_file hangs if file is empty


leonardoalt:
https://github.com/powdr-labs/powdr/blob/main/number/src/serialize.rs#L115

When the PIL file has no fixed columns (like https://github.com/powdr-labs/powdr/blob/main/test_data/asm/empty.asm), this function hangs while trying to read the file (which exists but it's empty).

Steps to reproduce:

```bash
cargo run pil test_data/asm/empty.asm -f -o output --field bn254
cargo run verification-key output/empty.pil --backend halo2 --field bn254 -d output
```

pacheco:
fixed by #1156

PR #1156, opened 2024-03-12
bugfix #1155: read_polys_file with 0 columns

pacheco:
Assert that `read_polys_file` is not called with no columns.
Now `read_poly_set` returns an `Option`, with `None` being returned when there are no columns of the type.

### DIFF
--- a/pipeline/src/pipeline.rs
+++ b/pipeline/src/pipeline.rs
@@ -27,7 +27,7 @@ use powdr_schemas::SerializedAnalyzed;
 
 use crate::{
     inputs_to_query_callback, serde_data_to_query_callback,
-    util::{read_poly_set, write_or_panic, FixedPolySet, WitnessPolySet},
+    util::{try_read_poly_set, write_or_panic, FixedPolySet, WitnessPolySet},
 };
 
 type Columns<T> = Vec<(String, Vec<T>)>;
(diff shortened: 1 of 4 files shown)


leonardoalt (commented):
Nice! Let's also add one or 2 tests in pipeline to prevent this from happening again

pipeline/src/util.rs
```diff
@@ -35,23 +35,28 @@ impl PolySet for WitnessPolySet {
     }
 }
 
+#[allow(clippy::type_complexity)]
 pub fn read_poly_set<P: PolySet, T: FieldElement>(
```
Schaeff, line 39:
```suggestion
pub fn try_read_poly_set<P: PolySet, T: FieldElement>(
```

pipeline/src/pipeline.rs
```diff
@@ -363,9 +363,15 @@ impl<T: FieldElement> Pipeline<T> {
...
 
-        assert_eq!(pil.degree.unwrap(), degree_fixed);
+        let fixed = if let Some((fixed, degree_fixed)) =
+            read_poly_set::<FixedPolySet, T>(&pil, directory, &name)
+        {
+            assert_eq!(pil.degree.unwrap(), degree_fixed);
+            fixed
+        } else {
+            vec![]
+        };
```
Schaeff, line 374: This could use `.map(...).unwrap_or_default()`

pipeline/src/util.rs
```diff
@@ -35,23 +35,28 @@ impl PolySet for WitnessPolySet {
...
+    }
+
     let fname = format!("{name}_{}", P::FILE_NAME);
 
-    read_polys_file(
+    Some(read_polys_file(
         &mut BufReader::new(File::open(dir.join(fname)).unwrap()),
         &column_names,
-    )
+    ))
```
Schaeff, line 59: This could use `(!column_names.is_empty).then(|| ...)`

pipeline/src/pipeline.rs
```diff
@@ -380,9 +386,15 @@ impl<T: FieldElement> Pipeline<T> {
...
 
-        assert_eq!(pil.degree.unwrap(), degree_witness);
+        let witness = if let Some((witness, degree_witness)) =
+            read_poly_set::<WitnessPolySet, T>(&pil, directory, &name)
+        {
+            assert_eq!(pil.degree.unwrap(), degree_witness);
+            witness
+        } else {
+            vec![]
+        };
```
Schaeff, line 397: Same here

leonardoalt (approved):

PR #1172, opened 2024-03-19
point halo2 to our forks so powdr lib uses do not need to patch it

leonardoalt:
Currently we patch our halo2 dependency to @georgwiese 's fork which applies a change we needed on top of their latest release 0.3.0. The problem with that is that patches do not propagate to users of powdr. This causes compilation errors on their side (eigen-zkvm, powdr-revme, etc).

So our current solution is to fork all halo2, halo2wrong and snark-verifier to point all of them to the same halo2 (ours), until halo2 0.3.1 is released and everyone updates everything.

These changes here make sure everything compiles properly for powdr users without any extra patches.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,8 +44,3 @@ overflow-checks = true
 panic = 'unwind'
 incremental = true         # This is true because target is cached
 codegen-units = 256
-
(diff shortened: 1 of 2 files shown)


georgwiese (approved):
LGTM

PR #1187, opened 2024-03-22
Function annotations

chriseth:
- [ ] see if this all works out
- [ ] split statements in BlockExpression to `LetStatement` (only with value) `ColumnDeclaration` (different syntax that allows for query) and `Constraint`.
- [ ] document
- [ ] test


std/utils.asm
```diff
@@ -25,6 +25,26 @@ let unchanged_until: expr, expr -> constr = |c, latch| (c' - c) * (1 - latch) =
 /// Evaluates to a constraint that forces `c` to be either 0 or 1.
 let force_bool: expr -> constr = |c| c * (1 - c) = 0;
 
+/// Creates a new witness column that is constrained to the values 0 and 1.
+let new_bool: -> expr = constr || {
```
leonardoalt, line 29:
```suggestion
let mk_bool: -> expr = constr || {
```

leonardoalt: probably just biased because SMT solvers do it like that, but I think this notation is also common in FP?

ast/src/parsed/mod.rs
```diff
@@ -694,6 +726,7 @@ pub enum FunctionDefinition {
     /// Array expression.
     Array(ArrayExpression),
     /// Prover query. The Expression usually is a LambdaExpression.
+    /// TODO can we replace this by a query-marked lambda expcession?
```
chriseth, line 729: Still want to see if this makes sense.

leonardoalt: still no vscode spell checker? :D

chriseth: It's removed code!

leonardoalt: I mean `expcession`

chriseth: I know, but I was able to replace it and deleted the comment.

book/src/pil/expressions.md
```diff
@@ -35,11 +35,39 @@ Elementary expressions are
...
 
+### Statement Blocks
+
+A ``{``-``}``-delimited block can be used everywhere where an expression is expected.
+
+It has the form ``{ <statement> ; <statement> ; ... ; <expression> }``,
+i.e. a sequence of statements followed by an expression and the value of the statement block
+is the value of the final expression.
+
+The statements can either be expressions or let statements: ``let x = ...`` / ``let x;`
```
leonardoalt, line 52:
```suggestion
The statements can either be expressions or let statements: ``let x = ...`` / ``let x;``
```
?

book/src/pil/expressions.md
```diff
@@ -35,11 +35,39 @@ Elementary expressions are
...
+
+Let statements without value create a new witness column and are only allowed inside [``constr``-functions](#constr-and-query-functions).
+
+Similarly, an expression at statement level can be used to create new constraints that are added to the global constraint set
+and this can only be done inside a [``constr``-functions](#constr-and-query-functions).
+
+Note that you can always create constraints and return them from a function, even in [pure function](#constr-and-query-functions).
+
+
+Example:
```
leonardoalt, line 65: Is this an example of the `Note...` par above?

book/src/pil/expressions.md
```diff
@@ -107,3 +119,38 @@ The following example illustrates how you can still use the generic language:
...
+using the `std::prover::eval` function.
+
+Both actions require a certain context to be available, which is not the case for example when
+the values of a fixed column are computed.
+
+A `query` function can only be used in the query or hint part of a witness column while `constr` functions
+can only be evaluated in the constraint part of a namespace or machine.
+
+You can define and call new `constr` functions inside a `constr` function and you can call and define
+new `query` functions inside `query` functions, but as soon as you enter a pure function, this is not possible any more.
```
leonardoalt, line 144:
```suggestion
new `query` functions inside `query` functions, but as soon as you enter a pure function, this is not possible anymore.
```

pil-analyzer/src/side_effect_checker.rs
```diff
@@ -0,0 +1,124 @@
...
+                let old_context = self.context;
+                let result = self.check(body);
+                self.context = old_context;
+                result
+            }
+            Expression::BlockExpression(statements, _expr) => {
+                for s in statements {
+                    match s {
+                        StatementInsideBlock::LetStatement(s) => {
+                            if s.value.is_none() && self.context != FunctionKind::Constr {
```
leonardoalt, line 64: Wouldn't it be cleaner to check that `s`'s type is `witness`? I guess it's equivalent to `s.value.is_none()` now, just wondering if it may change

pil-analyzer/src/side_effect_checker.rs
```diff
@@ -0,0 +1,124 @@
...
+                                return Err(format!(
+                                    "Tried to create a witness column in a {} context: {s}",
+                                    self.context
+                                ));
+                            }
+                        }
+                        StatementInsideBlock::Expression(expr) => {
```
leonardoalt, line 71: Why does this imply adding a constraint?

pil-analyzer/src/condenser.rs
```diff
@@ -344,6 +336,30 @@ impl<'a, T: FieldElement> SymbolLookup<'a, T> for Condenser<'a, T> {
...
+            Value::Array(items) => Box::new(items.iter().map(|item| match item.as_ref() {
+                Value::Identity(left, right) => (left, right),
+                _ => panic!("Expected constraint, but got {item}"),
+            })),
+            _ => panic!("Expected constraint but got {constraints}"),
+        };
+        for (left, right) in identities {
+            self.new_constraints
+                .push(Identity::from_polynomial_identity(
+                    0, // ID will be re-assigned later.
```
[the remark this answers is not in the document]
chriseth: No, please not.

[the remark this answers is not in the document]
chriseth: It is re-assigned in the same file.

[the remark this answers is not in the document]
chriseth: And all of them are fully-reassigned, not just those set to zero.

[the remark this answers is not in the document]
chriseth: Ok, I could also change the type of `new_constraints` to not contain the ID.

pil-analyzer/src/side_effect_checker.rs
```diff
@@ -0,0 +1,124 @@
...
+    analyzed::{
+        Expression, FunctionValueDefinition, Reference, Symbol, SymbolKind, TypedExpression,
+    },
+    parsed::{FunctionKind, LambdaExpression, StatementInsideBlock},
+};
+
+use lazy_static::lazy_static;
+
+/// Check that query functions are only referenced/defined in a query context
+/// and that constr functions are only referenced/defined in a constr context.
```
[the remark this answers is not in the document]
chriseth: Yes, pure functions are allowed everywhere.

pil-analyzer/src/side_effect_checker.rs
chriseth: It's the same rules as we have on the global level: If it's not a declaration, it must be a constraint:
```
let x; // declaration
let x = f(); // declaration
g(); // must result in a constraint that is added to the set
x = 7; // must result in a constraint that is added to the set
```

chriseth: Note that `StatementInsideBlock` is never the last component of a block, so in 
```
let x = constr || {
  let y = ...;
  let z =  ...;
  f();
  g()
}
```
The last line (`g()`) is the result of the function while `f()` must return a constraint

book/src/pil/expressions.md
chriseth: No, let me move it further up.

pil-analyzer/src/condenser.rs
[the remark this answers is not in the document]
leonardoalt: Yea Option was just the quick alternative I had when writing the comment, but it would be good to not rely on the value 0 to have different semantics

leonardoalt (approved):

PR #1188, opened 2024-03-22
Drop some artifacts during pipeline

leonardoalt:
This is similar in behavior to what the Pipeline used to do before the DAG refactor. We keep important artifacts such as `asm_string`, `pil_string`, `optimized_pil` and the columns, and drop the artifacts that are built along the way. They can still be computed on-the-fly if some module needs them.

This saves 50% RAM during execution in a heavy benchmark.


chriseth (approved):
Don't know much about the details, but looks good.

georgwiese (approved):
So it was not a single artifact, but more all of them? Do we understand why? For example, is the `ASMProgram` much larger that the ASM string? And if yes, is there a good reason for it?

leonardoalt:
@georgwiese a bit of both:  `constrained_machine_collection` and `linked_machine_graph` seemed to use 5GB each which sounds quite a lot, given the asm string is like 400MB. We should still investigate.

PR #1191, opened 2024-03-22
Fix cache for generic symbols.

chriseth:


chriseth:
This is only the test for now.

georgwiese (approved):
Nice!

PR #1192, opened 2024-03-22
simplify coprocessors

leonardoalt:
The idea here is to move the machines/instructions that are always needed by the RISCV machine to the fixed implementation.

These are Binary, Shift and SplitGL. The first 2 are used in the RISCV asm -> powdr-asm translation, and SplitGL is used by the powdr `mul` instruction.

After that we're going to refactor coprocessors to be more like syscalls.

### DIFF
--- a/riscv/src/coprocessors.rs
+++ b/riscv/src/coprocessors.rs
@@ -3,8 +3,6 @@ use std::{
     convert::TryFrom,
 };
 
-use powdr_number::{FieldElement, KnownField};
-
 type RuntimeFunctionImpl = (&'static str, fn() -> String);
 
 #[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq)]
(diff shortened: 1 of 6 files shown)


riscv/src/coprocessors.rs
```diff
@@ -270,12 +214,9 @@ pub fn call_every_submachine(coprocessors: &CoProcessors) -> Vec<String> {
...
     // https://github.com/powdr-labs/powdr/issues/548
     let mut calls = vec![];
-    if coprocessors.has(BINARY_COPROCESSOR.name) {
-        calls.push("x10 <== and(x10, x10);".to_string());
-    }
-    if coprocessors.has(SHIFT_COPROCESSOR.name) {
-        calls.push("x10 <== shl(x10, x10);".to_string());
-    }
+    calls.push("x10 <== and(x10, x10);".to_string());
+    calls.push("x10 <== shl(x10, x10);".to_string());
```
pacheco, line 218: since these moved to the compiler, maybe these also should, but this is quite hacky already and will probably change with following changes, so maybe fine as is

pacheco (approved):

PR #1193, opened 2024-03-23
Exporting files needed to execute zkevm-prover from starky.

lvella:
This PR is on hold while starky people investigates why the consttree.bin exported from starky is different from JS pil-stark.


leonardoalt:
@lvella is this still not working from the starky side?

leonardoalt:
Closing for now.

PR #1199, opened 2024-03-25
Introduce AllChildren iterator and avoid some clones.

chriseth:


ast/src/parsed/visitor.rs
```diff
@@ -176,3 +182,15 @@ impl<Expr: ExpressionVisitable<Expr>, C: Children<Expr>> ExpressionVisitable<Exp
             .try_for_each(|child| child.visit_expressions(f, o))
     }
 }
+
+impl<Ref> AllChildren<Expression<Ref>> for Expression<Ref> {
+    fn all_children(&self) -> Box<dyn Iterator<Item = &Expression<Ref>> + '_> {
+        Box::new(iter::once(self).chain(self.children().flat_map(|e| e.all_children())))
```
chriseth, line 188: It would also be nice to have `all_children_mut`, but for that, we probably would have to implement our own iterator for Expression. I also don't see how this can be done in safe code. The issue is:

the first call to `next()` has to return `&mut self`. Only after that, we should check what the children are. The problem is, how do we get back ownership of `&mut self`?

ast/src/parsed/visitor.rs
```diff
@@ -14,6 +14,12 @@ pub trait Children<O> {
     fn children_mut(&mut self) -> Box<dyn Iterator<Item = &mut O> + '_>;
 }
 
+pub trait AllChildren<O> {
+    /// Returns an iterator over all direct and indirect children of kind O in this object.
+    /// Visitation order is pre-visit.
```
leonardoalt, line 19:
```suggestion
    /// Pre-order visitor.
```

leonardoalt (approved):

PR #1205, opened 2024-03-27
Basic version of patterns.

chriseth:


ast/src/parsed/mod.rs
```diff
@@ -850,6 +822,47 @@ impl Children<Expression> for ArrayExpression {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub enum Pattern {
+    CatchAll,
+    #[schemars(skip)]
```
Schaeff, line 828: why skip?

chriseth: Because it's not implemented for the bigint type.

executor/src/constant_evaluator/mod.rs
```diff
@@ -176,11 +176,8 @@ mod test {
         let src = r#"
             let N = 8;
             namespace F(N);
-            pol constant LAST(i) { match i {
-                N - 1 => 1,
```
chriseth, line 180: We cannot match against values any more.

chriseth: No, this is just not how patterns work (I think). Rust allows constants, but only constants and not expressions containing constants.

ast/src/parsed/mod.rs
chriseth: What would that achieve? In Type we also don't group by elementary and non-elementary.

pil-analyzer/src/type_inference.rs
```diff
@@ -658,6 +659,58 @@ impl TypeChecker {
...
+            Pattern::Tuple(items) => Type::Tuple(TupleType {
+                items: items
+                    .iter()
+                    .map(|p| self.infer_type_of_pattern(p))
+                    .collect::<Result<_, _>>()?,
+            }),
+            Pattern::Array(items) => {
+                let item_type = self.new_type_var();
+                for item in items {
+                    self.expect_type_of_pattern(&item_type, item)?;
```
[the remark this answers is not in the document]
chriseth: Because array has one single base type and tuple has different types for each item.

I mean in the end, `expect` is also not perfect here, because for `[1, y, z]`, we will get the error that `y` is expected to have `FromLiteral` and we ignore what we match again, but I think it's good enough.

ast/src/parsed/mod.rs
leonardoalt: idk, just feels too specific, but I guess it's fine since those are the only literals

pil-analyzer/tests/parse_display.rs
```diff
@@ -564,3 +558,28 @@ fn let_inside_block_scoping_limited() {
...
+        ((_, y), _) => y,
+        (_, [2]) => 7,
+    };
+    ";
+    analyze_string::<GoldilocksField>(input).to_string();
+}
+
+#[test]
+#[should_panic = "Variable already defined: x"]
+fn patterns_shadowing() {
```
pacheco, line 578: maybe add tests for shadowing cases in #1227? If/when those change it's just a matter of updating test expectations

pil-analyzer/src/expression_processor.rs
```diff
@@ -201,8 +216,7 @@ impl<D: AnalysisDriver> ExpressionProcessor<D> {
             .collect::<Vec<_>>();
 
         let processed_expr = self.process_expression(expr);
-        self.local_variables = previous_local_vars;
-        self.local_variable_counter -= local_var_count;
+        self.reset_local_variables(vars);
```
pacheco, line 219: I think `store|reset_local_variables` could also be used in `process_function`

book/src/pil/expressions.md
```diff
@@ -46,28 +46,34 @@ Parentheses are allowed at any point to force precedence.
 A ``{``-``}``-delimited block can be used everywhere where an expression is expected.
 
 It has the form ``{ <statement> ; <statement> ; ... ; <expression> }``,
-i.e. a sequence of statements followed by an expression and the value of the statement block
-is the value of the final expression.
+i.e. a sequence of statements followed by an expression.
+The statements can either be expressions or let statements: ``let x = ...`` / ``let x;``
```
leonardoalt, line 50: what do non-let expressions do in that case if they're not the last statement?

chriseth: A bit further down:
```
Similarly, an expression at statement level (e.g. ``x * (x - 1) = 0;``) can be used to create new constraints that are added to the global constraint set
and this can only be done inside a [``constr``-functions](#constr-and-query-functions).
```

leonardoalt: ah right

leonardoalt (approved):

PR #1206, opened 2024-03-27
risc-v support for std Rust

leonardoalt:


lvella:
What is the goal here? How do we deal with `std::process`, threads, mutex, time and all the other std stuff?

leonardoalt:
We don't, these will all panic. The goal is to be able to not require `[no_std]` and optimistically support std code that doesn't use that stuff without having to mark it `no_std`. But this PR isn't it, this was just an experiment

PR #1207, opened 2024-03-27
Chore: Fix a small typo in book/src/hello-world.md

vuvoth:
# Changes 
Fix a small typo in hello-world.md. 

This doesn't create any breaking changes.

### DIFF
--- a/book/src/hello_world.md
+++ b/book/src/hello_world.md
@@ -24,12 +24,12 @@ We observe that witness generation fails, and no proof is created.
 # Setup & Verification
 
 The example above omits some important steps in proof generation: Setup and
-Verification key generation.  Some proof systems, such as Halo2 (and other
+Key generation.  Some proof systems, such as Halo2 (and other
 SNARKs), require a Setup to be performed before the proof. Such Setup can be
 specific to the program or universal, where its artifact is a binary usually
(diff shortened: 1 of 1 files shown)


leonardoalt:
Thanks @vuvoth , but I don't think the wording is better in the PR. Moreover the file path is supposed to be relative to the repository's root.

vuvoth:
I created the PR because I see the I inconsistency between a commands in example. I also think more sense if we call Key generation not Proving key generation in documents. Maybe this makes more sense maybe not 😅

book/src/hello_world.md
```diff
@@ -76,4 +76,6 @@ The proof can be verified by anyone via:
 powdr verify hello_world.asm --field bn254 --backend halo2 --vkey "vkey.bin" --params "params.bin" --proof "hello_world_proof.bin"
```
lvella, line 76: You mean inconsistency here? There is no path to `hello_world.asm`?

I agree this is inconsistent, but I think it is better to add the path to this one than to remove from the other, so that the user can just copy and paste and it should work.

leonardoalt:
I've implemented the path fixes here: https://github.com/powdr-labs/powdr/pull/1255

But I don't think the "Verification key" re-wording should go in, so I'm closing this PR for now. Thanks again @vuvoth !

PR #1211, opened 2024-03-27
zkEVM prover backend

lvella:
Similar to #1193, but in here I am just interested in having it working end-to-end, at least for a few cases, so that everybody can try it and build upon.


leonardoalt:
tests are failing

lvella:
> tests are failing

And the funny thing is, locally the broken test pass.

backend/src/estark/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+
+fn pil_hack_fix<'a, F: FieldElement>(
+    pil: &'a Analyzed<F>,
+    fixed: &'a [(String, Vec<F>)],
+) -> (PIL, Vec<(String, Vec<F>)>) {
+    let degree = pil.degree();
+
+    let mut pil: PIL = json_exporter::export(pil);
+
+    // TODO eStark provers requires a fixed column with the equivalent semantics
```
leonardoalt, line 49:
```suggestion
    // TODO eStark provers require a fixed column with the equivalent semantics
```

backend/src/estark/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+    StarkStruct {
+        nBits: n_bits,
+        nBitsExt: n_bits_ext,
+        nQueries: 2,
+        verificationHashType: "GL".to_owned(),
+        steps,
+    }
+}
+
+fn pil_hack_fix<'a, F: FieldElement>(
```
leonardoalt, line 41:
```suggestion
fn pil_first_step_hack<'a, F: FieldElement>(
```
or something a bit more descriptive?

backend/src/estark/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+
+                #[cfg(not(feature = "estark-polygon"))]
+                Ok(Vec::new())
+            }
+        } else {
+            // TODO: use a temporary directory to generate the proof.
+            //
+            // TODO: there is no point in using a temporary dir unless we find a
+            // way to return a meaningful proof to the caller. Preferably a
+            // unified one, compatible with starky.
```
leonardoalt, line 217: Their JSON proofs should be fine for this

leonardoalt: Why not use the temp dir, serialize the proof and return it? JSON should work

backend/src/estark/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+                return polygon_wrapper::prove_and_verify(
+                    &contraints_path,
+                    &stark_struct_path,
+                    &constants_path,
+                    &commits_path,
+                    output_dir,
+                );
+
+                #[cfg(not(feature = "estark-polygon"))]
+                Ok(Vec::new())
```
leonardoalt, line 210: Why doesn't it use the starky prover?

cli/src/main.rs
```diff
@@ -905,6 +905,7 @@ fn optimize_and_output<T: FieldElement>(file: &str) {
 mod test {
     use crate::{run_command, Commands, CsvRenderModeCLI, FieldArgument};
     use powdr_backend::BackendType;
+    use test_log::test;
```
leonardoalt, line 908: why?

pipeline/src/pipeline.rs
```diff
@@ -487,24 +487,11 @@ impl<T: FieldElement> Pipeline<T> {
         Ok(())
     }
 
-    fn maybe_write_constants(&self, constants: &[(String, Vec<T>)]) -> Result<(), Vec<String>> {
-        if let Some(path) = self.path_if_should_write(|name| format!("{name}_constants.bin"))? {
```
leonardoalt, line 491: where are these names given now?

pipeline/src/util.rs
```diff
@@ -48,9 +47,9 @@ pub fn try_read_poly_set<P: PolySet, T: FieldElement>(
         .collect();
 
     (!column_names.is_empty()).then(|| {
-        let fname = format!("{name}_{}", P::FILE_NAME);
+        let path = dir.join(P::FILE_NAME);
```
leonardoalt, line 50: why remove `name`?

pipeline/tests/asm.rs
```diff
@@ -341,22 +342,22 @@ fn read_poly_files() {
         // generate poly files
         let mut pipeline = Pipeline::<Bn254Field>::default()
             .from_file(resolve_test_file(f))
-            .with_output(tmp_dir.to_path_buf(), true);
+            .with_output(tmp_dir.to_path_buf(), true)
+            .with_backend(BackendType::EStarkDump);
```
leonardoalt, line 346: why with backend EStarkDump?

.github/workflows/pr-tests.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       run: git clone https://github.com/0xPolygonHermez/pilcom.git  && cd pilcom && npm install
     - uses: taiki-e/install-action@nextest
     - name: Run default tests
-      run: PILCOM=$(pwd)/pilcom/ cargo nextest run --archive-file tests.tar.zst --verbose
+      run: PILCOM=$(pwd)/pilcom/ RUST_BACKTRACE=1 RUST_LOG=info cargo nextest run --archive-file tests.tar.zst --verbose
```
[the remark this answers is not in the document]
lvella: Because I couldn't reproduce the problem locally. It was a debug commit.

backend/src/estark/mod.rs
lvella: This backend struct is used both by EStarkPolygon and EStarkDump. If feature "estark-polygon" is disabled, this can only be a EStarkDump, which doesn't do any proving, just dump the files.

cli/src/main.rs
lvella: Because there was a dependency but it wasn't being used. This allows for `RUST_LOG` to work on tests.

pipeline/src/pipeline.rs
lvella: They are not. The only thing that differentiates one output from another is the output dir.

pipeline/src/util.rs
lvella: Because these files are now written inside the backend. Either `EStarkDump` or `EStarkPolygon`. Backend already has output_dir. I think think we don't need another identifier string for the files.

pipeline/tests/asm.rs
lvella: Because no EStark files are being written by default now. If you want them, you have to use the `EStarkDump` backend.

leonardoalt:
@lvella are there any tests that are using the new prover?

lvella:
> @lvella are there any tests that are using the new prover?

No. Maybe I should put just one that is known to work in one of the slow subsets?

leonardoalt:
tests are failing

leonardoalt:
I think this PR is almost there, I'm running some tests locally.

For future PRs I think we should
- make sure all tests pass or at least understand why they're not passing
- re-use previously generated consttree/C files
- be able to use starky for consttree generation (I think we depend on starky here)

lvella:
My difficulty now is the deployment of the dependencies on the test run machines. I can bypass this difficulty entirely by not running the test on the PR.

lvella:
You can still run it locally.

leonardoalt:
Yea but that means it'll be constantly out of sync and  break silently

lvella:
Lets merge this, please.

leonardoalt:
> Lets merge this, please.

Well I was waiting for the tests to pass

book/src/installation.md
```diff
@@ -10,6 +10,19 @@ The easiest way to install both is with [`rustup.rs`](https://rustup.rs/).
 On Windows, you will also need a recent version of [Visual Studio](https://visualstudio.microsoft.com/downloads/),
 installed with the "Desktop Development With C++" Workloads option.
 
+If you want to enable `estark-polygon` feature, you also need the following
+runtime dependencies installed on the system:
+- `gcc`
+- `nlohmann-json3-dev`
+
+Yes, gcc and headers at runtime, this is not a typo.
```
leonardoalt, line 18: wdym `gcc and headers at runtime`? I think this sentence is unnecessary. If we want to explain why they're needed we could, but this sentence doesn't do that

riscv/Cargo.toml
```diff
@@ -11,6 +11,7 @@ repository = { workspace = true }
 [features]
 default = [] # complex-tests is disabled by default
 complex-tests = []
+estark-polygon = ["powdr-pipeline/estark-polygon"]
```
leonardoalt, line 14: this is kinda ugly, but fine for now

Issue #164, opened 2023-04-15
Provide log options for the commandline or at least print it in the help screen


chriseth:

chriseth:
No, we neither have a commandline option nor is the env var mentioned in the help screen. I think we should really have `--log <level>` or something.

PR #1212, opened 2024-03-28
Add log option in powdr command line

vuvoth:
# Context 
Now powdr can allow handle log in cli option. 

Example 

```bash
powdr --log DEBUG pil test_data/asm/book/hello_world.asm --field bn254 --force --inputs 0
``` 

--log value is LevelFilter 

```rust
pub enum LevelFilter {
    /// A level lower than all log levels.
    Off,
    /// Corresponds to the `Error` log level.
    Error,
    /// Corresponds to the `Warn` log level.
    Warn,
    /// Corresponds to the `Info` log level.
    Info,
    /// Corresponds to the `Debug` log level.
    Debug,
    /// Corresponds to the `Trace` log level.
    Trace,
} 
```

Solve #164

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -81,6 +81,10 @@ struct Cli {
     #[arg(long, hide = true)]
     markdown_help: bool,
 
+    #[arg(long)]
+    #[arg(default_value_t = LevelFilter::Info)]
+    log: LevelFilter,
+
     #[command(subcommand)]
     command: Option<Commands>,
 }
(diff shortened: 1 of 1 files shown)


leonardoalt (dismissed):

cli/src/main.rs
```diff
@@ -81,6 +81,11 @@ struct Cli {
     #[arg(long, hide = true)]
     markdown_help: bool,
 
+    /// Set log filter value [ Off, Error, Warn, Info, Debug, Trace ]
+    #[arg(long)]
+    #[arg(default_value_t = LevelFilter::Info)]
+    log: LevelFilter,
```
chriseth, line 87: I know I wrote the issue, but I think the parameter should rather be `--log-level`.

Also could you change this so that lowercase levels work? I don't want to write `--log-level DEBUG` but rather `--log-level debug`.

vuvoth: > I know I wrote the issue, but I think the parameter should rather be `--log-level`.

Yes, I update the PR for this :D

> Also could you change this so that lowercase levels work? I don't want to write `--log-level DEBUG` but rather `--log-level debug`.

The "clap" package very smart, we can --log-level DEBUG/debug/Debug, they can understand any.

chriseth (approved):

PR #1214, opened 2024-03-28
Patterns in let statements and function parameters

chriseth:


pil-analyzer/src/evaluator.rs
```diff
@@ -68,11 +68,20 @@ pub fn evaluate_function_call<'a, T: FieldElement>(
```
chriseth, line 68: The above could actually be an assert now.

book/src/pil/patterns.md
```diff
@@ -0,0 +1,75 @@
...
+> Note that PIL does not check that patterns in a match expression are exhaustive.
+
+## (Ir-)refutability
+
+A pattern is refutable if there is a value of the correct type that the pattern does not match.
+An example is the pattern `7` since it does not match all integers, or the patten `[x, ..]`, because it
+does not match the empty array.
+
+Refutable patterns are fine in match arms, because if the pattern does not match, the evaluator will
+just continue trying the next match arm, but they are disallowed in let statements an ind function
```
leonardoalt, line 52:
```suggestion
just continue trying the next match arm, but they are disallowed in let statements and in function
```

pil-analyzer/src/expression_processor.rs
```diff
@@ -174,20 +174,17 @@ impl<D: AnalysisDriver> ExpressionProcessor<D> {
...
-            params
-                .iter()
-                .zip(self.local_variable_counter..)
-                .map(|(p, i)| (p.clone(), i)),
-        );
-        self.local_variable_counter += params.len() as u64;
-
+        for param in params {
+            if !param.is_irrefutable() {
+                panic!("Function parameters must be irrefutable, but {param} is refutable.");
```
leonardoalt, line 184: we should really clean up these panics...

chriseth: You mean use Results instead? I think as long as we don't have source references in sub-expressions, we cannot display a nice error message anyway, so it's fine to use panic  for now

leonardoalt (approved):

pil-analyzer/src/expression_processor.rs
leonardoalt: yea I know, just meant one day

leonardoalt: image

PR #1220, opened 2024-04-01
Change riscv tests to only use rust crates

pacheco:
For single file rust tests, `powdr-riscv-runtime` refers to powdr's github `main` branch.
This makes it impossible to update the `riscv-runtime` crate without breaking tests.
Once we have crates published with semver, those should be usable again.


riscv/tests/riscv.rs
```diff
@@ -16,19 +16,20 @@ use powdr_riscv::{
 /// Compiles and runs a rust file with continuations, runs the full
```
lvella, line 16: The comment mentions rust file.

lvella:
In my opinion, even from command line the support for individual files should be removed.

lvella (approved):

PR #1224, opened 2024-04-01
Coprocessors refactor

pacheco:
- Refactor `CoProcessors` into a `Runtime` with submachines and syscalls
- Calls into prover functionality and submachines done via `ecall` instead of "call symbol replacement"
- Call `poseidon_gl` machine with riscv registers by using the stack (~50% speedup in many_chunks benchmark)
- New `riscv-syscalls` crate keeps the list of syscalls and related constants


riscv/src/continuations/bootloader.rs
```diff
@@ -165,11 +181,12 @@ submachine_init:
 // START OF BOOTLOADER
 
 // During the execution of the bootloader, registers are used as follows:
+// (NB. `SYSCALL_REGISTERS` are mapped into the `P*` registers)
```
leonardoalt, line 184: didn't we remove the P registers?

leonardoalt: ah I see now

riscv/src/compiler.rs
```diff
@@ -141,10 +137,6 @@ pub fn compile<T: FieldElement>(
     // Replace dynamic references to code labels
     replace_dynamic_label_references(&mut statements, &data_labels);
 
-    // Remove the riscv asm stub function, which is used
-    // for compilation, and will not be called.
-    statements = replace_coprocessor_stubs(statements, coprocessors).collect::<Vec<_>>();
```
leonardoalt, line 146: is this now done inside the runtime ?

riscv/src/compiler.rs
```diff
@@ -365,32 +343,6 @@ fn replace_dynamic_label_reference(
...
-        let mut filter_current = *filter_next;
-        *filter_next = predicate(&item);
-        // if the predicate says this line should be filtered, then
-        // the next one should be filtered as well.
-        filter_current |= *filter_next;
-        Some((filter_current, item))
-    })
-    .filter_map(|(filter, statement)| (!filter).then_some(statement))
-}
-
```
leonardoalt, line 382: Why were these removed?

riscv/src/compiler.rs
```diff
@@ -1416,44 +1297,29 @@ fn process_instruction(instr: &str, args: &[Argument], coprocessors: &CoProcesso
...
-                // supposed to use x6 to calculate the high bits of the
-                // destination address. Our implementation does not touch x6,
-                // but no sane program would rely on this behavior, so we are
-                // probably fine.
-                (Some(replacement), "call") => vec![replacement],
-                (Some(replacement), "tail") => {
-                    vec![replacement, "tmp1 <== jump_dyn(x1);".to_string()]
-                }
-                (Some(_), _) => unreachable!(),
-            }
```
leonardoalt, line 1446: Why is it fine to remove this?

riscv/src/compiler.rs
```diff
@@ -1416,44 +1297,29 @@ fn process_instruction(instr: &str, args: &[Argument], coprocessors: &CoProcesso
...
-            vec!["x10 <=X= ${ std::prover::Query::Input(std::convert::int(std::prover::eval(x10))) };".to_string()]
+            vec![
+                // push ra/x1 into stack
+                "x2 <=X= wrap(x2 - 4);".to_string(),
+                "mstore x2, x1;".to_string(),
+                // jump to to handler
+                "x1 <== jump(__ecall_handler);".to_string(),
+                // pop ra/x1
+                "x1, tmp1 <== mload(x2);".to_string(), // restore ra
+                "x2 <=X= wrap(x2 + 4);".to_string(),
```
leonardoalt, line 1316: Could have a helper for this

leonardoalt (commented):
Nice!

riscv/src/runtime.rs
```diff
@@ -0,0 +1,381 @@
...
+pub fn parse_function_statement(input: &str) -> FunctionStatement {
+    let ctx = ParserContext::new(None, input);
+    FUNCTION_STATEMENT_PARSER
+        .parse(&ctx, input)
+        .expect("invalid function statement")
+}
+
+struct SubMachine {
+    /// Full path to machine (e.g, `path::to::Machine`)
+    path: SymbolPath,
```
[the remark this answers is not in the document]
chriseth: In the the pil code, I'm only using full absolute paths as strings at some point as well. Would be important to ensure that it's really absolute paths, but as long as that is guaranteed, String is probably better (although I actually didn't check what happens with `path`).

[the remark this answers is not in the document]
leonardoalt: it gets inserted into the RISCV machine in `riscv/src/compiler.rs`

riscv/src/continuations/bootloader.rs
pacheco: will change `registers` for `variables` in the comment here

riscv/src/compiler.rs
pacheco: there is no more coprocessor stubs to replace: those `C` functions in the runtime became `ecall`s

riscv/src/compiler.rs
pacheco: they were not being used anymore, as there is no need for "stub replacement"

riscv/src/compiler.rs
pacheco: Before, special calls were done by replacing `call/tail special_func` with special handing code.
Now these are `ecall` for specific syscalls

riscv/src/runtime.rs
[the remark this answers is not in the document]
pacheco: My reasoning for using the parsers here was so that an error is signaled here instead of at a later time when trying to parse the full `.asm`

riscv/src/runtime.rs
```diff
@@ -0,0 +1,381 @@
...
+        r.add_submachine(
+            "std::shift::Shift",
+            None,
+            "shift",
+            [
+                "instr shl Y, Z -> X ~ shift.shl;",
+                "instr shr Y, Z -> X ~ shift.shr;",
+            ],
+            ["x10 <== shl(x10, x10);"],
+        );
```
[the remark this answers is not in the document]
pacheco: I think it makes sense to keep submachines here, as part of the risc-v asm runtime.
Maybe the risc-v instructions can be moved to the compiler, together with checks that the machines are part of the runtime?

riscv/src/runtime.rs
```diff
@@ -0,0 +1,381 @@
...
+                            None,
+                            "poseidon_gl",
+                            ["instr poseidon_gl ~ poseidon_gl.poseidon_permutation x10, x11, x12, x13, x14, x15, x16, x17, x6, x7, x28, x29 -> x10', x11', x12', x13';"],
+                            // init call
+                            ["poseidon_gl;",
+                             "x10 <=X= 0;",
+                             "x11 <=X= 0;",
+                             "x12 <=X= 0;",
+                             "x13 <=X= 0;",
+                            ]);
```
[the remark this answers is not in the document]
pacheco: you mean like this?
```
                            [
                                "poseidon_gl;",
                                "x10 <=X= 0;",
                                "x11 <=X= 0;",
                                "x12 <=X= 0;",
                                "x13 <=X= 0;",
                            ]);
```

riscv/src/runtime.rs
```diff
@@ -0,0 +1,381 @@
...
+        let setup = std::iter::once("tmp3 <=X= x10;".to_string());
+
+        // The poseidon instruction uses the first 12 SYSCALL_REGISTERS as input/output.
+        // The contents of memory are loaded into these registers before calling the instruction.
+        // These might be in used by the riscv machine, so we save the registers on the stack.
+        let save_register = |i| {
+            let reg = SYSCALL_REGISTERS[i];
+            [
+                // save register in stack
+                "x2 <=X= wrap(x2 - 4);".to_string(),
```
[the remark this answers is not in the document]
pacheco: Why would it not be ok to use `x2`?
I'm not completely sure of the need for the wrap, I followed similar code in the compiler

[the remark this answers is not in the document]
leonardoalt: I mean this is overwriting `x2`

riscv/src/compiler.rs
leonardoalt: this whole block of powdr asm

riscv/src/runtime.rs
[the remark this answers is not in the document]
leonardoalt: but that will only mean that the path is well formed, right? it doesn't actually look for the machine

riscv/src/runtime.rs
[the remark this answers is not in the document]
leonardoalt: yea probably fine as is

riscv/src/compiler.rs
pacheco: idk, looks ok to me

riscv-syscalls/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+/// These are all _caller_ saved registers.
+/// Thus, registers directly used as input/output to the ecall shouldn't need any special handling.
+/// The syscall implementation itself may use the other registers for temporary storage.
+/// * Since this usage is NOT visible from riscv asm, these must be explicitly saved and restored *
+/// For an example, see the `poseidon_gl` syscall implementation in the riscv runtime.
+pub static SYSCALL_REGISTERS: [&str; 14] = [
+    "x10", "x11", "x12", "x13", "x14", "x15", "x16", "x17", "x6", "x7", "x28", "x29", "x30", "x31",
+];
+
+// NB. Must be kept in sync with conversion trait implementations
```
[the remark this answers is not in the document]
pacheco: Nota bene (https://en.wikipedia.org/wiki/Nota_bene), its used as a "pay attention" marker, but I can remove/change it.

riscv/src/runtime.rs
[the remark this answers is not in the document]
pacheco: i've added the "has_submachine" checks to the compiler just to be safe

riscv/src/runtime.rs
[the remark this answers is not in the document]
pacheco: ah right, idk why cargo-fmt doesn't do it

riscv/src/compiler.rs
leonardoalt: lint isn't complaining so I guess it's fine

riscv/src/runtime.rs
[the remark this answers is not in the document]
leonardoalt: sounds good!

riscv/src/compiler.rs
leonardoalt: wait, ignore my previous comment, I confused the topic

leonardoalt: here I just meant this block of asm could be useful for other cases where we might need the same thing. But fine with waiting for that to happen

riscv/src/runtime.rs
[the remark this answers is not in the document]
pacheco: yes, its only checking that the asm strings here are well formed

riscv/src/compiler.rs
pacheco: i refactored the push/pop to stack to helper methods

riscv-runtime/src/hash.rs
```diff
@@ -0,0 +1,31 @@
...
+/// is less than the Goldilocks field.
+/// The unsafe version does not perform such checks.
+pub fn poseidon_gl(mut data: [u64; 12]) -> [u64; 4] {
+    for &n in data.iter() {
+        assert!(n < GOLDILOCKS);
+    }
+
+    unsafe {
+        asm!("ecall", in("a0") &mut data as *mut [u64; 12], in("t0") Syscall::PoseidonGL as u32);
+    };
```
Schaeff, line 20: Would it make sense to call `poseidon_gl_unsafe` here instead?

riscv/src/compiler.rs
```diff
@@ -444,7 +444,14 @@ fn preamble<T: FieldElement>(degree: u64, runtime: &Runtime, with_bootloader: bo
         "".to_string()
     };
 
-    let mul_instruction = mul_instruction::<T>();
+    for machine in ["binary", "shift"] {
```
leonardoalt, line 447:
```suggestion
    for machine in ["binary", "shift", "split_gl"] {
```
it's needed for the `mul` instruction

leonardoalt: ah you added it below already

riscv/src/runtime.rs
```diff
@@ -76,7 +78,7 @@ impl Runtime {
         };
 
         // Base submachines
-        // TODO: can/should the memory machine be part be here also?
+        // TODO: can/should the memory machine be part of the runtime also?
```
leonardoalt, line 81: when we move it out of compiler.rs, I think so

riscv/src/runtime.rs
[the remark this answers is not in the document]
leonardoalt: I still think this should be changed

riscv-runtime/src/hash.rs
leonardoalt: right, makes sense I think

riscv-runtime/src/hash.rs
```diff
@@ -0,0 +1,31 @@
...
+/// The safe version below also checks that each u64 element
+/// is less than the Goldilocks field.
+/// The unsafe version does not perform such checks.
+pub fn poseidon_gl(mut data: [u64; 12]) -> [u64; 4] {
+    for &n in data.iter() {
+        assert!(n < GOLDILOCKS);
+    }
+
+    unsafe {
+        asm!("ecall", in("a0") &mut data as *mut [u64; 12], in("t0") Syscall::PoseidonGL as u32);
```
chriseth, line 19: Why not implement `Into<u32>`?

riscv-runtime/src/hash.rs
```diff
@@ -0,0 +1,31 @@
...
+    };
+
+    [data[0], data[1], data[2], data[3]]
+}
+
+pub fn poseidon_gl_unsafe(mut data: [u64; 12]) -> [u64; 4] {
```
chriseth, line 25: It's a public function in the runtime, I think it should have documentation.

riscv/src/compiler.rs
```diff
@@ -1416,44 +1324,27 @@ fn process_instruction(instr: &str, args: &[Argument], coprocessors: &CoProcesso
...
+                // restore ra/x1
+                .chain(pop_register("x1"))
+                .collect()
         }
         "ebreak" => {
             assert!(args.is_empty());
-            // This is using x0 on purpose, because we do not want to introduce
-            // nondeterminism with this.
-            vec!["x0 <=X= ${ std::prover::Query::PrintChar(std::convert::int(std::prover::eval(x10))) };\n".to_string()]
+            // we don't use ebreak for anything, ignore
```
chriseth, line 1346: Shouldn't we panic? Is his used anywhere?

pacheco: yes, its used in the risc-v asm instruction IRC, for example

chriseth: Maybe at least prevent witgen from succeeding if this is executed?

chriseth: I would prefer removing/modifying the instruction test, because I think this can be dangerous.

pacheco: How would a `noop` in place of an `ebreak` be dangerous? This was previously used to call prover print_char.
The tests that use `ebreak` are those generated from the risc-v test suite...
Should I open an issue?

chriseth (approved):

PR #1231, opened 2024-04-03
sample adaptations

qwang98:

### DIFF
--- a/std/hash/keccaf.asm
+++ b/std/hash/keccaf.asm
@@ -1,0 +1,95 @@
+use std::array;
+
+// TODO
+let<T1, T2> routine: T1, T1, (T1, T1 -> T1), (T1, T1 -> T1), (T1, T1 -> T1) -> T2 =
+    |a, b, xor, and, or| a + b;
+
+// function definition
+let or: int, int -> int = |a, b| a | b;
+let and: int, int -> int = |a, b| a & b;
+let xor: int, int -> int = |a, b| a ^ b;
+
+// ln 9 - 12
+// TODO: check that these are good for word size 32
+let RHO: int[24] = [
+    1,  3,  6,  10, 15, 21, 28, 36, 45, 55, 2,  14,
+    27, 41, 56, 8,  25, 43, 62, 18, 39, 61, 20, 44
+];
+
+// ln 14 - 16
+// TODO: check that these are good for word size 32
+let PI: int[24] = [
+    10, 7,  11, 17, 18, 3, 5,  16, 8,  21, 24, 4,
+    15, 23, 19, 13, 12, 2, 20, 14, 22, 9,  6,  1
+];
+
+// ln 19 - 28
+// TODO: check that these are good for word size 32
+let RC: int[24] = [
+    0x0000000000000001, 0x0000000000008082, 0x800000000000808a,
+    0x8000000080008000, 0x000000000000808b, 0x0000000080000001,
+    0x8000000080008081, 0x8000000000008009, 0x000000000000008a,
+    0x0000000000000088, 0x0000000080008009, 0x000000008000000a,
+    0x000000008000808b, 0x800000000000008b, 0x8000000000008089,
+    0x8000000000008003, 0x8000000000008002, 0x8000000000000080,
+    0x000000000000800a, 0x800000008000000a, 0x8000000080008081,
+    0x8000000000008080, 0x0000000080000001, 0x8000000080008008
+];
+
+// ln 35 - 40
+// change endianness for a 32 bit number byte by byte
+// e.g. 0xaabbccdd -> 0xddccbbaa
+let swap_u32: int -> int = |x| 
+    or((or(and((x << 8), 0xff00ff00), and((x >> 8), 0x00ff00ff)) >> 16),
+    (or(and((x << 8), 0xff00ff00), and((x >> 8), 0x00ff00ff)) << 16)); 
+
+// compression function
+
+// ln 44
+let bc: int[5] = [0, 0, 0, 0, 0];
+
+// ln 45
+let t: int = 0;
+
+// ln 47 - 49
+let swap_u32_loop: int[25] -> int[25] = |st| array::map(array::new(25, |i| i), |i| swap_u32(st[i]));
+
+// ln 53 - 55
+let theta_bc: int[25], i -> int = |st, i| xor(st[i], xor(st[i + 5], xor(st[i + 10], xor(st[i + 15], st[i + 20]))));
+let theta_bc_loop: int[25] -> int[5] = |st| array::map(array::new(5, |i| i), |i| theta_bc(st, i));
+
+// ln 57 - 62
+let theta_t: int[5] -> int = |bc, i| xor(bc[(i + 4) % 5], rotl32(bc[(i + 1) % 5], 1));
+let theta_st: int[25], int, int, int -> int = |st, t, j, i| xor(st[(j * 5) + i], t);
+let theta_st_loop: int[25], int, int -> int[5] = |st, t, i| array::map(array::new(5, |i| i), |j| theta_st(st, t, j, i));
+let theta_t_loop: int[5], int[25], int -> int[5][5] = |bc, st, t| array::map(array::new(5, |i| i), |i| {
+    let t = theta_t(bc, i); // int
+    let st_partial = theta_st_loop(st, t, i); // int[5]
+    st_partial
+});
+// rearrange st returned in ln 60 - version 1: calls theta_t_loop
+let theta_t_loop_rearrange: int[5], int[25], int -> int[25] = |bc, st, t| {
+    let st_partials = theta_t_loop(bc, st, t);
+    let st = [
+        st_partials[0][0], st_partials[1][0], st_partials[2][0], st_partials[3][0], st_partials[4][0],
+        st_partials[0][1], st_partials[1][1], st_partials[2][1], st_partials[3][1], st_partials[4][1],
+        st_partials[0][2], st_partials[1][2], st_partials[2][2], st_partials[3][2], st_partials[4][2],
+        st_partials[0][3], st_partials[1][3], st_partials[2][3], st_partials[3][3], st_partials[4][3],
+        st_partials[0][4], st_partials[1][4], st_partials[2][4], st_partials[3][4], st_partials[4][4]
+    ];
+    st 
+};
+// rearrange st returned in ln 60 - version 2: needs to be called with theta_t_loop in another function
+let theta_t_loop_rearrange_v2: int[5][5] -> int[25] = |sp_partials| 
+    std::array::fold(std::array::new(5, |i| i), [], |initial, i| 
+        a + [sp_partials[0][i], sp_partials[1][i], sp_partials[2][i], sp_partials[3][i], sp_partials[4][i]]
+    );
+
+// ln 64
+let theta_t_reassign: int[25] -> int = |st| st[1];
+
+
+// main machine
+machine Main { 
+    let x; 
+}


std/hash/keccaf.asm
```diff
@@ -0,0 +1,95 @@
...
+let bc: int[5] = [0, 0, 0, 0, 0];
+
+// ln 45
+let t: int = 0;
+
+// ln 47 - 49
+let swap_u32_loop: int[25] -> int[25] = |st| array::map(array::new(25, |i| i), |i| swap_u32(st[i]));
+
+// ln 53 - 55
+let theta_bc: int[25], i -> int = |st, i| xor(st[i], xor(st[i + 5], xor(st[i + 10], xor(st[i + 15], st[i + 20]))));
```
chriseth, line 58: To improve readability, we could also introduce a helper function
```suggestion
let xor_mult: int[] -> int = |input| std::array::fold(input, 0, |x, y| xor(x, y));
let theta_bc: int[25], i -> int = |st, i| xor_mult([st[i], st[i + 5], st[i + 10], st[i + 15], st[i + 20]]);
```

std/hash/keccaf.asm
```diff
@@ -0,0 +1,95 @@
...
+let theta_bc_loop: int[25] -> int[5] = |st| array::map(array::new(5, |i| i), |i| theta_bc(st, i));
```
chriseth, line 59:
```suggestion
let theta_bc_loop: int[25] -> int[5] = |st| array::new(5, |i| theta_bc(st, i));
```

std/hash/keccaf.asm
```diff
@@ -0,0 +1,142 @@
...
+        st_partials[0][3], st_partials[1][3], st_partials[2][3], st_partials[3][3], st_partials[4][3],
+        st_partials[0][4], st_partials[1][4], st_partials[2][4], st_partials[3][4], st_partials[4][4]
+    ];
+    st 
+};
+// rearrange st returned in ln 60 - version 2: needs to be called with theta_t_loop in another function
+let theta_t_loop_rearrange_v2: int[5][5] -> int[25] = |sp_partials| 
+    array::fold(array::new(5, |i| i), [], |initial, i| 
+        a + [sp_partials[0][i], sp_partials[1][i], sp_partials[2][i], sp_partials[3][i], sp_partials[4][i]]
+    );
```
chriseth, line 90:
```suggestion
    array::new(25, |i| sp_partials[i %5][i / 5]);
```

std/hash/keccaf.asm
```diff
@@ -0,0 +1,142 @@
...
+// ln 57 - 62
+let theta_t: int[5] -> int = |bc, i| xor(bc[(i + 4) % 5], rotl32(bc[(i + 1) % 5], 1));
+let theta_st: int[25], int, int, int -> int = |st, t, j, i| xor(st[(j * 5) + i], t);
+let theta_st_loop: int[25], int, int -> int[5] = |st, t, i| array::map(array::new(5, |i| i), |j| theta_st(st, t, j, i));
+let theta_t_loop: int[5], int[25], int -> int[5][5] = |bc, st, t| array::map(array::new(5, |i| i), |i| {
+    let t = theta_t(bc, i); // int
+    let st_partial = theta_st_loop(st, t, i); // int[5]
+    st_partial
+});
+// rearrange st returned in ln 60 - version 1: calls theta_t_loop
```
chriseth, line 74: I don't understand, why do you need to rearrange?

Once the 5 bc values are computed in line 54, we can compute the new st values in line 60 in the following way

```
let theta = |bc, st| std::array::map_enumerated(st, |k, x| {
  let i = k % 5;
  let j = k / 5;
  let t = bc[(i + 4) % 5] ^ rotl64(bc[(i + 1) % 5], 1);
  xor(x, t)
});
```

qwang98: Agree on this. I just didn't know this kind of syntax for nested loop but now I do!

qwang98:
Thanks for the very helpful comments. I finished a first draft of the keccak helper functions. With `main` as the entry function. A few questions:
1. The `else if` syntax doesn't seem to work for one instance I'm using it.
2. Should we test on 64 bit words to make sure the outputs match first, before adapting to 32 bit words?

Next steps will be constructing the `routine`, potentially flattening the circuit, and write up the main machine with PIL columns. Let me know if my understanding is correct.

std/hash/keccaf.asm
qwang98: Fixed

std/hash/keccaf.asm
qwang98: Fixed

std/hash/keccaf.asm
qwang98: Fixed

chriseth:
Thanks a lot! I'll take another look.

1. The "else if" is not yet implemented, you have to use "else { if c { ... } }".
3. There is only the unlimited precision `int` type.
4. Yes, I think it's a good idea to have tests first, maybe even merge and then adapt to 32 bits.

2. and rest:  I think we should juts implement and test the keccakf function for now. It is not really clear yet how to pass teh input and output. Actually I think we can even remove the swap-part - since this only swaps endianness if I'm not mistaken.

I would say the next step is to test keccakf against a rust implementation (or also just some fixed example numbers).

std/hash/keccaf.asm
```diff
@@ -0,0 +1,163 @@
...
+// Note that neither t nor bc is needed as they are both helper variables
+// ln 52 - 55
+let xor_mult: int[] -> int = |input| std::array::fold(input, 0, |x, y| xor(x, y));
+let theta_bc: int[25] -> int = |st, i| |i| xor_mult([st[i], st[i + 5], st[i + 10], st[i + 15], st[i + 20]]);
+
+// ln 57 - 62
+let theta_st: int[25] -> int[25] = |st| array::map_enumerated(st, |idx, elem| {
+    let i = idx % 5;
+    let j = idx / 5;
+    let t = xor(theta_bc(st, (i + 4) % 5), rotl32(theta_bc(st, (i + 1) % 5), 1));
```
chriseth, line 63: Shouldn't this be rotl64?

std/hash/keccaf.asm
```diff
@@ -0,0 +1,163 @@
...
+    new_st + [new_st_j]
+});
+// rearrange st_j
+let rho_pi_rearrange: int[25] -> int[25] = |st| {
+    let rearranged_st = [
+        st[23], st[17], st[5], st[11], st[6], st[22],
+        st[1], st[8], st[21], st[0], st[2], st[16], 
+        st[15], st[19], st[12], st[7], st[3], st[4], 
+        st[14], st[18], st[9], st[20], st[13], st[10]
+    ];
```
chriseth, line 86: This is essentially the inverse of PI, right? Is there a way to compute it from PI?

std/hash/keccaf.asm
```diff
@@ -0,0 +1,163 @@
...
+let rho_pi: int[25], int -> int = |st, i| {
+    let t = if i == 0 { st[1] } else { st[PI[i - 1]] };
+    let new_st_j = rotl32(t, RHO[i]);
+    (new_st_j)
+};
+// collect st_j
+let rho_pi_loop: int[25] -> int[25] = |st| utils::fold(24, |i| i, [], |new_st, idx| {
+    let new_st_j = rho_pi(st, idx);
+    new_st + [new_st_j]
+});
```
chriseth, line 78: Isn't this
```suggestion
let rho_pi_loop: int[25] -> int[25] = |st| std::array::new(25, |i| rho_pi(st, i));
});
```
?

std/hash/keccaf.asm
```diff
@@ -0,0 +1,163 @@
...
+        st[15], st[19], st[12], st[7], st[3], st[4], 
+        st[14], st[18], st[9], st[20], st[13], st[10]
+    ];
+    rearranged_st
+};
+
+// ln 74 - 83
+// chi
+// TODO: make sure that modulus has the same precedence as multiplication
+let chi: int[25] -> int[25] = |st| array::map_enumerated(st, |idx, elem| {
```
chriseth, line 93: nice!

std/hash/keccaf.asm
```diff
@@ -0,0 +1,163 @@
...
+    let j = idx % 5;
+    xor(st[idx], and(not(st[i * 5 + (j + 1) % 5]), st[i * 5 + (j + 2) % 5]))
+});
+
+// ln 85 - 86
+// iota
+let iota: int[25], int -> int[25] = |st, r| array::new(24, |i| if i == 0 { xor(st[0], RC[r]) } else { st[i] } );
+
+// ln 51 - 87
+let r_loop: int[25] -> int[25] = |st| utils::fold(24, |i| i, st, |acc, i| iota(chi(rho_pi_rearrange(rho_pi_loop(theta_st(acc)))), r) );
```
chriseth, line 104:
```suggestion
let r_loop: int[25] -> int[25] = |st| utils::fold(24, |i| i, st, |acc, r| iota(chi(rho_pi_rearrange(rho_pi_loop(theta_st(acc)))), r) );
```

std/hash/keccaf.asm
```diff
@@ -0,0 +1,163 @@
...
+// TODO: make sure that modulus has the same precedence as multiplication
+let chi: int[25] -> int[25] = |st| array::map_enumerated(st, |idx, elem| {
+    let i = idx / 5;
+    let j = idx % 5;
+    xor(st[idx], and(not(st[i * 5 + (j + 1) % 5]), st[i * 5 + (j + 2) % 5]))
+});
+
+// ln 85 - 86
+// iota
+let iota: int[25], int -> int[25] = |st, r| array::new(24, |i| if i == 0 { xor(st[0], RC[r]) } else { st[i] } );
```
chriseth, line 101: maybe use map_enumerated?

std/hash/keccaf.asm
qwang98: There's a functional way of doing this in a double loop, but I think it's not as efficient as just hard coding an inverse vector constant, which is what I did.

std/hash/keccaf.asm
qwang98: Agreed.

std/hash/keccaf.asm
qwang98: Agreed.

qwang98:
All comments addressed.

A few updates:
1. I created the routine function with all binary function helpers (and, or, xor, not) as parameters in this draft PR (demo only): https://github.com/powdr-labs/powdr/pull/1266/files However, it doesn't look very clean and honestly I think it's better to not have them as parameters, but directly provided in the file. Let me know if you agree.
2. I tried to create a machine out of the asm code I wrote, so that I can refer to the machine in a separate test file. I tried to modify it off the poseidon example, but it's not very intuitive how I should proceed, given that the poseidon example mostly used raw PIL constraints, while I mostly used the asm language (if that's how it's called).

I think I'm at a bottleneck of not knowing how to set my code up for testing, so would be great to have a call on Thursday/Friday about these. I'll reach out separately.

std/hash/keccaf.asm
```diff
@@ -9,24 +9,33 @@ let<T1, T2> routine: T1, T1, (T1, T1 -> T1), (T1, T1 -> T1), (T1, T1 -> T1) -> T
...
 ];
 
 // ln 14 - 16
 // TODO: check that these are good for word size 32
-let PI: int[24] = [
+let PI: int[] = [
     10, 7,  11, 17, 18, 3, 5,  16, 8,  21, 24, 4,
     15, 23, 19, 13, 12, 2, 20, 14, 22, 9,  6,  1
 ];
+// inverse of PI using 1 index plus zero at the beginning
```
chriseth, line 24: Can you add a formal statement about the relation between the two?

chriseth:
Yes, the type checker unfortunately does not understand fixed-size arrays.

About testing: I think the proper place should be `pipeline/tests/asm.rs`. The current tests in that file do not do the same thing we should test, but there is a file `powdr_str.rs` in the same directory that calls a function `evaluate_integer_function`. I think a combination of that could allow you to run some tests on individual functions.

Another way of testing could also be to use assertions (`std::check::assert`). You can for example add a `test` function. Or we do it the rust way and add a `test` module inside that file and add helpers that call all functions inside the test module (we don't have annotations unfortunately).

chriseth:
I think until everything is working, we should not put this into the std directory, but just into a test file instead.

chriseth:
Are the tests running through?

qwang98:
> Are the tests running through?

Yes. I'm able to set up tests. It's just that they don't match existing implementations I'm testing against. I'm trying to debug it now.

pil-analyzer/src/evaluator.rs
```diff
@@ -356,7 +356,7 @@ impl<'a, T: Display> Display for Value<'a, T> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Value::Bool(b) => write!(f, "{b}"),
-            Value::Integer(x) => write!(f, "{x}"),
+            Value::Integer(x) => write!(f, "{:#x}", x),
```
qwang98, line 359: This is for easier debugging, so I'm leaving it here until we are ready to have it merged.

qwang98:
@chriseth @leonardoalt 

Changes since last time:
1. Passes all test vectors added to the pipeline.
2. Intermediate function tests in commit history, but removed for the review.

Caveat that this is the u64 version, so would still need another adaptation to make it compatible with Goldilocks field.

Questions:
1. Due to the unlimited number of bits integer, extra bitwise and masking is needed in wrapped bit rotations and endianness swaps to confine to 64 bits. Might be good to implement integer types with specific bits in the future.
2. Keccak currently works well alone. Not sure if the next step would be converting it to 32 bits or integrating it with other circuits using connects.

leonardoalt:
@qwang98 quick comment: can you fix fmt/lint so the tests run on CI?

chriseth:
Oh nice this is great! Can you move the files out of the std directory into a test directory, then we can merge this as a first step.

qwang98:
@chriseth 

Ready for review again! 

The function that we care the most about will be `keccakf_inner`, which takes input after padding and endianness swap.

I also kept test vectors for other functions such as `keccakf` and the full keccak `main`, because they have test vectors that are consistent with Zokrates, so we gain the assurance that the function is implemented correctly and understanding of how `keccakf_inner` is linked to the final keccak result.

pipeline/tests/asm.rs
```diff
@@ -454,3 +454,263 @@ fn test_failing_assertion() {
...
+                    panic!("Expected integer.");
+                }
+            }
+        } else {
+            panic!("Expected array.");
+        }
+    }
+
+    // Test keccakf_inner
+    let padded_endianness_swapped_input: Vec<u64> = vec![
```
chriseth, line 500: Why not `= padded_input.iter().map(|x| x.swap_bytes()).collect()`?

pipeline/tests/asm.rs
```diff
@@ -454,3 +454,263 @@ fn test_failing_assertion() {
...
+            let mut v = vec![0x2a; 399];
+            v.push(0x01);
+            v
+        },
+        // All zero input test vector
+        [0x00; 256].to_vec(),
+    ];
+
+    let expected = vec![
+        // ca85d1976d40dcb6ca3becc8c6596e83c0774f4185cf016a05834f5856a37f39
```
chriseth, line 673: What about
```suggestion
        vec!["ca85d1976d40dcb6ca3becc8c6596e83c0774f4185cf016a05834f5856a37f39", ... /* all the others */]
            .into_iter()
            .map(|x|
                (0..x.len())
                    .step_by(2)
                    .map(|i| u8::from_str_radix(&x[i..i + 2], 16).unwrap())
                    .collect()
             ).collect();
```

test_data/asm/keccak.asm
```diff
@@ -0,0 +1,200 @@
...
+use std::utils;
+
+// function definition
+let or: int, int -> int = |a, b| a | b;
+let and: int, int -> int = |a, b| a & b;
+let xor: int, int -> int = |a, b| a ^ b;
+let not: int -> int = |a| a ^ 0xffffffffffffffff; // bitwise not for 64 bits
+
+// ln 9 - 12
+// TODO: check that these are good for word size 32
```
chriseth, line 11: I think you can remove those

test_data/asm/keccak.asm
```diff
@@ -0,0 +1,200 @@
...
+];
+
+// ln 14 - 16
+// TODO: check that these are good for word size 32
+let PI: int[] = [
+    10, 7,  11, 17, 18, 3, 5,  16, 8,  21, 24, 4,
+    15, 23, 19, 13, 12, 2, 20, 14, 22, 9,  6,  1
+];
+
+// inverse of PI using 1 index plus zero at the beginning
```
chriseth, line 24: Please also add a more formal statement about the relation between PI and PI_INVERSE

test_data/asm/keccak.asm
```diff
@@ -0,0 +1,200 @@
...
+    0x000000008000808b, 0x800000000000008b, 0x8000000000008089,
+    0x8000000000008003, 0x8000000000008002, 0x8000000000000080,
+    0x000000000000800a, 0x800000008000000a, 0x8000000080008081,
+    0x8000000000008080, 0x0000000080000001, 0x8000000080008008
+];
+
+// ln 30 - 33
+// left rotation
+// 32 and 0xffffffff if u32
+let rotl64: int, int -> int = |x, n| and(or((x << n), (x >> (64 - n))), 0xffffffffffffffff);
```
chriseth, line 48: Shouldn't this move to the beginning of the file? Because I think this is one of the elementary gates.

test_data/asm/keccak.asm
```diff
@@ -0,0 +1,200 @@
...
+    let b_finalized = update_finalize_b(input,rate, delim); // 200 bytes
+
+    // ln 160
+    // note that the biggest W is 64, i.e. 512-bit keccak
+    // therefore, W should always be lower than byte length of b_finalized (200)
+    array::new(W, |i| b_finalized[i])
+};
+
+// main machine
+// TODO: not sure how to integrate these
```
chriseth, line 185: Please make sure that there are no TODOs left (at least those that are in scope here)

pipeline/tests/asm.rs
qwang98: Nice!

test_data/asm/keccak.asm
qwang98: Yes

test_data/asm/keccak.asm
qwang98: Done

test_data/asm/keccak.asm
qwang98: Agree

test_data/asm/keccak.asm
qwang98: Done

PR #1252, opened 2024-04-04
Constr as user-defined enum

chriseth:
Turns some of the built-in types into user types in the prelude.


chriseth:
Answer of the question: In `std::prelude` - https://github.com/powdr-labs/powdr/issues/1319

ast/src/analyzed/display.rs
```diff
@@ -41,6 +41,10 @@ impl<T: Display> Display for Analyzed<T> {
         for statement in &self.source_order {
             match statement {
                 StatementIdentifier::Definition(name) => {
+                    if self.auto_added_symbols.contains(name) {
+                        // Core symbol added automatically, no need to print.
```
chriseth, line 45: We should call them "prelude symbols"

chriseth: Or maybe the asm-prelude is something else than the pil core?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -133,6 +146,24 @@ impl PILAnalyzer {
...
     }
 
+    /// Adds core types if they are not present in the input.
+    /// These need to be present because the type checker relies on them.
+    fn core_types_if_not_present(&self) -> Option<PILFile> {
+        (!self.known_symbols.contains_key("std::prelude::Constr")).then(|| {
+            parse(
+                None,
+                "namespace std::prelude;
+    enum Constr {
```
chriseth, line 156: This overlaps with prelude.asm. We need the type here because otherwise, if we have a function
```
let x: -> expr = || 1;
let y = || x() = x();
```
Then the compiler will determine the type of `y` to be `-> std::prelude::Constr`, with the result that we cannot re-parse the printed and type-annotated source code.

book/src/pil/expressions.md
```diff
@@ -201,5 +201,5 @@ let square_of = query |x| { let v = std::prover::eval(x); v * v };
 let new_bool = constr |x| { let x = new_wit(); x * (x - 1) = 0; x };
 // This is a pure function that only returns a constraint, but does not add it
 // to the global set of constraints.
-let constrain_to_bool: expr -> constr = |x| x * (x - 1) = 0;
+let constrain_to_bool: expr -> Constr = |x| x * (x - 1) = 0;
```
pacheco, line 204: I was momentarily confused by the name `constrain_to_bool`, as we have a very similarly named type `Constr` and this follows the naming for "conversion" functions.
Maybe this could be called something else? Maybe `ensure_bool`? Or even `bool_constraint`?

leonardoalt (approved):

PR #1256, opened 2024-04-05
Halo2 single/aggr SNARK with Solidity verifier

leonardoalt:


book/src/hello_world.md
```diff
@@ -9,14 +9,14 @@ Let's write [a minimal VM](https://github.com/powdr-labs/powdr/blob/main/test_da
 Then let's generate a proof of execution for the valid prover input `0` (since for `0 + 1 - 1 == 0`)
 
 ```console
-powdr pil hello_world.asm --field bn254 --force --inputs 0 --prove-with halo2
```
Schaeff, line 12: Maybe we should have these in a shell script which we run in tests, and include fragments here the same way we do for powdr code snippets.

Schaeff: Can be another PR

halo2/src/prover.rs
```diff
@@ -137,11 +168,82 @@ impl<'a, F: FieldElement> Halo2Prover<'a, F> {
         Ok(proof)
     }
 
-    pub fn prove_aggr(
+    /// Generate a single proof for a given PIL using Keccak transcripts.
+    /// These proofs can be verified directly on Ethereum.
+    pub fn prove_snark_single(
```
georgwiese, line 173: There seems to be a lot of duplicated code with `prove_poseidon`. The only difference is the transcript. Maybe we could have a `prove<T>()` function that is called in both cases?

halo2/src/prover.rs
```diff
@@ -285,11 +433,84 @@ impl<'a, F: FieldElement> Halo2Prover<'a, F> {
         )
     }
 
+    pub fn verify_snark(&self, proof: &[u8], instances: &[Vec<F>]) -> Result<(), String> {
```
georgwiese, line 436: Same here, it's basically the same as `verify_poseidon`

backend/src/halo2_impl.rs
```diff
@@ -63,6 +83,20 @@ impl<'a, T: FieldElement> Backend<'a, T> for Halo2Prover<'a, T> {
 
         Ok(())
     }
+
+    fn export_ethereum_verifier(&self, output: &mut dyn io::Write) -> Result<(), Error> {
+        match self.proof_type() {
+            ProofType::Poseidon => Err(Error::BackendError(
+                "Ethereum verifier is only supported for snark_single and snark_aggr".to_string(),
+            )),
```
georgwiese, line 91:
```suggestion
            ProofType::Poseidon => Err(Error::NoEthereumVerifierAvailable),
```
No?

georgwiese (commented):
Nice! Had a first look through the code (not the book)

backend/src/pilstark/estark.rs
```diff
@@ -20,6 +20,23 @@ use starky::{
...
+    StarkBN,
+    SnarkBN,
+}
+
+impl From<BackendOptions> for ProofType {
+    fn from(options: BackendOptions) -> Self {
+        match options.as_str() {
+            "" | "stark_gl" => ProofType::StarkGL,
+            "stark_bn" => ProofType::StarkBN,
+            "snark_bn" => ProofType::SnarkBN,
```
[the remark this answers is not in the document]
leonardoalt: yes, at some point. Both Polygon-zkevm and estark (starky) have the recursion part that allows generating a BN snark out of a BN stark

halo2/src/prover.rs
```diff
@@ -12,14 +12,29 @@ use halo2_proofs::{
...
     transcript::{EncodedChallenge, TranscriptReadBuffer, TranscriptWriterBuffer},
 };
+
 use powdr_ast::analyzed::Analyzed;
 use powdr_executor::witgen::WitgenCallback;
 use powdr_number::{DegreeType, FieldElement, KnownField};
+
+// We use two different EVM verifier libraries.
+// 1. snark_verifier: supports single snark verification as well as aggregated proof verification.
+// However the generated smart contract code size is larger than the limit on Ethereum. This is mitigated in (2).
```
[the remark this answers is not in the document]
leonardoalt: Yes bad wording there, should have said "... is often larger ... "

halo2/src/prover.rs
```diff
@@ -49,50 +70,60 @@ pub struct Halo2Prover<'a, F> {
...
+    proof_type: ProofType,
 }
 
 fn degree_bits(degree: DegreeType) -> u32 {
     DegreeType::BITS - degree.leading_zeros() + 1
 }
 
 pub fn generate_setup(size: DegreeType) -> ParamsKZG<Bn256> {
-    ParamsKZG::<Bn256>::new(degree_bits(size))
+    // Halo2 does not like degree < 4, so we enforce a minimum of 4 here.
```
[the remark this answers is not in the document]
leonardoalt: yep

halo2/src/prover.rs
```diff
@@ -49,50 +70,60 @@ pub struct Halo2Prover<'a, F> {
...
         setup: Option<&mut dyn io::Read>,
+        proof_type: ProofType,
     ) -> Result<Self, io::Error> {
         Self::assert_field_is_bn254();
 
         let params = setup
             .map(|mut setup| ParamsKZG::<Bn256>::read(&mut setup))
             .transpose()?
-            .map(|mut params| {
-                params.downsize(degree_bits(analyzed.degree()));
```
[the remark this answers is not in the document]
leonardoalt: yea I think I actually have it back in my local unpushed version. Still running into some tests issues

backend/src/lib.rs
```diff
@@ -24,6 +24,11 @@ pub enum BackendType {
     PilStarkCli,
 }
 
+pub type BackendOptions = String;
```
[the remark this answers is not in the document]
leonardoalt: Why enum? Shouldn't it be a struct if anything other than a string? I had a struct at first, but the thing is that each backend has different options. I guess here an enum of different backend options could make sense, but I didn't want to commit to a particular fixed set of options for each backend, the idea being that the backend interface just receives a generic string, and each backend needs to know how to decode it. Otherwise the specific backend information leaks into the user side (eg main or pipeline)

pipeline/src/pipeline.rs
```diff
@@ -769,6 +772,7 @@ impl<T: FieldElement> Pipeline<T> {
 
         self.log("Optimizing pil...");
         let optimized = powdr_pilopt::optimize(analyzed_pil);
+        //let optimized = analyzed_pil.clone();
```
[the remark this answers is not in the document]
leonardoalt: yea debug leftover

halo2/src/prover.rs
leonardoalt: yep, that was on my TODO list. Just wanted to separate the interfaces first and then clean up the functions

halo2/src/prover.rs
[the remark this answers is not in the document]
leonardoalt: This was changed now to downsize only if the ProofType is Poseidon or SingleSnark

backend/src/halo2_impl.rs
```diff
@@ -16,8 +16,19 @@ impl<F: FieldElement> BackendFactory<F> for Halo2ProverFactory {
...
+        let proof_type = match options.as_str() {
+            "" | "poseidon" => ProofType::Poseidon,
+            "snark_single" => ProofType::SnarkSingle,
+            "snark_aggr" => ProofType::SnarkAggr,
+            _ => {
+                return Err(Error::BackendError(format!(
+                    "Invalid proof type: {options}"
+                )))
+            }
+        };
```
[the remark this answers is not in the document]
leonardoalt: Implemented this for Halo2 as well now. One annoying thing is that since Halo2 is a separate crate we don't have access to the BackendOptions type. Used String for now. I think we should move Halo2 from a crate to a module inside the `backend` crate.

leonardoalt:
This is ready for review. There are some improvements we can write in the backend crate but I'd rather get this merged first and do that after. I opened issues for the things we should do:
- https://github.com/powdr-labs/powdr/issues/1322
- https://github.com/powdr-labs/powdr/issues/1323
-

leonardoalt:
The one thing that's missing is how to get the accumulator publics that need to be passed to the aggregated proof's verifier. These come from the initial proof, so we could expose a way to retrieve these from a proof @georgwiese

backend/src/estark/starky_wrapper.rs
```diff
@@ -39,7 +41,18 @@ impl<F: FieldElement> BackendFactory<F> for Factory {
...
+        if verification_app_key.is_some() {
+            return Err(Error::NoAggregationAvailable);
+        }
+
+        let proof_type: ProofType = ProofType::from(options);
+        let hash_type = match proof_type {
+            ProofType::StarkGL => "GL",
+            ProofType::StarkBN => "BN",
+            ProofType::SnarkBN => "BN",
+        };
```
Schaeff, line 53: this is also in `estark/mod.rs` so could be better in a `ProofType` impl

book/src/hello_world_ethereum.md
```diff
@@ -0,0 +1,41 @@
...
+```
+{{#include ../../test_data/asm/book/hello_world.asm}}
+```
+
+```console
+powdr pil test_data/asm/book/hello_world.asm --field bn254 --inputs 0 --prove-with halo2 --backend-options "snark_single" -f --params params.bin
+```
+
+The extra parameter `backend-options` tells powdr to produce a single
+SNARK that uses Keccak. The option `-f` forces overwriting the compiled files
```
Schaeff, line 18: What do you mean by "that uses Keccak"?

leonardoalt: Keccak transcript as opposed to Poseidon transcript

book/src/hello_world_ethereum_aggregation.md
```diff
@@ -0,0 +1,88 @@
...
+that we've just generated and the recursion one we're going to generate
+later, let's rename them:
+```console
+mv vkey.bin vkey_app.bin
+mv hello_world_proof.bin hello_world_proof_app.bin
+```
+
+We can now generate a verification key for the Halo2 circuit that verifies our proofs recursively. This might take up to a minute.
+```console
+powdr verification-key test_data/asm/book/hello_world.asm --field bn254 --backend halo2 --backend-options "snark_aggr" --params params.bin --vkey-app vkey_app.bin
```
Schaeff, line 60: I tried running everything here and failed at this step. gist to reproduce [here](https://gist.github.com/Schaeff/aa7879e432eae31ac59f1001b561bb0e)
The first step was suspiciously quick.

Schaeff: Could it be that the setup size should be 2**22 instead of 22 here?

leonardoalt: Ah yes that's correct I think

book/src/hello_world_ethereum_aggregation.md
```diff
@@ -0,0 +1,88 @@
...
+
+First, we need a setup of "size" 22. This is the log2 of the maximum execution
+trace length of the recursion circuit.
+
+You can generate it using the command line below, or download a pre-computed
+one [here](https://drive.google.com/file/d/1xG_O_KqooiRR3QJno3tXIiNELSB2AQMH/view?usp=drive_link).
+
+This will take a couple minutes if you decide to compute it yourself:
+```console
+powdr setup 22 --backend halo2 --field bn254
```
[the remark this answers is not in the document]
Schaeff: Also "a couple of minutes" is an understatement on my machine (15 min)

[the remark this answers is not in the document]
leonardoalt: > Also "a couple of minutes" is an understatement on my machine (15 min)

skill issue

halo2/src/prover.rs
```diff
@@ -134,60 +194,119 @@ impl<'a, F: FieldElement> Halo2Prover<'a, F> {
...
+    /// One or more of these proofs can be aggregated by `prove_snark_aggr`.
+    pub fn prove_poseidon(
+        &self,
+        witness: &[(String, Vec<F>)],
+        witgen_callback: WitgenCallback<F>,
+    ) -> Result<Vec<u8>, String> {
+        assert!(matches!(self.proof_type, ProofType::Poseidon));
+
+        let proof = self.prove::<_, aggregation::PoseidonTranscript<NativeLoader, _>,  aggregation::PoseidonTranscript<NativeLoader, _>>(witness, witgen_callback)?;
+        Ok(proof.0)
```
georgwiese, line 210:
```suggestion
        let (proof, _) = self.prove::<_, aggregation::PoseidonTranscript<NativeLoader, _>,  aggregation::PoseidonTranscript<NativeLoader, _>>(witness, witgen_callback)?;
        Ok(proof)
```

halo2/src/prover.rs
```diff
@@ -134,60 +194,119 @@ impl<'a, F: FieldElement> Halo2Prover<'a, F> {
...
+    /// Generate a single proof for a given PIL using Keccak transcripts.
+    /// These proofs can be verified directly on Ethereum.
+    pub fn prove_snark_single(
         &self,
         witness: &[(String, Vec<F>)],
+        witgen_callback: WitgenCallback<F>,
+    ) -> Result<Vec<u8>, String> {
+        assert!(matches!(self.proof_type, ProofType::SnarkSingle));
+
+        let proof = self.prove::<_, EvmTranscript<_, _, _, _>, EvmTranscript<G1Affine, _, _, _>>(
```
georgwiese, line 222:
```suggestion
        let (proof, publics) = self.prove::<_, EvmTranscript<_, _, _, _>, EvmTranscript<G1Affine, _, _, _>>(
```

halo2/src/prover.rs
```diff
@@ -134,60 +194,119 @@ impl<'a, F: FieldElement> Halo2Prover<'a, F> {
...
 
-        let circuit_app = PowdrCircuit::new(self.analyzed, self.fixed).with_witness(witness);
-        let publics = vec![circuit_app.instance_column()];
+        let circuit_app = PowdrCircuit::new(self.analyzed, self.fixed)
+            .with_witgen_callback(witgen_callback)
+            .with_witness(witness);
 
-        assert_eq!(publics.len(), 1);
-        if !publics[0].is_empty() {
+        // TODO Support publics in the app snark
```
georgwiese, line 264:
```suggestion
```

I think I added it, but it's kind of redundant with the `unimplemented!` call just below ;)

halo2/src/prover.rs
```diff
@@ -205,28 +324,93 @@ impl<'a, F: FieldElement> Halo2Prover<'a, F> {
         }
 
         log::info!("Verifying aggregated proof in the EVM...");
-        aggregation::evm_verify(deployment_code, agg_circuit_with_proof.instances(), &proof);
+        evm_verify(deployment_code, agg_instances.clone(), &proof);
+
+        let output_dir = self.output_dir.ok_or("output_dir is None".to_owned())?;
```
georgwiese, line 329: I think we can handle `output_dir` of `None`?

halo2/src/prover.rs
```diff
@@ -205,28 +324,93 @@ impl<'a, F: FieldElement> Halo2Prover<'a, F> {
...
+            ProofType::SnarkAggr => self.verification_key_aggr(),
+        }
+    }
+
+    fn verification_key_single(&self) -> Result<VerifyingKey<G1Affine>, String> {
         let circuit = PowdrCircuit::new(self.analyzed, self.fixed);
         keygen_vk(&self.params, &circuit).map_err(|e| e.to_string())
     }
 
+    fn verification_key_aggr(&self) -> Result<VerifyingKey<G1Affine>, String> {
```
georgwiese, line 390:
```suggestion
    fn generate_verification_key_single(&self) -> Result<VerifyingKey<G1Affine>, String> {
        let circuit = PowdrCircuit::new(self.analyzed, self.fixed);
        keygen_vk(&self.params, &circuit).map_err(|e| e.to_string())
    }

    fn generate_verification_key_aggr(&self) -> Result<VerifyingKey<G1Affine>, String> {
```

halo2/src/prover.rs
```diff
@@ -307,7 +544,7 @@ fn gen_proof<
         .map(|instances| instances.as_slice())
         .collect_vec();
     let proof = {
-        let mut transcript = TW::init(Vec::new());
+        let mut transcript = TranscriptWriterBuffer::<_, G1Affine, _>::init(Vec::new());
```
georgwiese, line 547:
```suggestion
        let mut transcript = TW::init(Vec::new());
```

I don't think this change is necessary?

backend/src/estark/mod.rs
```diff
@@ -9,15 +9,32 @@ use std::{
...
+use crate::{Backend, BackendFactory, BackendOptions, Error, Proof};
 use powdr_ast::analyzed::Analyzed;
 
 use powdr_executor::witgen::WitgenCallback;
 use powdr_number::{buffered_write_file, write_polys_file, DegreeType, FieldElement};
 use serde::Serialize;
 use starky::types::{StarkStruct, Step, PIL};
 
-fn create_stark_struct(degree: DegreeType) -> StarkStruct {
+enum ProofType {
```
georgwiese, line 20: I don't understand these proof types yet!

backend/src/estark/mod.rs
```diff
@@ -158,8 +185,16 @@ impl<'a, F: FieldElement> EStarkFilesCommon<'a, F> {
         }
 
         // Write the stark struct JSON.
+        let hash_type = match self.proof_type {
```
georgwiese, line 188: I don't understand the hash type. GL and BN are not hash functions? Are these different variants of Poseidon?

backend/src/lib.rs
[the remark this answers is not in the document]
georgwiese: I think in both cases, right now, this is essentially specifying an enum variant of a proof type enum (but different enums). I don't understand the variants in the eStark case yet, but I guess my original comment was because I thought the proof type is not really specific to the backend.

georgwiese (commented):
Nice! Looks good in general, just a few questions here and there. The tutorial in the docs didn't work for me yet, but maybe I did something wrong on the way.

book/src/hello_world_ethereum.md
```diff
@@ -0,0 +1,41 @@
...
+generate proofs directly using a Keccak transcript instead of the Poseidon
+transcript of the previous example, which will enable us to verify proofs
+onchain. There are almost no differences from the CLI perspective.
+
+```
+{{#include ../../test_data/asm/book/hello_world.asm}}
+```
+
+```console
+powdr pil test_data/asm/book/hello_world.asm --field bn254 --inputs 0 --prove-with halo2 --backend-options "snark_single" -f --params params.bin
```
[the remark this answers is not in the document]
leonardoalt: Right, ill add instructions. Did you manage to install it and test after?

book/src/hello_world_ethereum_aggregation.md
```diff
@@ -0,0 +1,88 @@
+# Hello World on Ethereum via proof aggregation
+
+This example is yet another variation of the previous [Hello World on Ethereum](./hello_world_ethereum.md)
+example, still targeting verification on Ethereum but supporting more complex
+programs. As noted in the previous section, complex programs can lead to large
+Solidity verifiers that exceed the contract size limit on Ethereum. One
```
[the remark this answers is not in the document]
leonardoalt: I think you're in the other comment, it's rather large VMs aka number of columns

book/src/hello_world_ethereum_aggregation.md
```diff
@@ -0,0 +1,88 @@
...
+We can now generate a verification key for the Halo2 circuit that verifies our proofs recursively. This might take up to a minute.
+```console
+powdr verification-key test_data/asm/book/hello_world.asm --field bn254 --backend halo2 --backend-options "snark_aggr" --params params.bin --vkey-app vkey_app.bin
+```
+
+> Note that this verification key can only be used to verify recursive proofs that verify other proofs using the application's key `vkey_app.bin`.
+
+We can now generate the recursive proof (can take 3 or more minutes and use 28gb RAM):
+```console
+powdr prove test_data/asm/book/hello_world.asm --field bn254 --backend halo2 --backend-options "snark_aggr" --params params.bin --vkey vkey.bin --vkey-app vkey_app.bin --proof hello_world_proof_app.bin
```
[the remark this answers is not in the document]
leonardoalt: Hmmm let's investigate in a sync way. It also takes a while deriving the PK from the VK

halo2/src/prover.rs
```diff
@@ -48,51 +84,85 @@ pub struct Halo2Prover<'a, F> {
...
+    vkey_app: Option<VerifyingKey<G1Affine>>,
+    proof_type: ProofType,
+    output_dir: Option<&'a Path>,
 }
 
 fn degree_bits(degree: DegreeType) -> u32 {
     DegreeType::BITS - degree.leading_zeros() + 1
 }
 
+// TODO move to `backend` and sync with the similar function in estark
```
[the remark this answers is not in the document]
leonardoalt: I'd rather do it after we move halo2 into the backend crate

halo2/src/prover.rs
leonardoalt: Yea exactly that's what I meant here

halo2/src/prover.rs
leonardoalt: What do we use then?

halo2/src/prover.rs
leonardoalt: Will check

backend/src/estark/mod.rs
leonardoalt: It's similar to the halo2 case, but specific to what polygon/starky estark do to get a proof all the way to being verified on Ethereum:

- first proof is a GL STARK
- recursive proof that can take 1+ GL STARKS and verifies them in another GL STARK, so same as the previous one but has proofs as inputs
- recursive proof that can take a GL STARK and give a BN STARK
- recursive fflonk proof that takes a BN STARK and gives a BN SNARK

backend/src/estark/mod.rs
leonardoalt: Hash_type is just the name of the field in the StarkInfo struct, and these are the values it accepts

book/src/hello_world_ethereum.md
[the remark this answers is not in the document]
georgwiese: yes!

book/src/hello_world_ethereum_aggregation.md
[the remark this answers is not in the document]
georgwiese: Working now somehow, not sure what went wrong yesterday.

halo2/src/prover.rs
georgwiese: We don't write the publics. That's the only reason it's used, right?

Actually, I think the prover should not know about the output dir at all, and instead return the publics together with the proof. And then it is up to the pipeline to persist this or not.

backend/src/estark/mod.rs
georgwiese: Hmm, IDK, seems like very hard-coded way to encode this. What if I want to use a Halo2 proof instead of fflonk in the last step? It's also a bit redundant with the `--field` argument.

To do what you're describing I'd expect something like this:
```bash
# Initial proof
powdr prove ... --field gl --backend estark
# Compress a few times
powdr compress ... --field gl --backend estark
powdr compress ... --field gl --backend estark
powdr compress ... --field gl --backend estark
# Final STARK compression
powdr compress ... --field bn254 --backend estark
# Final SNARK
powdr compress ... --field bn254 --backend fflonk
```

But fine for this PR, we can figure out along the way what level of flexibility we really need!

leonardoalt: > What if I want to use a Halo2 proof instead of fflonk in the last step?

If you started with STARKS, you can't.

>  It's also a bit redundant with the --field argument.

This is CLI implementation detail. The field is important for the circuit, not the recursion part of the proof system.

> To do what you're describing I'd expect something like this:

Idk, I think I disagree here. I'm not sure `compress` should be a special command. It is also just a proof, but with different inputs.

Yea agree to do this later!

halo2/src/prover.rs
georgwiese: To me it would be a cleaner separation if the "backend" is only responsible for computing proofs and publics and returns both. You should be able to compute proofs without ever touching the disk. In the case of the Polygon eStark backend it might have to persist files, because it needs to invoke the prover in a different process and it expects file on disk. But that's an implementation detail of this specific backend.

PR #1261, opened 2024-04-05
Refactor ast

chriseth:


chriseth:
The goal here is to change all recursive structures to have the expression as a template argument. Once we do that, we can much more easily switch out expression under the hood, and for example, add a source reference to the expression or remove enum variants.

leonardoalt (approved):

Issue #1251, opened 2024-04-04
ASM: Move latch and operation_id from machine declaration


leonardoalt:
Currently a block machine is written like this:

```rust
machine Binary(latch, operation_id) {
    ...
    col witness operation_id;
    col fixed latch(i) { if (i % 4) == 3 { 1 } else { 0 } };
    ...
}
```

The idea is to remove them from the machine declaration and be an attribute, for example:

```rust
machine Binary {
    ...
    #[operation_id]
    col witness operation_id;
    
    #[latch]
    col fixed latch(i) { if (i % 4) == 3 { 1 } else { 0 } };

    ...
}
```

The reason for that is that we want to be able to pass shared machines to submachines, which would fit better a syntax that occupies that space:

```rust
machine Binary(mem: Memory) {
    ...
}
```

chriseth:
I think adding a annotation mechanism could complicate the whole thing, and while the header might not be an ideal place to put the latch and op id, I could imagine something like the following, just because it is much easier to implement right now:
```
machine Binary(arg1: int, arg2: MachineSomething) with latch: latch, operation_id: operation_id {
  ...
}
```

Schaeff:
What if instead of an annotation we could use a special stdlib/corelib function?

```
col witness operation_id;
std::link::operation_id(operation_id);
```

And later deriving that with `#[derive(operation_id)]`

chriseth:
Can this be closed, @pacheco ?

PR #1267, opened 2024-04-10
Machine properties using `with` syntax

pacheco:
This implements issue #1251.
Basically `machine Foo(a,b) { ... }` is now `machine Foo with latch: a, operation_id: b { ... }`


pacheco:
Which of the following options for declaring these machine properties seems nicer?
a) `latch: some_column`
b) `some_column: latch`
c) `latch = some_column`

Also, do we allow using simply `latch` as a shortcut for `latch: latch`?

chriseth:
Ah nice! I think it would make sense shorten `latch: latch` to just `latch`.

ast/src/asm_analysis/display.rs
```diff
@@ -71,16 +71,36 @@ impl Display for AnalysisASMFile {
...
+            .as_ref()
+            .map(|s| format!("degree: {s}"))
+            .into_iter()
+            .chain(self.latch.as_ref().map(|s| format!("latch: {s}")))
+            .chain(
+                self.operation_id
+                    .as_ref()
+                    .map(|s| format!("operation_id: {s}")),
+            )
+            // .chain(self.call_selectors
```
chriseth, line 85: What about these?

analysis/src/machine_check.rs
```diff
@@ -182,8 +166,10 @@ impl TypeChecker {
             }
         }
 
-        let latch = machine.arguments.latch;
-        let operation_id = machine.arguments.operation_id;
+        let degree = machine.properties.degree;
```
chriseth, line 169: Does it make sense to use `let MachineProperties { degree, latch, ...} = machine.properties;`?

ast/src/parsed/asm.rs
```diff
@@ -399,23 +400,111 @@ impl Machine {
...
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Default, Clone)]
+pub struct MachineProperties {
+    pub degree: Option<Expression>,
     pub latch: Option<String>,
     pub operation_id: Option<String>,
+    pub call_selectors: Option<String>,
+}
+
+impl MachineProperties {
+    pub fn names(&self) -> impl Iterator<Item = &String> {
```
chriseth, line 447: Maybe rename to `defined_names` or something?

ast/src/parsed/asm.rs
```diff
@@ -399,23 +400,111 @@ impl Machine {
...
+                    if props.latch.is_some() {
+                        return Err("`latch` already defined");
+                    };
+                    if let Expression::Reference(r) = value {
+                        if let Some(id) = r.try_to_identifier() {
+                            props.latch = Some(id.clone());
+                            continue;
+                        }
+                    };
+                    return Err("`latch` machine property expects a column name");
```
chriseth, line 475:
```suggestion
                    return Err("`latch` machine property expects a local column name");
```

ast/src/parsed/asm.rs
```diff
@@ -399,23 +400,111 @@ impl Machine {
...
+        let mut props: Self = Default::default();
+        for (name, value) in prop_list {
+            match name.as_str() {
+                "degree" => {
+                    if props.degree.is_some() {
+                        return Err("`degree` already defined");
+                    };
+                    props.degree = Some(value);
+                }
+                "latch" => {
```
chriseth, line 465: Can these three cases be combined?

ast/src/parsed/display.rs
```diff
@@ -36,23 +36,35 @@ impl Display for ModuleStatement {
...
+                                .as_ref()
+                                .map(|s| format!("call_selectors: {s}")),
                         )
+                        .join(", ");
+                    match (args.is_empty(), props.is_empty()) {
+                        (true, true) => write!(f, "machine {name} {m}"),
+                        (true, false) => write!(f, "machine {name} with {props} {m}"),
+                        (false, true) => write!(f, "machine {name}({args}) {m}"),
+                        (false, false) => write!(f, "machine {name}({args}) with {props} {m}"),
                     }
```
chriseth, line 68: Matter of style, but I would prefer
```suggestion
                    let props = if props.is_empty() { props } else { format!(" with {props}")) };
                    let args = if args.is_empty() { args } else { format!("({args})")) };
                    write!(f, "machine {name}{args}{props} {m}")
```

parser/src/powdr.lalrpop
```diff
@@ -255,20 +255,31 @@ GenericTypedName: (String, Option<TypeScheme<Expression>>) = {
 // ---------------------------- ASM part -----------------------------
 
 MachineDefinition: SymbolDefinition = {
-    "machine" <name:Identifier> <arguments:MachineArguments> "{" <statements:(MachineStatement)*> "}" => SymbolDefinition { name, value: Machine { arguments, statements}.into() }
+    "machine" <name:Identifier> <arguments:MachineArguments> "{" <statements:(MachineStatement)*> "}" => SymbolDefinition { name, value: Machine { arguments, properties: Default::default(), statements}.into() },
+    "machine" <name:Identifier> <arguments:MachineArguments> "with" <properties:MachineProperties> "{" <statements:(MachineStatement)*> "}" => SymbolDefinition { name, value: Machine { arguments, properties, statements}.into() },
```
chriseth, line 259:
```suggestion
    "machine" <name:Identifier> <arguments:MachineArguments> <properties:("with" <MachineProperties>)?> "{" <statements:(MachineStatement)*> "}" => SymbolDefinition { name, value: Machine { arguments, properties: properties.unwrap_or_default(), statements}.into() },
```

ast/src/parsed/asm.rs
```diff
@@ -399,23 +400,111 @@ impl Machine {
...
+impl MachineArguments {
+    pub fn names(&self) -> impl Iterator<Item = &String> {
+        self.0.iter().map(|p| &p.name)
+    }
+}
+
+impl TryFrom<Vec<Param>> for MachineArguments {
+    type Error = &'static str;
+
+    fn try_from(params: Vec<Param>) -> std::result::Result<Self, Self::Error> {
```
chriseth, line 428: What about duplicate names? Where is that caught?

pacheco: didn't really do anything about arguments besides parsing, will be handled in a separate PR

ast/src/asm_analysis/display.rs
pacheco: ops!

analysis/src/machine_check.rs
pacheco: mmm that gave me a slightly different idea: just embed the `MachineProperties` object directly

pacheco: or maybe not, will go with your suggestion

pacheco: ill keep your suggestion for this PR and try what I said in a separate PR

ast/src/parsed/asm.rs
pacheco: the issue is the Error needs to be a static string for bubbling up with lalrpop parsing AFAIR...
I thought about validating this in `machine_check` instead of parsing time, for more flexibility, will note this and try it in a follow up PR

parser/src/powdr.lalrpop
pacheco: good catch!

chriseth (approved):

PR #1268, opened 2024-04-12
Fix missing feature in starky

eigmax:
The field missed the feature `AVX512`, the fix in starky is https://github.com/0xEigenLabs/eigen-zkvm/commit/737b1a062fc33b4ebbbffa9f74f654e58d76740b

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,3 +45,7 @@ overflow-checks = true
 panic = 'unwind'
 incremental = true         # This is true because target is cached
 codegen-units = 256
+
+
+[features]
+avx512 = []
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -21,7 +21,7 @@ strum = { version = "0.24.1", features = ["derive"] }
 log = "0.4.17"
 serde_json = "1.0"
 thiserror = "1.0.43"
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -45,3 +45,7 @@ overflow-checks = true
 panic = 'unwind'
 incremental = true         # This is true because target is cached
 codegen-units = 256
+
+
+[features]
+avx512 = []
```
leonardoalt, line 51: shouldn't this rather be a feature of the `backend` and `powdr` crates instead of the whole repo?

backend/Cargo.toml
```diff
@@ -9,6 +9,7 @@ repository = { workspace = true }
 
 [features]
 halo2 = ["dep:powdr-halo2"]
+avx512 = ["starky/avx512"]
```
leonardoalt, line 12:
```suggestion
starky-avx512 = ["starky/avx512"]
```

I think this name would be better because it signals that this is important for starky, and it's not powdr that's using avx itself

Cargo.toml
eigmax: Yes, you are right. But I am wondering if we can support the avx512 here, cause powdr uses the stable channel, but the feature `stdarch_x86_avx512` is not available there.

PR #1269, opened 2024-04-12
Fix missing feature in starky

eigmax:
- [x] support AVX512 acceleration for starky.  https://github.com/0xEigenLabs/eigen-zkvm/commit/6f52173775bc3f5f8046d56411eee4cd36e81686 .  The feature `stdarch_x86_avx512` is not available in Rust's stable channel.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.77"
+channel = "nightly"
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -9,6 +9,7 @@ repository = { workspace = true }
 
 [features]
 halo2 = ["dep:powdr-halo2"]
+starky-avx512 = ["starky/avx512"]
 
 [dependencies]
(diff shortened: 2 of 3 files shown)


rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.77"
+channel = "nightly"
```
Schaeff, line 2: Is this change required?

eigmax: Yes, this is required by AVX512.  And that's why I try to merge it into powdr-labs:avx

eigmax: There is another options:  creating a new rust-toolchain in backend,  I am not sure which one is better.

leonardoalt (approved):

leonardoalt:
@eigmax the tests are still failing due to `feature` being used with a stable channel. For the purposes of this branch we can also make the whole repo `nightly` so you can run some experiments.

leonardoalt:
@eigmax ok I'll just merge this PR then regardless of CI

Issue #962, opened 2024-01-26
Expression Display: Reduce the number of parentheses


chriseth:
The Display trait for Expression adds parentheses in order to ensure correct parsing order. These are not always necessary. The Display traits for TypeName and Type have an example how parentheses can be avoided: They have to be added by the display method of the outer element which needs to ask the inner element if parentheses are needed to parse it ambiguously. Due to operator precedence, this could be a bit more tricky, but we could supply the current precedence in the `needs_parentheses` call. In any case, even removing some parentheses would help already.

Here is a description of the general procedure: https://stackoverflow.com/questions/6277747/pretty-print-expression-with-as-few-parentheses-as-possible

aminlatifi:
@chriseth @leonardoalt It would be my pleasure to contribute to this project,
As I look into `ast/src/parsed/display.rs` file, particularly to `format_type_with_parentheses` function, the parenthesis generation can be skipped simply by not calling this function and just passing the value itself.

(https://github.com/powdr-labs/powdr/blob/3f771638467a8bd527805239ac614dd56e65f11c/ast/src/parsed/display.rs#L737-L743)

Actually, the `needs_parentheses` function is only called in `format_type_with_parentheses` method, and there is no complex process. 
If you agree I can decommission `format_type_with_parentheses` in places where no parenthesis is needed (like a list of types? or enum variant?!)

Please let me know if I am mistaken about anything. Furthermore, can you guide me to produce examples in which extra parentheses are generated? I am new to powdr.

aminlatifi:
I made #1270. I tried to be in accord with @chriseth guidance.
I appreciate it if one of you could do the review.

leonardoalt:
Was this implemented in https://github.com/powdr-labs/powdr/pull/1289 ?

chriseth:
Pretty much, yes. I think there is still something left at the higher level, but I think we can close this.

PR #1270, opened 2024-04-14
Reduce Number of Expression Parentheses

aminlatifi:
Adds binary operation precedence support to avoid unnecessary parentheses in expression printed format

- #962


ast/src/parsed/display.rs
```diff
@@ -539,6 +539,52 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
     format!("{}", list.into_iter().format(", "))
 }
 
+impl<E: Display> Expression<E> {
+    pub fn binary_op_precedence(&self) -> ExpressionPrecedence {
```
chriseth, line 543: I think it might be better to add this to the impl for BinaryOperator - and you can put it directly in the mod.rs, it's not just related to Display.

ast/src/parsed/display.rs
```diff
@@ -539,6 +539,52 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+                // &&
+                LogicalAnd => 10,
+                // ||
+                LogicalOr => 11,
+                // .. ..=
+                // ??
+                // = += -= *= /= %= &= |= ^= <<= >>=
+                Identity => 12,
+            },
+            _ => 0,
```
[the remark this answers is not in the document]
aminlatifi: Always add or not add? Here I don't add parentheses normally.

ast/src/parsed/display.rs
```diff
@@ -549,7 +595,13 @@ impl<Ref: Display> Display for Expression<Ref> {
             Expression::Tuple(items) => write!(f, "({})", format_list(items)),
             Expression::LambdaExpression(lambda) => write!(f, "{}", lambda),
             Expression::ArrayLiteral(array) => write!(f, "{array}"),
-            Expression::BinaryOperation(left, op, right) => write!(f, "({left} {op} {right})"),
+            Expression::BinaryOperation(left, op, right) => write!(
+                f,
+                "{} {} {}",
+                format_expression_with_precedence(left, self.binary_op_precedence()),
```
chriseth, line 601: Hm, we could make this a member function of Expression and call it `format_with_precedence`, then this monster would get a bit shorter.

ast/src/parsed/display.rs
```diff
@@ -864,4 +916,124 @@ mod tests {
         assert_eq!(p.clone().join(s.clone()).to_string(), "::abc::y");
         assert_eq!(SymbolPath::from(p.join(s)).to_string(), "::abc::y");
     }
+
+    mod binary_op_parentheses {
```
chriseth, line 920: I think the tests are much better readable if the AST is parsed instead of constructed.
You can take a look at parser/lib/src.rs. Maybe it makes sense to create a new file in the test directory inside the parser crate, though.
Then the tests would look like this:
```
#[test]
fn add_mul_paren() {
        let input = "let t = ((x + y) * z);"
        let printed = format!("{}", parse(None, input).unwrap_err_to_stderr());
        assert_eq!("let t = (x + y) * z;", printed);
}
```

chriseth (changes requested):
This looks great already! I mainly have some naming and testing suggestions.

ast/src/parsed/display.rs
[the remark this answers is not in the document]
chriseth: I mean if you have two expressions that both have precedence zero, then this code does not add parentheses.

ast/src/parsed/display.rs
```diff
@@ -549,7 +595,13 @@ impl<Ref: Display> Display for Expression<Ref> {
             Expression::Tuple(items) => write!(f, "({})", format_list(items)),
             Expression::LambdaExpression(lambda) => write!(f, "{}", lambda),
             Expression::ArrayLiteral(array) => write!(f, "{array}"),
-            Expression::BinaryOperation(left, op, right) => write!(f, "({left} {op} {right})"),
+            Expression::BinaryOperation(left, op, right) => write!(
```
chriseth, line 598: Before this change, this always had parentheses. Now the outer structure never has parentheses. Shouldn't we rather always add parentheses unless we know the precedence of the outer structure?

ast/src/parsed/display.rs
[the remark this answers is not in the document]
chriseth: I think it's better to add parentheses unless we are 100% sure they are not needed.

ast/src/parsed/display.rs
aminlatifi: I added tests there but kept this one as a reference. We can add further tests in parser/tests/parentheses_test.rs

ast/src/parsed/mod.rs
```diff
@@ -589,6 +591,38 @@ pub enum BinaryOperator {
...
+            // == != < > <= >=
+            Equal | NotEqual | Less | Greater | LessEqual | GreaterEqual => 9,
+            // &&
+            LogicalAnd => 10,
+            // ||
+            LogicalOr => 11,
+            // .. ..=
+            // ??
+            // = += -= *= /= %= &= |= ^= <<= >>=
+            Identity => 12,
```
chriseth, line 621: Should have 9 as well.

aminlatifi (commented):
I did rebase to main branch, but it has made unexpected changes!
BTW, @chriseth the associativity issue is fixed.

aminlatifi:
Closed in favor of https://github.com/powdr-labs/powdr/pull/1289

PR #1271, opened 2024-04-15
remove unaligned memory check in double sorted memory machine in witgen

leonardoalt:
This is only needed for RISCV, but witgen is more general than RISCV and should allow memory accesses anywhere, especially since we have the std memory machine.

The RISCV executor also checks this so it won't go unchecked.

### DIFF
--- a/executor/src/witgen/machines/double_sorted_witness_machine.rs
+++ b/executor/src/witgen/machines/double_sorted_witness_machine.rs
@@ -2,7 +2,6 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use std::iter::once;
 
 use itertools::Itertools;
-use num_traits::Zero;
 
 use super::{FixedLookup, Machine};
 use crate::witgen::affine_expression::AffineExpression;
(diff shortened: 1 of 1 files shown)


lvella:
This doesn't look like a RISCV issue to me. We implement our memory machine in words of 32 bits, so it makes sense it only supports accesses aligned to 4 bytes. I find it a reasonable limitation that must be dealt with at zkVM level, but if we want to support unaligned access, we would need to implement an adapter machine, like Polygon Hermez did.

leonardoalt:
No, the RISCV implementation implements it as words of 32-bits, the std machine itself implements memory as field elements, so this assertion is over constraining use cases that don't require 4-byte alignment.

lvella:
Ah, ok. Yes, we can see the memory as an arbitrary map that the pil uses however it likes. But then this constraint should be probably moved into RISCV implementation.

leonardoalt:
But that's what I mean, the riscv-executor already has this same assertion, so I think that's fine

leonardoalt:
Yes it's enforced at the assembly instruction level

PR #1275, opened 2024-04-16
add estark-polygon as features of other crates

leonardoalt:


lvella:
Lets have an approval on #1211 before rebasing and merging this one.

leonardoalt:
I was going to approve it earlier but I didn't because it was missing these lol

lvella:
I'll cherry pick this commit after I squash the other. I must squash the other, right?

leonardoalt:
Don't have to, since we have auto-squash now at merge time. So also no need to cherry pick this commit, can also merge this PR as is and then squash the other one when ready to merge

PR #1278, opened 2024-04-17
Running format before lint and using pr-tests for example.

lvella:
Format runs faster, so it is better if it fails immediately.

The example run is not fine because it is not using the cache nor results from the build job.


chriseth:
I'm actually wondering if we shouldn't move `fmt` and `clippy` to their own test runs so that we still get the test failures. But I'm also fine with this.

chriseth (dismissed):

lvella:
I am not sure I am fine with this.

The version approved increased the time of test_quick job, due to building with the optimized profile.

The current version increases the build job time by 50 seconds, which is worse than the 2 min it saves in the test_quick job, as the latter run in parallel with slower jobs.

chriseth:
We could also just have a separate job for the examples.

chriseth:
We could have one job that runs clippy, fmt, examples and another job that just compiles

leonardoalt:
what's happening with this PR?

lvella:
It seems I was trying to make the PR tests fail early on format errors, but I managed to make them slower.

leonardoalt:
@lvella so what do we do with this PR?

PR #1289, opened 2024-04-19
Reduce Number of Expression Parentheses

aminlatifi:
Adds binary operation precedence support to avoid unnecessary parentheses in expression printed format

- #962


ast/src/parsed/mod.rs
```diff
@@ -633,6 +635,51 @@ pub enum BinaryOperator {
...
+            // == != < > <= >=
+            Equal | NotEqual | Less | Greater | LessEqual | GreaterEqual => 9,
+            // &&
+            LogicalAnd => 10,
+            // ||
+            LogicalOr => 11,
+            // .. ..=
+            // ??
+            // = += -= *= /= %= &= |= ^= <<= >>=
+            Identity => 12,
```
chriseth, line 671: This should be 9

ast/src/parsed/mod.rs
```diff
@@ -633,6 +635,51 @@ pub enum BinaryOperator {
...
+        }
+    }
+
+    pub fn associativity(&self) -> BinaryOperatorAssociativity {
+        match self {
+            BinaryOperator::Pow => BinaryOperatorAssociativity::Right,
```
chriseth, line 677: I'm not sure if this is correct. I'm also not sure if your grammar is correct. Can you check that `x**y**2` parses the same way as `x**(y**2)`?

chriseth: If not we have to change the grammar.

ast/src/parsed/display.rs
```diff
@@ -892,4 +939,124 @@ mod tests {
         assert_eq!(p.clone().join(s.clone()).to_string(), "::abc::y");
         assert_eq!(SymbolPath::from(p.join(s)).to_string(), "::abc::y");
     }
+
+    mod binary_op_parentheses {
```
chriseth, line 943: Can you turn all these tests into the other form please and then remove this? These tests are very verbose and we need to modify all of them if we change the structs.

ast/src/parsed/display.rs
```diff
@@ -574,7 +621,7 @@ impl<Ref: Display> Display for Expression<Ref> {
             Expression::Tuple(items) => write!(f, "({})", format_list(items)),
             Expression::LambdaExpression(lambda) => write!(f, "{}", lambda),
             Expression::ArrayLiteral(array) => write!(f, "{array}"),
-            Expression::BinaryOperation(left, op, right) => write!(f, "({left} {op} {right})"),
+            Expression::BinaryOperation(_, _, _) => self.format_binary_operation(f),
```
chriseth, line 624: I think we need to put parentheses here, because we do not know the outer precedence. You can try testing `-(x + y)`.

ast/src/parsed/display.rs
```diff
@@ -564,6 +564,53 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+        match self {
+            Expression::BinaryOperation(_, op, _) => Some(op.precedence()),
+            _ => None,
+        }
+    }
+
+    pub fn format_binary_operation(&self, f: &mut Formatter<'_>) -> Result {
+        match self {
+            Expression::BinaryOperation(left, op, right) => {
+                let use_left_parenthes = match left.precedence() {
```
chriseth, line 578:
```suggestion
                let use_left_parentheses = match left.precedence() {
```

ast/src/parsed/display.rs
```diff
@@ -564,6 +564,53 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+                    Some(left_precedence) => {
+                        left_precedence > op.precedence()
+                            || (left_precedence == op.precedence()
+                                && op.associativity() == BinaryOperatorAssociativity::Right)
+                    }
+                    None => false,
```
chriseth, line 584: Shouldn't this be `true`? We don't know the precedence of the inner, so we should better put parentheses.

Maybe you are saying that it is not a binary operation, so we don't have to put parentheses, but then we should cross-check the Display function for Expression that it never puts parentheses for anything that returns None here.

ast/src/parsed/display.rs
```diff
@@ -564,6 +564,53 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
     format!("{}", list.into_iter().format(", "))
 }
 
+impl<E: Display> Expression<E> {
+    pub fn precedence(&self) -> Option<ExpressionPrecedence> {
+        match self {
+            Expression::BinaryOperation(_, op, _) => Some(op.precedence()),
```
chriseth, line 570: I think we should implement this also for unary operation and LambdaExpression, because those also need parentheses.

ast/src/parsed/display.rs
chriseth: Ok, I think the other way to fix this is to implement the precedence-aware formatting for unary operation and lambda expression.

chriseth: It might be sufficient to implement the `precedence()` function for them.

chriseth:
I think this is looking very good already! I got myself confused during the review.

ast/src/parsed/mod.rs
aminlatifi: @chriseth Do you mean it must be 9 and others starting from 9 right now be shifted by one?!

aminlatifi: According to https://doc.rust-lang.org/reference/expressions.html#expression-precedence
It's at the end of the list. But as I look at it now, I missed the point that it's right associative.

ast/src/parsed/mod.rs
aminlatifi: We forgot to include identity (= += -= *= /= %= &= |= ^= <<= >>=) as right associative.
Also there is an imporatnt thing, the associativey of some fiels is "Require parentheses" ! we must handle it as well
https://doc.rust-lang.org/reference/expressions.html#expression-precedence

aminlatifi: I found this in pilcom source code, I assume power is left assosiative and we must do the same
https://github.com/0xPolygonHermez/pilcom/blob/36bfb6ab4e0f1bae3bbeda11cc6305ede153f6b6/src/pil_parser.jison#L51
@chriseth can you double check?

ast/src/parsed/display.rs
aminlatifi: I fixed the issue for unary operations, but couldn't find an example for lambda, can you help me on this?

ast/src/parsed/mod.rs
chriseth: What matters is not the precedence in the rust documentation but the precedence as implemented by our grammar: https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L469

While our grammar mostly aligns with that of rust, the identity is one point where it diverges, because Identity is not an assignment but another kind of operator.

ast/src/parsed/mod.rs
chriseth: It's a good thing that the parser cannot parse `x = y <= z`. The reason is because `=` and `<=` have the same precedence. Please add tests for `(x = y) <= z` and `x = (y <= z)`.

ast/src/parsed/mod.rs
```diff
@@ -633,6 +635,71 @@ pub enum BinaryOperator {
...
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub enum BinaryOperatorAssociativity {
+    Left,
+    Right,
+    RequireParentheses,
+}
+
+const UNARY_EXPRESSION_PRIORITY_OFFSET: ExpressionPrecedence = 0; // All unary operators have the same precedence.
```
chriseth, line 645: Not sure this is a good idea, I think it distracts. Shouldn't we just stick to the table in the rust documentation?

ast/src/parsed/mod.rs
```diff
@@ -633,6 +635,71 @@ pub enum BinaryOperator {
...
+const BINARY_EXPRESSION_PRIORITY_OFFSET: ExpressionPrecedence = 10; // Binary operators will have a lower precedence than unary operators.
+
+impl UnaryOperator {
+    pub fn precedence(&self) -> ExpressionPrecedence {
+        use UnaryOperator::*;
+        UNARY_EXPRESSION_PRIORITY_OFFSET
+            + match self {
+                Minus | LogicalNot => 1,
```
chriseth, line 653: I think these should be swapped: https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L555

ast/src/parsed/mod.rs
```diff
@@ -633,6 +635,71 @@ pub enum BinaryOperator {
...
+            }
+    }
+}
+
+impl BinaryOperator {
+    pub fn precedence(&self) -> ExpressionPrecedence {
+        use BinaryOperator::*;
+        BINARY_EXPRESSION_PRIORITY_OFFSET
+            + match self {
+                // Unary - * ! & &mut
```
chriseth, line 664:
```suggestion
                // **
```

ast/src/parsed/mod.rs
```diff
@@ -633,6 +635,71 @@ pub enum BinaryOperator {
...
+    }
+
+    pub fn associativity(&self) -> BinaryOperatorAssociativity {
+        use BinaryOperator::*;
+        use BinaryOperatorAssociativity::*;
+        match self {
+            Identity => Right,
+            Equal | NotEqual | Less | Greater | LessEqual | GreaterEqual => RequireParentheses,
+            // .. ..= => RequireParentheses,
+            _ => Left,
```
chriseth, line 698: What about Pow? Is it correct that it is left-associative?

parser/tests/parentheses_test.rs
```diff
@@ -0,0 +1,68 @@
...
+            ("(-x) + y * (!z);", "-x + y * !z;"),
+            ("(x * y) * z;", "x * y * z;"),
+            ("(x / y) / z;", "x / y / z;"),
+            ("x ** (y ** z);", "x ** (y ** z);"),
+            ("(x ** (y ** z));", "x ** (y ** z);"),
+            // Observe associativity
+            ("x * (y * z);", "x * (y * z);"),
+            ("x / (y / z);", "x / (y / z);"),
+            ("(x ** y) ** z;", "x ** y ** z;"),
+            ("((x ** y) ** z);", "x ** y ** z;"),
```
chriseth, line 37: I think you should rather test `x ** (y ** z)` than this one.

chriseth: And I think indeed, exponentiation is right-associative.

ast/src/parsed/display.rs
chriseth: LambdaEperssion (called "closure" in the rust documentation) should have something like 13 - but I'm not sure if this is still true with the offsets. It should be higher than any binary operation.

Tests would be for example

```
let x = 1 + |i| i + 2;
let x = 1 + (|i| i) + 2;
let x = 1 + |i| (i + 2);
let x = (1 + |i| i) + 2;
let x = (1 + |i| (i + 2));
let x = (1 + |i| i + 2);
```

although you should also test some other binary operators, especially those with high precedence numbers.

ast/src/parsed/mod.rs
aminlatifi: I found something in polygon pil that says it's left associative if I am not mistaken 
>>> I found this in pilcom source code, I assume power is left assosiative and we must do the same
https://github.com/0xPolygonHermez/pilcom/blob/36bfb6ab4e0f1bae3bbeda11cc6305ede153f6b6/src/pil_parser.jison#L51
@chriseth can you double check?

Can you double check?

ast/src/parsed/display.rs
```diff
@@ -564,6 +564,90 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+
+    pub fn format_binary_operation(&self, f: &mut Formatter<'_>) -> Result {
+        match self {
+            Expression::BinaryOperation(left, op, right) => {
+                let use_left_parentheses = left.require_parentheses()
+                    || match left.precedence() {
+                        Some(left_precedence) => {
+                            left_precedence > op.precedence()
+                                || (left_precedence == op.precedence()
+                                    && op.associativity() == BinaryOperatorAssociativity::Right)
```
chriseth, line 618: Do we really need the `require_parentheses` function? Isn't it just
```suggestion
                let use_left_parentheses = match left.precedence() {
                        Some(left_precedence) => {
                            left_precedence > op.precedence()
                                || (left_precedence == op.precedence()
                                    && op.associativity() != BinaryOperatorAssociativity::Left)
```

ast/src/parsed/display.rs
```diff
@@ -574,14 +658,8 @@ impl<Ref: Display> Display for Expression<Ref> {
...
-            Expression::UnaryOperation(op, exp) => {
-                if op.is_prefix() {
-                    write!(f, "{op}{exp}")
-                } else {
-                    write!(f, "{exp}{op}")
-                }
-            }
+            Expression::BinaryOperation(_, _, _) => self.format_binary_operation(f),
+            Expression::UnaryOperation(_, _) => self.format_unary_operation(f),
             Expression::IndexAccess(index_access) => write!(f, "{index_access}"),
```
chriseth, line 663: Since we don't put any parentheses at the very outside of a group of binary operations, I fear we also have to include IndexAccess and FunctionCall now. Because `(a + b)(2)` will now be formatted as `a + b(2)` and `(a + b)[2]` also as `a + b[2]`.

ast/src/parsed/display.rs
```diff
@@ -564,6 +564,90 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+                } else {
+                    write!(f, "{}{}", _exp_string, op)
+                }
+            }
+            _ => panic!("Format with precedence called on non-unary operation."),
+        }
+    }
+
+    pub fn format_binary_operation(&self, f: &mut Formatter<'_>) -> Result {
+        match self {
```
chriseth, line 611: If you make this a free function and move the `left, op, right` to the parameters, we can save the match.

chriseth: Same with unary.

ast/src/parsed/display.rs
aminlatifi: Generally we won't make anything wrong if we do it in the way I implemented. But in the way you suggested and I think it is like what it was before, I thought there would be a case that the operation precedence be higher than the left expression and the parentheses be skipped. 
So I added that as future proof.

ast/src/parsed/display.rs
aminlatifi: I will test

ast/src/parsed/display.rs
chriseth: Not ethat I changed `== Right` to `!= Left`

ast/src/parsed/mod.rs
chriseth: I think exponentiation only makes sense as right-associative. It is fine to break compatibility with polygon-pil there. How would you understand `x**2**16`? And please double-check either with the grammar or how it is parsed on an example.

ast/src/parsed/display.rs
aminlatifi: In our case, since all operations with `RequireParentheses` associativity have the same precedence, my version and your version work the same. But if we had a case which two parentheses required operations with different precedence to be next to each other, there would be a problem.
Overall, I think the current version is more expressive and readable.

aminlatifi: I may be wrong, but the scenario I imagined is in the future we have two kind of operation with `RequireParentheses` associativity and different precedence. In that case we may not put any parentheses. @chriseth  Do you think this can be probable?

ast/src/parsed/display.rs
aminlatifi: Three points about lambda expression,
1. Expression like "let x = 1 + |i| i + 2;" cannot be parsed. It seems the parser expects a lambad expression always have parentheses.
2. In formatting Lambda Expression always parentheses are added
https://github.com/powdr-labs/powdr/blob/930f24ac4368ab9e74013ec0cf04bf920c70ec4a/ast/src/parsed/display.rs#L652
3. Our implementation of adding precedence to lambda expression causes this extra parenthesis!
```
Diff < left / right > :
<let x = 1 + (|i| i) + 2;
>let x = 1 + ((|i| i)) + 2;
```
To solve 3, I changed LambdaExpression precedence to `None`, then it worked.
@chriseth Let me know wdyt?

ast/src/parsed/display.rs
chriseth: It is totally fine to not put parentheses in that case since the precedence is different.

ast/src/parsed/display.rs
chriseth: Sure, if the other examples still work, then this seems to be the solution.

ast/src/parsed/display.rs
```diff
@@ -564,6 +552,121 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+        } else {
+            write!(
+                f,
+                "({})({})",
+                fun_call.function,
+                format_list(&fun_call.arguments)
+            )
+        }
+    }
+    pub fn format_index_access(
```
chriseth, line 658: Looks like this can just be the `impl<E: Display> Display for IndexAccess<E> {`, can't it?

ast/src/parsed/display.rs
```diff
@@ -564,6 +552,121 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+        index_access: &IndexAccess<Expression<E>>,
+        f: &mut Formatter<'_>,
+    ) -> Result {
+        if let None = index_access.array.precedence() {
```
chriseth, line 662: Do you have a test for `(|i| i) [j]`?

ast/src/parsed/mod.rs
```diff
@@ -352,6 +352,8 @@ pub enum Expression<Ref = NamespacedPolynomialReference> {
     BlockExpression(Vec<StatementInsideBlock<Self>>, Box<Self>),
 }
 
+pub type ExpressionPrecedence = u16;
```
chriseth, line 355: We don't need to save space here, I think this can just be a regular u64.

parser/tests/parentheses_test.rs
```diff
@@ -0,0 +1,89 @@
...
+            test_paren(&test_case);
+        }
+    }
+    #[test]
+    fn test_lambda_ex_parentheses() {
+        let test_cases: Vec<TestCase> = vec![
+            // ("let x = 1 + |i| i + 2;", "let x = 1 + |i| i + 2;"),
+            ("let x = 1 + (|i| i) + 2;", "let x = 1 + (|i| i) + 2;"),
+            ("let x = (1 + (|i| i)) + 2;", "let x = 1 + (|i| i) + 2;"),
+            // ("let x = 1 + |i| (i + 2);", "let x = 1 + |i| (i + 2);"),
```
chriseth, line 59: Can you make these work by adding additional parentheses to the input?

ast/src/parsed/mod.rs
```diff
@@ -638,6 +640,78 @@ pub enum BinaryOperator {
...
+            Add | Sub => 5,
+            // << >>
+            ShiftLeft | ShiftRight => 6,
+            // &
+            BinaryAnd => 7,
+            // ^
+            BinaryXor => 8,
+            // |
+            BinaryOr => 9,
+            // = += -= *= /= %= &= |= ^= <<= >>=
```
chriseth, line 685: This is not the right comment. Identity is `=`. Please also combine this with the next line

analysis/README.md
```diff
@@ -6,17 +6,17 @@ This crate is where most of the compilation pipeline lives. It takes a parse tre
 
 We define two types of machines: virtual machines and constrained machines. Constrained machines are the lower level kind of machine. They have a notion of blocks through a latch and an operation_id. Virtual machines are a higher level type of machines. For each machine type, we provide the elements which appear in that type.
 
-|                         | Virtual | Constrained |
```
chriseth, line 9: Could you please undo these changes? There are also some newly introduced or removed newlines below.

parser/tests/parentheses_test.rs
```diff
@@ -81,6 +83,14 @@ mod test {
                 "(a | ((b * (c << (d + e))) & (f ^ g))) = (h * ((i + g)));",
                 "a | b * (c << d + e) & (f ^ g) = h * (i + g);",
             ),
+            (
+                "instr_or { 0, X, Y, Z } is (main_bin.latch * main_bin.sel[0]) { main_bin.operation_id, main_bin.A, main_bin.B, main_bin.C };",
+                "instr_or { 0, X, Y, Z } is main_bin.latch * main_bin.sel[0] { main_bin.operation_id, main_bin.A, main_bin.B, main_bin.C };",
```
chriseth, line 88: Can you also add this one as input (as a second test)?

pil-analyzer/tests/parse_display.rs
```diff
@@ -21,7 +21,7 @@ namespace std::convert(65536);
 namespace T(65536);
     col fixed first_step = [1] + [0]*;
     col fixed line(i) { i };
-    let ops: int -> bool = (|i| ((i < 7) && (6 >= -i)));
+    let ops: int -> bool = (|i| (i < 7) && (6 >= -i));
```
chriseth, line 24:
```suggestion
    let ops: int -> bool = |i| i < 7 && 6 >= -i;
```

aminlatifi: Greater and greater than equal have RequireParentheses associativity

ast/src/parsed/mod.rs
```diff
@@ -638,6 +640,78 @@ pub enum BinaryOperator {
...
+            Identity => 10,
+            // == != < > <= >=
+            Equal | NotEqual | Less | Greater | LessEqual | GreaterEqual => 10,
```
chriseth, line 688:
```suggestion
            // = == != < > <= >=
            Identity | Equal | NotEqual | Less | Greater | LessEqual | GreaterEqual => 10,
```

pil-analyzer/tests/parse_display.rs
chriseth: Maybe then this is one example of what I was talking about: You only need parentheses with "RequireParentheses" when the other expression has the same precedence.

aminlatifi: Fixed.

ast/src/parsed/display.rs
```diff
@@ -580,6 +588,88 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+        }
+    }
+
+    pub fn format_unary_operation(
+        &self,
+        op: &UnaryOperator,
+        exp: &Expression<E>,
+        f: &mut Formatter<'_>,
+    ) -> Result {
+        let _exp_string = if let (Some(_precision), Some(_exp_precision)) =
```
chriseth, line 615:
```suggestion
        let exp_string = if let (Some(predecence), Some(inner_precedence)) =
```

ast/src/parsed/display.rs
```diff
@@ -580,6 +588,88 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+            (self.precedence(), exp.precedence())
+        {
+            if _precision < _exp_precision {
+                format!("({})", exp)
+            } else {
+                format!("{}", exp)
```
chriseth, line 621: Actually I think this part can be better done using a match:
```
let exp_string = match (self.precedence(), exp.precedence()) {
    (Some(precedence), Some(inner_precedence)) if precedence < inner_precedence => format!("({exp})"),
    _ => format!("{exp}")
};
``

File: parser/tests/parentheses_test.rs [hunk 27]
```diff
@@ -0,0 +1,105 @@
...
+            ("(-x) + y * (!z);", "-x + y * !z;"),
+            ("(x * y) * z;", "x * y * z;"),
+            ("(x / y) / z;", "x / y / z;"),
+            ("x ** (y ** z);", "x ** y ** z;"),
+            ("(x ** (y ** z));", "x ** y ** z;"),
+            ("(x - (y + z));", "x - (y + z);"),
+            // Observe associativity
+            ("x * (y * z);", "x * (y * z);"),
+            ("x / (y / z);", "x / (y / z);"),
+            ("x ** (y ** z);", "x ** y ** z;"),
```
chriseth, line 38: This is the same test as above. What we don't have is `(x ** y) ** z`, do we?

ast/src/parsed/display.rs
```diff
@@ -587,6 +595,76 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
     format!("{}", list.into_iter().format(", "))
 }
 
+impl<E: Display> Expression<E> {
+    pub fn precedence(&self) -> Option<ExpressionPrecedence> {
+        match self {
+            Expression::UnaryOperation(op, _) => Some(op.precedence()),
+            Expression::BinaryOperation(_, op, _) => Some(op.precedence()),
+            _ => None,
```
chriseth, line 603: Shouldn't you define precedence also for array and lambda expression?

aminlatifi: It's not required with our configuration in my view. They have more precedence than others and are postfix.

ast/src/parsed/display.rs
```diff
@@ -329,15 +329,23 @@ impl<T: Display> Display for Params<T> {
...
+            write!(f, "({})[{}]", self.array, self.index)
+        }
     }
 }
 
-impl<E: Display> Display for FunctionCall<E> {
+impl<E: Display> Display for FunctionCall<Expression<E>> {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
-        write!(f, "{}({})", self.function, format_list(&self.arguments))
+        if self.function.precedence().is_none() {
```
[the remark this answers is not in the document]
aminlatifi: I added them and also checked whether an expression is postfix in displaying function call and index access.

ast/src/parsed/display.rs
```diff
@@ -331,7 +331,7 @@ impl<T: Display> Display for Params<T> {
 
 impl<E: Display> Display for IndexAccess<Expression<E>> {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
-        if self.array.precedence().is_none() {
+        if self.array.precedence().is_none() || self.array.is_postfix() {
```
chriseth, line 334: I don't think we should do it like that. If you implement precedence for IndexAccess and FunctionCall, this should just work as the others: Parentheses if the precedence of the inner is higher than the current, otherwise no parentheses.

chriseth: Hm, ok, maybe this is the way to go, sorry :)

ast/src/parsed/display.rs
```diff
@@ -587,6 +595,84 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+            Some(right_precedence) => {
+                right_precedence > op.precedence()
+                    || (right_precedence == op.precedence()
+                        && op.associativity() != BinaryOperatorAssociativity::Right)
+            }
+            None => false,
+        };
+
+        let left_string = if use_left_parentheses {
+            format!("({})", left)
```
chriseth, line 662: same here and in many other places:
```suggestion
            format!("({left})")
```

ast/src/parsed/display.rs
```diff
@@ -659,14 +659,14 @@ impl<E: Display> Expression<E> {
...
         };
         let right_string = if use_right_parentheses {
-            format!("({})", right)
+            format!("({right})")
         } else {
-            format!("{}", right)
+            format!("{right}")
         };
 
         write!(f, "{} {} {}", left_string, op, right_string)
```
chriseth, line 672: here as well

parser/src/lib.rs
```diff
@@ -125,6 +125,30 @@ pub fn unescape_string(s: &str) -> String {
     result
 }
 
+use powdr_ast::parsed::{PILFile, PilStatement};
+fn pil_statement_clear_source_ref(stmt: &mut PilStatement) {
```
chriseth, line 129: Why did you move this out of the test module?

chriseth: Ah I see. Can you create a new file `test_utils.rs` or something and move it there?

ast/src/parsed/display.rs
```diff
@@ -640,9 +632,15 @@ impl<E: Display> Expression<E> {
         right: &Expression<E>,
         f: &mut Formatter<'_>,
     ) -> Result {
+        let parenthes_on_precedence = match op {
```
chriseth, line 635: `parentheses_on_precedence`

or maybe just `force_parentheses` or inline `op == BinaryOperator::Pow`?

chriseth (approved):

PR #1307, opened 2024-04-23
avoid `split_off` in `combine`

pacheco:
delay the `split_off` decision to the `combine` method, many times we don't need a new `Vec`


chriseth (commented):
Ah, I also considered that actually. Thanks!

PR #1311, opened 2024-04-24
Supply file path to parser.

chriseth:


pipeline/src/pipeline.rs
```diff
@@ -554,15 +554,17 @@ impl<T: FieldElement> Pipeline<T> {
...
-
-                let parsed_asm = powdr_parser::parse_asm(None, asm_string).unwrap_or_else(|err| {
-                    match path.as_ref() {
-                        Some(path) => eprintln!("Error parsing .asm file: {}", path.display()),
-                        None => eprintln!("Error parsing .asm file:"),
-                    }
-                    err.output_to_stderr();
-                    panic!();
-                });
+                let path_str = path.as_ref().map(|p| p.display().to_string());
```
lvella, line 557: I don't think `display()` should be used here. It is better to `to_str()?`.

If the file name is not Unicode, it is better for it to fail here than fail with `file "fo�.txt" not found`.

chriseth: I'm pretty sure that the value is constrained to valid utf8 already, but I can check again. Also note that `display()` was already used in the error reporting below.

lvella: If it is already a valid UTF-8, `to_str().unwrap()` is better in both cases. In this line, it avoids an allocation.

Anyway, in error reporting it is fine to use `display()`, because there is no better way to actually display the invalid string. But if you are planning to actually open the file, then it isn't.

lvella (approved):

PR #1312, opened 2024-04-24
Implement Eq for string.

chriseth:


lvella (approved):
My suggestion is optional, so approved.

lvella (approved):

PR #1313, opened 2024-04-24
fix pipeline callback chain

leonardoalt:


pacheco (commented):
Couldn't query QueryCallback return just `Result<T>` instead of `Result<Option<T>>`?
Doesn't seem like there's any situation where `Ok(None)` is returned.

leonardoalt:
I'd prefer to keep the changes minimal in this PR, since changing the type will affect more places. We can re-consider whether `Ok(None)` is useful later on

pacheco (approved):
Ok, can you make an issue for it?

leonardoalt:
Done: https://github.com/powdr-labs/powdr/issues/1314

PR #1316, opened 2024-04-25
Add publics method to pipeline

Schaeff:
This seemed to me like it was missing.

### DIFF
--- a/pipeline/src/test_util.rs
+++ b/pipeline/src/test_util.rs
@@ -1,6 +1,5 @@
 use powdr_ast::analyzed::Analyzed;
 use powdr_backend::BackendType;
-use powdr_executor::witgen::extract_publics;
 use powdr_number::{buffered_write_file, BigInt, Bn254Field, FieldElement, GoldilocksField};
 use powdr_pil_analyzer::evaluator::{self, SymbolLookup};
 use std::path::PathBuf;


pipeline/src/pipeline.rs
```diff
@@ -827,6 +828,12 @@ impl<T: FieldElement> Pipeline<T> {
         Ok(self.artifact.witness.as_ref().unwrap().clone())
     }
 
+    pub fn publics(&self) -> Result<Vec<(String, T)>, Vec<String>> {
+        let pil = self.optimized_pil()?;
```
Schaeff, line 832: Is `optimized_pil` the right method to call here? And should this method take a mutable reference and compute anything that isn't there yet?

leonardoalt: The `compute_*` methods take a `&mut self` and actually compute whatever is needed. `optimized_pil` takes an immutable ref an assumes all the needed deps have been computed.

But don't you need the witness here too?

pipeline/src/pipeline.rs
```diff
@@ -827,6 +828,12 @@ impl<T: FieldElement> Pipeline<T> {
...
+        let witness = self.witness()?;
```
leonardoalt, line 833: witness has optimized_pil as dependency, so if you assume the witness exists then optimized_pil should exist too

pipeline/src/pipeline.rs
Schaeff: Yes I get it on the next line

pipeline/src/pipeline.rs
```diff
@@ -827,6 +828,12 @@ impl<T: FieldElement> Pipeline<T> {
         Ok(self.artifact.witness.as_ref().unwrap().clone())
     }
 
+    pub fn publics(&self) -> Result<Vec<(String, T)>, Vec<String>> {
```
chriseth, line 831: Wait but what does this do? Aren't publics always inputs?

leonardoalt: This retrieves the publics from the witness. It's useful for backends that need to pass the publics explicitly for the prover or a verifier

leonardoalt: The user passes the publics to powdr/pipeline, but for powdr they're just part of the witness table, with some copy constraints.

chriseth: Maybe this should be documented? I would not have expected a full witness computation just from asking for the publics.

leonardoalt: But thats not it, its just for when you already have the witness but just want to know what the publics are

leonardoalt (commented):
I think the current assumptions here are fine

leonardoalt (approved):

leonardoalt (approved):

PR #1317, opened 2024-04-25
increase continuations bootloader's page size

leonardoalt:
Without this powdr-revme runs out of Poseidon rows sometimes.


riscv/src/continuations/bootloader.rs
```diff
@@ -9,7 +9,7 @@ use super::memory_merkle_tree::MerkleTree;
 pub const MEMORY_SIZE_LOG: usize = 32;
 
 /// Page size is 1KB
```
chriseth, line 11:
```suggestion
/// Page size is 2KB
```

leonardoalt: done

chriseth (approved):

Issue #1287, opened 2024-04-18
Detect non-exhaustiveness of patterns in match


chriseth:
We already have a check if a pattern is refutable, we should also check that match patterns are exhaustive.

PR #1318, opened 2024-04-25
Non exhaustiveness of patterns

gzanitti:
This PR implements non-exhaustiveness check for matchs (issue #1287). Based on the `usefulness` implementation proposed in the rustc_pattern_analysis[1] module of the Rust compiler. A match is exhaustive iff the wildcard `_` pattern is not useful with respect to  the patterns in the match. The values returned by usefulness are used to tell the user which values are missing ([] where the match is exhaustive).

Still in draft until I finish analyzing some edge cases.

[1] https://doc.rust-lang.org/nightly/nightly-rustc/rustc_pattern_analysis/usefulness/index.html


ast/src/parsed/mod.rs
```diff
@@ -914,6 +914,111 @@ impl Pattern {
             Pattern::Tuple(p) => p.iter().all(|p| p.is_irrefutable()),
         }
     }
+
+    pub fn is_fully_catch_all(&self) -> bool {
```
chriseth, line 918: What's the difference to `is_irrefutable`?

gzanitti: I think theoretically there should be no difference, but why do you only consider irrefutable an array with ellipsis? The case `[_,_,_]` (where all cases are catchalls) isn't irrefutable as well?

gzanitti: Oh, I see, you could use it to match the array size, my bad:

    let vec = vec![1, 2, 3];

    match vec.as_slice() {
        [_,_] => println!("2!"),
        [_, _, _] => println!("3"),
        _ => println!("Other"),
    }

gzanitti: Fixed. Thanks!

chriseth:
Ah nice! I asked because I was not sure if the other function is maybe different. One small note: I think currently an enum variant is always marked as refutable, but with this checker, I think we should be more exact: It is irrefutable if and only if it is the only variant of the enum.

ast/src/parsed/mod.rs
```diff
@@ -947,3 +1039,132 @@ pub struct TypedExpression<Ref = NamespacedPolynomialReference, E = Expression<R
     pub e: Expression<Ref>,
     pub type_scheme: Option<TypeScheme<E>>,
 }
+
+#[cfg(test)]
```
chriseth, line 1043: I'm wondering if it would be better to move the pattern stuff into its own file since it seems to grow quite a bit :)

chriseth:
> Cool! Just to confirm:
> 
> ```
> enum Name {
>     Variant1,
>     Variant2,
> }
> ```
> 
> is represented as `Enum(Name, [Variant1, Variant2])`, right?


Not sure I understood correctly but given that enum declaration, we could use it in a pattern like this:
```
match x {
  Name::Variant1 => ...,
  Name::Variant2 => ...,
}
```
The second parameter of `Pattern::Enum` is the data fields of an enum if it has them, i.e. `Name::Variant1(x, 2) => ...`.

If you consider the following enum definition:
```
enum SingleVarEnum {
  Var(int, int)
}
```
Then we could do the following because the pattern is irrefutable:
```
let SingleVarEnum::Var(x, y) = m;
```
Of course we still need to check that the inner patterns are all refutable.

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,358 @@
...
+            Pattern::Number(_) | Pattern::String(_) | Pattern::Enum(_, _) => false,
+            Pattern::Array(items) => {
+                // Only "[..]"" is irrefutable
+                items == &vec![Pattern::Ellipsis]
+            }
+            Pattern::Tuple(p) => p.iter().all(|p| p.is_irrefutable()),
+        }
+    }
+
+    // Specialize a pattern based on a "constructor" pattern passed as a parameter.
```
chriseth, line 48: Can you use triple-slash and explain at least in some words what this does, so one can get a general idea without having to open that link?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,358 @@
...
+        match (constructor, self) {
+            //(Pattern::CatchAll, _) => Some([].to_vec()),
+            (Pattern::CatchAll, Pattern::CatchAll) => Some([].to_vec()),
+
+            (Pattern::Number(y), Pattern::Number(x)) => {
+                if y == x {
+                    Some([].to_vec())
+                } else {
+                    None
+                }
```
chriseth, line 60:
```suggestion
                (y == x).then_some(].to_vec())
```

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -46,6 +47,7 @@ fn analyze<T: FieldElement>(files: Vec<PILFile>) -> Analyzed<T> {
     let mut analyzer = PILAnalyzer::new();
     analyzer.process(files);
     analyzer.side_effect_check();
+    analyzer.match_exhaustiveness_check();
```
chriseth, line 50: I think the errors we get are better if we first do type checking, because we assume the types to be correct in the exhaustiveness check.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -174,6 +176,71 @@ impl PILAnalyzer {
         }
     }
 
+    pub fn match_exhaustiveness_check(&self) {
+        for (name, (_, value)) in &self.definitions {
```
chriseth, line 180: You should also iterate over the identities.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -174,6 +176,71 @@ impl PILAnalyzer {
...
+        if patterns.is_empty() || patterns.iter().any(|p| p.is_irrefutable()) {
+            return vec![];
+        }
+
+        let mut expanded_patterns = patterns.to_vec();
+        expanded_patterns.push(new_pattern.clone());
+        let constructors = expanded_patterns.clone(); //self.constructors_from_patterns(expanded_patterns.as_slice());
+
+        for constructor in &constructors {
+            let specialized_new_pattern = new_pattern.specialize(constructor);
```
chriseth, line 213: Doesn't specialize need to return both a new constructor and a new pattern?

chriseth:
No worries, take your time, it's not easy navigating in a new codebase.

pil-analyzer/src/pil_analyzer.rs
gzanitti: No, specialize only "removes a layer" of the pattern if the constructor is the same. 

```
specialize(pattern = [1,2,3], cons = [_,_,_,_]) => (1 ,2 ,3) 
```
because they are both `[]`.

also unspecialize(...) is literally specialize(...)<sup>-1</sup> so
```
unspecialize(pattern = (1,2,3), cons = [_,_,_,_] ) => [1,2,3]
```

gzanitti:
It seems to be working. I may add a few more tests, but overall it is ready for review.

By the way, it looks like this match
https://github.com/powdr-labs/powdr/blob/930f24ac4368ab9e74013ec0cf04bf920c70ec4a/linker/src/lib.rs#L504 is not exhaustive and that's why the tests are failing. 
How should I go about getting someone to look at it? Should I open an issue? Just a message on matrix at any time?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -174,6 +176,99 @@ impl PILAnalyzer {
...
+                if witnesses.is_empty() {
+                    panic!("Function {name} is not exhaustive")
+                }
+            }
+        }
+
+        for id in &self.identities {
+            let patterns = id
+                .all_children()
+                .filter_map(|e| {
```
chriseth, line 203: Can you try to re-use the code from above? Maybe there is a way to use an iterator through both structures.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -174,6 +176,99 @@ impl PILAnalyzer {
...
 
+    /// Check that all match expressions are exhaustive.
+    pub fn match_exhaustiveness_check(&self) {
+        for (name, (_, value)) in &self.definitions {
+            let Some(value) = value else { continue };
+            for e in value.all_children() {
+                let patterns = if let Expression::MatchExpression(_, arms) = e {
+                    arms.iter()
+                        .map(|arm| PatternTuple {
+                            patterns: vec![arm.pattern.clone()],
```
chriseth, line 187: Is there a way to avoid the clone? In the end you are passing a const ref.

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
+use powdr_number::BigInt;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use std::iter::{empty, once};
+
+use super::{asm::SymbolPath, visitor::Children};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct PatternTuple {
```
chriseth, line 10: Can you document a little bit what this is beyond just a vector of patterns? Also please make clear that this is not part of the AST.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -174,6 +176,99 @@ impl PILAnalyzer {
...
+                        })
+                        .collect::<Vec<_>>()
+                } else {
+                    continue;
+                };
+                let witnesses = Self::usefulness(&patterns, Pattern::CatchAll.into());
+                if witnesses.is_empty() {
+                    panic!("Function {name} is not exhaustive")
```
chriseth, line 195: You can print the whole match expression until we have source references. Can you print the witness as well?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -174,6 +176,99 @@ impl PILAnalyzer {
...
+                if witnesses.is_empty() {
+                    panic!("Identity is not exhaustive")
+                }
+            }
+        }
+    }
+
+    /// Check if a new pattern is useful in a set of patterns.
+    /// A pattern is useful if it is not covered by the other patterns.
+    /// If a pattern is useful, it is returned as a witness.
```
chriseth, line 228: But why does this return a vector of PatternTuple?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -174,6 +176,99 @@ impl PILAnalyzer {
...
+    pub fn usefulness(patterns: &[PatternTuple], new_pattern: PatternTuple) -> Vec<PatternTuple> {
+        let mut witnesses = HashSet::new();
```
chriseth, line 230: Can move further down.

pil-analyzer/src/pil_analyzer.rs
chriseth: Also the new pattern is a PatternTuple...

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -174,6 +176,99 @@ impl PILAnalyzer {
...
+        {
+            return vec![];
+        }
+
+        let mut expanded_patterns = patterns.to_vec();
+        expanded_patterns.push(new_pattern.clone());
+        let constructors = expanded_patterns.clone();
+
+        for constructor in &constructors {
+            let specialized_new_pattern = if !new_pattern.is_empty() {
```
chriseth, line 244: Might be my personal preference but please avoid negation if possible (and swap the cases instead)

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -174,6 +176,99 @@ impl PILAnalyzer {
...
+                        .filter_map(|pattern| pattern.specialize(constructor))
+                        .collect::<Vec<_>>();
+
+                    let specialized_usefull = Self::usefulness(&specialized_results, v);
+
+                    for witness in specialized_usefull {
+                        if let Some(v) = constructor.unspecialize(witness) {
+                            witnesses.insert(v);
+                        }
+                    }
```
chriseth, line 263: maybe
```suggestion
                    witnesses.extend(specialized_usefull.into_iter().flat_map(|witness| constructor.unspecialize(witness));
```
?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
...
+impl PatternTuple {
+    pub fn is_empty(&self) -> bool {
+        self.patterns.is_empty()
+    }
+
+    pub fn is_irrefutable(&self) -> bool {
+        if self.patterns.len() != 1 {
+            return false;
+        }
+        self.patterns[0].is_irrefutable()
```
chriseth, line 23:
```suggestion
        if self.patterns.len() == 1 {
            self.patterns[0].is_irrefutable()
        } else {
            false;
        }

```

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
...
+pub struct PatternTuple {
+    pub patterns: Vec<Pattern>,
+}
+
+impl PatternTuple {
+    pub fn is_empty(&self) -> bool {
+        self.patterns.is_empty()
+    }
+
+    pub fn is_irrefutable(&self) -> bool {
```
chriseth, line 19: Can you document this? I cannot follow its implementation at all :D

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
...
+        if self.patterns.is_empty() || self.patterns.len() != constructor.patterns.len() {
+            return None;
+        }
+
+        let first = self.patterns.first().unwrap();
+
+        let mut patterns = self.patterns.clone();
+        let rest = patterns.drain(1..).collect::<Vec<Pattern>>();
+        let specialized = first.specialize(&constructor.patterns[0]);
+        match specialized {
```
chriseth, line 40: This could be a `.map(...)`

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
...
+    /// "one layer" (the constructor) and returning internal patterns.
+    pub fn specialize(&self, constructor: &PatternTuple) -> Option<PatternTuple> {
+        if self.patterns.is_empty() || self.patterns.len() != constructor.patterns.len() {
+            return None;
+        }
+
+        let first = self.patterns.first().unwrap();
+
+        let mut patterns = self.patterns.clone();
+        let rest = patterns.drain(1..).collect::<Vec<Pattern>>();
```
chriseth, line 38: This is only needed in the Some case below

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
...
+        match self {
+            Pattern::Ellipsis => unreachable!(),
+            Pattern::CatchAll | Pattern::Variable(_) => true,
+            Pattern::Number(_) | Pattern::String(_) | Pattern::Enum(_, _) => false,
+            Pattern::Array(items) => {
+                // Only "[..]"" is irrefutable
+                items == &vec![Pattern::Ellipsis]
+            }
+            Pattern::Tuple(p) => p.iter().all(|p| p.is_irrefutable()),
+        }
```
chriseth, line 165: Can you add a TODO or something that this should also return true for enum variants that are the only ones in the enum?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    /// Specialize a pattern based on a "constructor" pattern passed as a parameter.
+    /// Based on https://doc.rust-lang.org/nightly/nightly-rustc/rustc_pattern_analysis/usefulness/index.html#specialization.
+    /// If the constructor is shared with the pattern, specialize the pattern removing
```
chriseth, line 170: Can you explain a bit more here? For example what does returning `None` mean? Or what is an empty pattern tuple?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
...
+                }
+                match (patterns1, patterns2) {
+                    (Some(_), Some(_)) => {
+                        let pats = patterns2.as_ref().unwrap();
+                        Some(PatternTuple {
+                            patterns: pats.clone(),
+                        })
+                    }
+
+                    _ => None,
```
chriseth, line 234: is this correct? If the enum variant doesn't take data and the pattern also does not have fields, shouldn't this be an empty pattern tuple?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
...
+                Some(PatternTuple { patterns })
+            }
+            None => None,
+        }
+    }
+
+    /// Unspecialize a pattern tuple based on a "constructor" pattern tuple passed as a parameter.
+    /// Based on https://doc.rust-lang.org/nightly/nightly-rustc/rustc_pattern_analysis/usefulness/index.html##undoing-specialization.
+    /// Construct a pattern based in the constructor using the values in the tuple as parameters
+    /// If the constructor takes fewer parameters than there are in the tuple, the tuple is extended with the remaining parameters.
```
chriseth, line 52: Again, what does it mean to return None here?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,411 @@
...
+
+        let constructor = self.patterns.first().unwrap();
+        if tuple.patterns.is_empty() {
+            Some(PatternTuple {
+                patterns: vec![constructor.clone()],
+            })
+        } else {
+            match constructor {
+                Pattern::CatchAll => {
+                    let mut patterns = tuple.patterns.clone();
```
chriseth, line 66: Do you really need to clone here?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
+use powdr_number::BigInt;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use std::iter::{empty, once};
+
+use super::{asm::SymbolPath, visitor::Children};
+
+/// When a pattern is specialised, in some cases the algorithm will return
```
chriseth, line 9: Maybe it makes sense to move this a bit down so that Pattern is the first element in this file?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+}
+
+impl PatternTuple {
+    pub fn is_empty(&self) -> bool {
+        self.patterns.is_empty()
+    }
+
+    /// Check for irrefutable patterns in pattern tuples. Since pattern tuples can contain
+    /// internal patterns, irrefutability only applies in cases of unique patterns (len == 1).
+
```
chriseth, line 27: Does this work fine if you have an empty line without comment?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+    /// Example:
+    /// If (_, _) was obtained from specialization, it is not true that the pattern is irrefutable
+    /// because having more than one element implies that the evaluation is performed on internal patterns.
+    /// If, on the other hand, the pattern tuple is `(Pattern::Tuple(_,_), )``, irrefutability applies in the standard way.
+    pub fn is_irrefutable(&self) -> bool {
```
chriseth, line 32: Hm, I cannot find any use of this function...

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+        } else {
+            false
+        }
+    }
+
+    /// Specialize a pattern tuple based on a "constructor" pattern tuple passed as a parameter.
+    /// Based on https://doc.rust-lang.org/nightly/nightly-rustc/rustc_pattern_analysis/usefulness/index.html#specialization.
+    /// If the constructor is shared with the pattern, specialize the pattern removing
+    /// "one layer" (the constructor) and returning internal patterns.
+    /// Return None if the pattern can't be specialized with the constructor or if the pattern is empty.
```
chriseth, line 44: I still don't really understand this routine, but maybe adding the following items to the docstring helps:

We are only interested in the outermost type-constructor of the `constructor` pattern and ignore the fields and the inner patterns.

And maybe also a variant of the following:
```
matches!(c(v_1, .., v_n), p)
<=> specialize(c, p) returns something
    && matches!((v_1, .., v_n), specialize(c, p))
```

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+                Pattern::Enum(path1, _) => Some(PatternTuple {
+                    patterns: vec![Pattern::Enum(path1.clone(), Some(tuple.patterns))],
+                }),
+                _ => None,
+            }
+        }
+    }
+}
+
+impl Default for PatternTuple {
```
chriseth, line 142: Isn't that the same as `derive(Default)`?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+            }
+            Pattern::Tuple(p) => p.iter().all(|p| p.is_irrefutable()),
+        }
+    }
+
+    /// Specialize a pattern based on a "constructor" pattern passed as a parameter.
+    /// Based on https://doc.rust-lang.org/nightly/nightly-rustc/rustc_pattern_analysis/usefulness/index.html#specialization.
+    /// If the constructor is shared with the pattern, specialize the pattern removing
+    /// "one layer" (the constructor) and returning internal patterns.
+    /// Return None if the pattern can't be specialized using the constructor.
```
chriseth, line 194: Ok, I think this part I understand now. Let me rephrase it in my own words and then we can see how we want to combine the docs.
```suggestion
    /// We are only interested in the type constructor part of `constructor` and ignore the inner patterns and fields.
    /// If the constructor does not match the pattern, returns None.
    /// If it matches, returns a vector of corresponding constructor patterns for the inner fields of the pattern.
```

And then maybe add some formal statement like on https://doc.rust-lang.org/nightly/nightly-rustc/rustc_pattern_analysis/usefulness/index.html#specialization

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+    pub fn specialize(&self, constructor: &Self) -> Option<PatternTuple> {
+        match (constructor, self) {
+            (_, Pattern::CatchAll) => Some(PatternTuple { patterns: vec![] }),
+            (Pattern::CatchAll, _) => None,
```
chriseth, line 198: I'm wondering, shouldn't this panic? CatchAll is not a constructor, is it?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+                Some(PatternTuple {
+                    patterns: constructors,
+                })
+            }
+            (_, Pattern::Variable(var)) => Some(PatternTuple {
+                patterns: [].to_vec(),
+            }),
+
+            (Pattern::Enum(path1, patterns1), Pattern::Enum(path2, patterns2)) => {
+                if path1 != path2 {
```
chriseth, line 245: Could be pulled into the match pattern

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+                    return None;
+                }
+                match (patterns1, patterns2) {
```
chriseth, line 248:
```suggestion
                match (cons_patterns, patterns) {
```

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+                    (_, Some(_)) => {
+                        let pats = patterns2.as_ref().unwrap();
+                        Some(PatternTuple {
+                            patterns: pats.clone(),
+                        })
```
chriseth, line 253:
```suggestion
                        Some(PatternTuple {
                            patterns: patterns2.as_ref().unwrap().clone(),
                        })
```
(or maybe even `.cloned()` cold work here?)

chriseth: Oh wait, why don't you use `Some(patterns)` instead of `Some(_)` one line above?

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+                if cons_patterns.len() == patterns.len() =>
+            {
+                Some(PatternTuple {
+                    patterns: patterns.to_vec(),
+                })
+            }
+            (Pattern::Array(cons_patterns), Pattern::Array(patterns)) => {
+                let max_len = cons_patterns.len();
+                let mut constructors = Vec::new();
+                if patterns.contains(&Pattern::Ellipsis) {
```
chriseth, line 216:
```suggestion
                if let Some(ellipsis_inde) = patterns.iter().position(|p| p ==&Pattern::Ellipsis) {
```

ast/src/parsed/pattern.rs
```diff
@@ -0,0 +1,432 @@
...
+
+            (Pattern::Tuple(cons_patterns), Pattern::Tuple(patterns))
+                if cons_patterns.len() == patterns.len() =>
+            {
+                Some(PatternTuple {
+                    patterns: patterns.to_vec(),
+                })
+            }
+            (Pattern::Array(cons_patterns), Pattern::Array(patterns)) => {
+                let max_len = cons_patterns.len();
```
chriseth, line 214: cons_len? Also please pull it into the block where it is used.

ast/src/parsed/pattern.rs
chriseth: Ok it seems I still have not understood it :)
The returned patterns correspond to the sub-items of the constructor, not of the pattern, right?

ast/src/parsed/pattern.rs
chriseth, line 216: Can we assert that the constructor does not contain an ellipsis?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -381,6 +423,61 @@ impl PILAnalyzer {
     fn driver(&self) -> Driver {
         Driver(self)
     }
+
+    /// Check if a new pattern is useful in a set of patterns.
```
chriseth, line 427: What is the new pattern here?

PR #1325, opened 2024-04-26
Restore outputing commits and constants in the pipeline.

lvella:


backend/src/estark/mod.rs
```diff
@@ -131,45 +121,43 @@ impl<'a, F: FieldElement> EStarkFilesCommon<'a, F> {
...
-    fn write_files(
-        &self,
-        output_dir: &Path,
-        witness: &[(String, Vec<F>)],
-    ) -> Result<ProverInputFilePaths, Error> {
+    fn write_files(&self, output_dir: &Path) -> Result<ProverInputFilePaths, Error> {
+        let constants_name = if self.patched_constants.is_some() {
+            "constants_estark.bin"
+        } else {
+            "constants.bin"
```
leonardoalt, line 147: in this case, won't this have been written already?

pipeline/src/pipeline.rs
```diff
@@ -920,6 +915,10 @@ impl<T: FieldElement> Pipeline<T> {
         self.output_dir.as_ref().map(|p| p.as_ref())
     }
 
+    pub fn force_overwrite(&self) -> bool {
```
leonardoalt, line 918: this sounds like you're telling pipeline to force overwrite things. Maybe `is_force_overwrite`?

lvella: fixed

lvella:
Finally...

PR #1329, opened 2024-04-29
The prelude

chriseth:


chriseth:
Still needs documentation.

std/prelude.asm
```diff
@@ -0,0 +1,9 @@
+/// Like in Rust, all symbols from this file are available in all modules,
+/// meaning that a failed symbol resolution always re-tries relative to this module.
+
+/// The Option type.
+use std::utils::Option;
+
+use std::convert::expr as to_expr;
+use std::convert::fe as to_fe;
+use std::convert::int as to_int;
```
georgwiese, line 9: Why the rename? So that it's not confused with the type?

std/prelude.asm
```diff
@@ -0,0 +1,9 @@
+/// Like in Rust, all symbols from this file are available in all modules,
```
georgwiese, line 1: Can we add `true` and `false` as well?

book/src/asm/modules.md
```diff
@@ -6,4 +6,8 @@ powdr exposes a module system to help organise and reuse code.
 {{#include ../../../test_data/asm/book/modules.asm}}
 ```
 
-Note that a module can't be called `std`, as this name is reserved for an upcoming powdr standard library.
\ No newline at end of file
+Note that a module can't be called `std`, as this name is reserved for an upcoming powdr standard library.
```
leonardoalt, line 9:
```suggestion
Note that a module can't be called `std`, as this name is reserved for the powdr standard library.
```

book/src/asm/modules.md
```diff
@@ -6,4 +6,8 @@ powdr exposes a module system to help organise and reuse code.
...
+
+Similar to Rust, any reference that cannot be resolved is lookup up once more in ``std::prelude``.
```
leonardoalt, line 11:
```suggestion
Similar to Rust, any reference that cannot be resolved is looked up once more in ``std::prelude``.
```
?

book/src/asm/modules.md
```diff
@@ -6,4 +6,8 @@ powdr exposes a module system to help organise and reuse code.
...
+This module exposes basic types like ``Option``. This means that you can use `Option` anywhere
```
leonardoalt, line 12:
```suggestion
This module exposes basic types such as ``Option``. This means that you can use `Option` anywhere
```

georgwiese (approved):
LGTM, but two questions.

book/src/pil/declarations.md
```diff
@@ -24,4 +24,10 @@ Examples:
...
 ```rust
 {{#include ../../../test_data/pil/book/declarations.pil:declarations}}
-```
\ No newline at end of file
+```
+
+Name lookup is performed as follows:
+
+Lookup is performed starting from the current namespace, going up to the root component by component
+and the first match is used. If the lookups all fail, a last attempt is done
```
leonardoalt, line 32:
```suggestion
where the first match is used. If all lookups fail, a last attempt is done
```

book/src/asm/modules.md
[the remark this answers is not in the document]
chriseth: I didn't write that :)

book/src/asm/modules.md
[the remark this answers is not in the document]
chriseth: Why?

std/prelude.asm
chriseth: Yes

chriseth: We can also rename the builtin function

book/src/asm/modules.md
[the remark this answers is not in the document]
georgwiese: Are two ticks doing something else than one tick?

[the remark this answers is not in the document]
chriseth: No but in some dialects of markdown they are required.

PR #1218, opened 2024-03-31
Circuit compiler

chriseth:

PR #1337, opened 2024-04-30
Keccak Rust Wrappers

qwang98:
Input for syscall is one memory pointer to the state array of 25 gl fe. Output is calculated in place.

All functions outside of keccakf are executed in Rust. Might need to delete everything except keccakf from keccakf.asm (including the padding, updating, and byte <-> u64 conversions).

Not tested. Should I wait till all infrastructure needed are merged? Technically can also do the following for the machine and I think I can test it already?

```
pol commit input_state[25];
pol commit output_state[25];
operation keccakf_permutation<0> input_state[0], input_state[1], ..., input_state[24] -> output_state[0], output_state[1], ..., output_state[24]
array::zip(input_state, output_state, |input_state, output_state| (keccakf(input_state) - output_state = 0);
```


riscv/src/runtime.rs
```diff
@@ -148,6 +148,48 @@ impl Runtime {
...
+                "instr keccakf ~ keccakf.keccakf {};", // operation
+                instr_register_params(0, 25, 25) // 25 registers for the states, covered by 26 registers of riscv
+            )], 
+            0,
+            // zero out output registers
+            // TODO: I did this similar to what poseidon did, 
+            // but I'm concerned that any preexisting values in the registers aren't saved.
+            // Can I simply leave this blank?
+            (0..25).map(|i| format!("{} <=X= 0;", reg(i))),
+        );
```
qwang98, line 166: Not sure if this is correct.

test_data/asm/keccakf.asm
```diff
@@ -163,6 +163,8 @@ let main: int, int[], int -> int[] = |W, input, delim| {
 };
 
 // main machine
-machine Main { 
+machine KeccakF { 
     let x;
+    let y;
+    operation keccakf<0> x -> y;
```
qwang98, line 169: Just a placeholder, but I think we need 25 columns for input and 25 columns for output right?

I mean in the syscall it's just one memory pointer for both the input and the output but the sub machine should be different right?

riscv/src/runtime.rs
qwang98, line 166: init_call field. Not sure if this is correct.

test_data/asm/keccakf.asm
qwang98, line 169: Just a placeholder, but I think we need 25 columns for input and 25 columns for output right?

I mean in the syscall it's just one memory pointer for both the input and the output but the sub machine should be different right?

riscv/src/runtime.rs
```diff
@@ -148,6 +148,48 @@ impl Runtime {
         self.syscalls.contains_key(&s)
     }
 
+    pub fn with_keccak(mut self) -> Self {
+        self.add_submachine(
+            "test_data::asm::keccakf::KeccakF",
```
leonardoalt, line 153: I don't think this is accessible, you need to add a new machine to `std`.

test_data/asm/keccakf.asm
leonardoalt: I think it's easier to leave this machine/test unchanged and make a new almost-empty machine in `std/hash`

riscv/src/runtime.rs
```diff
@@ -148,6 +148,48 @@ impl Runtime {
...
+            None,
+            "keccakf",
+            [format!(
+                "instr keccakf ~ keccakf.keccakf {};", // operation
+                instr_register_params(0, 25, 25) // 25 registers for the states, covered by 26 registers of riscv
```
leonardoalt, line 158: It should be just 2 numbers here, the memory addresses of the input and output. The machine will use memory directly.

riscv/src/runtime.rs
```diff
@@ -148,6 +148,48 @@ impl Runtime {
...
+            )],
+            0,
+            // zero out output registers
+            // TODO: I did this similar to what poseidon did,
+            // but I'm concerned that any preexisting values in the registers aren't saved.
+            // Can I simply leave this blank?
```
leonardoalt, line 164: I think this is actually not needed because the machine will use memory directly.

riscv/src/runtime.rs
```diff
@@ -148,6 +148,48 @@ impl Runtime {
...
+            // reuse x10 as input to the keccakf machine instruction.
+            std::iter::once("tmp3 <=X= x10;".to_string())
+            // The keccakf instruction uses registers 0..25 as input/output.
+            // The memory field elements are loaded into these registers before calling the instruction.
+            // They might be in use by the riscv machine, so we save the registers on the stack.
+            .chain((0..25).flat_map(|i| push_register(&reg(i))))
+            .chain((0..25).flat_map(|i| load_gl_fe("tmp3", i as u32 * 8, &reg(i))))
+            .chain(std::iter::once("keccakf;".to_string()))
+            .chain((0..25).flat_map(|i| store_gl_fe("tmp3", i as u32 * 8, &reg(i))))
+            // After copying the result back into memory, we restore the original register values.
```
leonardoalt, line 181: I'm confused by this, why is this passing so many registers?

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
     poseidon_gl_unsafe(data)
 }
+
+/// Calls the keccakf machine
+/// Return value is placed in the same array as the input
+/// This is unsafe because it does not check if the u64 elements fit the Goldilocks field.
+pub fn keccakf_unsafe(mut data: [u64; 25]) -> [u64; 25] {
+    unsafe {
+        // syscall input: memory pointer to state array
+        asm!("ecall", in("a0") &mut data as *mut [u64; 25], in("t0") u32::from(Syscall::KeccakF));
```
leonardoalt, line 37: Nice!

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
 
     poseidon_gl_unsafe(data)
 }
+
+/// Calls the keccakf machine
+/// Return value is placed in the same array as the input
+/// This is unsafe because it does not check if the u64 elements fit the Goldilocks field.
+pub fn keccakf_unsafe(mut data: [u64; 25]) -> [u64; 25] {
```
leonardoalt, line 34: Maybe it makes sense to take the input by reference, allocate a new array and return that?

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+    keccakf_unsafe(data)
+}
+
+// Output number of bytes for keccak-256 (32 bytes)
+const W: usize = 32;
+
+/// Keccak function that calls the keccakf machine
+/// Input is a byte array of arbitrary length and a delimiter byte
+/// Output is a byte array of length W
+pub fn keccak(data: &[u8], delim: u8) -> [u8; W] {
```
leonardoalt, line 57: Nice!

riscv/src/runtime.rs
qwang98: It currently resides in the test folder only till the machine is complete, but I agree that eventually it'll be in `std`.

riscv/src/runtime.rs
qwang98: I see. I can add an output pointer parameter as well. I just wonder if it'll be easier to actually just use one pointer for both input and output, as this compression/mixing function is run multiple times iteratively on the output of itself during a single keccak run.

riscv/src/runtime.rs
leonardoalt: is this path accessible though?

riscv/src/runtime.rs
leonardoalt: yes

leonardoalt: and the reason why we need 2 pointers, one for input and one for output, is that we can do both the input reads and the output writes in a single row

riscv/src/runtime.rs
leonardoalt: yes that's what we do now also with Poseidon, but the new machines will be able to use memory directly, so we only need to pass the 2 pointers via registers

riscv/src/runtime.rs
pacheco: AFAIU, we'll need a separate output pointer due to the workings of the memory machine (to allow reading/writing in the same "row")

riscv-runtime/src/hash.rs
pacheco: Yeah, these values are quite large, maybe it should take a slice and return a `Vec`?

pacheco: Or maybe take two parameters? `input` as an immutable slice, and `output` as a mutable slice (then the caller can decide how to allocate space).

pacheco: The keccak loop could allocate two arrays and alternate them as input/output for example.

riscv/src/runtime.rs
```diff
@@ -148,6 +148,48 @@ impl Runtime {
...
+            [format!(
+                "instr keccakf ~ keccakf.keccakf {};", // operation
+                instr_register_params(0, 25, 25) // 25 registers for the states, covered by 26 registers of riscv
+            )],
+            0,
+            // zero out output registers
+            // TODO: I did this similar to what poseidon did,
+            // but I'm concerned that any preexisting values in the registers aren't saved.
+            // Can I simply leave this blank?
+            (0..25).map(|i| format!("{} <=X= 0;", reg(i))),
```
pacheco, line 165: This parameter only exists as a workaround for a current limitation in witgen:
any included machines must be called at least once.

So what this should be doing is calling the `keccakf` instruction and then zeroing out the output registers (as they should be zero when the actual program starts). 
Now, I'm not completely sure how this should work when the input parameters are actually memory addresses (what address should we use? and we probably need to zero out memory after the call).

Or maybe this need of calling machines at least once could be fixed in witgen? Gotta ask @georgwiese

georgwiese: Yeah, this get's pretty complicated, unfortunately, because this call would do memory operations that we have to "undo". If we're not using continuations, we could just pass any memory pointers and then zero the resulting region again. But in the presence of continuations it's even weirder, because there are constraints that only paged-in pages can be accessed, so what if there are none (because this chunk doesn't otherwise access memory)?

Probably fixing the requirement for witgen is the best solution. #548 tracks it, but it's also not straight-forward...

riscv/src/runtime.rs
qwang98: Not super sure but I'll test it out.

riscv/src/runtime.rs
qwang98: Understood!

riscv/src/runtime.rs
qwang98: Got it, and these two pointers are really both "input" argument pointers, i.e. the Rust wrapper will take two arguments, each corresponding to a pointer in the register? It's just that the function will operate on the array pointed by the pointer for the input, and put the results into the array pointed by the other pointer for the output?

riscv-runtime/src/hash.rs
qwang98: Makes sense. Thanks!

riscv/src/runtime.rs
qwang98: But why should it call Keccakf first before zeroing out the output registers?

> This parameter only exists as a workaround for a current limitation in witgen: any included machines must be called at least once.
> 
> So what this should be doing is calling the `keccakf` instruction and then zeroing out the output registers (as they should be zero when the actual program starts). Now, I'm not completely sure how this should work when the input parameters are actually memory addresses (what address should we use? and we probably need to zero out memory after the call).
> 
> Or maybe this need of calling machines at least once could be fixed in witgen? Gotta ask @georgwiese

qwang98: Given some updates are needed. I'll leave it as is with a TODO for now.

riscv/src/runtime.rs
leonardoalt: yes! As @pacheco also said in some other comment, the advantage of taking the output pointer also as parameter is that the user is then responsible for allocating the output array and our library doesn't have to do that

test_data/asm/keccakf.asm
qwang98: Added a placeholder in std/hash.

riscv/src/runtime.rs
qwang98: Done.

riscv/src/runtime.rs
qwang98: Done.

riscv/src/runtime.rs
qwang98: Done.

riscv-runtime/src/hash.rs
qwang98: Done.

riscv/src/runtime.rs
qwang98: Done.

qwang98:
Ready for review again.

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
 
     poseidon_gl_unsafe(data)
 }
+
+/// Calls the keccakf machine
```
pacheco, line 31: minor: function comments here could use periods at the end of sentences

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+
+// Output number of bytes for keccak-256 (32 bytes)
+const W: usize = 32;
+
+/// Keccak function that calls the keccakf machine
+/// Input is a byte array of arbitrary length and a delimiter byte
+/// Output is a byte array of length W
+pub fn keccak(data: &[u8], delim: u8) -> [u8; W] {
+    let mut b_input = [0u8; 200];
+    let mut b_output = [0u64; 25];
```
pacheco, line 58: Can both variables here be `[0u8; 200]`, and then use unsafe to cast as `[0u64; 25]` when calling keccakf?
I don't think there's any endianess issue here right?
Then, inside the loop, you could work with references to these and just swap them at each iteration... this would avoid the copying/conversions

riscv/src/runtime.rs
```diff
@@ -148,6 +148,29 @@ impl Runtime {
...
     }
 
+    pub fn with_keccak(mut self) -> Self {
+        self.add_submachine(
+            "std::machines::hash::keccakf::KeccakF",
+            None,
+            "keccakf",
+            [format!(
+                "instr keccakf ~ keccakf.keccakf {};", // operation
+                instr_register_params(0, 2, 2)         // 2 pointers as both input and output
```
pacheco, line 158: I think the `keccakf` has 2 inputs and 0 outputs (it won't write to any register, it only modifies memory)

riscv/src/runtime.rs
```diff
@@ -148,6 +148,29 @@ impl Runtime {
...
+            )],
+            0,
+            std::iter::once("keccakf;".to_string()) // must be called at least once
+                .chain((0..2).map(|i| format!("{} <=X= 0;", reg(i)))), // zero out output registers
```
pacheco, line 162: I think this code here needs to do the following (for now, as it won't work with continuations, per @georgwiese 's comment):
- point the first two registers to distinct memory addresses
- call `keccakf`
- zero out memory pointed out by second register (i.e., output written by keccakf)

std/machines/hash/keccakf.asm
```diff
@@ -0,0 +1,8 @@
+// TODO: placeholder for Rust wrapper
+machine KeccakF with
+    latch: LATCH,
+{ 
+    let x;
+    let y;
+    operation keccakf x -> y;
```
pacheco, line 7: i think this will be `keccakf x, y ->`, that is, it takes two memory addresses as input (the second one will be written to by the machine)

test_data/asm/keccakf.asm
```diff
@@ -163,6 +163,10 @@ let main: int, int[], int -> int[] = |W, input, delim| {
 };
 
 // main machine
-machine Main { 
+machine KeccakF with
+    latch: LATCH,
+{ 
     let x;
+    let y;
+    operation keccakf x -> y;
```
pacheco, line 171: same here

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+        asm!("ecall", in("a0") input as *const [u64; 25], in("a1") output as *mut [u64; 25], in("t0") u32::from(Syscall::KeccakF));
+    }
+}
+
+/// Calls the keccakf machine
+/// This function will panic if any of the u64 elements doesn't fit the Goldilocks field.
+pub fn keccakf(input: &[u64; 25], output: &mut [u64; 25]) {
+    for &n in input.iter() {
+        assert!(n < GOLDILOCKS);
+    }
```
georgwiese, line 46: I don't think that's needed!

I think this code has been copied from the Poseidon implementation. But in the case of Poseidon, the data (passed as `[u64; 12]` is interpreted as 12 field elements, so we have to make sure it is in the right range (otherwise, the hash of `[0, ...]` and `[0xffffffff00000001, ...]` would be the same!).

But in the case of Keccak, the input is just seen as a block of 1600 bits (represented as an array of 50 32-Bit words, so it can't overflow the ~64-Bit Goldilocks field).

riscv-runtime/src/hash.rs
georgwiese: `keccakf` could also just receive a byte array, right? I think the only reason `u64` is convenient is because of the assertion above, but this should be removed anyway (see comment above).

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+/// Input is a byte array of arbitrary length and a delimiter byte
+/// Output is a byte array of length W
+pub fn keccak(data: &[u8], delim: u8) -> [u8; W] {
+    let mut b_input = [0u8; 200];
+    let mut b_output = [0u64; 25];
+    let rate = 200 - (2 * W);
+    let mut pt = 0;
+
+    // update
+    for i in 0..data.len() {
```
georgwiese, line 63: Question to Rust / compiler experts: Would the compiler be smart enough here to know that it can do the XORs at the word level, not byte? Or would we be 4x more efficient by casting `data` to `&[usize]`?

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+        b_input[pt] = b_input[pt] ^ data[i];
+        pt = (pt + 1) % rate;
+        if pt == 0 {
+            keccakf(&from_bytes(b_input), &mut b_output);
+            b_input.copy_from_slice(&to_bytes(b_output));
```
georgwiese, line 68: We could get rid of this by having two buffers and changing the order every second call.

An alternative would be to pass only one pointer for input and output, changing the buffer in-place. But that would be more complicated to implement (we'd have to make sure that the writing happens *after* the reading, and not at the same time step). For the first version, I'd prefer using two buffers.

std/machines/hash/keccakf.asm
georgwiese: It also needs the current time step! The main machine already has a `STEP` fixed column that can be passed as an argument.

georgwiese (commented):
Very cool! :)

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+    keccakf_unsafe(input, output)
+}
+
+// Output number of bytes for keccak-256 (32 bytes)
+const W: usize = 32;
+
+/// Keccak function that calls the keccakf machine
+/// Input is a byte array of arbitrary length and a delimiter byte
+/// Output is a byte array of length W
+pub fn keccak(data: &[u8], delim: u8) -> [u8; W] {
```
chriseth, line 56: What is the purpose of `delim`? Should we fix it to the right value used in ethereum's keccak?

riscv-runtime/src/hash.rs
chriseth: I mean shouldn't we then just use `[u32; 50]` for the keccakf api?

riscv-runtime/src/hash.rs
chriseth: I think keccakf uses u64 / u32 in its API to have an endianness guarantee (and maybe also alignment).

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +27,85 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+    let mut b_output = [0u64; 25];
+    let rate = 200 - (2 * W);
+    let mut pt = 0;
+
+    // update
+    for i in 0..data.len() {
+        b_input[pt] = b_input[pt] ^ data[i];
+        pt = (pt + 1) % rate;
+        if pt == 0 {
+            keccakf(&from_bytes(b_input), &mut b_output);
```
chriseth, line 67: Can we avoid the from_bytes / to_bytes somehow?

riscv/src/runtime.rs
```diff
@@ -148,6 +148,29 @@ impl Runtime {
...
+        );
+
+        // The keccakf syscall has a two arguments passed on x10 and x11,
+        // the memory address of the 25 field element input array
```
chriseth, line 166: 50 field elements

riscv-runtime/src/hash.rs
qwang98: Done.

riscv/src/runtime.rs
qwang98: Done. Was a bit worried about syntax like `x, y ->` with no output, but guess that's fine.

riscv/src/runtime.rs
qwang98: What do you mean pointing the first two registers to distinct memory addresses? Do you mean randomized or I thought they will be specified by input arguments `x10` and `x11`?

std/machines/hash/keccakf.asm
qwang98: > i think this will be `keccakf x, y ->`, that is, it takes two memory addresses as input (the second one will be written to by the machine)

Agreed and done.

test_data/asm/keccakf.asm
qwang98: Done.

riscv-runtime/src/hash.rs
qwang98: Yes. Can do.

riscv-runtime/src/hash.rs
qwang98: It's a single byte xor'ed at the end of the input. Mind pointing me to the right value used in Ethereum's Keccak? (couldn't find it from https://github.com/ethereum/eth-hash/tree/main) I assume you mean a hardcoded value so the user don't need to provide it.

riscv-runtime/src/hash.rs
qwang98: I mean this only depends on the input and output format of the user, so we can avoid it if we only expose the user to provide u32 numbers.

riscv/src/runtime.rs
qwang98: Alrighty, I'll make all input and output u32 then.

std/machines/hash/keccakf.asm
qwang98: `instr keccakf X, Y -> ~ keccakf X, Y, STEP;`

qwang98: Main machine -> keccak .asm

qwang98:
Implement all comments. Ready for a final review.

Changes:
1. Rust wrapper takes two memory pointers and uses reinterpret casting.
2. Rust wrapper tested correct using tiny keccak instead of unfinished syscall, which can't be done yet.

Tested using: `cargo run -r --bin powdr-rs compile riscv/tests/riscv_data/keccak -o output && RUST_BACKTRACE=full cargo test --release  --package powdr-riscv --test riscv -- keccak --exact --show-output --ignored`

riscv/src/runtime.rs
```diff
@@ -148,6 +148,33 @@ impl Runtime {
         self.syscalls.contains_key(&s)
     }
 
+    pub fn with_keccak(mut self) -> Self {
+        self.add_submachine(
+            "std::machines::hash::keccakf::KeccakF",
+            None,
+            "keccakf",
+            [format!(
+                "instr keccakf {} ~ keccakf.keccakf {}, {}, STEP ->;", // instr keccakf X, Y -> ~ keccakf X, Y, STEP ->; (sub calls asm)
```
qwang98, line 157: Please make sure I'm doing this correctly.

std/machines/hash/keccakf.asm
```diff
@@ -0,0 +1,8 @@
+// TODO: placeholder for Rust wrapper
+machine KeccakF with
+    latch: LATCH,
+{ 
+    let x;
+    let y;
+    operation keccakf x, y ->;
+}
```
qwang98, line 8: The operation should really be `x, y, STEP ->`. Not sure how to "import" the `STEP` fixed column from the main machine here.

Doesn't really matter at this point though given it's a placeholder.

test_data/asm/keccakf.asm
```diff
@@ -163,6 +163,10 @@ let main: int, int[], int -> int[] = |W, input, delim| {
...
 
 // main machine
-machine Main { 
+machine KeccakF with
+    latch: LATCH,
+{ 
     let x;
+    let y;
+    operation keccakf x, y ->;
 }
```
qwang98, line 172: Same here.

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +29,69 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+}
+
+// Output number of bytes for keccak-256 (32 bytes)
+const W: usize = 32;
+
+/// Keccak function that calls the keccakf machine
+/// Input is a byte array of arbitrary length and a delimiter byte
+/// Output is a byte array of length W
+pub fn keccak(data: &[u8], delim: u8) -> [u8; W] {
+    let mut b_toggle = [[0u8; 200]; 2];
```
pacheco, line 63: I think this can be simpler (no need for the toggle)
```
    let mut b = [[0u8; 200]; 2];
    let [mut b_input, mut b_output] = &mut b;
...
    // then, to swap:
    core::mem::swap(&mut b_input, &mut b_output);
```

pacheco: then, if keccakf takes normal references (& / &mut) I think you can just do:
`keccakf(mem::transmute(&mut b_input), mem::transmute(&mut b_output));`

std/machines/hash/keccakf.asm
qwang98: Done.

test_data/asm/keccakf.asm
qwang98: Done.

riscv/tests/riscv_data/keccak/src/lib.rs
```diff
@@ -1,19 +1,38 @@
...
+//     hasher.finalize(&mut output);
+//     assert_eq!(
+//         output,
+//         [
+//             96, 41, 143, 120, 204, 11, 71, 23, 11, 167, 156, 16, 170, 56, 81, 215, 100, 139, 217,
+//             111, 47, 142, 70, 161, 157, 188, 119, 124, 54, 251, 12, 0,
+//         ],
+//     );
+// }
+
```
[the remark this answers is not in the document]
qwang98: Done.

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +29,69 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+    // TODO: uncomment this chunk after syscall implemented
+    // unsafe {
+        // // syscall inputs: memory pointer to input array and memory pointer to output array
+        // asm!("ecall", in("a0") input, in("a1") output, in("t0") u32::from(Syscall::KeccakF));
+    // }
+
+    // TODO: delete the following testing only chunk which uses tiny_keccak once syscall implemented
+    unsafe {
+        // Convert the input pointer to a mutable reference
+        let input_slice = &mut *(input as *mut [u64; 25]);
```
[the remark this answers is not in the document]
qwang98: Agree. This is just using tiny keccak as a test coverage though, which will be eventually removed.

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +29,69 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+/// Calls the keccakf machine
+/// Return value is placed in the output array.
+pub fn keccakf(input: *const [u64; 25], output: *mut [u64; 25]) {
+    // TODO: uncomment this chunk after syscall implemented
+    // unsafe {
+        // // syscall inputs: memory pointer to input array and memory pointer to output array
+        // asm!("ecall", in("a0") input, in("a1") output, in("t0") u32::from(Syscall::KeccakF));
+    // }
+
+    // TODO: delete the following testing only chunk which uses tiny_keccak once syscall implemented
```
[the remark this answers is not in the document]
qwang98: Done and commented out everything.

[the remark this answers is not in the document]
qwang98: > and this would be just `unimplemented` for now

I think the tiny keccak part is still needed just for the test in riscv.rs to pass.

riscv-runtime/src/hash.rs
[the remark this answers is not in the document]
qwang98: This part is now also sound after we changed the keccakf signature.

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +29,69 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
 
     poseidon_gl_unsafe(data)
 }
+
+/// Calls the keccakf machine
+/// Return value is placed in the output array.
+pub fn keccakf(input: *const [u64; 25], output: *mut [u64; 25]) {
```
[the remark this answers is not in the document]
qwang98: Yes and it's done.

qwang98:
All comments addressed and ready for review again. @pacheco

riscv/src/runtime.rs
```diff
@@ -148,6 +148,33 @@ impl Runtime {
...
+                instr_register_params(0, 2, 0), // one pointer for input and one pointer for output (both as input arguments)
+                reg(0),
+                reg(1),
+            )],
+            0,
+            std::iter::once(format!("{} <=X= 0x100;", reg(0))) // filler value for input pointer
+                .chain(std::iter::once(format!("{} <=X= 0x300;", reg(1)))) // filler value for output pointer (at least 200 bytes away)
+                .chain(std::iter::once("keccakf;".to_string())) // must be called at least once
+                .chain((0..50).flat_map(|i| store_word(&reg(1), i as u32 * 4, "x0"))), // zero out 200 bytes following output pointer
```
chriseth, line 166: Can you explain why this is needed? Also where do the hard-coded values `0x100` and `0x300` come from?

qwang98: This is the `init` call, which is required as all submachine must be called at least once. Hardcoded values are arbitrary input and output address values as long as they are at least 200 bytes away from each other, because the input and output are both 200 bytes. These are set up so that we can call `keccakf`. Finally we need to zero out the output values.

.gitignore
```diff
@@ -11,3 +11,6 @@ Cargo.lock
 
 riscv/runtime/target
 cargo_target/
+riscv-runtime/target
+riscv/tests/riscv_data/keccak/target
+output/
```
leonardoalt, line 16: Please remove, this is a user option

.gitignore
```diff
@@ -11,3 +11,6 @@ Cargo.lock
 
 riscv/runtime/target
```
leonardoalt, line 12: this can actually be removed

.gitignore
```diff
@@ -11,3 +11,6 @@ Cargo.lock
...
 cargo_target/
+riscv-runtime/target
+riscv/tests/riscv_data/keccak/target
```
leonardoalt, line 15:
```suggestion
riscv/tests/riscv_data/**/target
```

riscv-runtime/Cargo.toml
```diff
@@ -11,6 +11,7 @@ repository = "https://github.com/powdr-labs/powdr"
 serde = { version = "1.0", default-features = false, features = ["alloc", "derive", "rc"] }
 serde_cbor = { version = "0.11.2", default-features = false, features = ["alloc"] }
 powdr-riscv-syscalls = { path = "../riscv-syscalls" }
+tiny-keccak = { version = "2.0.2", features = ["keccak"] }
```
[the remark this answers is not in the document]
leonardoalt: I don't think this should be here. Why do we need this dependency here for tests?

riscv-runtime/src/hash.rs
leonardoalt: small comment but this sentence is still missing the ending period

riscv-runtime/src/hash.rs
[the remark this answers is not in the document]
leonardoalt: which test is that? The keccak test already imports tiny keccak itself, we shouldn't ship tiny keccak with the riscv-runtime crate

riscv-runtime/src/hash.rs
```diff
@@ -27,3 +29,63 @@ pub fn poseidon_gl(data: [u64; 12]) -> [u64; 4] {
...
+        output.copy_from_slice(&input_copy);
+    }
+}
+
+// Output number of bytes for keccak-256 (32 bytes)
+const W: usize = 32;
+
+/// Keccak function that calls the keccakf machine
+/// Input is a byte array of arbitrary length and a delimiter byte
+/// Output is a byte array of length W
```
leonardoalt, line 60: Please add periods at the end of sentences in all comments

riscv/src/runtime.rs
leonardoalt: what is this related to?

riscv/tests/riscv_data/keccak/src/lib.rs
```diff
@@ -1,19 +1,32 @@
 #![no_std]
 
-use tiny_keccak::{Hasher, Keccak};
+extern crate alloc;
```
leonardoalt, line 3: I don't think this test should be modified like this, I think we should add a new test instead. It's still good to have a test that uses native tiny keccak.

.gitignore
qwang98: Done

.gitignore
qwang98: Done

.gitignore
qwang98: Done

riscv-runtime/src/hash.rs
qwang98: Done

riscv-runtime/src/hash.rs
qwang98: Discussed and would be out of scope of this PR for now.

riscv/src/runtime.rs
qwang98: Ignore my comment for this. It should still be u64, but 25 field elements for the input and 25 field elements for the output, so a total of 50 field elements. I modified the comments to make this clear.

riscv-runtime/Cargo.toml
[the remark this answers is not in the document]
qwang98: This is because currently the keccak.asm isn't ready and depends on Chris's circuit compiler #1218 to be merged first, so we put a placeholder keccak.asm. To test the Rust wrapper, we instead use the tiny-keccak.

riscv-runtime/src/hash.rs
qwang98: Done

riscv/tests/riscv_data/keccak/src/lib.rs
qwang98: Alrighty. I'm restoring this file entirely and removed all tests using our Rust wrapper because it still depends on `keccak.asm` which is road blocked by Chris' circuit compiler.

I'm moving the tests for Rust wrapper to another draft PR, so that we can merge after all road blocks are cleared.

riscv-runtime/src/hash.rs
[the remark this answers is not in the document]
qwang98: It's the test here https://github.com/powdr-labs/powdr/blob/main/riscv/tests/riscv_data/keccak/src/lib.rs, which I'm now moving to a separate draft PR. I'm also leaving this as `unimplemented!` for now, and the contents here are also moved to the same draft PR.

riscv-runtime/Cargo.toml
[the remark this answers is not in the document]
qwang98: I'm removing this now and leaving the `keccakf` Rust wrapper as unimplemented.

qwang98:
Split roadblocked contents to #1408. This PR is ready for a final review. @leonardoalt

riscv/tests/riscv_data/keccak/Cargo.toml
```diff
@@ -8,5 +8,6 @@ edition = "2021"
 [dependencies]
 tiny-keccak = { version = "2.0.2", features = ["keccak"] }
 powdr-riscv-runtime = { path = "../../../../riscv-runtime" }
+hex-literal = "0.3.1"
```
leonardoalt, line 11: I think this can also go?

leonardoalt (commented):
Sorry for being pedantic, but I think we could still add another riscv test (like a sibling of riscv/tests/riscv_data/keccak) that does call the new syscall but has a dummy assert, just so that's "ready to go"

riscv/tests/riscv_data/keccak/Cargo.toml
qwang98: Done

qwang98:
> Sorry for being pedantic, but I think we could still add another riscv test (like a sibling of riscv/tests/riscv_data/keccak) that does call the new syscall but has a dummy assert, just so that's "ready to go"

Done. Note that the new syscall will return an error when run because it's unimplemented, so I left those out as comments that can be easily uncommented when we have the syscall ready (currently roadblocked by the circuit compiler).

qwang98:
@leonardoalt Ready for a final review.

riscv-syscalls/src/lib.rs
```diff
@@ -54,4 +54,6 @@ syscalls!(
     (5, EcAdd, "ec_add"),
     (6, EcDouble, "ec_double"),
     (7, KeccakF, "keccakf"),
+    (8, Mod256, "mod_256"),
+    (9, Halt, "halt"),
```
leonardoalt, line 58: These 2 seem unrelated to this PR?

riscv/src/runtime.rs
```diff
@@ -281,17 +281,17 @@ impl Runtime {
             "std::machines::hash::keccakf::KeccakF",
             None,
             "keccakf",
+            vec![],
             [format!(
-                "instr keccakf {} ~ keccakf.keccakf {}, {}, STEP ->;", // instr keccakf X, Y -> ~ keccakf X, Y, STEP ->; (sub calls asm)
-                instr_register_params(0, 2, 0), // one pointer for input and one pointer for output (both as input arguments)
+                "instr keccakf link ~> keccakf.keccakf({}, {}, STEP);",
```
leonardoalt, line 286: I think this needs to be changed again because of registers in memory. I think instruction `keccakf` should now receive register ids via assignment registers and then use links to load memory addresses from those addresses. The `poseidon` instruction works that way now, you can check what it does, it should be the same I think.

riscv/src/runtime.rs
```diff
@@ -281,17 +281,17 @@ impl Runtime {
...
-                reg(1),
+                reg(1)
             )],
-            0,
+            2,
             std::iter::once(format!("{} <=X= 0x100;", reg(0))) // filler value for input pointer
                 .chain(std::iter::once(format!("{} <=X= 0x300;", reg(1)))) // filler value for output pointer (at least 200 bytes away)
                 .chain(std::iter::once("keccakf;".to_string())) // must be called at least once
-                .chain((0..50).flat_map(|i| store_word(&reg(1), i as u32 * 4, "x0"))), // zero out 200 bytes following output pointer
+                .chain((0..50).flat_map(|i| store_word(11, i as u32 * 4, "x0"))), // zero out 200 bytes following output pointer
```
leonardoalt, line 294: this probably needs to change too

riscv-syscalls/src/lib.rs
qwang98: Yeah. These are just changes from merge main and I think it appears as additions because you set review display to the latest 6 commits. They won't appear as changes if you display all commits.

riscv/src/runtime.rs
qwang98: Done

riscv/src/runtime.rs
qwang98: Done

qwang98:
Ready for review again. @leonardoalt

riscv/src/runtime.rs
```diff
@@ -281,29 +281,69 @@ impl Runtime {
...
+                    tmp4_col * 4 = tmp2_col,
+                    // make sure the factors fit in 32 bits
+                    tmp3_col = X_b1 + X_b2 * 0x100 + X_b3 * 0x10000 + X_b4 * 0x1000000,
+                    tmp4_col = Y_b5 + Y_b6 * 0x100 + Y_b7 * 0x10000 + Y_b8 * 0x1000000
+                }
+            "#
+            .to_string()],
+            0,
+            std::iter::once("set_reg 10, 0x100;".to_string()) // filler value for input pointer
+                .chain(std::iter::once("set_reg 11, 0x300;".to_string())) // filler value for output pointer (at least 200 bytes away)
```
leonardoalt, line 301: what are these filler values actually?

qwang98: These are memory addresses for the registers used in the init call. They need to be at least 200 bytes apart as that's the length of the input (25 64-bit values) and the output (25 64-bit values).

leonardoalt: I think we could use the same address for input and output. Also, we need to change this for continuations, but fine for now, I'll approve.

leonardoalt (approved):

PR #1342, opened 2024-05-03
Rebase main into starky avx512

eigmax:


lvella:
Hi @eigmax, I took a look at the git graph comparing `main`, this branch and `avx` branch, and things are not ideal, because `main` changes have been squashed into `avx`. I think as long `avx` exists independently, we should use simple merges. This is probably the fault of how our repo is configured to disallow PRs with merge commits, but if you are okay with it, I'll fix it locally and make a push to the AVX branch.

PR #1343, opened 2024-05-03
Run tests on estark-polygon backend

lvella:


leonardoalt:
@lvella what's missing here?

lvella:
The tests run, but verification fails for some of them. I don't know why. Either we have not configured zkevm-prover correctly, it has a bug, or some of our compiled problems are inherently incompatible with it.

leonardoalt:
Closing for now.

Issue #1293, opened 2024-04-20
Add source location to expression


chriseth:
For improved error messages, the Expression type should have a source reference.

There are two options to implement that:
1) Add it as a field to every expression variant and have a trait that can be used to retrieve the source ref.
2) Add an intermediate type such that Expression is a struct that contains a source ref and the current Expression type.

I'm slightly leaning towards option 1, but am also open for option 2.

In the future, we might want to have different expression types that might contain more or less variants, but I think this is compatible with both ways to do it.

We might also want to have a source ref for AlgebraicExpression. For that, it might be better to implement option 1, so that the handling is uniform using the trait.

gzanitti:
I'm leaning towards option one as well. Even in that case, if Expression enumerates parsed expressions and the internal structs represent the expression itself, I would directly rewrite them as 

```
pub enum Expression<Ref = NamespacedPolynomialReference> {
    Reference(Ref, SourceRef),
    PublicReference(PublicRefence, SourceRef),
    Number(Number, SourceRef),
    String(String_, SourceRef),
    Tuple(Tuple, SourceRef),
    LambdaExpression(LambdaExpression<Self>, SourceRef),
    ArrayLiteral(ArrayLiteral<Self>, SourceRef),
    BinaryOperation(BinaryOperation<Self>, SourceRef),
    UnaryOperation(UnaryOperator<Self>, SourceRef),
    IndexAccess(IndexAccess<Self>),
    FunctionCall(FunctionCall<Self>),
    FreeInput(FreeInput<Self>, SourceRef),
    MatchExpression(MatchExpression<Self>, SourceRef),
    IfExpression(IfExpression<Self>, SourceRef),
    BlockExpression(BlockExpression<Self>, SourceRef),
}
```

where every entry (especially Number, String, BinaryOperation, etc) becomes
```
    ExpressionEnumerated(ExpressionStruct, SourceRef)
```

Let's take advantage of the fact that we are going to add breaking changes in the open PRs of everyone in the team anyway :joy: 

What do you think?

chriseth:
This has been implemented mostly.

PR #1345, opened 2024-05-03
Source location in expressions

gzanitti:
This PR solves issue #1293.
I am using the SourceRef defined in the AST, but it could be changed. It might even be more convenient to use Option<SourceRef> so that we don't have to use SourceRef::unknown() in some cases. Let me know what you think.

All comments are welcome :)


executor/src/constant_evaluator/mod.rs
```diff
@@ -69,10 +70,19 @@ fn generate_values<T: FieldElement>(
...
-                });
+                index_expr = Expression::IndexAccess(
+                    IndexAccess {
+                        array: e.clone().into(),
+                        index: Box::new(Expression::Number(
+                            Number {
+                                value: index.into(),
+                                type_: None,
+                            },
+                            SourceRef::unknown(),
```
gzanitti, line 81: I'm not sure about these two unknowns, should I take the value from e? Or does it not make sense in this context?

executor/src/constant_evaluator/mod.rs
```diff
@@ -69,10 +70,19 @@ fn generate_values<T: FieldElement>(
             };
             let index_expr;
             let e = if let Some(index) = index {
-                index_expr = Expression::IndexAccess(IndexAccess {
-                    array: e.clone().into(),
-                    index: Box::new(Expression::Number(index.into(), None)),
-                });
+                index_expr = Expression::IndexAccess(
+                    SourceRef::unknown(),
+                    IndexAccess {
```
chriseth, line 75: I'm wondering if it  makes sense to implement Into from IndexAccess to Expression, where the source ref is set to unknown (and of course for all other structs inside expression). Could that lead to us not setting the source ref even though it is known?

ast/src/parsed/mod.rs
```diff
@@ -332,29 +333,163 @@ impl<Expr> Children<Expr> for SelectedExpressions<Expr> {
...
-    IfExpression(IfExpression<Self>),
-    BlockExpression(Vec<StatementInsideBlock<Self>>, Box<Self>),
+    Reference(SourceRef, Ref),
+    PublicReference(SourceRef, String),
+    Number(SourceRef, Number),
+    String(SourceRef, String),
+    Tuple(SourceRef, Vec<Self>),
+    LambdaExpression(SourceRef, LambdaExpression<Self>),
+    ArrayLiteral(SourceRef, ArrayLiteral<Self>),
+    BinaryOperation(SourceRef, BinaryOperation<Self>),
```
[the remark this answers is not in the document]
gzanitti: Sure!

parser/src/lib.rs
```diff
@@ -249,20 +258,164 @@ mod test {
 
     fn pil_statement_clear_source_ref(stmt: &mut PilStatement) {
         match stmt {
-            PilStatement::Include(s, _)
```
chriseth, line 252: Could this function use Children trait?

parser/src/powdr.lalrpop
```diff
@@ -450,8 +450,8 @@ BoxedExpression: Box<Expression> = {
 }
 
 LambdaExpression: Box<Expression> = {
-    <kind:FunctionKind> "||" <body:BoxedExpression> => Box::new(Expression::LambdaExpression(LambdaExpression{kind, params: vec![], body})),
-    <kind:FunctionKind> "|" <params:ParameterList> "|" <body:BoxedExpression> => Box::new(Expression::LambdaExpression(LambdaExpression{kind, params, body})),
+    <start:@L> <kind:FunctionKind> "||" <body:BoxedExpression> => Box::new(Expression::LambdaExpression(ctx.source_ref(start), LambdaExpression{kind, params: vec![], body})),
```
chriseth, line 453: I think we should also store the end of the expression, but that can be done in another PR.

ast/src/parsed/mod.rs
```diff
@@ -332,29 +333,163 @@ impl<Expr> Children<Expr> for SelectedExpressions<Expr> {
...
+            (Expression::MatchExpression(_, a), Expression::MatchExpression(_, b)) => a == b,
+            (Expression::IfExpression(_, a), Expression::IfExpression(_, b)) => a == b,
+            (Expression::BlockExpression(_, a), Expression::BlockExpression(_, b)) => a == b,
+            _ => false,
+        }
+    }
+}
+
+pub trait SourceInfo {
+    fn get_source(&self) -> &SourceRef;
```
chriseth, line 380: Shouldn't this just be `source(&self)`? Also it's not really the source, it's a source reference, but maybe we can drop that distinction.

ast/src/parsed/mod.rs
```diff
@@ -332,29 +333,163 @@ impl<Expr> Children<Expr> for SelectedExpressions<Expr> {
...
+    fn get_source_mut(&mut self) -> &mut SourceRef;
+    fn set_source(&mut self, source: SourceRef);
```
chriseth, line 382: I think this is not used at all. I think in the rare situation where we need it, we can just use `*x.get_source_mut() = s;` (or `*x.source_mut()  = s;`

parser/src/powdr.lalrpop
```diff
@@ -514,7 +514,7 @@ BinaryAndOp: BinaryOperator = {
 }
 
 BitShift: Box<Expression> = {
-    BitShift BitShiftOp Sum => Box::new(Expression::BinaryOperation(<>)),
+    <start:@L> <left:BitShift> <op:BitShiftOp> <right:Sum> => Box::new(Expression::BinaryOperation(ctx.source_ref(start), BinaryOperation { left, op, right })),
```
chriseth, line 517: I'm wondering if we could simplify this into
```suggestion
    <start:@L> <left:BitShift> <op:BitShiftOp> <right:Sum> => Box::new((BinaryOperation { left, op, right }).into().with_source(ctx.source_ref(start))),
```

if we have Into for conversion between the Expression sub-structs into Expression (and unknown source) and then a generic function on Expression:

```
impl Expression {
  pub fn with_source(&mut self, s: SourceRef) -> Self {
    *(self.source_mut()) = s;
    self
  }
}
```

gzanitti:
Hey @chriseth, ended up separating all the structs in a set of PRs to make them easy to review (#1351 #1352 #1353 #1354 #1355). There is also a PR that adds the SourceRefs on top of them (#1357). I will close this.

Issue #1323, opened 2024-04-26
Move Halo2 crate to a module inside the backend crate


leonardoalt:

leonardoalt:
Once this is done, we can replace `String` by `BackendOptions` in Halo2

PR #1347, opened 2024-05-04
Move halo2 crate to a backend module

leonardoalt:
Fixes https://github.com/powdr-labs/powdr/issues/1323


backend/src/halo2/mock_prover.rs
```diff
@@ -0,0 +1,44 @@
...
+    mock_prover.verify().map_err(|errs| {
+        // Using err.emit() is the only way to get a pretty error message,
+        // so we print it here even though the error might be caught by the caller.
+        for err in errs {
+            err.emit(&mock_prover);
+            eprintln!();
+        }
+        "Circuit was not satisfied".to_string()
+    })
+}
```
leonardoalt, line 44: This file was entirely moved but I removed the mock prover tests because we already run all PIL tests with the mock prover.

backend/src/halo2/prover.rs
```diff
@@ -49,33 +51,6 @@ use std::{
...
-    fn from(s: String) -> Self {
-        match s.as_str() {
-            "" | "poseidon" => Self::Poseidon,
-            "snark_single" => Self::SnarkSingle,
-            "snark_aggr" => Self::SnarkAggr,
-            _ => panic!("Invalid proof type: {s}"),
-        }
-    }
-}
-
```
leonardoalt, line 78: This was moved into the new `halo2/src/mod.rs`

PR #1351, opened 2024-05-06
BinaryExpression struct in Expressions

gzanitti:
This PR is part of issue #1345.
In particular, it adds the struct BinaryOperation to Expressions to homogenise the structure before including source references.


ast/src/parsed/display.rs
```diff
@@ -679,7 +683,7 @@ impl<Ref: Display> Display for Expression<Ref> {
             Expression::Tuple(items) => write!(f, "({})", format_list(items)),
             Expression::LambdaExpression(lambda) => write!(f, "{lambda}"),
             Expression::ArrayLiteral(array) => write!(f, "{array}"),
-            Expression::BinaryOperation(left, op, right) => {
+            Expression::BinaryOperation(BinaryOperation { left, op, right }) => {
                 Expression::format_binary_operation(left, op, right, f)
```
chriseth, line 687: We should implement Display for BinaryOperation and "call" it here.

gzanitti: I think it's ready. I had to rearrange the new `precedence` function/trait a bit. When we move on to the PR of `unary operations`, I'll finish moving the remaining function (`format_unary_operation`). `Main` has already been merged, so the `Number` struct is already included in this PR as well.

Let me know what you think :)

ast/src/parsed/mod.rs
```diff
@@ -701,6 +720,22 @@ impl Precedence for BinaryOperator {
...
+
+impl<E> Precedence for Expression<E> {
+    fn precedence(&self) -> Option<ExpressionPrecedence> {
+        match self {
+            Expression::UnaryOperation(op, _) => op.precedence(),
+            Expression::BinaryOperation(BinaryOperation {
+                left: _,
+                op,
+                right: _,
+            }) => op.precedence(),
```
Schaeff, line 737:
```suggestion
            Expression::BinaryOperation(BinaryOperation {
                op,
                ..
            }) => op.precedence(),
```

or 

```suggestion
            Expression::BinaryOperation(operation) => operation.op.precedence(),
```

chriseth (approved):

PR #1355, opened 2024-05-06
BlockExpression struct in Expressions

gzanitti:
This PR is part of issue https://github.com/powdr-labs/powdr/pull/1345.
In particular, it adds the struct BlockExpression to Expressions to homogenize the structure before including source references.

PR #1354, opened 2024-05-06
MatchExpression struct in Expressions

gzanitti:
This PR is part of issue https://github.com/powdr-labs/powdr/pull/1345.
In particular, it adds the struct MatchExpression to Expressions to homogenise the structure before including source references.

PR #1353, opened 2024-05-06
Number struct in Expressions

gzanitti:
This PR is part of issue https://github.com/powdr-labs/powdr/pull/1345.
In particular, it adds the struct Number to Expressions to homogenize the structure before including source references.

PR #1352, opened 2024-05-06
UnaryExpression struct in Expressions

gzanitti:
This PR is part of issue https://github.com/powdr-labs/powdr/pull/1345.
In particular, it adds the struct UnaryOperation to Expressions to homogenise the structure before including source references.

PR #1357, opened 2024-05-07
SourceRef in expressions

gzanitti:
This PR solves issue https://github.com/powdr-labs/powdr/issues/1293.

Includes PRs that modify some Expression in order to match fields. It is recommended to review/merge those first: 
- #1351 
- #1352
- #1353
- #1354
- #1355

All comments are welcome :)


lvella:
Let's merge the dependency PRs before undrafting this one.

ast/src/parsed/mod.rs
```diff
@@ -332,24 +332,119 @@ impl<Expr> Children<Expr> for SelectedExpressions<Expr> {
...
+    IndexAccess(SourceRef, IndexAccess<Self>),
+    FunctionCall(SourceRef, FunctionCall<Self>),
+    FreeInput(SourceRef, Box<Self>),
+    MatchExpression(SourceRef, MatchExpression<Self>),
+    IfExpression(SourceRef, IfExpression<Self>),
+    BlockExpression(SourceRef, BlockExpression<Self>),
+}
+
+// Comparison function for expressions that ignore source information.
+impl<Ref: PartialEq> PartialEq for Expression<Ref> {
```
chriseth, line 356: Do you know where this is needed?

ast/src/parsed/mod.rs
```diff
@@ -332,24 +332,119 @@ impl<Expr> Children<Expr> for SelectedExpressions<Expr> {
...
+            ArrayLiteral(src, _) => *src = new_src,
+            BinaryOperation(src, _) => *src = new_src,
+            UnaryOperation(src, _) => *src = new_src,
+            IndexAccess(src, _) => *src = new_src,
+            FunctionCall(src, _) => *src = new_src,
+            FreeInput(src, _) => *src = new_src,
+            MatchExpression(src, _) => *src = new_src,
+            IfExpression(src, _) => *src = new_src,
+            BlockExpression(src, _) => *src = new_src,
+        }
```
chriseth: *self.source_reference_mut() = new_src;

(actually implement that in the trait)

ast/src/parsed/mod.rs
```diff
@@ -386,17 +481,25 @@ pub struct BinaryOperation<E = Expression<NamespacedPolynomialReference>> {
...
+            SourceRef::unknown(),
+            BinaryOperation {
+                left: Box::new(left),
+                op,
+                right: Box::new(right),
+            },
+        )
+    }
+
+    pub fn with_source_reference(&mut self, s: SourceRef) -> &mut Self {
```
chriseth, line 480: Shouldn't this take and return by value?

ast/src/parsed/mod.rs
gzanitti: I remember this particular case:

https://github.com/powdr-labs/powdr/blob/e2310b1b66a4c85fddee5d8d954a170f42e8b4b8/linker/src/lib.rs#L39

But also many matches against numbers started to fail in tests.

parser/src/powdr.lalrpop
```diff
@@ -450,8 +450,8 @@ BoxedExpression: Box<Expression> = {
 }
 
 LambdaExpression: Box<Expression> = {
-    <kind:FunctionKind> "||" <body:BoxedExpression> => Box::new(Expression::LambdaExpression(LambdaExpression{kind, params: vec![], body})),
-    <kind:FunctionKind> "|" <params:ParameterList> "|" <body:BoxedExpression> => Box::new(Expression::LambdaExpression(LambdaExpression{kind, params, body})),
+    <start:@L> <kind:FunctionKind> "||" <body:BoxedExpression> => Box::new(Expression::LambdaExpression(ctx.source_ref(start), LambdaExpression{kind, params: vec![], body})),
```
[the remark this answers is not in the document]
gzanitti: > Could be `LambdaExpression{kind, praams: vec![], body}.into().with_source_ref(ctx.source_ref(start))`

I tried to implement it that way, but Rust asks me to give a fully qualified path to specify the expected types, and the result is much uglier.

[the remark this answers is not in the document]
chriseth: What about writing a parser helper function?
```
fn to_expr_with_source_ref<T: Into<Expression>>(inner_expr: T, offset: usize) -> Box<Expression> {
  let mut e = Expression::from(inner_expr);
  e.set_source_ref(offset);
  Box::new(e)
}
```
Also if this is the only use of `set_source_ref`, I would prefer to inline its current definition, because we should not need to set the source ref that often, I think...

parser/src/lib.rs
```diff
@@ -43,6 +44,17 @@ impl ParserContext {
...
     }
+
+    pub fn to_expr_with_source_ref<T: Into<Expression>>(
+        &self,
+        inner_expr: T,
+        offset: usize,
+    ) -> Box<Expression> {
+        let mut expr = inner_expr.into();
+        let sr = self.source_ref(offset);
+        *expr.source_reference_mut() = sr;
```
chriseth, line 55: Why the local variable?

ast/src/parsed/display.rs
```diff
@@ -598,28 +598,37 @@ fn format_list<L: IntoIterator<Item = I>, I: Display>(list: L) -> String {
...
+                writeln!(f, "match {scrutinee} {{")?;
+                write_items_indented(f, arms)?;
+                write!(f, "}}")
             }
-            Expression::IfExpression(e) => write!(f, "{e}"),
-            Expression::BlockExpression(block_expr) => {
-                write!(f, "{block_expr}")
+            Expression::IfExpression(_, e) => write!(f, "{e}"),
+            Expression::BlockExpression(_, BlockExpression { statements, expr }) => {
+                if statements.is_empty() {
```
chriseth, line 624: Looks like an old version as well.

ast/src/parsed/mod.rs
```diff
@@ -332,24 +332,97 @@ impl<Expr> Children<Expr> for SelectedExpressions<Expr> {
...
+    BinaryOperation(SourceRef, BinaryOperation<Self>),
+    IndexAccess(SourceRef, IndexAccess<Self>),
+    FunctionCall(SourceRef, FunctionCall<Self>),
+    FreeInput(SourceRef, Box<Self>),
+    MatchExpression(SourceRef, MatchExpression<Self>),
+    IfExpression(SourceRef, IfExpression<Self>),
+    BlockExpression(SourceRef, BlockExpression<Self>),
+}
+
+// Comparison function for expressions that ignore source information.
```
chriseth, line 355:
```suggestion
/// Comparison function for expressions that ignore source information.
```

ast/src/parsed/mod.rs
```diff
@@ -332,24 +332,97 @@ impl<Expr> Children<Expr> for SelectedExpressions<Expr> {
...
+            (Expression::ArrayLiteral(_, a), Expression::ArrayLiteral(_, b)) => a == b,
+            (Expression::BinaryOperation(_, a), Expression::BinaryOperation(_, b)) => a == b,
+            (Expression::UnaryOperation(_, a), Expression::UnaryOperation(_, b)) => a == b,
+            (Expression::IndexAccess(_, a), Expression::IndexAccess(_, b)) => a == b,
+            (Expression::FunctionCall(_, a), Expression::FunctionCall(_, b)) => a == b,
+            (Expression::FreeInput(_, a), Expression::FreeInput(_, b)) => a == b,
+            (Expression::MatchExpression(_, a), Expression::MatchExpression(_, b)) => a == b,
+            (Expression::IfExpression(_, a), Expression::IfExpression(_, b)) => a == b,
+            (Expression::BlockExpression(_, a), Expression::BlockExpression(_, b)) => a == b,
+            _ => false,
```
chriseth, line 374: I think this could be quite dangerous. If we add something to Expression, we will probably not extend this implementation.

Is there a different way of doing this? We probably don't want to clone and set the source ref, but maybe some simple macro?

ast/src/parsed/mod.rs
```diff
@@ -332,24 +332,97 @@ impl<Expr> Children<Expr> for SelectedExpressions<Expr> {
...
+pub trait SourceReference {
+    fn source_reference(&self) -> &SourceRef;
+    fn source_reference_mut(&mut self) -> &mut SourceRef;
+}
+
+impl<E> SourceReference for Expression<E> {
+    fn source_reference(&self) -> &SourceRef {
+        use Expression::*;
+        match self {
+            Reference(src, _) => src,
```
chriseth, line 388: Maybe a macro would be useful here as well :)

ast/src/parsed/mod.rs
```diff
@@ -442,34 +512,25 @@ pub struct MatchExpression<E = Expression<NamespacedPolynomialReference>> {
 
 impl<Ref> From<MatchExpression<Expression<Ref>>> for Expression<Ref> {
     fn from(match_expr: MatchExpression<Expression<Ref>>) -> Self {
-        Expression::MatchExpression(match_expr)
+        Expression::MatchExpression(SourceRef::unknown(), match_expr)
     }
 }
 
-impl<E> Children<E> for MatchExpression<E> {
```
chriseth, line 449: This might also be a merge conflict resolution error.

ast/src/parsed/mod.rs
```diff
@@ -520,32 +581,40 @@ impl<R> Expression<R> {
...
-            Expression::Tuple(v) => v.iter(),
-            Expression::LambdaExpression(lambda) => lambda.children(),
-            Expression::ArrayLiteral(ArrayLiteral { items }) => items.iter(),
-            Expression::BinaryOperation(BinaryOperation { left, right, .. }) => {
+            Expression::Reference(_, _)
+            | Expression::PublicReference(_, _)
+            | Expression::String(_, _) => empty(),
+            Expression::Number(_, _) => empty(),
+            Expression::Tuple(_, v) => v.iter(),
+            Expression::LambdaExpression(_, LambdaExpression { body, .. }) => once(body.as_ref()),
```
chriseth, line 589: This should call children on the inner lambda expression

ast/src/parsed/mod.rs
```diff
@@ -559,32 +628,40 @@ impl<R> Expression<R> {
...
-            Expression::Tuple(v) => v.iter_mut(),
-            Expression::LambdaExpression(lambda) => lambda.children_mut(),
-            Expression::ArrayLiteral(ArrayLiteral { items }) => items.iter_mut(),
-            Expression::BinaryOperation(BinaryOperation { left, right, .. }) => {
+            Expression::Reference(_, _)
+            | Expression::PublicReference(_, _)
+            | Expression::String(_, _) => empty(),
+            Expression::Number(_, _) => empty(),
+            Expression::Tuple(_, v) => v.iter_mut(),
+            Expression::LambdaExpression(_, LambdaExpression { body, .. }) => once(body.as_mut()),
```
chriseth, line 636: Same here, also for the others.

ast/src/parsed/display.rs
gzanitti: Oh, sorry, I must have forgotten some branch of the match during the merge, how embarrassing. It won't happen again :sweat:

linker/src/lib.rs
```diff
@@ -230,13 +230,10 @@ mod test {
     };
     use powdr_number::{BigUint, FieldElement, GoldilocksField};
 
+    use crate::{link, DEFAULT_DEGREE};
```
chriseth, line 233: The changes in this file seem unrelated.

parser/src/test_utils.rs
```diff
@@ -20,6 +24,30 @@ pub fn pil_statement_clear_source_ref(stmt: &mut PilStatement) {
         | PilStatement::EnumDeclaration(s, _) => *s = SourceRef::unknown(),
     }
 }
+
+fn pil_expression_clear_source_ref(expr: &mut powdr_ast::parsed::Expression) {
+    use powdr_ast::parsed::Expression::*;
+    match expr {
```
chriseth, line 30: Can't you call `source_ref_mut()` here?

ast/src/parsed/mod.rs
gzanitti: I ended up implementing all the comparison inside a macro and using all the variants to make the match exhaustive (without using wildcards), in case we add a new variant to Expressions, the match will complain.

What do you think?

ast/src/parsed/mod.rs
```diff
@@ -353,29 +353,42 @@ pub enum Expression<Ref = NamespacedPolynomialReference> {
...
-            _ => false,
+macro_rules! impl_partial_eq_for_expression {
+    ($($variant:ident),*) => {
+        impl<Ref: PartialEq> PartialEq for Expression<Ref> {
+            fn eq(&self, other: &Self) -> bool {
+                match (self, other) {
+                    $(
+                        (Expression::$variant(_, a), Expression::$variant(_, b)) => a == b,
+                    )*
+                    // This catches the case where variants are different and returns false
```
chriseth, line 364: Nice!

ast/src/parsed/mod.rs
```diff
@@ -353,29 +353,42 @@ pub enum Expression<Ref = NamespacedPolynomialReference> {
...
+                    $(
+                        (Expression::$variant(_, _), _) => false,
+                    )*
+                }
+            }
         }
     }
 }
 
+impl_partial_eq_for_expression!(
```
chriseth, line 374: I think the macro I mentioned could be used here as well

chriseth (approved):

Issue #1360, opened 2024-05-08
Figure out why clippy does not warn on format!("-{}", x) any more.


chriseth:
Clippy used to warn that `format!("-{}", x)` should be replaced by `format!("-{x}")` but it does not do that any more. How can we re-enable it?

leonardoalt:
we're on 1.77 though

PR #1362, opened 2024-05-08
Re-enable non-inlined format args

chriseth:
Fixes https://github.com/powdr-labs/powdr/issues/1360


leonardoalt (approved):

leonardoalt:
some more missing looks like?

leonardoalt (approved):

PR #1364, opened 2024-05-08
Attempting to fake std unsing risc0 target.

lvella:
Should support std, but only if you don't use it.


lvella:
Superseded by #1641

PR #1366, opened 2024-05-10
Set syntax hilighting for asm and pil files.

chriseth:


georgwiese (approved):
Cool!

.gitattributes
```diff
@@ -0,0 +1,2 @@
+**/*.asm linguist-language=Rust
```
leonardoalt, line 1: for `asm` maybe `asm` is better?

PR #1368, opened 2024-05-10
Improve logging.

chriseth:


pipeline/src/pipeline.rs
```diff
@@ -705,13 +705,13 @@ impl<T: FieldElement> Pipeline<T> {
     }
 
     fn compute_analyzed_pil_from_parsed_pil_file(&mut self) -> Result<Analyzed<T>, Vec<String>> {
-        self.log("Analyzing pil...");
```
chriseth, line 708: There is tons of stuff that is not "analyzing pil" inside "compute_parsed_pil_file".

leonardoalt (approved):

PR #1370, opened 2024-05-10
Add degree checks.

chriseth:


std/machines/binary.asm
```diff
@@ -8,9 +8,6 @@ machine Binary with
     // Allow this machine to be connected via a permutation
     call_selectors: sel,
 {
-
-    // lower bound degree is 262144
```
chriseth, line 12: I'm counting 196608 - am I doing something wrong?

leonardoalt: This one I really thought was 2**18, but yea also don't see why. Maybe next power of 2

std/array.asm
```diff
@@ -29,6 +29,9 @@ let<T1, T2> fold: T1[], T2, (T2, T1 -> T2) -> T2 = |arr, initial, folder| std::u
 // use this function to flatten an array of arrays.
 let<T: Add + FromLiteral> sum: T[] -> T = |arr| fold(arr, 0, |a, b| a + b);
 
+/// Returns the product of the array elements.
+let<T: Mul + FromLiteral> product: T[] -> T = |arr| fold(arr, 0, |a, b| a * b);
```
leonardoalt, line 33:
```suggestion
let<T: Mul + FromLiteral> product: T[] -> T = |arr| fold(arr, 1, |a, b| a * b);
```
?

chriseth: Good one!

leonardoalt (approved):

PR #1377, opened 2024-05-14
Dual license

leonardoalt:
Followed https://github.com/arnavyc/dual-licensed-mit-apache

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,7 +1,5 @@
 MIT License
 
-Copyright (c) 2023,2024 powdr labs and contributors
-
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
 in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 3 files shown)


LICENSE-MIT
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 chriseth
+Copyright (c) 2023,2024 powdr labs and contributors
```
lvella, line 3: Will we remember to update this every year?

lvella (approved):

LICENSE-MIT
leonardoalt: idk, you suggested the year in the chat

lvella: I didn't make up the year in the copyright notice, it is already there in the previous version.

In theory it server to count when the copyright expires. In practice it is silly because this is publicly git versioned.

lvella: I am in favor of removing this line. The only question is if GitHub will still identify the license.

leonardoalt: we can try and fix otherwise

leonardoalt: done

chriseth:
Just for correctness sake, since my username is mentioned explicitly in the old licence document:

I explicitly approve this change.

PR #1390, opened 2024-05-21
Generic `write` syscall that can be read from the host

leonardoalt:


riscv-runtime/src/input.rs
```diff
@@ -1,45 +0,0 @@
-use core::arch::asm;
-
-extern crate alloc;
```
leonardoalt, line 3: This file became `io.rs` + new changes

pipeline/src/lib.rs
```diff
@@ -108,15 +170,20 @@ pub fn handle_simple_queries_callback<'a, T: FieldElement>() -> impl QueryCallba
...
-                        .map_err(|e| format!("Invalid char to print: {e}"))?
-                        as char
-                );
+            "Output" => {
+                assert_eq!(data.len(), 2);
+                let fd = data[0]
+                    .parse::<u32>()
+                    .map_err(|e| format!("Invalid fd: {e}"))?;
+                let byte = data[1]
+                    .parse::<u8>()
```
pacheco, line 179: I'm a bit confused, the syscall writes `u32`, how come here it is parse `u8` for the data?

pipeline/src/lib.rs
```diff
@@ -9,13 +9,75 @@ pub mod test_util;
...
 pub use pipeline::Pipeline;
 
 pub use powdr_backend::{BackendType, Proof};
 use powdr_executor::witgen::QueryCallback;
 
 use powdr_number::FieldElement;
 
+#[derive(Clone)]
+pub struct HostContext {
+    pub data: Arc<Mutex<BTreeMap<u32, Vec<u8>>>>,
```
pacheco, line 26: will the `Input` syscall also be modified to access this instead?
also, maybe rename `data` => `file_descriptors` or `file_data`?

pacheco: also, maybe add a small comment describing the purpose of this struct

pipeline/src/pipeline.rs
```diff
@@ -126,13 +126,19 @@ pub struct Pipeline<T: FieldElement> {
...
 }
 
+use super::HostContext;
+
 impl<T> Default for Pipeline<T>
 where
     T: FieldElement,
 {
     fn default() -> Self {
+        let ctx = HostContext::new();
```
pacheco, line 140: i was thinking, maybe the constructor could already return `(HostContext, QueryCallback)`, one doesn't make sense without the other, and the `query_callback` method won't ever be called more than once, i think.

riscv-runtime/src/io.rs
```diff
@@ -0,0 +1,67 @@
...
+
+pub fn read_slice(fd: u32, data: &mut [u32]) {
+    for (i, d) in data.iter_mut().enumerate() {
+        unsafe {
+            asm!("ecall", lateout("a0") *d, in("a0") fd, in("a1") (i+1) as u32, in("t0") u32::from(Syscall::DataIdentifier))
+        };
+    }
+}
+
+pub fn read_data_len(fd: u32) -> usize {
```
pacheco, line 26: maybe add some docs to these functions?

riscv/tests/riscv_data/output/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+    x: i32,
+    y: i32,
+}
+
+#[no_mangle]
+pub fn main() {
+    let input = read_word(0);
+
+    write_word(42, input);
+    write_slice(43, &[input, input * 2, input * 3]);
```
pacheco, line 17: nice!

std/prover.asm
```diff
@@ -7,8 +7,9 @@ let eval: expr -> fe = [];
 enum Query {
     /// Query a prover input element by index.
     Input(int),
-    /// Print a character on stdout.
-    PrintChar(int),
+	/// Writes a byte (second argument) to a file descriptor (first argument).
```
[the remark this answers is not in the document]
leonardoalt: yes

pipeline/src/lib.rs
leonardoalt: this data here is only to print chars, so this must be a char

pipeline/src/lib.rs
leonardoalt: > will the Input syscall also be modified to access this instead?

Yea that's a good question, I'm not sure yet.

leonardoalt:
Fixed all review comments

pipeline/src/lib.rs
leonardoalt: the point is that the main ways we use this are to either print it to stdout, or write into a Deserializable struct, so in practice we always use u8. We can change it to make it more granular, but I'd do it in a follow-up PR

pacheco: I mean, why isn't the method simply called `write_u8`?

leonardoalt: yea good point. I'll make the distinction more precise

leonardoalt: pushed

riscv-runtime/src/io.rs
```diff
@@ -0,0 +1,70 @@
...
+/// Reads the length of the data first at index 0, then the data itself.
+pub fn read_data_len(fd: u32) -> usize {
+    let mut out: u32;
+    unsafe {
+        asm!("ecall", lateout("a0") out, in("a0") fd, in("a1") 0, in("t0") u32::from(Syscall::DataIdentifier))
+    };
+    out as usize
+}
+
+/// Writes a single u32 to the file descriptor fd.
```
pacheco, line 37: comment was missed

riscv-runtime/src/io.rs
```diff
@@ -0,0 +1,70 @@
...
+pub fn write_u8(fd: u32, w: u8) {
```
pacheco, line 38: very minor, but `w` was from "word" right?

leonardoalt: yep, changed it to `byte`

riscv-runtime/src/io.rs
leonardoalt: fixed

pacheco (dismissed):

pacheco (approved):

PR #1393, opened 2024-05-21
Allow the type checker to accept empty values

gzanitti:


chriseth:
Oh I wouldn't introduce a new type, this should just be the empty tuple.

chriseth:
I think we need to change the interaction between the type inference and the side effect checker a little.

Because now, statements without assignment are valid everywhere, so we cannot really make this an error in the side effect checker in non-constr functions any more (so the `"Tried to add a constraint in a {} context: {expr}",`-error needs to go without direct replacement).

Then I think we need to change `statement_type` to something like `constr_function_statement_type` (and keep it at the type passed in from outside).

Now whenever we enter the body of a LambdaExpression in type inference, we need to set a flag depending on its `kind`: When it is `constr`, we allow both the empty tuple and `constr_function_statement_type`. When it is anything else, we only allow the empty tuple. At the very outside of type inference, we start out as if it was a `constr` function.

pil-analyzer/src/type_unifier.rs
```diff
@@ -116,6 +119,13 @@ impl Unifier {
                     .zip(args2)
                     .try_for_each(|(a1, a2)| self.unify_types(a1, a2))
             }
+            (Type::Array(ArrayType { length: None, .. }), Type::Tuple(a2))
```
chriseth, line 122: I think we don't need this, at least it should not be handled in the unifier.

pil-analyzer/tests/condenser.rs
```diff
@@ -65,7 +65,7 @@ fn new_witness_column_name_clash() {
 #[test]
 fn create_constraints() {
     let input = r#"namespace N(16);
-    let force_bool: expr -> Constr = |c| c * (1 - c) = 0;
+    let force_bool: expr -> () = |c| c * (1 - c) = 0;
```
chriseth, line 68: This should stay as it is, it returns the constraint and does not add it to the global set.

pil-analyzer/tests/parse_display.rs
```diff
@@ -315,16 +315,16 @@ fn function_type_display() {
 #[test]
 fn expr_and_identity() {
     let input = r#"namespace N(16);
-    let f: expr, expr -> Constr[] = |x, y| [x = y];
```
chriseth, line 318: All these should also not be modified.

std/check.asm
```diff
@@ -9,4 +9,4 @@ let panic: string -> ! = [];
 /// prover-internal consistency.
 /// The panic message is obtained by calling the function `reason`.
 /// Returns an empty array on success, which allows it to be used at statement level.
-let assert: bool, (-> string) -> Constr[] = |condition, reason| if !condition { panic(reason()) } else { [] };
\ No newline at end of file
+let assert: bool, (-> string) -> () = |condition, reason| if !condition { panic(reason()) } else { [] };
```
chriseth, line 12:
```suggestion
let assert: bool, (-> string) -> () = |condition, reason| if !condition { panic(reason()) } else { () };
```

std/utils.asm
```diff
@@ -29,10 +29,10 @@ let<T> unwrap_or_else: Option<T>, (-> T) -> T = |o, f| match o {
 /// `c` is allowed to change.
 ///
 /// Note: `latch` needs to be equal to `1`, and not merely non-zero for `c` to be able to change.
-let unchanged_until: expr, expr -> Constr = |c, latch| (c' - c) * (1 - latch) = 0;
```
chriseth, line 32: These should stay as they are as well.

pil-analyzer/src/type_unifier.rs
gzanitti: Yes, I know, it was just a little hack to better understand what was going on :+1:

pil-analyzer/src/type_inference.rs
```diff
@@ -688,16 +705,60 @@ impl<'a> TypeChecker<'a> {
...
+                        .map_err(|err| {
+                            format!(
+                                "Error checking expression {expr}:\nExpected type: {}\nInferred type: {}\n{err}",
+                                self.type_into_substituted(expected_type.clone()),
+                                self.type_into_substituted(inferred_type.clone())
+                            )
+                        })
+                } else {
+                    Err(format!(
+                        "Error Constr type inside Pure/Query function {expr}:\nExpected type: {}\nInferred type: {}\n",
```
chriseth, line 739: I’m not sure I understand this correctly, but the type checker should not care about the Constr type. It is totally fine to use that type in any function. One thing we need to make sure is that is that the type of a statement can only be the empty tuple, except inside functions of kind constr, where it is also allowed to have the type passed in from the pil analyser (which is Constr, but the type inference does not know of care about the concrete type)

pil-analyzer/src/type_inference.rs
```diff
@@ -456,17 +457,21 @@ impl<'a> TypeChecker<'a> {
...
+            self.expect_type(&expected_type.ty, expr, None)
         }
     }
 
     /// Process an expression and return the type of the expression.
-    fn infer_type_of_expression(&mut self, e: &mut Expression) -> Result<Type, String> {
+    fn infer_type_of_expression(
+        &mut self,
+        e: &mut Expression,
+        kind: Option<FunctionKind>,
```
chriseth, line 473: I think it would be better to store that inside TypeChecker instead of dragging it around all the time. We also store local variable types, and this is kind of similar.

pil-analyzer/src/type_inference.rs
```diff
@@ -74,6 +76,7 @@ impl<'a> TypeChecker<'a> {
             declared_type_vars: Default::default(),
             unifier: Default::default(),
             last_type_var: Default::default(),
+            lambda_kind: None,
```
chriseth, line 79: At the outer context, it should be  `constr`

pil-analyzer/src/type_inference.rs
```diff
@@ -526,21 +529,17 @@ impl<'a> TypeChecker<'a> {
...
                 let old_len = self.local_var_types.len();
                 let result = params
                     .iter()
                     .map(|p| self.infer_type_of_pattern(p))
                     .collect::<Result<Vec<_>, _>>()
                     .and_then(|param_types| {
-                        Ok((param_types, self.infer_type_of_expression(body)?))
+                        self.lambda_kind = Some(*kind);
+                        let body_type = self.infer_type_of_expression(body)?;
+                        self.lambda_kind = None;
```
chriseth, line 541: lambda expressions can be nested, so this should restore it to the previous value (just like with `local_var_types`)

pil-analyzer/src/type_inference.rs
```diff
@@ -624,22 +623,59 @@ impl<'a> TypeChecker<'a> {
...
+                };
                 for statement in statements {
                     match statement {
                         StatementInsideBlock::LetStatement(LetStatementInsideBlock {
                             pattern,
                             value,
                         }) => {
                             let value_type = if let Some(value) = value {
-                                self.infer_type_of_expression(value)?
+                                let statement_type = self.infer_type_of_expression(value)?;
```
chriseth, line 638: Wait I just realize that this is the case where we have `let pattern = value;`. This means there should not be any change at all. We only need a change for `value;`-statements. And I think without a grammar change, they can only occur at the top level currently.

pil-analyzer/src/type_inference.rs
```diff
@@ -624,22 +623,59 @@ impl<'a> TypeChecker<'a> {
...
+                        StatementInsideBlock::Expression(expr) => match self.lambda_kind {
+                            Some(FunctionKind::Constr) => {
+                                self.expect_type_with_flexibility(&self.statement_type, expr).or_else(|err| {
+                                    self.expect_type_with_flexibility(&expected_empty_tuple, expr).map_err(|_err2| {
+                                        format!("Invalid expression ({expr}) inside Constr function:\n{err}")
+                                    })
+                                })?;
+                            }
+                            _ => {
+                                self.expect_type_with_flexibility(&expected_empty_tuple, expr).map_err(|err| {
```
chriseth, line 674: We don't need `_with_flexibility` here.

pil-analyzer/src/type_inference.rs
```diff
@@ -624,22 +623,59 @@ impl<'a> TypeChecker<'a> {
...
                                 Type::Expr
                             };
                             self.expect_type_of_pattern(&value_type, pattern)?;
                         }
-                        StatementInsideBlock::Expression(expr) => {
-                            self.expect_type_with_flexibility(self.statement_type, expr)?;
-                        }
+                        StatementInsideBlock::Expression(expr) => match self.lambda_kind {
+                            Some(FunctionKind::Constr) => {
+                                self.expect_type_with_flexibility(&self.statement_type, expr).or_else(|err| {
```
chriseth, line 667: I don't feel comfortable with recovering from errors inside the type inference. It might be that type unification did some changes that we then would have to undo. It's better to infer the type and if it is the empty tuple, accept that, otherwise call `expect...` for the other type.

chriseth:
In the loop, we say that the expressions can only be `Constr` or `Constr[]`. We should also allow `()`.

pil-analyzer/src/type_inference.rs
```diff
@@ -456,8 +463,8 @@ impl<'a> TypeChecker<'a> {
                 })
                 .map_err(|err| {
                     format!(
-                        "Expression is expected to evaluate to {} or ({})[]:\n  {expr}:\n{err}",
```
chriseth, line 459: Are you sure about this change?

pil-analyzer/src/type_inference.rs
```diff
@@ -638,17 +643,70 @@ impl<'a> TypeChecker<'a> {
...
     }
 
+    fn check_kind_in_expression(
+        &mut self,
+        expr: &mut powdr_ast::parsed::Expression<Reference>,
+    ) -> Result<(), String> {
+        let empty_tuple = Type::Tuple(TupleType { items: vec![] });
+        match self.lambda_kind {
+            FunctionKind::Constr => {
+                if self.expect_type(&self.statement_type.ty, expr).is_ok() {
```
chriseth, line 665: If possible, call `infer_type` and then check if the type equals `()`, otherwise try the other type

pil-analyzer/src/type_inference.rs
```diff
@@ -420,7 +425,18 @@ impl<'a> TypeChecker<'a> {
         expressions: &mut [(&mut Expression, ExpectedType)],
     ) -> Result<(), String> {
         for (e, expected_type) in expressions {
-            self.expect_type_with_flexibility(expected_type, e)?;
+            let result = self.expect_type_with_flexibility(expected_type, e);
```
chriseth, line 428: I think it would be better to do this change inside `expect_type_with_flexibility`.

As I said in other comments: Once type checking fails, you cannot re-check the same expression with a different expected type because it has run unification which adds substitutions and trait constraints on types and so on. If you then re-run it with a different expectation, these substitutions are still there.

pil-analyzer/src/type_inference.rs
```diff
@@ -638,17 +652,78 @@ impl<'a> TypeChecker<'a> {
                             self.expect_type_of_pattern(&value_type, pattern)?;
                         }
                         StatementInsideBlock::Expression(expr) => {
-                            self.expect_type_with_flexibility(self.statement_type, expr)?;
+                            self.check_kind_in_expression(expr)?
```
chriseth, line 655: I think this should just be
```suggestion
                            self.expect_type_with_flexibility(self.statement_type(), expr)?;
```
where the `statement_type` function returns the correct expected type depending on the current context / function kind.

pil-analyzer/src/type_inference.rs
```diff
@@ -638,17 +652,78 @@ impl<'a> TypeChecker<'a> {
...
                         }
                     }
                 }
                 let result = self.infer_type_of_expression(expr);
+                let result = self.check_kind(result?);
```
chriseth, line 660: I don't think this line should be here.

pil-analyzer/src/type_inference.rs
```diff
@@ -649,6 +664,81 @@ impl<'a> TypeChecker<'a> {
...
+            self.constr_function_statement_type.clone()
+        } else {
+            ExpectedType {
+                ty: Type::Tuple(TupleType { items: vec![] }),
+                allow_array: false,
+            }
+        }
+    }
+
+    fn check_kind_and_expression(
```
gzanitti, line 678: I think it's probably cleaner to separate this function into two: One for expressions at the end of blocks, and one for isolated expressions (which can be LambaExpressions with their own kind).

pil-analyzer/src/type_inference.rs
```diff
@@ -667,6 +672,18 @@ impl<'a> TypeChecker<'a> {
...
+    fn statement_type(&self) -> ExpectedType {
+        if self.lambda_kind == FunctionKind::Constr {
+            self.constr_function_statement_type.clone()
+        } else {
+            ExpectedType {
+                ty: Type::Tuple(TupleType { items: vec![] }),
+                allow_array: false,
+                allow_empty: false,
+            }
+        }
```
chriseth, line 684:
```suggestion
            Type::Tuple(TupleType { items: vec![] }).into()
```

pil-analyzer/tests/types.rs
```diff
@@ -516,3 +516,46 @@ fn enum_too_many_fields() {
     ";
     type_check(input, &[]);
 }
+
+#[test]
```
chriseth, line 520: These don't really test statements, do they?

pil-analyzer/src/type_inference.rs
```diff
@@ -667,6 +672,18 @@ impl<'a> TypeChecker<'a> {
         })
     }
 
+    fn statement_type(&self) -> ExpectedType {
```
chriseth, line 675:
```suggestion
    /// Returns the type expected at statement level, given the current function context.
    fn statement_type(&self) -> ExpectedType {
```

pil-analyzer/src/type_inference.rs
```diff
@@ -40,20 +40,22 @@ pub fn infer_types(
...
         ExpectedType {
             ty,
             allow_array: false,
+            allow_empty: false,
         }
     }
 }
 
 struct TypeChecker<'a> {
     /// The expected type for expressions at statement level in block expressions.
```
chriseth, line 57:
```suggestion
    /// The expected type for expressions at statement level in block expressions inside a constr function.
```

pil-analyzer/src/type_inference.rs
```diff
@@ -40,20 +40,22 @@ pub fn infer_types(
 pub struct ExpectedType {
     pub ty: Type,
     pub allow_array: bool,
+    pub allow_empty: bool,
```
chriseth, line 43:
```suggestion
    /// If true, arrays of `ty` are also allowed.
    pub allow_array: bool,
    /// If true, the empty tuple is also allowed.
    pub allow_empty: bool,
```

pil-analyzer/src/type_inference.rs
```diff
@@ -470,10 +472,15 @@ impl<'a> TypeChecker<'a> {
...
-        } else {
-            self.expect_type(&expected_type.ty, expr)
-        }
+            } else {
+                let expected_type = if ty == Type::Tuple(TupleType { items: vec![] }) {
+                    Type::Tuple(TupleType { items: vec![] })
+                } else {
+                    expected_type.ty.clone()
+                };
+                self.expect_type(&expected_type, expr)
```
chriseth, line 481: I think we shouldn't call expect_type after calling infer_type on the same expression.

Essentially I would say this should be changed to

```
        self.infer_type_of_expression(expr).and_then(|ty| {
            let ty = self.type_into_substituted(ty);
            let expected_type = if expected_type.allow_array && matches!(ty, Type::Array(_)) {
              ...
            } else if expected_type.allow_empty && ty == Type::Tuple(...)) {
              ...
            } else {
              expected_type.ty.clone()
            };
            self.unifier.unify_types(...)
        })
```

pil-analyzer/src/condenser.rs
```diff
@@ -238,7 +238,14 @@ impl<'a, T: FieldElement> Condenser<'a, T> {
         if identity.kind == IdentityKind::Polynomial {
             let expr = identity.expression_for_poly_id();
             evaluator::evaluate(expr, self)
-                .and_then(|expr| self.add_constraints(expr, identity.source.clone()))
+                .and_then(|expr| {
+                    if let Value::Tuple(items) = expr.as_ref() {
+                        assert_eq!(items.len(), 0);
```
chriseth, line 243:
```suggestion
                        assert_eq!(items.is_empty());
```

pil-analyzer/src/type_inference.rs
```diff
@@ -441,31 +448,29 @@ impl<'a> TypeChecker<'a> {
...
-                    Type::Array(ArrayType {
-                        base: Box::new(expected_type.ty.clone()),
-                        length: None,
-                    })
-                } else {
-                    expected_type.ty.clone()
-                };
+        self.infer_type_of_expression(expr).and_then(|ty| {
+            let ty = self.type_into_substituted(ty);
+            let expected_type = if expected_type.allow_array & matches!(ty, Type::Array(_)) {
```
chriseth, line 453:
```suggestion
            let expected_type = if expected_type.allow_array && matches!(ty, Type::Array(_)) {
```

pil-analyzer/src/type_inference.rs
```diff
@@ -441,31 +448,29 @@ impl<'a> TypeChecker<'a> {
...
+                Type::Array(ArrayType {
+                    base: Box::new(expected_type.ty.clone()),
+                    length: None,
+                })
+            } else if expected_type.allow_empty & (ty == Type::Tuple(TupleType { items: vec![] })) {
```
chriseth, line 458:
```suggestion
            } else if expected_type.allow_empty && (ty == Type::Tuple(TupleType { items: vec![] })) {
```

pil-analyzer/src/type_inference.rs
```diff
@@ -441,31 +448,29 @@ impl<'a> TypeChecker<'a> {
...
-                let ty = self.type_into_substituted(ty);
-                let expected_type = if matches!(ty, Type::Array(_)) {
-                    Type::Array(ArrayType {
-                        base: Box::new(expected_type.ty.clone()),
-                        length: None,
-                    })
-                } else {
-                    expected_type.ty.clone()
-                };
+        self.infer_type_of_expression(expr).and_then(|ty| {
```
chriseth, line 451: I think this could now also just be `let ty = self.infer_type_of_expression(expr)?;`

pil-analyzer/src/type_inference.rs
```diff
@@ -448,14 +448,31 @@ impl<'a> TypeChecker<'a> {
         expected_type: &ExpectedType,
         expr: &mut Expression,
     ) -> Result<(), Error> {
+        if let Expression::Number(
```
chriseth, line 451: Can you pull that into a helper function that is also called from expect_type?

chriseth (approved):

PR #1394, opened 2024-05-23
profiler for powdr asm compiled from rust

pacheco:
Rust profiler (#1174) implemented by tracking the execution of the `riscv-executor`.

### DIFF
--- a/cli-rs/src/main.rs
+++ b/cli-rs/src/main.rs
@@ -235,13 +235,8 @@ fn run_command(command: Commands) {
             generate_flamegraph,
             generate_callgrind,
         } => {
-            call_with_field!(execute::<field>(
-                Path::new(&file),
-                split_inputs(&inputs),
-                Path::new(&output_directory),
-                continuations,
-                witness,
-                ProfilerOptions {
+            let profiling = if generate_callgrind || generate_flamegraph {
+                Some(ProfilerOptions {
                     file_stem: Path::new(&file)
                         .file_stem()
                         .and_then(OsStr::to_str)
@@ -249,7 +244,17 @@ fn run_command(command: Commands) {
                     output_directory: output_directory.clone(),
                     flamegraph: generate_flamegraph,
                     callgrind: generate_callgrind,
-                }
+                })
+            } else {
+                None
+            };
+            call_with_field!(execute::<field>(
+                Path::new(&file),
+                split_inputs(&inputs),
+                Path::new(&output_directory),
+                continuations,
+                witness,
+                profiling
             ))
         }
     };
@@ -320,7 +325,7 @@ fn execute<F: FieldElement>(
     output_dir: &Path,
     continuations: bool,
     witness: bool,
-    profiler_opt: ProfilerOptions,
+    profiling: Option<ProfilerOptions>,
 ) -> Result<(), Vec<String>> {
     let mut pipeline = Pipeline::<F>::default()
         .from_file(file_name.to_path_buf())
@@ -328,7 +333,7 @@ fn execute<F: FieldElement>(
 
     let bootloader_inputs = if continuations {
         pipeline = pipeline.with_prover_inputs(inputs.clone());
-        powdr_riscv::continuations::rust_continuations_dry_run(&mut pipeline, profiler_opt.clone())
+        powdr_riscv::continuations::rust_continuations_dry_run(&mut pipeline, profiling.clone())
     } else {
         vec![]
     };
@@ -351,7 +356,7 @@ fn execute<F: FieldElement>(
                 pipeline.data_callback().unwrap(),
                 &[],
                 powdr_riscv_executor::ExecMode::Fast,
-                profiler_opt,
+                profiling,
             );
             log::info!("Execution trace length: {}", trace.len);
         }
--- a/cli-rs/src/main.rs
+++ b/cli-rs/src/main.rs
@@ -247,8 +247,8 @@ fn run_command(command: Commands) {
                         .and_then(OsStr::to_str)
                         .map(String::from),
                     output_directory: output_directory.clone(),
-                    generate_flamegraph,
-                    generate_callgrind,
+                    flamegraph: generate_flamegraph,
+                    callgrind: generate_callgrind,
                 }
             ))
         }
--- a/cli-rs/src/main.rs
+++ b/cli-rs/src/main.rs
@@ -9,7 +9,9 @@ use log::LevelFilter;
 
 use powdr_number::{BigUint, Bn254Field, FieldElement, GoldilocksField};
 use powdr_pipeline::Pipeline;
+use powdr_riscv_executor::ProfilerOptions;
 
+use std::ffi::OsStr;
 use std::io;
 use std::{borrow::Cow, io::Write, path::Path};
 use strum::{Display, EnumString, EnumVariantNames};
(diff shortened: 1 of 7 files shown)


riscv-executor/src/lib.rs
```diff
@@ -1073,6 +1138,7 @@ pub fn execute<F: FieldElement>(
     inputs: &Callback<F>,
     bootloader_inputs: &[Elem<F>],
     mode: ExecMode,
+    profiler_opt: ProfilerOptions,
```
leonardoalt, line 1141: Should this maybe be an Option?

riscv-executor/src/lib.rs
```diff
@@ -979,20 +1003,33 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
     }
 }
 
+/// return true if the expression is a jump instruction
+fn expr_is_jump(e: &Expression) -> bool {
```
leonardoalt, line 1007:
```suggestion
fn is_jump(e: &Expression) -> bool {
```
arg is already Expr

cli-rs/src/main.rs
```diff
@@ -122,6 +124,16 @@ enum Commands {
         #[arg(short, long)]
         #[arg(default_value_t = false)]
         witness: bool,
+
+        /// Generate a flamegraph plot of the execution ("[file].svg")
+        #[arg(long)]
+        #[arg(default_value_t = false)]
+        generate_flamegraph: bool,
```
leonardoalt, line 131:
```suggestion
        flamegraph: bool,
```
maybe simpler?

cli-rs/src/main.rs
```diff
@@ -122,6 +124,16 @@ enum Commands {
...
+
+        /// Generate callgrind file of the execution ("[file].callgrind")
+        #[arg(long)]
+        #[arg(default_value_t = false)]
+        generate_callgrind: bool,
```
leonardoalt, line 136:
```suggestion
        callgrind: bool,
```
maybe simpler?

riscv-executor/src/profiler.rs
```diff
@@ -0,0 +1,346 @@
...
+
+use rustc_demangle::demangle;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Call<'a> {
+    from: Loc<'a>,
+    target: Loc<'a>,
+}
+
+/// Risc-v asm profiler.
```
leonardoalt, line 19:
```suggestion
/// RISCV asm profiler.
```

riscv-executor/src/profiler.rs
```diff
@@ -0,0 +1,346 @@
...
+/// Tracks the self-cost of functions and cummulative cost specific function calls (i.e., callgrind style).
```
leonardoalt, line 20:
```suggestion
/// Tracks the self-cost of functions and the cumulative cost of specific function calls (i.e., callgrind style).
```
?

riscv-executor/src/profiler.rs
```diff
@@ -0,0 +1,346 @@
...
+    location_stats: BTreeMap<Loc<'a>, usize>,
+    /// (count, cummulative cost) of calls
+    call_stats: BTreeMap<Call<'a>, (usize, usize)>,
+    /// stack sampling format for FlameGraph
+    folded_stack_stats: BTreeMap<Vec<&'a str>, usize>,
+}
+
+#[derive(Default, Clone)]
+pub struct ProfilerOptions {
+    pub output_directory: String,
```
leonardoalt, line 44: In other crates we try to keep FS access to a minimal, and rather return results that CLI then writes to the FS. I guess we could do the same here without really disrupting the architecture?

leonardoalt: or you could take an `impl Write` here instead of the file path

riscv-executor/src/profiler.rs
```diff
@@ -0,0 +1,346 @@
...
+        }
+    }
+
+    /// profiling only starts once "__runtime_start" is reached
+    pub fn running(&self) -> bool {
+        !self.call_stack.is_empty()
+    }
+
+    /// function at the top of the call stack
+    pub fn curr_function(&self) -> Option<&'a str> {
```
leonardoalt, line 185:
```suggestion
    pub fn current_function(&self) -> Option<&'a str> {
```

riscv-executor/src/profiler.rs
```diff
@@ -0,0 +1,346 @@
...
+            self.write_flamegraph(&path);
+        }
+        if self.options.generate_callgrind {
+            path.set_extension("callgrind");
+            self.write_callgrind(&path);
+        }
+    }
+
+    /// profiling only starts once "__runtime_start" is reached
+    pub fn running(&self) -> bool {
```
leonardoalt, line 180:
```suggestion
    pub fn is_running(&self) -> bool {
```

riscv-executor/src/profiler.rs
```diff
@@ -0,0 +1,346 @@
...
+            })
+            .collect();
+        let mut options = Default::default();
+        let file = File::create(path).unwrap();
+        let w = BufWriter::new(file);
+        inferno::flamegraph::from_lines(&mut options, lines.iter().map(|s| s.as_str()), w).unwrap();
+    }
+
+    /// calculate totals and write out results
+    pub fn execution_finished(&mut self) {
```
leonardoalt, line 163:
```suggestion
    pub fn finish(&mut self) {
```
maybe?

riscv-executor/src/profiler.rs
```diff
@@ -0,0 +1,346 @@
...
+
+        if self
+            .return_pc_stack
+            .last()
+            .is_some_and(|saved_pc| *saved_pc == target_pc)
+        {
+            // "return" from current function
+            let (done_call, cost) = self.call_stack.pop().unwrap();
+            self.return_pc_stack.pop();
+            // add to calls cummulative cost and to parent call
```
leonardoalt, line 306:
```suggestion
            // add to calls cumulative cost and to parent call
```

leonardoalt: wdym by `to parent call`?

riscv-executor/src/profiler.rs
```diff
@@ -0,0 +1,346 @@
...
+                *curr_cost += cost;
+            }
+        } else {
+            let target = self.location_at(target_pc).unwrap();
+            let curr_function = self.curr_function().unwrap();
+            if target.function != curr_function {
+                // "tail call": replace the current call in the stack
+                let (done_call, cost) = self.call_stack.pop().unwrap();
+
+                // add to calls cummulative cost and to parent call
```
leonardoalt, line 318: same here

riscv-executor/src/profiler.rs
```diff
@@ -0,0 +1,346 @@
...
+            } else {
+                // "control flow" (or "tail call" to self, if that is a thing), don't think this needs special handling
+            }
+        }
+    }
+}
+
+fn format_function_name(name: &str) -> String {
+    if let Some(prefix) = name.find("___ZN") {
+        // format!("{}_{}", &name[0..4], demangle(&name[prefix + 2..]))
```
leonardoalt, line 341: what about this?

leonardoalt (commented):
Nice! Mostly small comments

riscv-executor/src/profiler.rs
pacheco: we update two costs here (e.g., lets say this is `bar` calling `foo` from line 20):
- cumulative cost of this specific call (`foo -> bar` at line 20)
- running cost of caller (`foo`)

leonardoalt (approved):

PR #1396, opened 2024-05-27
Store contents in SourceRef and change to offset instead of line/col.

chriseth:


backend/src/estark/json_exporter/mod.rs
```diff
@@ -561,4 +593,41 @@ mod test {
         compare_export_file_ignore_idq_hex("rom.pil");
         compare_export_file_ignore_idq_hex("main.pil");
     }
+
+    #[test]
+    fn line_calc() {
```
chriseth, line 598: These tests also come from lines.rs

leonardoalt (approved):

PR #1405, opened 2024-05-29
Implementing a more generic version of `jalr`.

lvella:
Still not the full generic version: only supports zero offset.


leonardoalt (approved):

leonardoalt:
Conflicts?

lvella:
Can't be conflicts, it built and executed. Maybe the VM crashed?

PR #1408, opened 2024-05-29
Keccak Rust Wrapper Finalize

qwang98:
Currently the `keccakf` rust wrapper is road blocked by `keccak.asm`, which can only be completed after merging #1218. This PR is to finalize #1337, which currently leaves `keccakf` as `unimplemented()!` and also complements the rust wrapper tests, which currently only uses tiny keccak.

This PR should be merged once we have `keccak.asm` ready. Additionally, we also need to write an executor (witness generator) for keccak.


leonardoalt:
I think we actually finished this in other PRs.

PR #1409, opened 2024-05-30
Check docs in CI

Schaeff:


leonardoalt:
I think we recently removed this for some reason

PR #1410, opened 2024-05-30
modul asm only

qwang98:
Only the asm file changes that's ready to merge. To be paired with another PR that will be updated and merged later: #1404


georgwiese (approved):
Nice! 🚀

PR #1416, opened 2024-06-04
PIL -> rust compiler

chriseth:
This is a (yet unused) component that takes an `Analyzed` and a list of names and turns these names into `fn(u64) -> u64` to be called directly.

Not all syntax is supported yet (and no built-in functions).

The first goal is not be able to compile all constant functions.

The benchmarks show that `sqrt` is roughly 10 times faster than the evaluator.


pil-analyzer/tests/types.rs
```diff
@@ -704,7 +704,7 @@ fn trait_user_defined_enum_wrong_type() {
     }
 
     let n: int = 7;
-    let r1 = Convert::convert(n);
+    let r1: int = Convert::convert(n);
```
chriseth, line 707:
```suggestion
    let r1 = Convert::convert(n);
```

jit-compiler/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+        return Ok(Default::default());
+    };
+
+    let glue_code = generate_glue_code(&successful_symbols, analyzed)?;
+
+    let (dir, lib_path) = call_cargo(&format!("{glue_code}\n{}\n", codegen.compiled_symbols()))?;
+    let metadata = fs::metadata(&lib_path).unwrap();
+
+    log::info!(
+        "Loading library with size {} MB...",
```
chriseth, line 45:
```suggestion
        "Loading library of size {} MB...",
```

georgwiese:
Haven't review the code yet, but this doesn't run on my Mac, because the compiler does not output a `libpowdr_jit_compiled.so` file (I think the Mac-equivalent is `libpowdr_jit_compiled.dylib`, which does exist).

jit-compiler/tests/execution.rs
```diff
@@ -0,0 +1,39 @@
...
+    powdr_jit_compiler::compile(&analyzed, &[symbol]).unwrap()[symbol]
+}
+
+#[test]
+fn identity_function() {
+    let f = compile("let c: int -> int = |i| i;", "c");
+
+    assert_eq!(f(10), 10);
+}
+
```
gzanitti, line 17: I'm not sure if this is the right place, but I would like some test to fail until more types are supported.

```suggestion
#[test]
#[should_panic = "Only (int -> int) functions and columns are supported, but requested  c: int -> bool"]
fn invalid_function() {
    let _ = compile("let c: int -> bool = |i| true;", "c");
}
```

gzanitti (changes requested):

jit-compiler/src/compiler.rs
```diff
@@ -0,0 +1,133 @@
...
+        .join("target")
+        .join("release")
+        .join(format!("libpowdr_jit_compiled.{extension}"));
+    Ok((dir, lib_path.to_str().unwrap().to_string()))
+}
+
+/// Loads the given library and creates funtion pointers for the given symbols.
+pub fn load_library(path: &str, symbols: &[&str]) -> Result<SymbolMap, String> {
+    let c_path = CString::new(path).unwrap();
+    let lib = unsafe { dlopen(c_path.as_ptr(), RTLD_NOW) };
```
lvella, line 113: I don't think Windows have a `dlopen()` function like the other two other systems who are more POSIX compliant. If supporting Windows is needed (and I suspect Windows support in Powdr is broken in more ways than this) we should use a library that abstracts the differences between operating systems, like [libloading](https://crates.io/crates/libloading).

jit-compiler/src/compiler.rs
```diff
@@ -0,0 +1,133 @@
...
+            ));
+        }
+
+        // TODO we should use big int instead of u64
+        let name = escape_symbol(sym);
+        glue.push_str(&format!(
+            r#"
+            #[no_mangle]
+            pub extern fn {}(i: u64) -> u64 {{
+                u64::try_from({name}(ibig::IBig::from(i))).unwrap()
```
lvella, line 54: It would be nice if ibig was dynamically linked both on powdr and on the JITed library, this would ensure that the same runtime version of the library was used on both, reducing the JITed code size and build time, and allowing the type to be used directly on this interface.

lvella: Which brings another potential issue: since no extern "C" is being used, the ABI is guaranteed to be compatible only if the exact same version who built powdr also builds the library.

georgwiese:
Mac is working now 🥳

jit-compiler/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+mod compiler;
+
+use std::{collections::HashMap, fs};
+
+use codegen::CodeGenerator;
+use compiler::{call_cargo, generate_glue_code, load_library};
+use powdr_ast::analyzed::Analyzed;
+use powdr_number::FieldElement;
+
+pub type SymbolMap = HashMap<String, fn(u64) -> u64>;
```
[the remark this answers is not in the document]
chriseth: Yes, for the prover functions. We'll see how many we need.

jit-compiler/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    requested_symbols: &[&str],
+) -> Result<SymbolMap, String> {
+    log::info!("JIT-compiling {} symbols...", requested_symbols.len());
+
+    let mut codegen = CodeGenerator::new(analyzed);
+    let successful_symbols = requested_symbols
+        .iter()
+        .filter_map(|&sym| {
+            if let Err(e) = codegen.request_symbol(sym) {
+                log::warn!("Unable to generate code for symbol {sym}: {e}");
```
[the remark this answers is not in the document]
chriseth: Yes because we can always run the interpreter.

jit-compiler/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    log::info!(
+        "Loading library of size {} MB...",
+        metadata.len() as f64 / 1000000.0
+    );
+
+    let result = load_library(&lib_path, &successful_symbols);
+    log::info!("Done.");
+
+    drop(dir);
```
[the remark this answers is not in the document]
chriseth: It's a temporary directory that is deleted when `dir` goes out of scope. An explicit `drop` keeps it alive longer.

jit-compiler/src/compiler.rs
```diff
@@ -0,0 +1,133 @@
...
+};
+use powdr_number::FieldElement;
+
+use crate::{codegen::escape_symbol, SymbolMap};
+
+// TODO make this depend on T
+
+const PREAMBLE: &str = r#"
+#![allow(unused_parens)]
+//type FieldElement = powdr_number::GoldilocksField;
```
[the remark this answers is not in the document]
chriseth: There is an another PR that implements it.

jit-compiler/src/codegen.rs
```diff
@@ -0,0 +1,350 @@
...
+                let right = self.format_expr(right)?;
+                match op {
+                    BinaryOperator::ShiftLeft => {
+                        format!("(({left}).clone() << u32::try_from(({right}).clone()).unwrap())")
+                    }
+                    _ => format!("(({left}).clone() {op} ({right}).clone())"),
+                }
+            }
+            Expression::UnaryOperation(_, UnaryOperation { op, expr }) => {
+                format!("({op} ({}).clone())", self.format_expr(expr)?)
```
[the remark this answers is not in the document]
chriseth: I think so, everything is in parentheses - or am I missing something?

jit-compiler/src/codegen.rs
```diff
@@ -0,0 +1,350 @@
...
+    #[test]
+    fn fun_calls() {
+        let result = compile(
+            "let c: int -> int = |i| i + 20; let d = |k| c(k * 20);",
+            &["c", "d"],
+        );
+        assert_eq!(
+            result,
+            "fn c(i: ibig::IBig) -> ibig::IBig { ((i).clone() + (ibig::IBig::from(20_u64)).clone()) }\n\
+            \n\
```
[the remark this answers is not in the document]
chriseth: The advantage of using `\n\` is that you can indent the following line. Raw strings would not allow that IIUC.

jit-compiler/src/compiler.rs
chriseth: Yeah, I would not rely on the fact that we have exactly the same rust compiler at runtime as at compile time. Will change to extern c.

jit-compiler/src/lib.rs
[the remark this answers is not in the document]
chriseth: It's a temporary directory that is deleted when the reference is dropped. We need to keep it alive at least until the dlopen call.

jit-compiler/tests/execution.rs
```diff
@@ -0,0 +1,45 @@
...
+#[test]
+fn sqrt() {
+    let f = compile(
+        "
+        let sqrt_rec: int, int -> int = |y, x|
+        if y * y <= x && (y + 1) * (y + 1) > x {
+            y
+        } else {
+            sqrt_rec((y + x / y) / 2, x)
+        };
```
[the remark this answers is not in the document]
chriseth: We have multiple copies of that. It's [Heron's method](https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Heron's_method)

jit-compiler/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+itertools = "0.13"
+
+[dev-dependencies]
+powdr-pil-analyzer.workspace = true
+pretty_assertions = "1.4.0"
+test-log = "0.2.12"
+env_logger = "0.10.0"
+
+[lints.clippy]
+uninlined_format_args = "deny"
```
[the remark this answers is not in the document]
chriseth: We have that setting in all crates.

jit-compiler/src/compiler.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+    let mut result = HashMap::new();
+    for sym in symbols {
+        let extern_sym = extern_symbol_name(sym);
+        let sym_cstr = CString::new(extern_sym).unwrap();
+        let fun_ptr = unsafe { dlsym(lib, sym_cstr.as_ptr()) };
+        if fun_ptr.is_null() {
+            return Err(format!("Failed to load symbol: {fun_ptr:?}"));
+        }
+        let fun = unsafe { std::mem::transmute::<*mut c_void, fn(u64) -> u64>(fun_ptr) };
```
lvella, line 125: Since you change to `extern "C"` on the definition, it should be needed here, too...

jit-compiler/src/lib.rs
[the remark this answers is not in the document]
lvella: But the effect of drop here must be exactly the same as letting the function return. Did you mean `std::mem::forget()`, or even `dir.release()`?

If it is working with the drop, it should be working without it as well, and there is something else going on that allows dlopen to find the file.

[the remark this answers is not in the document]
chriseth: Ah, didn't see `dir.release()`. But at least `dir.release()` must be called, otherwise (as far as I understand), the compiler can drop the value any time even before the function returns.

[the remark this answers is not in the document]
georgwiese: Yeah, I tried just removing this line and it still works.

jit-compiler/src/compiler.rs
```diff
@@ -0,0 +1,133 @@
...
+        "so"
+    };
+    let lib_path = dir
+        .join("target")
+        .join("release")
+        .join(format!("libpowdr_jit_compiled.{extension}"));
+    Ok((dir, lib_path.to_str().unwrap().to_string()))
+}
+
+/// Loads the given library and creates funtion pointers for the given symbols.
```
georgwiese, line 110:
```suggestion
/// Loads the given library and creates function pointers for the given symbols.
```

jit-compiler/src/compiler.rs
```diff
@@ -0,0 +1,133 @@
...
+crate-type = ["cdylib"]
+
+[dependencies]
+ibig = { version = "0.3.6", features = [], default-features = false }
+"#;
+
+/// Compiles the given code and returns the path to the
+/// temporary directory containing the compiled library
+/// and the path to the compiled library.
+pub fn call_cargo(code: &str) -> Result<(Temp, String), String> {
```
georgwiese, line 80: Not a big deal, but you could make a `RustCompiler` struct that keeps the tmp dir and provides a `compile(&mut self, &str) -> Result<(), String>` and `lib_path(&self) -> String`. Then, the caller wouldn't have to know about the tmp dir.

But maybe that's overkill if the function is only `pub(crate)` and the only user is the `compile()` function, which would be the only function used outside this crate.

georgwiese (commented):
Cool!

pipeline/benches/evaluator_benchmark.rs
```diff
@@ -114,5 +114,37 @@ fn evaluator_benchmark(c: &mut Criterion) {
...
+        let mut pipeline = Pipeline::default().from_asm_string(code, None);
+        pipeline.compute_analyzed_pil().unwrap().clone()
+    };
+
+    let sqrt_fun = powdr_jit_compiler::compile(&sqrt_analyzed, &["sqrt"]).unwrap()["sqrt"];
+
+    for x in [879882356, 1882356, 1187956, 56] {
+        group.bench_with_input(format!("sqrt_{x}"), &x, |b, &x| {
+            b.iter(|| {
+                let y = (x as u64) * 112655675;
```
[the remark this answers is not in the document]
chriseth: I actually don't remember...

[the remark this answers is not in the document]
chriseth: Ah right. because the plain numbers are too small to see a difference in runtime across the sizes.

jit-compiler/src/codegen.rs
```diff
@@ -0,0 +1,350 @@
...
+        IndexAccess, LambdaExpression, Number, StatementInsideBlock, UnaryOperation,
+    },
+};
+use powdr_number::FieldElement;
+
+pub struct CodeGenerator<'a, T> {
+    analyzed: &'a Analyzed<T>,
+    requested: HashSet<String>,
+    failed: HashMap<String, String>,
+    symbols: HashMap<String, String>,
```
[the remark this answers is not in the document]
chriseth: I actually combined all three of them...

jit-compiler/src/lib.rs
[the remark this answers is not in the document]
lvella: I don't understand how it works, either, but this drop is a nop 😎

jit-compiler/src/compiler.rs
```diff
@@ -104,28 +110,34 @@ pub fn call_cargo(code: &str) -> Result<(Temp, String), String> {
...
+) -> Result<HashMap<String, LoadedFunction>, String> {
+    let library = Arc::new(
+        unsafe { libloading::Library::new(path) }
+            .map_err(|e| format!("Error loading library at {path}: {e}"))?,
+    );
+    symbols
+        .iter()
+        .map(|&sym| {
+            let extern_sym = extern_symbol_name(sym);
+            let function = *unsafe { library.get::<fn(u64) -> u64>(extern_sym.as_bytes()) }
```
lvella, line 132: I am unsure about the syntax, but you have to tell somehow this function is extern "C".

jit-compiler/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+    };
+
+    let glue_code = generate_glue_code(&successful_symbols, analyzed)?;
+
+    let lib_file = call_cargo(&format!("{glue_code}\n{}\n", codegen.compiled_symbols()))?;
+    let metadata = fs::metadata(&lib_file.path).unwrap();
+
+    log::info!(
+        "Loading library of size {} MB...",
+        metadata.len() as f64 / 1000000.0
```
lvella, line 60: Maybe use 1024 * 1024 instead of 1000000?

jit-compiler/src/lib.rs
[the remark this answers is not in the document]
chriseth: It actually did not work for me in the sense that `let (dir, ..) = ...` will lead to a clippy error because `dir` is unused. And if I remove `dir`, then the directory will be deleted.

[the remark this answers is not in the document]
lvella: Ah, that explains the need for it: so that dir is used and clippy does not complain. Alternatively you can try naming it _dir.

pipeline/benches/evaluator_benchmark.rs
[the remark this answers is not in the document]
georgwiese: So there is nothing specific about any of these numbers, right? Maybe you could add a comment, so we don't wonder next time :)

jit-compiler/src/compiler.rs
```diff
@@ -104,28 +110,35 @@ pub fn call_cargo(code: &str) -> Result<(Temp, String), String> {
...
+    let library = Arc::new(
+        unsafe { libloading::Library::new(path) }
+            .map_err(|e| format!("Error loading library at {path}: {e}"))?,
+    );
+    symbols
+        .iter()
+        .map(|&sym| {
+            let extern_sym = extern_symbol_name(sym);
+            let function =
+                *unsafe { library.get::<unsafe extern "C" fn(u64) -> u64>(extern_sym.as_bytes()) }
```
lvella, line 133: I am not sure I agree the loaded function should be `unsafe`. We just wrote its code and built it ourselves, and its rust code is perfectly safe.

jit-compiler/src/lib.rs
```diff
@@ -1,22 +1,36 @@
...
 
-pub type SymbolMap = HashMap<String, fn(u64) -> u64>;
+/// Wrapper around a dynamically loaded function.
+/// Prevents the dynamically loaded library to be unloaded while the function is still in use.
+#[derive(Clone)]
+pub struct LoadedFunction {
+    #[allow(dead_code)]
+    library: Arc<libloading::Library>,
+    function: unsafe extern "C" fn(u64) -> u64,
+}
```
lvella, line 19: How about this instead of using `Arc`?
```
pub struct SymbolMap {
    library: libloading::Library
}
```

and then:
```
impl SymbolMap {
    fn get<'a>(&'a self, name: &str) -> Result<libloading::Symbol<'a, extern "C" fn(u64) -> u64>>
    {
        // most of the code currently inside `load_library()`,
        // i.e. the code to get the function
    }
}
```

Looks more elegant, but I can think of two drawbacks:

- The error checking for symbol availability is deferred until `get()` (which should not be a problem if we are loading the library we just built, as the symbol is expected to be there);
- The symbol search is linear (which should not be a problem if each function is retrieved just a few times).

jit-compiler/src/compiler.rs
```diff
@@ -104,28 +110,35 @@ pub fn call_cargo(code: &str) -> Result<(Temp, String), String> {
...
+                    .map_err(|e| format!("Error accessing symbol {sym}: {e}"))?;
+            let fun = LoadedFunction {
+                library: library.clone(),
+                function,
+            };
+            Ok((sym.to_string(), fun))
+        })
+        .collect::<Result<_, String>>()
```
lvella, line 141: Maybe this entire block should go inside an `SymbolMap::get()`.

jit-compiler/src/lib.rs
lvella: According to ChatGPT, symbol lookup is also indexed on macOS and Windows.

chriseth: I don't really want to leak the weird Symbol thing with it's lifetime parameter. In the tests, we are retrieving a single function and it's much more fun to just store or return it and don't care about lifetime.

chriseth: Also what are the issues about Arc? It's a tiny atomic update for each clone / drop, right? I don't think they should happen often.

jit-compiler/src/compiler.rs
chriseth: You mean we should delay accessing the symbol? I think it's better to find them right away first. The issue is that for fixed column creation, I would like to have a proper function pointer that has the least cost associated to it because we call it 2**20 times, and I don't want to search for it 2**20 times.

jit-compiler/src/lib.rs
lvella: There is no issue with Arc. If you prefer this way, as long as the raw function pointer doesn't leak out of the struct, this is fine. One extra pointer per function won't make any difference.

jit-compiler/src/compiler.rs
lvella: I don't think this would be any different. In the current way, at some point you will have to retrieve the `LoadedFunction` from the `HashMap` returned by `load_library()`. In the way I propose, you will instead retrieve the `libloading::Symbol` from the `SymbolMap` returned by `load_library()` and still will have a proper function pointer to call, just that it is lifetime bound to the `SymbolMap` instead of being freestanding (due to the lack of Arc).

lvella: Depending how it is used, the lifetime can be an inconvenience, sure, in this case your solution is better.

chriseth: At some point, we will store the type of the symbol. I think at latest then we can't really expose the Symbol stuff to the user either.

Another point I would like to make: If we return a Symbol to the user, there is no protection against dereferencing the symbol. Once the symbol is dereferenced into a function, it can be freely copied, and the association with the library is gone. That's why I think not exposing the function type to the user at all is the best way.

pacheco (approved):

jit-compiler/src/codegen.rs
```diff
@@ -0,0 +1,363 @@
...
+                match self.generate_code(&name) {
+                    Ok(code) => {
+                        to_insert = Ok(Some(code));
+                        to_return = Ok(());
+                    }
+                    Err(err) => {
+                        to_insert = Err(err.clone());
+                        to_return = Err(err);
+                    }
+                }
```
Schaeff, line 52:
```suggestion
                let (to_insert, to_return) =
                match self.generate_code(&name) {
                    Ok(code) => {
                        (Ok(Some(code)), Ok(()))
                    }
                    Err(err) => {
                        (Err(err.clone()), Err(err))
                    }
                };
```

chriseth: The next PR heavily modifies this again anyway, so I would just keep it for now: https://github.com/powdr-labs/powdr/pull/1837/files#diff-adff81349be67708164c673cd6da23098e721c006cd10cf61add179bb33b0679R37

Schaeff: Fine

PR #1417, opened 2024-06-04
List type bounds.

chriseth:


std/machines/arith.asm
```diff
@@ -210,6 +210,12 @@ machine Arith with
     col witness s_13(i) query Query::Hint(fe(select_limb(s_hint(), 13)));
     col witness s_14(i) query Query::Hint(fe(select_limb(s_hint(), 14)));
     col witness s_15(i) query Query::Hint(fe(select_limb(s_hint(), 15)));
+    prover || {
```
chriseth, line 213: Haha that shouldn't be here.

leonardoalt (approved):

PR #1422, opened 2024-06-05
Refactoring powdr asm generator to make it source agnostic.

lvella:
Code gen has a generic interface that is now able to take RISC-V programs from other sources besides assembly (e.g., from the upcoming ELF executable).


asm-utils/src/data_storage.rs
```diff
@@ -7,23 +7,23 @@ use crate::{
     utils::{alignment_size, next_aligned},
 };
 
-pub enum SingleDataValue<'a> {
+pub enum SingleDataValue {
     Value(u32),
-    LabelReference(&'a str),
-    Offset(&'a str, &'a str),
+    LabelReference(String),
+    Offset(String, String),
```
lvella, line 13: This change won't make new String copies, the original Strings are being moved now.

asm-utils/src/data_storage.rs
```diff
@@ -7,23 +7,23 @@ use crate::{
     utils::{alignment_size, next_aligned},
 };
 
-pub enum SingleDataValue<'a> {
+pub enum SingleDataValue {
```
pacheco, line 10: could be nice to have a small comment on why this type exists and a comment on each variant about which situations they come from

asm-utils/src/data_storage.rs
pacheco: nice

riscv/src/asm_translate/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+
+    fn source_files_info(&self) -> impl Iterator<Item = SourceFileInfo> {
+        self.file_ids.iter().map(|(id, dir, file)| SourceFileInfo {
+            id: *id as u32,
+            dir,
+            file,
+        })
+    }
+
+    fn initial_mem(&self) -> impl Iterator<Item = MemEntry> {
```
pacheco, line 51: i think the usage of this method will be more clear as `take_initial_mem(&mut)` and dropping the `Cell` usage

riscv/src/asm_translate/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+    assemblies: BTreeMap<String, String>,
+    runtime: &Runtime,
+    with_bootloader: bool,
+) -> String {
+    let asm_program = compile_internal(assemblies);
+
+    code_gen::translate_program::<F>(&asm_program, runtime, with_bootloader)
+}
+
+fn compile_internal(mut assemblies: BTreeMap<String, String>) -> AsmProgram {
```
pacheco, line 189: maybe `compile_asm`?

lvella: This is just so that the big function don't have generic parameters (so that the generated code is smaller). So I think this function should be name `<name_of_the_public_function>_internal`

riscv/src/lib.rs
```diff
@@ -13,19 +13,13 @@ use powdr_number::FieldElement;
 use serde_json::Value as JsonValue;
 use std::fs;
 
-use crate::compiler::{FunctionKind, Register};
 pub use crate::runtime::Runtime;
 
-pub mod compiler;
+pub mod asm_translate;
```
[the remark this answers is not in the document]
lvella: done

riscv/src/code_gen.rs
```diff
@@ -63,135 +48,117 @@ impl fmt::Display for FunctionKind {
...
+    fn source_files_info(&self) -> impl Iterator<Item = SourceFileInfo>;
+
+    fn initial_mem(&self) -> impl Iterator<Item = MemEntry>;
+    fn executable_statements(
+        &'a self,
+    ) -> impl Iterator<Item = Statement<Self::Label, Self::InstructionArgs>>;
+    fn start_function(&self) -> &str;
+}
+
+pub fn translate_program<'a, F: FieldElement>(
```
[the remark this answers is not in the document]
lvella: Done.

riscv/src/code_gen.rs
```diff
@@ -812,47 +645,8 @@ fn memory(with_bootloader: bool) -> String {
...
-            };
-            ret.extend(processed_instr.into_iter().map(|s| "  ".to_string() + &s));
-            ret
-        }
-    }
-}
-
-trait Args {
-    type Error;
+pub trait Args {
```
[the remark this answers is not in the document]
lvella: Done.

asm-utils/src/data_storage.rs
lvella: Done.

riscv/src/asm_translate/mod.rs
lvella: Done.

PR #1427, opened 2024-06-06
Process pc lookup first.

chriseth:


executor/src/witgen/vm_processor.rs
```diff
@@ -314,9 +314,29 @@ impl<'a, 'b, 'c, T: FieldElement, Q: QueryCallback<T>> VmProcessor<'a, 'b, 'c, T
         identities: &mut CompletableIdentities<'a, T>,
     ) -> Result<Constraints<&'a AlgebraicReference, T>, Vec<EvalError<T>>> {
         let mut outer_assignments = vec![];
+
+        // The PC lookup fills most of the columns and enables hints thus it should be run first.
+        // We find it as largest plookup identity.
```
leonardoalt, line 319:
```suggestion
        // We find it as largest lookup identity.
```
I think we should stop calling things plookup

PR #1428, opened 2024-06-06
Flat lookup

chriseth:
I think this might not have been worth it.


leonardoalt:
shall we close the PR and keep the branch for now?

PR #1431, opened 2024-06-07
Doing RISC-V initializations on runtime lib, with linker support.

lvella:
Improves compatibility with vanilla RISC-V development pipeline.


riscv-runtime/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "powdr-riscv-runtime"
 description = "powdr runtime provider for RISCV programs"
-version = "0.1.0-alpha.2"
+version = "0.2.0-alpha.0"
```
lvella, line 4: Is this an incompatible change that warrants a version bump? I am not sure...

leonardoalt: the current version in the file is the next to be released, I think it's fine to keep it like that

asm-utils/src/reachability.rs
```diff
@@ -141,13 +141,20 @@ fn extract_replacements<R: Register, F: FunctionOpKind>(
...
-                {
-                    Some((from.to_string(), to.to_string()))
-                } else {
-                    panic!();
+                match &args[..] {
+                    [Argument::Expression(Expression::Symbol(from)), Argument::Expression(Expression::Symbol(to))] =>
+                    {
+                        Some((from.to_string(), to.to_string()))
+                    },
+                    [Argument::Expression(Expression::Symbol(_)), Argument::Expression(Expression::Number(_))] => {
```
leonardoalt, line 149: what exactly falls here and needs to be handled elsewhere?

riscv-runtime/Cargo.toml
lvella: Reverted.

asm-utils/src/data_parser.rs
```diff
@@ -142,6 +160,18 @@ pub fn extract_data_objects<R: Register, F: FunctionOpKind>(
                         ));
                     }
                 }
+                (
+                    ".set",
+                    [Argument::Expression(Expression::Symbol(label)), Argument::Expression(Expression::Number(value))],
+                ) => {
```
lvella, line 166: The other kind of `.set` directive is handled here. But I changed the comment to be less mysterious.

asm-utils/src/reachability.rs
lvella: Clarified comment.

lvella: But this function handle aliases: making one label alias for another. The variant of `.set` directive I am using now is not an alias, so it is not handled here, but can't fail here either, because it is valid.

leonardoalt (approved):

PR #1434, opened 2024-06-07
Build powdr rust programs to executables instead of libraries.

lvella:
This is for future compatibility with translating program from ELFs.


riscv/tests/riscv_data/many_chunks_memory/src/main.rs
```diff
@@ -1,6 +1,8 @@
+#![no_main]
 #![no_std]
 
 extern crate alloc;
+extern crate powdr_riscv_runtime;
```
leonardoalt, line 5: Why is this needed if it's also a dependency?

leonardoalt (approved):

PR #1435, opened 2024-06-07
Allow blocks to be empty

gzanitti:
This PR builds on top of #1393.
It mainly modifies the grammar by changing the way SelectedExpressions are declared, to allow blocks to be empty.


chriseth:
Is this ready? (Asking because it's in draft mode)

gzanitti:
> Is this ready? (Asking because it's in draft mode)

Sorry, I forgot to change the status when we merge the other PR on empty blocks. It's ready now.

analysis/README.md
```diff
@@ -6,17 +6,17 @@ This crate is where most of the compilation pipeline lives. It takes a parse tre
 
 We define two types of machines: virtual machines and constrained machines. Constrained machines are the lower level kind of machine. They have a notion of blocks through a latch and an operation_id. Virtual machines are a higher level type of machines. For each machine type, we provide the elements which appear in that type.
 
-|                         | Virtual | Constrained |
```
chriseth, line 9: Please undo this change.

analysis/README.md
```diff
@@ -99,6 +99,7 @@ ASM to PIL has two steps: ROM generation and reduction to constrained.
 ##### ROM generation
 
 Rom generation generates a single ROM for each virtual machine using the following process:
+
```
chriseth, line 102: Same with all the newline changes here - I assume they were by accident.

ast/src/analyzed/display.rs
```diff
@@ -267,7 +267,35 @@ impl Display for RepeatedArray {
     }
 }
 
-impl Display for Identity<Expression> {
+impl Display for parsed::SelectedExpressions<Expression> {
```
chriseth, line 270: Shouldn't this be in src/parsed/display.rs?

ast/src/analyzed/mod.rs
```diff
@@ -625,21 +628,51 @@ impl PublicDeclaration {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct SelectedExpressions<Expr> {
+    pub selector: Option<Expr>,
+    pub expressions: Vec<Expr>,
+}
+
+impl<Expr> Default for SelectedExpressions<Expr> {
```
chriseth, line 637: Isn't this the derived implementation?

chriseth: Oh, it doesn't work because Expr doesn't need to implement Default, right?

analysis/README.md
gzanitti: Yes, probably the autoformatting when saving did its thing. I'll fix it ASAP

ast/src/analyzed/mod.rs
```diff
@@ -625,21 +628,51 @@ impl PublicDeclaration {
...
     /// For a simple polynomial identity, the selector contains
     /// the actual expression (see expression_for_poly_id).
-    pub left: SelectedExpressions<Expr>,
-    pub right: SelectedExpressions<Expr>,
+    pub left: SelectedExpressions,
+    pub right: SelectedExpressions,
 }
 
-impl<Expr> Identity<Expr> {
+impl<T> Identity<SelectedExpressions<AlgebraicExpression<T>>> {
```
chriseth, line 669: Why not for generic `Expr`, i.e. `SelectedExpressions<Expr>`?

book/src/pil/types.md
```diff
@@ -241,13 +240,15 @@ References to columns have type `expr` and `expr` also implements `FromLiteral`,
 which means that literal numbers can be used in contexts where `expr` is expected.
 
 Example:
+
```
chriseth, line 243: In this files are also some auto-changes.

executor/src/witgen/global_constraints.rs
```diff
@@ -102,15 +102,17 @@ impl<T: FieldElement> RangeConstraintSet<&AlgebraicReference, T> for GlobalConst
     }
 }
 
+type SelExpressions<T> = SelectedExpressions<Expression<T>>;
```
chriseth, line 105: I think in witgen, there should be something like
`type Identity<T> = Identity<SelectedExpressions<Expression<T>>>`

chriseth: Maybe this even makes sense in some parts of analyzed...

parser/src/powdr.lalrpop
```diff
@@ -622,19 +621,16 @@ MatchArm: MatchArm = {
...
 
-BlockExpression: Box<Expression> = {
-    <start:@L> "{" <statements:StatementInsideBlock+> <expr:BoxedExpression> "}" <end:@R> => ctx.to_expr_with_source_ref(BlockExpression{statements, expr}, start, end),
-}
 
-BracedExpression: Box<Expression> = {
-    <start:@L> "{" <statements:StatementInsideBlock+> <expr:BoxedExpression> "}" <end:@R> => ctx.to_expr_with_source_ref(BlockExpression{statements, expr}, start, end),
-    <start:@L> "{" <expr:BoxedExpression> "}" <end:@R> => ctx.to_expr_with_source_ref(BlockExpression{statements: vec![], expr}, start, end),
+BlockExpression: Box<Expression> = {
+    <start:@L> "{" <statements:StatementInsideBlock*> "}" <end:@R> => ctx.to_expr_with_source_ref(BlockExpression{statements, expr: Box::new(Expression::Tuple(ctx.source_ref(start, end), vec![]))}, start, end),
```
chriseth, line 632: I'm not sure I would implement the tuple at the syntax level. Is it much more difficult to make `expr` and `Option<Expression>`?

chriseth: Also the source reference is misleading.

pil-analyzer/src/condenser.rs
```diff
@@ -280,18 +283,14 @@ impl<'a, T: FieldElement> Condenser<'a, T> {
 
     fn condense_selected_expressions(
         &mut self,
-        sel_expr: &'a SelectedExpressions<Expression>,
-    ) -> SelectedExpressions<AlgebraicExpression<T>> {
-        SelectedExpressions {
+        sel_expr: &'a powdr_ast::parsed::SelectedExpressions<Expression>,
+    ) -> powdr_ast::analyzed::SelectedExpressions<AlgebraicExpression<T>> {
```
chriseth, line 287: is the `powdr_ast` prefix needed?

pil-analyzer/src/expression_processor.rs
```diff
@@ -84,6 +84,23 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
             .collect()
     }
 
+    pub fn process_vec_into_array_literal(&mut self, exprs: Vec<parsed::Expression>) -> Expression {
+        let unknown = SourceRef::unknown();
+        let src = exprs
+            .first()
+            .map(|e| e.source_reference())
```
chriseth, line 91: Could you make the start the min of all starts and the end the max of all ends?

pil-analyzer/src/statement_processor.rs
```diff
@@ -340,17 +340,19 @@ where
...
                 IdentityKind::Connect,
-                SelectedExpressions {
+                powdr_ast::parsed::SelectedExpressions {
                     selector: None,
-                    expressions: self
-                        .expression_processor(&Default::default())
-                        .process_expressions(left),
+                    expressions: Box::new(
+                        self.expression_processor(&Default::default())
+                            .process_vec_into_array_literal(left),
```
chriseth, line 347: Is there a way to handle connect in the same way as `in`? I know it doesn't have a selector but that's the only difference, right?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -287,8 +288,16 @@ impl PILAnalyzer {
                     if let Some(selector) = &mut part.selector {
                         expressions.push((selector, Type::Expr.into()))
                     }
-                    for e in &mut part.expressions {
-                        expressions.push((e, Type::Expr.into()))
+                    if let Expression::ArrayLiteral(_, ArrayLiteral { items }) =
```
chriseth, line 291: It's fine to use `x $ f() in y $ g()` now, as long as `f()` and `g()` have the type `Expr[]`. So you shouldn't iterate through the items here but just request an array as type for `part.expressions`.

ast/src/analyzed/mod.rs
gzanitti: The problem here is that now there are two SelectedExpression (`parsed` and `analyzed`) and in some cases functions specific to SelectedExpressions, such as `contains_next_ref`, are call, so you end up with a lot more duplicate code.

Although maybe a bit "uglier", I think this solution is cleaner.

ast/src/analyzed/mod.rs
gzanitti: It worked, but it was only used in the right SelectedExpression in the Identity impl. I found it more practical to delete this implementation and hardcode the default expression. Thanks!

backend/src/estark/json_exporter/mod.rs
```diff
@@ -508,93 +508,93 @@ mod test {
...
-        }
-    }
-
-    fn compare_export_file_ignore_idq_hex(file: &str) {
-        let (mut json_out, mut pilcom_parsed) = generate_json_pair(file);
-        normalize_idq_and_hex(&mut json_out);
-        normalize_idq_and_hex(&mut pilcom_parsed);
-        assert_eq!(json_out, pilcom_parsed);
-    }
+    // fn normalize_idq_and_hex(v: &mut JsonValue) {
```
chriseth, line 511: Let's just remove these tests.

executor/src/witgen/generator.rs
```diff
@@ -24,9 +24,9 @@ struct ProcessResult<'a, T: FieldElement> {
 }
 
 pub struct Generator<'a, T: FieldElement> {
-    connecting_identities: BTreeMap<u64, &'a Identity<Expression<T>>>,
+    connecting_identities: BTreeMap<u64, &'a Identity<SelectedExpressions<Expression<T>>>>,
```
chriseth, line 27: Same in this file.

executor/src/witgen/global_constraints.rs
```diff
@@ -102,15 +103,17 @@ impl<T: FieldElement> RangeConstraintSet<&AlgebraicReference, T> for GlobalConst
     }
 }
 
+type Identity<T> = AnalyzedIdentity<SelectedExpressions<Expression<T>>>;
```
chriseth, line 106: Please move this to a more central point. Could be at the root of this crate or even at some exeternal-facing point of the podr_ast::analyzed module.

executor/src/witgen/identity_processor.rs
```diff
@@ -126,7 +127,7 @@ impl<'a, 'b, 'c, T: FieldElement, Q: QueryCallback<T>> IdentityProcessor<'a, 'b,
     /// Returns the updates.
     pub fn process_identity(
         &mut self,
-        identity: &'a Identity<Expression<T>>,
+        identity: &'a Identity<SelectedExpressions<Expression<T>>>,
```
chriseth, line 130: Should use the typedef as well.

executor/src/witgen/processor.rs
```diff
@@ -27,15 +27,15 @@ pub struct OuterQuery<'a, 'b, T: FieldElement> {
...
+    pub connecting_identity: &'a Identity<SelectedExpressions<Expression<T>>>,
     /// The left side of the connecting identity, evaluated.
     pub left: Left<'a, T>,
 }
 
 impl<'a, 'b, T: FieldElement> OuterQuery<'a, 'b, T> {
     pub fn new(
         caller_rows: &'b RowPair<'b, 'a, T>,
-        connecting_identity: &'a Identity<Expression<T>>,
+        connecting_identity: &'a Identity<SelectedExpressions<Expression<T>>>,
```
chriseth, line 38: Same here, all the files in this crate should only need to use `Identity<T>`.

parser/tests/parentheses_test.rs
```diff
@@ -0,0 +1,136 @@
+#[cfg(test)]
```
chriseth, line 1: I think this file was moved.

pil-analyzer/src/evaluator.rs
```diff
@@ -749,7 +749,12 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
             Expression::BlockExpression(_, BlockExpression { statements, expr }) => {
                 self.op_stack
                     .push(Operation::TruncateLocals(self.local_vars.len()));
-                self.op_stack.push(Operation::Expand(expr));
+                match expr {
+                    Some(expr) => {
+                        self.op_stack.push(Operation::Expand(expr));
+                    }
+                    None => {}
```
chriseth, line 756: Do you have tests covering this line? Shouldn't it at least push an empty tuple or something?

pil-analyzer/src/expression_processor.rs
```diff
@@ -362,3 +381,17 @@ struct LocalVariableState {
     pub local_variables: HashMap<String, u64>,
     pub local_variable_counter: u64,
 }
+
+fn combine_source_refs(refs: Vec<&SourceRef>) -> SourceRef {
```
chriseth, line 385: Maybe this was not such a good suggestion of mine. What happens if the source refs are from different files? Maybe it's best to keep it at Unknown. Sorry!

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -287,8 +288,18 @@ impl PILAnalyzer {
...
+                            part.expressions.as_mut(),
+                            Type::Array(ArrayType {
+                                base: Box::new(Type::Expr),
+                                length: None,
+                            })
+                            .into(),
+                        ))
+                    } else {
+                        panic!("Selected expressions need to be provided as arrays.\n  Expected [{}].\n  Found: {}",
+                        part.expressions, part.expressions)
```
chriseth, line 302: Formatting looks off.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -287,8 +288,18 @@ impl PILAnalyzer {
                     if let Some(selector) = &mut part.selector {
                         expressions.push((selector, Type::Expr.into()))
                     }
-                    for e in &mut part.expressions {
-                        expressions.push((e, Type::Expr.into()))
+                    if let Expression::ArrayLiteral(_, _) = part.expressions.as_ref() {
```
chriseth, line 291: Why only for array literals? This should work for any expression. We can now also test things like `x $ f() in g()`.

chriseth:
I think this is almost ready. Sorry for the delay in reviewing. I think this can still be extended, but we can do it later on.

pil-analyzer/src/evaluator.rs
gzanitti: My guess here was that it made no difference to expand an empty expression, hence the empty None case. Also, there is the problem that creating an expression at this point is complex because of the lifetimes.


I'll see what I can do to solve it, thanks for your comment.

pil-analyzer/src/pil_analyzer.rs
gzanitti: I seem to remember that we had talked about this kind of expressions now always being of the form x $ [f()] in [g()], but no problem, I can fix that. The only thing I need to know (if I understand correctly), is how to treat those expressions that are not arrays here:

https://github.com/powdr-labs/powdr/blob/7e3a5c70913537f6351064762eb82ff24abd0199/pil-analyzer/src/condenser.rs#L293

because this function expects &sel_expr.expressions to evaluate to an array (which doesn't happen anymore in the example you gave me).

pil-analyzer/src/evaluator.rs
gzanitti: This test should cover the case you mention, or am I wrong?

https://github.com/powdr-labs/powdr/blob/524d579544eb4d3f2370a5d8af0cb1011ded5546/pil-analyzer/tests/condenser.rs#L208

gzanitti:
> I think this is almost ready. Sorry for the delay in reviewing. I think this can still be extended, but we can do it later on.

You don't have to apologise, it's a bit my fault because it took me a while to get into rhythm and understand a couple of things. It shouldn't happen anymore :)

pil-analyzer/src/evaluator.rs
chriseth: I don't know. Can you put a panic inside there to see if it is hit?

Also can you add a test for the following (also just curious what is happening and if we should disallow it):
```
let x = {};
```

pil-analyzer/src/pil_analyzer.rs
chriseth: Sorry, I should have expanded the example a little: My idea was that `f()` and `g()` evaluate to arrays. So you don't have to support tuples or anything else in `condense_to_array_of_algebraic_expressions`.

What I meant was that you should not expect here in line 291 that part.expressions is syntactically an array literal - only that its type is an array of expr.

pil-analyzer/src/evaluator.rs
chriseth: Oh or maybe also something a bit more complicated:
```
let f = |_| 7;
let g = f({});
```

gzanitti: Fixed. To avoid the lifetime issue, what I did was to change the None cases to the empty tuple directly in the statement processing, so at this point, the None case should not be reachable (I added some tests, including this one you mention above).

chriseth: Are you sure this is a good solution? I don't really like modifying the AST in such a way. You don't need to deal with the lifetime of expressions here. For an example, see the ::Number case. I think the following should do it:

```
 self.value_stack.push(Value::Tuple(vec![]))
```

pil-analyzer/src/condenser.rs
```diff
@@ -322,7 +321,7 @@ impl<'a, T: FieldElement> Condenser<'a, T> {
                     _ => panic!("Expected expression but got {item}"),
                 })
                 .collect(),
-            _ => panic!("Expected array of algebraic expressions, but got {result}"),
+            _ => panic!("Expected array/tuple of algebraic expressions or an algebraic expression, but got {result}"),
```
chriseth, line 324:
```suggestion
            _ => panic!("Expected array of algebraic expressions or an algebraic expression, but got {result}"),
```

pil-analyzer/tests/types.rs
```diff
@@ -516,3 +516,13 @@ fn enum_too_many_fields() {
     ";
     type_check(input, &[]);
 }
+
```
chriseth, line 519: Can you add another test that checks types for empty match arms, empty if/else blocks and just empty blocks inside fuctions? - or at least blocks with empty expression part.

chriseth (approved):

PR #1437, opened 2024-06-10
Fix instructions to run `vec_median`

### LINKED ISSUE #1387: Documentation inconsistency - Rust subcommand
Changed happend in the commit 101829b23075f20aeb8e1883dcc3a4bfb547891c caused this command to not be executable anymore
https://github.com/powdr-labs/powdr/blob/cbc5dd118e0098fc21b65569fc834715c7ccc9a5/riscv/tests/riscv_data/vec_median/src/lib.rs#L11

Schaeff:
Closes #1387, thanks @onurinanc

### DIFF
--- a/riscv/tests/riscv_data/vec_median/src/lib.rs
+++ b/riscv/tests/riscv_data/vec_median/src/lib.rs
@@ -8,7 +8,8 @@
 //! vector [15,75,6,5,1,4,7,3,2,9,2] and compare the result with the expected
 //! value of 5:
 //! ```
-//! cargo run --release rust riscv/tests/riscv_data/vec_median -o tmp -f -i 5,11,15,75,6,5,1,4,7,3,2,9,2
+//! cargo run -r --bin powdr-rs compile riscv/tests/riscv_data/vec_median -o tmp
+//! cargo run -r --bin powdr -- pil tmp/vec_median.asm -o tmp -i 5,11,15,75,6,5,1,4,7,3,2,9,2
 //! ```
 
 #![no_std]


riscv/tests/riscv_data/vec_median/src/lib.rs
```diff
@@ -8,7 +8,8 @@
 //! vector [15,75,6,5,1,4,7,3,2,9,2] and compare the result with the expected
 //! value of 5:
 //! ```
-//! cargo run --release rust riscv/tests/riscv_data/vec_median -o tmp -f -i 5,11,15,75,6,5,1,4,7,3,2,9,2
+//! cargo run -r --bin powdr-rs compile riscv/tests/riscv_data/vec_median -o tmp
+//! cargo run -r --bin powdr-rs execute tmp/vec_median.asm -i 5,11,15,75,6,5,1,4,7,3,2,9,2
```
lvella, line 12: These execution instructions are semantically different from the original one. To match the original behavior (i.e. create a ZK proof), the second line should read:
```
cargo run -r --bin powdr -- pil tmp/vec_median.asm -i 5,11,15,75,6,5,1,4,7,3,2,9,2
```

lvella (changes requested):

riscv/tests/riscv_data/vec_median/src/lib.rs
Schaeff: I can't see a proof file in `tmp` after running that, is it normal?

leonardoalt: there's no proof being requested in any of these commands

- `powdr-rs execute` in the PR just runs the RISCV executor
- Lucas' suggestion doesn't call `prove-with`

But the original test also didn't request a proof, so I guess it's fine

Schaeff: Ok so PR is fine without this change?

leonardoalt: hmm I would replace the second line by Lucas' suggestion to keep the same behavior as before

lvella: Ah, I was sure it would generate a proof with some default prover. Maybe this behavior changed a long time ago?

lvella (approved):

PR #1438, opened 2024-06-10
Use goldilocks from plonky2

chriseth:


leonardoalt:
Not a blocker, but why P2 and not P3?

leonardoalt (approved):

leonardoalt:
linter fails

leonardoalt (approved):

PR #1444, opened 2024-06-11
Avoid parentheses in `AlgebraicExpression` display

### LINKED ISSUE #1429: Implement Display function for BinaryOperation for AlgebraicExpression in a similar way as for Expression's BinaryOperation
We recently re-implemented the Display function for BinaryOperation so that it does not add parentheses unless needed, the same should be done for AlgebraicExpression.

Schaeff:
- Introduce structs in the analysed AST
- Port parenthesis minimisation from parsed AST to analysed AST 

Closes #1429


pil-analyzer/tests/parse_display.rs
```diff
@@ -61,7 +61,7 @@ namespace T(65536);
     col fixed p_read_X_pc = [0, 0, 0, 0, 0, 0, 0, 0, 0] + [0]*;
     col fixed p_reg_write_X_A = [0, 0, 0, 1, 0, 0, 0, 1, 0] + [0]*;
     col fixed p_reg_write_X_CNT = [1, 0, 0, 0, 0, 0, 0, 0, 0] + [0]*;
-    { T.pc, T.reg_write_X_A, T.reg_write_X_CNT } in (1 - T.first_step) { T.line, T.p_reg_write_X_A, T.p_reg_write_X_CNT };
+    { T.pc, T.reg_write_X_A, T.reg_write_X_CNT } in 1 - T.first_step { T.line, T.p_reg_write_X_A, T.p_reg_write_X_CNT };
```
Schaeff, line 64: I'm unsure about this one

Schaeff: But it's allowed in the parser before this change so it's unrelated to this PR.

ast/Cargo.toml
```diff
@@ -18,9 +18,13 @@ serde = { version = "1.0", default-features = false, features = ["alloc", "deriv
 schemars = { version = "0.8.16", features = ["preserve_order"]}
 serde_cbor = "0.11.2"
 auto_enums = "0.8.5"
+test-log = "0.2.12"
```
chriseth, line 21: Isn't this a dev dependency if it's only used in tests?

ast/src/analyzed/mod.rs
```diff
@@ -801,13 +803,119 @@ pub enum AlgebraicExpression<T> {
...
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct AlgebraicBinaryOperation<T> {
+    pub left: Box<AlgebraicExpression<T>>,
+    pub op: AlgebraicBinaryOperator,
+    pub right: Box<AlgebraicExpression<T>>,
+}
+impl<T> AlgebraicBinaryOperation<T> {
+    fn new(
```
chriseth, line 817: Do we need this if all attributes are `pub`?

chriseth: Ah it boxes - ok!

ast/src/analyzed/display.rs
```diff
@@ -378,3 +437,79 @@ impl Display for PolynomialReference {
...
+        let (input, expected) = &(&wrap(input), &wrap(expected));
+        let analyzed = analyze_string::<GoldilocksField>(input);
+        let printed = analyzed.to_string();
+
+        assert_eq!(expected.trim(), printed.trim());
+    }
+
+    #[test]
+    fn test_binary_op_parentheses() {
+        let test_cases: Vec<TestCase> = vec![
```
chriseth, line 479: Can you add a test that shows associativity of `**` and `*` is handled correctly?

chriseth:
Sorry, should have been clearer: I mainly meant `x**y**z`, `(x**y)**z` and `x**(y**z)`. (but the other now tests are also good)

Also one test is failing.

Schaeff:
I added a test for `(x ** y) ** z` and `x ** (y ** z)`, but if I understand correctly they are not produced by the analyzer, and it makes sense: `AlgebraicExpression` is a polynomial language, so these would not be legal. If that is the case, why not restrict the rhs of `**` to being a literal?

Also, assuming these are legal, I'm confused why in the parsed AST, we simplify `(x * y) * z` to `x * y * z` by left-assoc but not `x ** (y ** z)` to `x ** y ** z` by right assoc.

chriseth:
Ah ok, nice! Yeah, I guess the main problem is that we don't have a AlgebraicExpression parser directly, and for generic expressions we do want to support things like `x**(2**k)` where `k` is a constant (even if we probably won't have large exponents).

chriseth (approved):

Issue #1286, opened 2024-04-18
traits in pil


chriseth:
sub-issue of https://github.com/powdr-labs/powdr/issues/1016

Tasks:
- [x] parse traits
- [x] parse impls
- [ ] extend trait bounds to traits with multiple type variables (maybe using "where" clauses, but potentially not out-of-place)
- [ ] check that impls are unique (by trying to pairwise unify the type var arguments)
- [x] resolve trait impls on non-generic arguments

Implementation notes:

Trait bounds in the type inference need to be extended to `Trait<T1, T2, ...>`, not just `T: Trait` (which is shorthand for `Trait<T>`). This means substitutions need to happen inside all known trait bounds.

As soon as substitutions results in a trait bound having no variables any more, we check if there is an implementation for that concrete trait (it has to be unique since we checked before that every implementation is unique).

For generic symbols, we just add the remaining trait bounds to the signature.

For non-generic symbols, all trait bounds should have been substituted to bounds on concrete types already, so there is nothing more to do.

One open problem remains: After we have found a concrete implementation for a reference to a trait function, we should probably store that implementation in the AST so that the evaluator does not have to discover it. We can store it in the AST but I don't see a way to also print this pointer/reference. Maybe it's fine to keep it only in the AST since we will always run type inference on the code before we run the evaluator (at least on generic code).

PR #1450, opened 2024-06-11
Traits in PIL

gzanitti:
Impl of #1286.


analysis/src/machine_check.rs
```diff
@@ -337,6 +337,18 @@ impl TypeChecker {
                                 Item::TypeDeclaration(enum_decl),
                             );
                         }
+                        asm::SymbolValue::TraitDeclaration(trait_decl) => {
+                            res.insert(
+                                ctx.clone().with_part(&name),
+                                Item::TraitDeclaration(trait_decl),
+                            );
+                        }
+                        asm::SymbolValue::TraitImplementation(trait_impl) => {
```
chriseth, line 346: I think we have to move this out of the `SymbolDefinition` item into its own ModuleStatement because it does not define a name.

ast/src/parsed/display.rs
```diff
@@ -553,6 +557,68 @@ impl Display for FunctionDefinition {
...
+        write!(
+            f,
+            "trait {name} {type_vars} {{\n{methods}}}",
+            name = self.name,
+            methods = indent(self.methods.iter().map(|m| format!("{m},\n")).format(""), 1)
+        )
+    }
+}
+
+impl Display for TraitMethod<Expression> {
```
chriseth, line 576: I think since we do not have an implicit Self type, this should be TraitFunction instead.

ast/src/parsed/mod.rs
```diff
@@ -138,6 +140,18 @@ impl PilStatement {
                         .map(move |v| (name, Some(&v.name), SymbolCategory::TypeConstructor)),
                 ),
             ),
+            PilStatement::TraitDeclaration(_, TraitDeclaration { name, .. }) => {
+                Box::new(once((name, None, SymbolCategory::Type)))
+            }
+            PilStatement::TraitImplementation(_, TraitImplementation { name, methods, .. }) => {
+                Box::new(
+                    once((name, None, SymbolCategory::Value)).chain(
```
chriseth, line 148: I think this might be a type instead of a value

ast/src/parsed/mod.rs
```diff
@@ -138,6 +140,18 @@ impl PilStatement {
...
+                        methods
+                            .iter()
+                            .map(move |m| (name, Some(&m.name), SymbolCategory::Value)),
```
chriseth, line 151: Please update the documentation (it only talks about enums returning two items)

ast/src/parsed/mod.rs
```diff
@@ -306,6 +323,51 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
...
+    fn children(&self) -> Box<dyn Iterator<Item = &Expression> + '_> {
+        Box::new(self.methods.iter().flat_map(|m| m.body.children()))
+    }
+    fn children_mut(&mut self) -> Box<dyn Iterator<Item = &mut Expression> + '_> {
+        Box::new(self.methods.iter_mut().flat_map(|m| m.body.children_mut()))
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct ImplMethod<Expression> {
```
chriseth, line 366: I think this can be more general, it's just a named item of a `{}`-list.

importer/src/path_canonicalizer.rs
```diff
@@ -879,6 +910,55 @@ fn check_type(
...
+            ))
+        },
+    )?;
+
+    trait_impl
+        .methods
+        .iter()
+        .try_for_each(|ImplMethod { name, body }| {
+            check_expression(location, body, state, &HashSet::default()).map_err(|err| {
+                // Transformar el error e incluir el nombre del método
```
chriseth, line 932: :D

parser/src/lib.rs
```diff
@@ -529,4 +529,36 @@ namespace N(2);
         let printed = format!("{}", parse(Some("input"), input).unwrap_err_to_stderr());
         assert_eq!(expected.trim(), printed.trim());
     }
+
+    #[test]
+    fn parse_trait() {
+        let input = r#"
+    trait Add<T> {
+        add: T, T -> T,
+    }"#;
```
chriseth, line 538: Can you also add something with multiple parameters like
```
trait Iterator<S, I> {
  next: S -> (S, Option<I>),
}
```

parser/src/lib.rs
```diff
@@ -529,4 +529,36 @@ namespace N(2);
...
+
+        let printed = format!("{}", parse(Some("input"), input).unwrap_err_to_stderr());
+        assert_eq!(expected.trim(), printed.trim());
+    }
+
+    #[test]
+    fn parse_impl() {
+        let input = r#"
+    impl<T> Iterator<ArrayIterator<T>, T> {
```
chriseth, line 552: Can you add `T: Add, Q` or something?

parser/src/powdr.lalrpop
```diff
@@ -684,6 +694,44 @@ PatternList: Vec<Pattern> = {
     => vec![]
 }
 
+// ---------------------------- Trait/Impl Declaration -----------------------------
+
+TraitDeclaration: TraitDeclaration<Expression> = {
+    "trait" <name:Identifier> <type_vars:("<" <TypeVarBounds> ">")?> "{" <methods:TraitMethods> "}" => TraitDeclaration { name, type_vars: type_vars.unwrap_or_default(), methods }
```
chriseth, line 700: I'm wondering, maybe it is easier to not allow restricting the type vars in the trait declaration but only in the implementation?

parser/src/powdr.lalrpop
```diff
@@ -684,6 +694,44 @@ PatternList: Vec<Pattern> = {
...
+TraitMethod: TraitMethod<Expression> = {
+    <name:Identifier> ":" <params:TypeTermList> "->" <value:TypeTermBox> => TraitMethod { name, _type: Type::Function(FunctionType{params, value}) }
+}
+
+ImplDeclaration: TraitImplementation<Expression> = {
+    "impl" <type_vars:("<" <TypeVarBounds> ">")?> <name:Identifier> <trait_vars:("<" <TraitTypeList> ">")> "{" <methods:ImplMethods> "}" => TraitImplementation { name, type_vars: type_vars.unwrap_or_default(), trait_vars, methods }
+}
+
+// This should be represented in a better way
+TraitTypeList: Vec<(String, TypeBounds)> = {
```
chriseth, line 717: I think it's `TypeTermList`, isn't it?

ast/src/analyzed/display.rs
```diff
@@ -438,6 +455,50 @@ impl Display for PolynomialReference {
...
+                    Default::default()
+                } else {
+                    let formatted_elements: Vec<String> =
+                        scheme.types.iter().map(|t| format!("{t}")).collect();
+                    format!("<{}>", formatted_elements.join(", "))
+                }
+            });
+
+        let s = format!(
+            "impl{type_vars} {trait_name}{trait_vars} {{\n{methods}}}",
```
chriseth, line 484: There is the function `format_type_scheme_around_name` that could help here.

gzanitti: I don't think it will work, because trait_vars is a vector (and type_scheme has only one element).

This difference is the result of this separation between TypeScheme and TraitScheme:
https://github.com/powdr-labs/powdr/blob/e904efd693fccc3de3de7f40db8c438fe1e4e3e6/ast/src/parsed/types.rs#L355

Both could be unified if I model the current TypeScheme with a single element vector (ty), but I felt it would add a lot of code changes. What do you think?

chriseth:
I see that you are starting to work on the type checking part. I think this PR is already too big to be properly reviewed at the current stage. Could you maybe split out just the parsing of the trait declarations (not even the impls), so that we can get that merged already?

PR #1453, opened 2024-06-14
Translating RISC-V programs directly from ELF files.

lvella:
There are many advantages in using standard assemblers and linkers, like:
- maturity;
- more complete support of assembly language, which in turn allows for support to more high level languages and compilers;
- link-time optimizations;
- no need to deal with language edge cases (our assembler and linker deals with a number of rust stuff, but it would need even more to support `std`).

But it comes with a cost, as we need to lift references to text data back into labels, (because powdr operates on a higher abstraction level), and to do that, we need the ELF file to either be a PIE (Position Independent Executable), or to still have the linkage relocation tables (option `--emit-relocs` of GNU and LLVM linkers).


lvella:
TODO:
- [ ] Add an option in the `compile` command to select either asm path or elf path;
- [ ] Run all tests via elf path, too;
- [ ] Bring elf to feature parity with asm, in particular, load debug symbols;
- [ ] Add at least one specific test translating from PIE elf, since there is the code path to support it.

riscv/src/lib.rs
```diff
@@ -18,10 +18,10 @@ pub use crate::runtime::Runtime;
 pub mod asm;
 mod code_gen;
 pub mod continuations;
+pub mod elf;
 pub mod runtime;
 
-/// Compiles a rust file all the way down to PIL and generates
-/// fixed and witness columns.
+/// Compiles a rust file to Pordr asm.
```
leonardoalt, line 24:
```suggestion
/// Compiles a rust file to Powdr asm.
```

riscv/src/elf.rs
```diff
@@ -0,0 +1,1213 @@
...
+
+struct ElfProgram {
+    symbol_table: SymbolTable,
+    data_map: BTreeMap<u32, Data>,
+    text_labels: BTreeSet<u32>,
+    instructions: Vec<HighLevelInsn>,
+    entry_point: u32,
+}
+
+pub fn elf_translate<F: FieldElement>(
```
chriseth, line 41: What does it translate to? The return type is `String`, so it's a lot of guessing :)

riscv/src/elf.rs
```diff
@@ -0,0 +1,1213 @@
...
+    instruction::{Extensions, Instruction as Ins, OpcodeKind as Op},
+    Isa,
+};
+
+use crate::{
+    code_gen::{self, InstructionArgs, MemEntry, Register, RiscVProgram, Statement},
+    Runtime,
+};
+
+struct ElfProgram {
```
chriseth, line 33: Private below pub?

riscv/src/elf.rs
```diff
@@ -0,0 +1,1213 @@
...
+                rd: None,
+                rs1: None,
+                rs2: None,
+            } => Ok(()),
+            _ => Err(format!("Expected: no args, got {:?}", self.args)),
+        }
+    }
+}
+
+struct AddressMap<'a>(BTreeMap<u32, &'a program_header::ProgramHeader>);
```
chriseth, line 515: Is the key an address? What is the value?

riscv/src/elf.rs
```diff
@@ -0,0 +1,1213 @@
...
+        if let (ReadOrWrite::Write(refs), HighLevelImmediate::CodeLabel(addr)) =
+            (&mut self.referenced_text_addrs, &result.args.imm)
+        {
+            refs.insert(*addr);
+        }
+
+        Some(result)
+    }
+
+    fn map_one(&mut self, insn: MaybeInstruction) -> HighLevelInsn {
```
chriseth, line 847: Map one what? Or is it the literal 1?

lvella: This is a trait implementation. See trait `TwoOrOneMapper`.

riscv/src/elf.rs
lvella: Sorry, I don't understand what taboo I violated here. This struct should go below the pub function?

riscv/src/code_gen.rs
```diff
@@ -1024,47 +1038,46 @@ fn process_instruction<A: InstructionArgs + ?Sized>(
             )
         }
         "sw" => {
-            let (r1, r2, off) = args.rro()?;
-            vec![format!("mstore {r2} + {off}, {r1};")]
+            let (r2, r1, off) = args.rro()?;
+            vec![format!("mstore {r1} + {off}, {r2};")]
```
lvella, line 1042: These register renames is to make them in line with the instruction definition.

riscv/src/elf.rs
lvella: Added comments.

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+        }
+    }
+}
+
+/// Index the symbols by their addresses.
+struct SymbolTable(HashMap<u32, String>);
+
+impl SymbolTable {
+    fn new(elf: &Elf) -> SymbolTable {
+        // TODO: read the symbols from the debug information to be more comprehensive.
```
leonardoalt, line 244: still in this PR or later? If later, please open an issue

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+        let mut deduplicator = HashMap::new();
+        for sym in elf.syms.iter() {
+            if sym.st_name == 0 || (sym.st_type() != STT_OBJECT && sym.st_type() != STT_FUNC) {
```
leonardoalt, line 247: can you add a comment about what this if is filtering?

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+
+    /// Get the symbol or a default label formed from the address value.
+    fn get_as_string(&self, addr: u32) -> String {
+        self.get(addr).into_owned()
+    }
+}
+
+impl RiscVProgram for ElfProgram {
+    fn take_source_files_info(&mut self) -> impl Iterator<Item = crate::code_gen::SourceFileInfo> {
+        // TODO: read the source files from the debug information.
```
leonardoalt, line 282: here or later?

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+                    opc: Op::ADDI,
+                    rd: Some(rd_addi),
+                    rs1: Some(rs1_addi),
+                    imm: Some(lo),
+                    ..
+                },
+            ) if rd_lui == rs1_addi => {
+                // Sometimes, in non-PIE code, this pair of instructions is used
+                // to load an address into a register. We must check if this is
+                // the case, and if the address points to the text section, so
```
leonardoalt, line 713:
```suggestion
                // the case, and if the address points to the text section,
```

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+                    self.composed_immediate(*hi, *lo, *rd_lui, *rd_addi, insn2_addr, is_address)?;
+
+                HighLevelInsn {
+                    op,
+                    args,
+                    original_address,
+                }
+            }
+            (
+                // All other double instructions we can lift starts with auipc.
```
leonardoalt, line 726:
```suggestion
                // All other double instructions we can lift start with auipc.
```

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+                        panic!(
+                            "Unexpected instruction after AUIPC: {insn2:?} at {original_address:08x}"
+                        );
+                    }
+                }
+            }
+            _ => return None,
+        };
+
+        // TODO: implement here other kinds of RISC-V fusions as optimization.
```
leonardoalt, line 846: open issue?

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+        let Some(insn) = insn.insn else {
+            return HighLevelInsn {
+                op: "unimp",
+                args: Default::default(),
+                original_address,
+            };
+        };
+
+        let mut imm = match insn.opc {
+            // All jump instructions that have the immediate as an address
```
leonardoalt, line 868:
```suggestion
            // All jump instructions that have an address as immediate
```
?

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+                };
+            }
+            // All other instructions, which have the immediate as a value
+            _ => match insn.imm {
+                Some(imm) => HighLevelImmediate::Value(imm),
+                None => HighLevelImmediate::None,
+            },
+        };
+
+        // For some reason, atomic instructions comes with the immediate set to
```
leonardoalt, line 912:
```suggestion
        // For some reason, atomic instructions come with the immediate set to
```

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+        // zero instead of None (maybe to mimic assembly syntax? Who knows). Fix
+        // this.
```
leonardoalt, line 914: fix in this PR?

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+        if let Extensions::A = insn.extension {
+            assert!(matches!(imm, HighLevelImmediate::Value(0)));
+            imm = HighLevelImmediate::None;
+        }
+
+        // TODO: lift other instructions to their pseudoinstructions,
```
leonardoalt, line 920: TODO in this PR?

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+            advance = 4;
+            let insn = u32::from_le_bytes(
+                self.remaining_data[0..4]
+                    .try_into()
+                    .expect("Not enough bytes to complete a 32-bit instruction"),
+            )
+            .decode(Isa::Rv32)
+            .unwrap_or_else(|_| {
+                // TODO: maybe instead of failing we should just emit `unimp`.
+                // This way we would support the garbage default GNU binutils
```
leonardoalt, line 1013:
```suggestion
                // This way we would support the default GNU binutils
```
No need for that

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+                insn: match bin_instruction.decode(Isa::Rv32) {
+                    Ok(c_insn) => Some(to_32bit_equivalent(c_insn)),
+                    Err(raki::decode::DecodingError::IllegalInstruction) => {
+                        // Although not a real RISC-V instruction, sometimes 0x0000
+                        // is used on purpose as an illegal instruction (it even has
+                        // its own mnemonic "unimp"), so we support it here.
+                        // Otherwise, there is something more fishy going on, and we
+                        // panic.
+
+                        // TODO: we should just emit `unimp` for every unknown.
```
leonardoalt, line 1045: why not done?

riscv/Cargo.toml
```diff
@@ -24,12 +24,14 @@ powdr-pipeline.workspace = true
...
 
+goblin = { version = "0.8" }
 lazy_static = "1.4.0"
-itertools = "^0.10"
+itertools = "0.13"
 lalrpop-util = { version = "^0.19", features = ["lexer"] }
 log = "0.4.17"
 mktemp = "0.5.0"
 num-traits = "0.2.15"
+raki = { git = "https://github.com/powdr-labs/raki.git", branch = "patch-1" }
```
leonardoalt, line 34: Did you open a PR upstream? The change seemed small enough

cli-rs/src/main.rs
```diff
@@ -62,13 +62,17 @@ enum Commands {
         #[arg(long)]
         coprocessors: Option<String>,
 
+        /// Convert from the executable ELF file instead of the assembly.
+        #[arg(short = 'e', long)]
+        #[arg(default_value_t = false)]
+        via_elf: bool,
```
leonardoalt, line 68:
```suggestion
        elf: bool,
```
Looks nicer on the CLI to use `--elf`

leonardoalt (commented):
Many but mostly small comments

riscv/tests/instruction_tests/sources/rvc.S
```diff
@@ -0,0 +1,132 @@
...
+  RVC_TEST_CASE (35, ra, 0, \
+        la t0, 1f; \
+        li ra, 0; \
+        c.jr t0; \
+        c.j 2f; \
+      1:c.j 1f; \
+      2:j fail; \
+      1:)
+
+// Powdr note: tests 36 and 37 were removed, as they did arithmetic on text addresses, which we don't support.
```
pacheco, line 107: does it make sense to comment them out instead of removing? also, 38 and 39 are not present here either

riscv/tests/riscv_data/dynamic_relocation/dynamic_relocation.s
```diff
@@ -0,0 +1,53 @@
+    .section .data
+    .align 4
+    .global data_section
+
```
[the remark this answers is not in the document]
lvella: If you call ChatGPT "manually"...

riscv/tests/instructions.rs
```diff
@@ -1,17 +1,29 @@
...
     use powdr_backend::BackendType;
     use powdr_number::GoldilocksField;
     use powdr_riscv::asm::compile;
     use powdr_riscv::Runtime;
     use test_log::test;
 
-    fn run_instruction_test(assembly: &str, name: &str) {
+    fn run_instruction_test(path: &Path, name: &str) {
+        // Test from ELF path:
+        verify_riscv_asm_file(path, &Runtime::base(), false);
```
pacheco, line 15: nitpick as this is only used here i think: have `elf` in the function name somehow?

riscv/tests/instructions.rs
```diff
@@ -1,17 +1,29 @@
...
+
+        if name == "rvc" {
+            // "rvc" test is not via assembly path
```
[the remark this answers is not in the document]
lvella: Because C extension mnemonics are not supported by our assembler. Also, see changes the riscv/tests/README.md

riscv/tests/instruction_tests/sources/rvc.S
```diff
@@ -0,0 +1,132 @@
+# See LICENSE for license details.
+
+#*****************************************************************************
+# rvc.S
+#-----------------------------------------------------------------------------
+#
+# Test RVC corner cases.
```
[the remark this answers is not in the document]
lvella: Mostly from the official testsuite, but I left some comments on the modifications I had to make.

riscv/src/elf.rs
lvella: This return the powdr asm contents.

riscv/src/elf.rs
lvella: Address to program section.

riscv/tests/common/mod.rs
```diff
@@ -35,3 +40,60 @@ pub fn verify_riscv_asm_string<S: serde::Serialize + Send + Sync + 'static>(
     );
     verify_pipeline(pipeline, backend).unwrap();
 }
+
+fn find_assembler() -> &'static str {
+    let options = ["riscv64-elf-as", "riscv64-unknown-elf-as"];
```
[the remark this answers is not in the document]
lvella: I am sure.

riscv/src/elf.rs
lvella: This is already big enough, so I think one later.

riscv/src/elf.rs
lvella: I was pretty sure there was an issue already, but I couldn't find it. Created a new one: #1540

cli-rs/src/main.rs
lvella: I was thinking into replacing this with `--asm`, and set ELF as default. This way it would see more usage and we could be more confident about removing the asm path.

riscv/src/code_gen.rs
```diff
@@ -539,16 +537,18 @@ fn memory(with_bootloader: bool) -> String {
...
+    /// Special case used in amo* instructions, returning rd, rs2, rs1
+    fn rrr2(&self) -> Result<(Register, Register, Register), Self::Error>;
     fn ri(&self) -> Result<(Register, u32), Self::Error>;
     fn rr(&self) -> Result<(Register, Register), Self::Error>;
-    fn rrl(&self) -> Result<(Register, Register, String), Self::Error>;
-    fn rl(&self) -> Result<(Register, String), Self::Error>;
+    fn rrl(&self) -> Result<(Register, Register, impl AsRef<str>), Self::Error>;
+    fn rl(&self) -> Result<(Register, impl AsRef<str>), Self::Error>;
     fn rro(&self) -> Result<(Register, Register, u32), Self::Error>;
-    fn rrro(&self) -> Result<(Register, Register, Register, u32), Self::Error>;
```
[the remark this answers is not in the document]
lvella: `rrro` stands for "register, register, register, offset", but there is no RISC-V instruction that encodes this much information (there is not enough space in the 32 bits). It existed because the assembly syntax for these instructions allowed for writing `amoadd.d rd, rs2, 0(rs1)`, but there is no place to put that `0` in the actual encoding (if I try to assemble with any other value I get an error).

So, at first I replaced `rrro()` with `rrr()`, and changed it to accept both `None` and 0 offset. Then I figured out that the order of the registers expected by these instructions was different from other instructions using `rrr()`, so I created `rrr2()` (ok, the name could be better) to say that this is an instruction that expects 3 registers, but in a different order from `rrr()`.

riscv/src/lib.rs
```diff
@@ -132,50 +172,140 @@ pub fn compile_riscv_asm<T: FieldElement>(
...
 ];
 
-pub fn compile_rust_crate_to_riscv_asm(
+pub fn compile_rust_crate_to_riscv(
     input_dir: &str,
     output_dir: &Path,
-) -> BTreeMap<String, String> {
-    // We call cargo twice, once to get the build plan json, so we know exactly
-    // which object file to use, and once to perform the actual building.
+) -> (Option<PathBuf>, BTreeMap<String, PathBuf>) {
```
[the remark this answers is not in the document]
lvella: If the user crate is a lib instead of an executable, there is no final elf to translate, but this is still supported by the assembly path.

riscv/tests/instruction_tests/README.md
```diff
@@ -9,65 +9,61 @@ kind of arithmetic over `.text` label and addresses, nor alignment or spacing
 directives in `.text` sections. Most unsupported instructions are related to
 this limitation.
 
+One consequence of this limitation is that all the supported uses of `auipc` and
+`lui`, when referring to a `.text` label, is when they can be fused with the
+next instruction to make use of the loaded address immediately (e.g. `addi` or
```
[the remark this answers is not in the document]
lvella: I'll try to write it better. But using `auipc` and `lui` to load a `.text` address is only supported if it can be fused with the next instruction. It works fine if they are used to load a data address.

riscv/tests/instruction_tests/sources/rvc.S
```diff
@@ -0,0 +1,132 @@
...
+RVTEST_CODE_BEGIN
+
+  .align 2
+  .option push
+  .option norvc
+
+  #define RVC_TEST_CASE(n, r, v, code...) \
+    TEST_CASE (n, r, v, .option push; .option rvc; code; .align 2; .option pop)
+
+  // Powdr note: there was a test case from the original riscv-tests here,
```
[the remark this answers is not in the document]
lvella: Yes, because it worked.

riscv/tests/instruction_tests/sources/rvc.S
lvella: 38 and 39 were already missing. Maybe they are RISC-V 64 specific?

riscv/tests/instructions.rs
lvella: But this also executes the test from the assembly path.

riscv/tests/riscv.rs
```diff
@@ -356,22 +384,45 @@ fn verify_riscv_crate_with_data<S: serde::Serialize + Send + Sync + 'static>(
...
+fn verify_riscv_crate_from_both_paths<S: serde::Serialize + Send + Sync + 'static>(
+    case: &str,
+    inputs: Vec<GoldilocksField>,
+    runtime: &Runtime,
+    data: Option<Vec<(u32, S)>>,
+    backend: BackendType,
+) {
     let temp_dir = Temp::new_dir().unwrap();
-    let riscv_asm = powdr_riscv::compile_rust_crate_to_riscv_asm(
+    println!("Directory: {}", temp_dir.display());
```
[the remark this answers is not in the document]
lvella: fixed

riscv/tests/instructions.rs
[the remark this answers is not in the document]
lvella: fixed

riscv/tests/instruction_tests/README.md
```diff
@@ -9,65 +9,61 @@ kind of arithmetic over `.text` label and addresses, nor alignment or spacing
 directives in `.text` sections. Most unsupported instructions are related to
 this limitation.
 
+One consequence of this limitation is that all the supported uses of `auipc` and
+`lui`, when referring to a `.text` label, is when they can be fused with the
```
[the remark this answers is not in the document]
lvella: I rewrote this part.

riscv/tests/instruction_tests/README.md
```diff
@@ -9,65 +9,61 @@ kind of arithmetic over `.text` label and addresses, nor alignment or spacing
...
-directly the assembly file, not the binary format. All the alignment and precise
-spacing of the `.text` section in this test doesn't make sense for Powdr.
+- `rvc`
 
-We just compile to riscv32imac instead of riscv32ima because the later is not
-supported by rust.
+The RISC-V "C" test is only supported when converting from the ELF executable,
+and only a modified version of the test is supported, because the original
+inserts data in the middle of the text section, which breaks our expected
+separation between data and code address space.
```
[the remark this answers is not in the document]
lvella: Done.

riscv/src/elf.rs
lvella: done

riscv/src/elf.rs
lvella: I removed this TODO because we already have one on the topic of debug symbols

riscv/src/elf.rs
lvella: The fix is the code following. I reworded.

riscv/src/elf.rs
lvella: No. This is big enough.

riscv/src/elf.rs
lvella: It can make the generated code very big and may mask other issues. I reworded the TODO.

lvella: Also, there are unanswered questions: should this data be readable from the program itself? What about the valid instructions?

cli-rs/src/main.rs
lvella: I have changed to ELF being the default.

riscv/Cargo.toml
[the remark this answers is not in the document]
lvella: done

riscv/src/asm/mod.rs
```diff
@@ -91,13 +89,17 @@ impl InstructionArgs for [Argument] {
 
     fn rrr(&self) -> Result<(Register, Register, Register), &'static str> {
         match self {
-            [Argument::Register(r1), Argument::Register(r2), Argument::Register(r3)] => {
-                Ok((*r1, *r2, *r3))
-            }
+            [Argument::Register(r1), Argument::Register(r2), Argument::Register(r3)
+            | Argument::RegOffset(None | Some(Expression::Number(0)), r3)] => Ok((*r1, *r2, *r3)),
```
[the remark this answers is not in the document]
lvella: See https://github.com/powdr-labs/powdr/pull/1453#discussion_r1666773962.

But indeed, for consistency `rr` should be able to handle the 0 case when decoding the instruction `lr.w rd, 0(rs1)`, but we have encountered this case in handwritten assembly (the test for this instruction, `lrsc.S`, omits the 0), and the compiler doesn't place the zero there, either.

I fixed it.

riscv/tests/riscv_data/vec_median/src/main.rs
```diff
@@ -21,8 +21,9 @@ use alloc::vec::Vec;
 use powdr_riscv_runtime::io::read_u32;
 use powdr_riscv_runtime::print;
 
+/// entry point called by the runtime
 #[no_mangle]
-fn main() {
```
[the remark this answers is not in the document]
lvella: This is not needed, and I don't know why I changed it. Reverted.

riscv/src/lib.rs
```diff
@@ -99,16 +120,35 @@ pub fn compile_riscv_asm_bundle<T: FieldElement>(
...
 
     fs::write(powdr_asm_file_name.clone(), &powdr_asm).unwrap();
     log::info!("Wrote {}", powdr_asm_file_name.to_str().unwrap());
 
     Some((powdr_asm_file_name, powdr_asm))
 }
 
-/// Compiles a riscv asm file all the way down to PIL and generates
-/// fixed and witness columns.
+#[allow(clippy::print_stderr)]
```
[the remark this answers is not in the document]
lvella: The comment was outdated, but I have no idea why where this line came from. I am removing it.

riscv/src/lib.rs
```diff
@@ -37,51 +39,70 @@ pub fn compile_rust<T: FieldElement>(
...
+        )
+    }
 }
 
-#[allow(clippy::print_stderr)]
-pub fn compile_riscv_asm_bundle<T: FieldElement>(
+fn compile_program<P>(
     original_file_name: &str,
-    riscv_asm_files: BTreeMap<String, String>,
+    input_program: P,
```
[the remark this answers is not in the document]
lvella: I don't think it is a good idea. There are a lot of functions with more or less this same order of arguments.

riscv/src/lib.rs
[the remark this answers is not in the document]
lvella: So, yes.

riscv/src/lib.rs
```diff
@@ -132,50 +172,140 @@ pub fn compile_riscv_asm<T: FieldElement>(
...
+                );
+            }
+        }
     }
-    assemblies
+
+    (executable, assemblies)
+}
+
+const CARGO_TARGET_DIR: &str = "cargo_target";
```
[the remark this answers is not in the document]
lvella: fixed

riscv/tests/riscv.rs
```diff
@@ -252,6 +267,20 @@ fn two_sums_serde() {
     );
 }
 
+const DYNAMIC_RELOCATION_S: &str = "tests/riscv_data/dynamic_relocation/dynamic_relocation.s";
+
+#[ignore = "Too slow"]
+#[test]
+fn dynamic_relocation_pie() {
```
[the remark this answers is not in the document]
lvella: Done. Also renamed.

riscv/tests/riscv_data/dynamic_relocation/dynamic_relocation.s
```diff
@@ -0,0 +1,53 @@
+    .section .data
```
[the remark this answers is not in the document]
lvella: Done. Also renamed.

riscv/tests/instruction_tests/sources/rvc.S
lvella: I restored the removed tests, but commented out.

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+struct ElfProgram {
+    symbol_table: SymbolTable,
+    data_map: BTreeMap<u32, Data>,
+    text_labels: BTreeSet<u32>,
+    instructions: Vec<HighLevelInsn>,
+    entry_point: u32,
+}
+
+/// Generates a Powdr Assembly program from a RISC-V 32 executable ELF file.
+pub fn elf_translate<F: FieldElement>(
```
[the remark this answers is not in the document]
lvella: Done.

riscv/src/code_gen.rs
```diff
@@ -79,10 +77,10 @@ pub trait RiscVProgram {
...
         &mut self,
-    ) -> impl Iterator<Item = Statement<impl AsRef<str>, Self::Args>>;
+    ) -> impl Iterator<Item = Statement<impl AsRef<str>, impl InstructionArgs>>;
 
     /// The name of the function that should be called to start the program.
-    fn start_function(&self) -> &str;
+    fn start_function(&self) -> impl AsRef<str>;
 }
 
 /// Translates a RISC-V program to POWDR ASM.
```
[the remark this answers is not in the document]
lvella: It is still true...

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+    // Keep a list of referenced text addresses, so we can generate the labels.
+    let mut referenced_text_addrs = BTreeSet::from([elf.entry as u32]);
+
+    // Find the text addresses referenced from text sections and load the data sections.
+    let mut data_map = BTreeMap::new();
+    for (&addr, &p) in address_map.0.iter() {
+        let section_data = &file_buffer[p.p_offset as usize..(p.p_offset + p.p_filesz) as usize];
+
+        // Test if executable
+        if p.p_flags & 1 == 1 {
```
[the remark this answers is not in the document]
lvella: There is no such method. But I improved it by using the named constant.

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+                imm,
+            },
+            original_address,
+        }
+    }
+}
+
+/// Find all the references to text addresses in the instructions and add them
+/// to the set.
+fn search_text_addrs(
```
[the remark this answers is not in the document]
lvella: The reason to have the `TwoOrOneMapper` trait is to be able to write the algorithm in `try_map_two_by_two()` independently of types, so it is was easier to write and reason about.

I don't see how to get rid of `ReadOrWrite` enum without replacing with something very similar. The lifting algorithm has to run twice, once just collecting the references, and another making use of them.

riscv/src/elf.rs
```diff
@@ -0,0 +1,1229 @@
...
+    /// Get the symbol or a default label formed from the address value.
+    fn get(&self, addr: u32) -> Cow<str> {
+        self.0
+            .get(&addr)
+            .map(|name| Cow::Borrowed(name.as_str()))
+            .unwrap_or_else(|| Cow::Owned(format!("L{addr:08x}")))
+    }
+
+    /// Get the symbol or a default label formed from the address value.
+    fn get_as_string(&self, addr: u32) -> String {
```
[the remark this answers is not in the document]
lvella: Fixed.

riscv/src/lib.rs
```diff
@@ -99,16 +120,35 @@ pub fn compile_riscv_asm_bundle<T: FieldElement>(
...
+        output_dir,
+        force_overwrite,
+        runtime,
+        with_bootloader,
+        asm::compile::<T>,
+    )
+}
+
+/// Compiles a riscv asm file to Powdr asm.
 pub fn compile_riscv_asm<T: FieldElement>(
```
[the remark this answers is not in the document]
lvella: Removed.

riscv/src/code_gen.rs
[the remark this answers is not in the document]
pacheco: i see, i think name is fine, can you add a small comment?

riscv/tests/instructions.rs
pacheco: ah right, i got confused by the `asm` in the name, but asm here is from the input not output

riscv/src/elf.rs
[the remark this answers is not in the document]
pacheco: https://github.com/m4b/goblin/blob/617775898fdfd0b843b2e937c80e455cf6f2a637/src/elf/program_header.rs#L150
isn't this it?

leonardoalt:
CI failing

riscv/src/lib.rs
```diff
@@ -132,50 +172,140 @@ pub fn compile_riscv_asm<T: FieldElement>(
...
+}
+
 macro_rules! as_ref [
     ($t:ty; $($x:expr),* $(,)?) => {
         [$(AsRef::<$t>::as_ref(&$x)),+]
     };
 ];
 
-pub fn compile_rust_crate_to_riscv_asm(
+pub fn compile_rust_crate_to_riscv(
```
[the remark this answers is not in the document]
lvella: Done.

riscv/src/code_gen.rs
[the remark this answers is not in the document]
lvella: There are comments on both `rrr()` and `rrr2()` explaining the difference between them.

riscv/src/elf.rs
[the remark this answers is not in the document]
lvella: Well, rust-analyzer didn't find it, but I'll try again.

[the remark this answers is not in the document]
lvella: Done.

riscv/src/elf.rs
lvella: Moved to below pub just in case that is what you meant.

lvella (commented):
a

riscv/tests/instruction_tests/README.md
```diff
@@ -57,13 +57,11 @@ These instructions are not yet implemented.
...
 ## From the "C" (compressed) extension (rv32uc):
 
 - `rvc`
 
-The RISC-V "C" test is only supported when converting from the ELF executable,
-and only a modified version of the test is supported, because the original
-inserts data in the middle of the text section, which breaks our expected
-separation between data and code address space.
+The RISC-V "C" test file is only partially supported when converting from the
+ELF executable. Some tests that we don't support were disable.
```
leonardoalt, line 67:
```suggestion
ELF executable. Some tests that we don't support were disabled.
```

riscv/src/lib.rs
```diff
@@ -198,10 +173,25 @@ macro_rules! as_ref [
     };
 ];
 
-pub fn compile_rust_crate_to_riscv(
-    input_dir: &str,
-    output_dir: &Path,
-) -> (Option<PathBuf>, BTreeMap<String, PathBuf>) {
+pub struct CompilationResult {
+    pub executable: Option<PathBuf>,
+    assemblies: BTreeMap<String, PathBuf>,
```
leonardoalt, line 178: why just make one `pub` and the other not?

cli-rs/src/main.rs
```diff
@@ -359,9 +359,14 @@ fn compile_riscv_asm<F: FieldElement>(
         None => powdr_riscv::Runtime::base(),
     };
 
-    powdr_riscv::compile_riscv_asm::<F>(
+    powdr_riscv::compile_riscv_asm_bundle::<F>(
```
leonardoalt, line 362: is elf still the default compilation path here?

riscv/src/lib.rs
lvella: because there is no methog to get the `pub` one

cli-rs/src/main.rs
lvella: There is no ELF path in this function, this only from assembly.

leonardoalt:
The other thing that's lost are labels, which I assume are useful for the profiler? @pacheco 

For example, for the `trivial` test we get via asm:

```rust
    main:
    trivial_dash_be3334ad2e9c9ee9___dot_Lfunc_begin0:
    .debug loc 338 7 17;
    tmp1 <== jump_dyn(x1);
```

and via elf:

```rust
    main:
    tmp1 <== jump_dyn(x1);
```

leonardoalt (approved):

Issue #711, opened 2023-10-20
Improve bit constraint detection


chriseth:
We currently don't properly detect bit/range constraints for the following cases:
```
let bytes = |i| i % 256;
Z * (Z - 1) * (Z - 2) * (Z - 3) = 0;
{ X } in { bytes };
{ X * 2**6 } in { bytes };
```

Both `X` and `Z` should be bit-constrained to `0x2` and value-constrained to `[0,4]`.

We currently properly detect cases of `Y * (Y - 1) = 0`, so we should extend that detection to the general case of products over `X - a_k`.

For the second case, we have to first derive a bit-constraint on `{ X } in { bytes }` and then use that to refine the bit-constraint using `{ X * 2**6 } in { bytes };`

chriseth:
https://github.com/powdr-labs/powdr/pull/720 implements access to already determined global constraints on fixed columns. So this could be used to implement deriving bit constraints on `{ X * 2**6 } in { bytes }` even if the lookup is conditional.

PR #1461, opened 2024-06-23
Improved bit constraint detection

aminlatifi:
- Ref #711

### DIFF
--- a/executor/src/witgen/global_constraints.rs
+++ b/executor/src/witgen/global_constraints.rs
@@ -266,7 +266,21 @@ fn propagate_constraints<T: FieldElement>(
     (known_constraints, remove)
 }
 
-pub fn get_roots<T: FieldElement>(
+/// Get simple expression roots for a given expression.
+///
+/// This function recursively evaluates an algebraic expression and attempts
+/// to extract roots from it. The roots are returned as a vector of evaluation
+/// values containing references to algebraic expressions and field elements.
+///
+/// # Parameters
+///
+/// - `expr`: A reference to an `Expression` of type `T`.
+///
+/// # Returns
+///
+/// An `Option` containing a vector of `EvalValue<&AlgebraicReference, T>`
+/// if the roots can be successfully extracted, otherwise `None`.
+pub fn find_algebraic_roots<T: FieldElement>(
     expr: &Expression<T>,
 ) -> Option<Vec<crate::witgen::EvalValue<&AlgebraicReference, T>>> {
     match expr {
@@ -275,8 +289,8 @@ pub fn get_roots<T: FieldElement>(
             op: AlgebraicBinaryOperator::Mul,
             right,
         }) => {
-            let right_root = get_roots(right)?;
-            let mut left_roots = get_roots(left)?;
+            let right_root = find_algebraic_roots(right)?;
+            let mut left_roots = find_algebraic_roots(left)?;
 
             if !left_roots.is_empty() && !right_root.is_empty() {
                 if let ([(id1, Constraint::Assignment(_))], [(id2, Constraint::Assignment(_))]) = (
@@ -295,19 +309,20 @@ pub fn get_roots<T: FieldElement>(
         Expression::Number(_) => Some(vec![]),
         _ => {
             let symbolic_ev = SymbolicEvaluator;
-            let _root = ExpressionEvaluator::new(symbolic_ev)
+            let root = ExpressionEvaluator::new(symbolic_ev)
                 .evaluate(expr)
                 .ok()
                 .and_then(|l| l.solve().ok())?;
 
-            match _root.constraints[..] {
+            match root.constraints[..] {
                 [] => Some(vec![]),
-                [(id, Constraint::Assignment(_))] if id.is_witness() => Some(vec![_root]),
+                [(id, Constraint::Assignment(_))] if id.is_witness() => Some(vec![root]),
                 _ => None,
             }
         }
     }
 }
+
 /// Tries to find "X * (1 - X) = 0"
 fn is_binary_constraint<T: FieldElement>(
     expr: &Expression<T>,
@@ -330,7 +345,7 @@ fn is_binary_constraint<T: FieldElement>(
         right: _,
     }) = expr
     {
-        if let Some(roots) = get_roots(expr) {
+        if let Some(roots) = find_algebraic_roots(expr) {
             // For setting mask range constaints at least two roots is needed.
             if roots.len() > 1 {
                 let poly_id = roots[0].constraints[0].0.poly_id;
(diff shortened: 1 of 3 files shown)


executor/src/witgen/global_constraints.rs
```diff
@@ -268,8 +266,52 @@ fn propagate_constraints<T: FieldElement>(
     (known_constraints, remove)
 }
 
+pub fn get_roots<T: FieldElement>(
```
chriseth, line 269: Can you document this please?

executor/src/witgen/global_constraints.rs
```diff
@@ -268,8 +266,52 @@ fn propagate_constraints<T: FieldElement>(
...
+    expr: &Expression<T>,
+) -> Option<Vec<crate::witgen::EvalValue<&AlgebraicReference, T>>> {
```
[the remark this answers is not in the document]
chriseth: Ah ok you are also checking the variable, but then shouldn't it be `Option<(&AlgebraicReference, vec<T>)>`?

executor/src/witgen/global_constraints.rs
```diff
@@ -268,8 +266,52 @@ fn propagate_constraints<T: FieldElement>(
...
+                }
+            }
+
+            left_roots.extend(right_root);
+            Some(left_roots)
+        }
+        Expression::Number(_) => Some(vec![]),
+        _ => {
+            let symbolic_ev = SymbolicEvaluator;
+            let _root = ExpressionEvaluator::new(symbolic_ev)
```
chriseth, line 298: Please don't user underscores like that, we won't get "unused" warnings if you do it.

executor/src/witgen/global_constraints.rs
```diff
@@ -268,8 +266,67 @@ fn propagate_constraints<T: FieldElement>(
     (known_constraints, remove)
 }
 
+/// Get simple expression roots for a given expression.
```
chriseth, line 269: Please tell me if the following is correct:

Tries to return the set of roots of a given expression in one variable.

If `Some(vec![r1, r2, ... ,rn])` is returned, then the expression is equivalent to `(x - r1)*(x - r2) * ... * (x - rn)`.

Please also change this function to return a tuple of the variable an the vector of roots. Or do you think this return type is better suited in some way?

executor/src/witgen/global_constraints.rs
```diff
@@ -268,8 +266,67 @@ fn propagate_constraints<T: FieldElement>(
...
+///
+/// # Parameters
+///
+/// - `expr`: A reference to an `Expression` of type `T`.
+///
+/// # Returns
+///
+/// An `Option` containing a vector of `EvalValue<&AlgebraicReference, T>`
+/// if the roots can be successfully extracted, otherwise `None`.
+pub fn find_algebraic_roots<T: FieldElement>(
```
chriseth, line 283: About the name of the function - maybe a bit nitpicking: A root is always algebraic, so I think `find_roots` should be ok.

executor/src/witgen/global_constraints.rs
chriseth: Oh, the above is actually not true, it should be equivalent to a nonzero multiple of the product.

executor/src/witgen/global_constraints.rs
```diff
@@ -268,8 +266,67 @@ fn propagate_constraints<T: FieldElement>(
...
+        Expression::Number(_) => Some(vec![]),
+        _ => {
+            let symbolic_ev = SymbolicEvaluator;
+            let root = ExpressionEvaluator::new(symbolic_ev)
+                .evaluate(expr)
+                .ok()
+                .and_then(|l| l.solve().ok())?;
+
+            match root.constraints[..] {
+                [] => Some(vec![]),
```
chriseth, line 318: I think in this case, the expression evaluates to zero and this is a very important specila case: While the numbers returned are all roots in that case, the list is not complete (all numbers are roots).

executor/src/witgen/global_constraints.rs
```diff
@@ -214,10 +214,8 @@ fn propagate_constraints<T: FieldElement>(
     let mut remove = false;
     match identity.kind {
         IdentityKind::Polynomial => {
-            if let Some(p) = is_binary_constraint(identity.expression_for_poly_id()) {
-                assert!(known_constraints
-                    .insert(p, RangeConstraint::from_max_bit(0))
-                    .is_none());
+            if let Some((p, constraint)) = is_binary_constraint(identity.expression_for_poly_id()) {
```
chriseth, line 217: Now to use the root-finding function in the most general case (that includes the binary one), you should do the following:

Set the constraint to from_range(min, max) and set `remove` to true if all values from min to max occur in the list.

executor/src/witgen/global_constraints.rs
aminlatifi: So you mean we must return None and don't make any constraint?

executor/src/witgen/global_constraints.rs
aminlatifi: > If Some(vec![r1, r2, ... ,rn]) is returned, then the expression is equivalent to (x - r1)*(x - r2) * ... * (x - rn).

This one is correct. If I understood correctly, you need to support (x-r1)*(x-r2)*...*(y-r'1)*(y-r'2)*...
and return [(x, [r1,r2,...]),(y,[r'1, r'2,...])]

executor/src/witgen/global_constraints.rs
aminlatifi: Done,

executor/src/witgen/global_constraints.rs
aminlatifi: I didn't get the it well, can you give more details? You mean I don't use `from_mask` and use `from_range` instead? and don't try to make a binary mask which covers all numbers?

executor/src/witgen/global_constraints.rs
chriseth: No, we cannot support that, for the pattern to match, there can only even be one variable.

executor/src/witgen/global_constraints.rs
chriseth: Yes

executor/src/witgen/global_constraints.rs
chriseth: exactly

executor/src/witgen/global_constraints.rs
```diff
@@ -266,7 +266,21 @@ fn propagate_constraints<T: FieldElement>(
     (known_constraints, remove)
 }
 
-pub fn get_roots<T: FieldElement>(
+/// Get simple expression roots for a given expression.
+///
+/// This function recursively evaluates an algebraic expression and attempts
```
chriseth, line 271: This function is very complex, so its description has to be water-tight. It has to define what a root is and so on. I think a wording in the from of "if X is returned, it means the expression is equivalent to ..." would be helpful since it leaves little room for misunderstanding. Also, I don't think we need to explain _how_ the function works, just what it returns.

leonardoalt:
We're moving away from runtime witgen solving in favor of JIT, so I'm closing this for now. Thanks for the contribution!

PR #1467, opened 2024-06-25
merge compatible links from different instructions

pacheco:
Merge compatible `link`s into a single permutation/lookup.
We only consider merging links from different instructions, as a single instruction can be active at a time.
Links with next references are ignored due to a limitation in witgen (left a TODO so its easily fixed upon witgen support)


airgen/src/lib.rs
```diff
@@ -295,11 +309,128 @@ impl<'a> ASMPILConverter<'a> {
...
+                };
+
+                // TODO: temporary solution to a witgen limitation: avoid merging links that use next references.
+                // Remove this when witgen supports it (and update related tests).
+                use powdr_ast::parsed::{
+                    visitor::ExpressionVisitable, UnaryOperation, UnaryOperator::Next,
+                };
+                use Expression::UnaryOperation as EUnaryOperation;
+                if link.from.params.inputs_and_outputs().any(|p| {
+                    p.expr_any(|e| matches!(e, EUnaryOperation(_, UnaryOperation { op: Next, .. })))
```
chriseth, line 367:
```suggestion
                    p.expr_any(|e| e.contains_next_ref())
```

chriseth: But please check if expr_any already traverses the tree (beacuse contains_next_ref also does)

airgen/src/lib.rs
```diff
@@ -295,11 +309,128 @@ impl<'a> ASMPILConverter<'a> {
...
+        // idea here is:
+        // - group by LinkInfo
+        // - inside each group, separate links into sets of with mutually exclusive flags (that is, from different instructions)
+        let mut grouped_links: BTreeMap<LinkInfo, Vec<BTreeMap<Expression, Link>>> =
+            Default::default();
+        for link_def in defs.iter() {
+            let link = self.handle_link_def(link_def.clone());
+
+            // group compatible links
+            if link.from.instr_flag.is_some() {
```
chriseth, line 352: Can't you use `filter()` on these? It's quite a long if in an already long function.

airgen/src/lib.rs
```diff
@@ -295,11 +309,128 @@ impl<'a> ASMPILConverter<'a> {
...
+
+        // resulting links
+        let mut links = vec![];
+
+        // idea here is:
+        // - group by LinkInfo
+        // - inside each group, separate links into sets of with mutually exclusive flags (that is, from different instructions)
+        let mut grouped_links: BTreeMap<LinkInfo, Vec<BTreeMap<Expression, Link>>> =
+            Default::default();
+        for link_def in defs.iter() {
```
chriseth, line 348: I have the impression that this would be better readable if you
- partition the links into groupable and non-groupable, this could be as simple as mapping all links into pairs of `(Option<LinkIndo>, Link)`, where non-groupable are mapped to None in the first component.
- then you collect that into a BTreeMap as you already do and then
- process the None-group by just returning it and
- process the other groups by merging

airgen/src/lib.rs
```diff
@@ -295,11 +309,128 @@ impl<'a> ASMPILConverter<'a> {
...
     }
+
+    /// Process each link and then combine compatible links.
+    /// Links can be merged iff:
+    /// - they originate from the same machine instance
+    /// - they target the same instance.operation
+    /// - they are of the same kind (permutation/lookup)
+    /// - their flags are mutually exclusive
+    /// Right now we only consider links from different instructions,
+    /// as a single instruction can be active at a time
```
chriseth, line 328: Could you maybe (can also be inside the function) explain what merging links does (with an example)?

ast/src/asm_analysis/mod.rs
```diff
@@ -64,10 +64,12 @@ pub struct InstructionDefinitionStatement {
 }
 
 #[derive(Clone, Debug)]
-pub struct LinkDefinitionStatement {
+pub struct LinkDefinition {
     pub source: SourceRef,
-    /// the flag which activates this link. Should be boolean.
-    pub flag: Expression,
+    /// the instruction flag, if this is an instruction link. Should be boolean.
```
chriseth, line 69: Can you explain what the difference between the two flags is?

linker/src/lib.rs
```diff
@@ -185,7 +187,9 @@ fn process_link(link: Link) -> PilStatement {
     } else {
         // plookup lhs is `flag { operation_id, inputs, outputs }`
         let lhs = SelectedExpressions {
-            selector: Some(from.flag),
+            selector: from
+                .instr_flag
+                .map_or(Some(from.link_flag.clone()), |f| Some(f * from.link_flag)),
```
chriseth, line 192: This is the same code as above (and I think I have seen similar code somewhere else) - can this be a function?

linker/src/lib.rs
```diff
@@ -672,7 +679,7 @@ machine Main {
     pol constant p_read_X_pc = [0]*;
     pol constant p_reg_write_X_A = [0]*;
     { pc, reg_write_X_A, instr_add5_into_A, instr__jump_to_operation, instr__reset, instr__loop, instr_return, X_const, X_read_free, read_X_A, read_X_pc } in { p_line, p_reg_write_X_A, p_instr_add5_into_A, p_instr__jump_to_operation, p_instr__reset, p_instr__loop, p_instr_return, p_X_const, p_X_read_free, p_read_X_A, p_read_X_pc };
-    instr_add5_into_A { 0, X, A' } in main_vm.latch { main_vm.operation_id, main_vm.x, main_vm.y };
+    instr_add5_into_A * 1 { 0, X, A' } in main_vm.latch { main_vm.operation_id, main_vm.x, main_vm.y };
```
chriseth, line 682: Can these multiplications by 1 be avoided (at least in the Display routine)?

ast/src/asm_analysis/mod.rs
pacheco: In the end, they are a single multiplied flag. I had to split them because I need to know this link comes from an instruction, and what that instruction is (the optional flag serves this purpose).

linker/src/lib.rs
pacheco: I actually thought about modifying display, but this is best left to the optimizer I think: the other side of the multiplication by 1 could be anything.
Another option is having the link flag be an Option, but then it needlessly complicates the rest of code with option handling.
Maybe the "combine flags" function you suggested above would be a better place.

airgen/src/lib.rs
pacheco: Actually looked for such a function and didn't find it! Didn't find it because it only exists for `AlgebraicExpression` and this is a `parsed::Expression`... I'll create it

linker/src/lib.rs
pacheco: done the above, worked nicely

ast/src/asm_analysis/mod.rs
chriseth: Can you add the explanation to the code comment?

chriseth (dismissed):

chriseth (approved):

PR #1479, opened 2024-06-27
Allow clippy etc into estark backend

Schaeff:
I don't know if this is the right way to fix it, but I'm tired of CI failing because my local checks ignored estark as it's behind a feature. @lvella any thoughts?

### DIFF
--- a/backend/src/estark/polygon_wrapper.rs
+++ b/backend/src/estark/polygon_wrapper.rs
@@ -1,4 +1,4 @@
-use std::{fs, path::PathBuf};
+use std::path::PathBuf;
 
 use powdr_ast::analyzed::Analyzed;
 use powdr_executor::witgen::WitgenCallback;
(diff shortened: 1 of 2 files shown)


backend/src/estark/polygon_wrapper.rs
```diff
@@ -33,6 +33,7 @@ impl<F: FieldElement> BackendFactory<F> for Factory {
     }
 }
 
+#[allow(dead_code)]
```
leonardoalt, line 36: hmmmm this looks rather hacky.

lvella:
I am not sure I understand the problem. If clippy is failing, this means you changed stuff inside the module `polygon_wrapper`, right? Then you can't run `cargo clippy --feature=estark-polygon` because you are not on a x86-64?

If so, this PR just moves the problem a little deeper: instead of the whole module being not covered, now it is the `prove()` function that is not covered... is it still worth it?

Schaeff:
I think you summed it up well. I think it's not worth it, closing.

PR #1482, opened 2024-06-28
Plonky3 secure publics

topanisto:
Securely implementing public inputs in Plonky3! No longer requires the use of fixed columns by constraining every public value with 3 columns-- an incrementor starting from the row index of each exposed witness, an inverse to calculate the inverse of incrementor (0 if incrementor is 0), and a selector column constrained by a series of `is_zero` constraints to be 1 at the row index and 0 everywhere else.


plonky3/src/circuit_builder.rs
```diff
@@ -1,12 +1,12 @@
...
-//! implemented as extra witness columns in the execution trace.
+//! Support for public values without the use of fixed columns.
 //!
 //! Namely, given ith public value pub[i] corresponding to a witness value in
 //! row j of column Ci, a corresponding selector column Pi is constructed to
-//! constrain Pi * (pub[i] - Ci) on every row. Pi is precomputed in the trace
-//! to be 0 everywhere and 1 in row j.
+//! constrain Pi * (pub[i] - Ci) on every row. Pi is constrained to be 1 at the
+//! evaluation index and 0 everywhere else by two additional columns acting as a
+//! decrementor from the evaluation index, and the inverse of the decrementor row.
```
Schaeff, line 9: I think this would be explained better using pil code :)

plonky3/src/circuit_builder.rs
```diff
@@ -232,12 +240,34 @@ impl<'a, T: FieldElement, AB: AirBuilderWithPublicValues<F = Val>> Air<AB> for P
...
-        let local = matrix.row_slice(0);
+        let (local, next) = (matrix.row_slice(0), matrix.row_slice(1));
 
-        // constraining Pi * (Ci - pub[i]) = 0
         publics.iter().zip(pi_moved).enumerate().for_each(
-            |(index, ((_, col_id, _), public_value))| {
-                let selector = local[self.analyzed.commitment_count() + index];
+            |(index, ((_, col_id, row_id), public_value))| {
+                let (incrementor, inverse, selector) = (
+                    local[self.analyzed.commitment_count() + 3 * index],
```
Schaeff, line 248: let's have named variables for these. Let's say you introduce `x` in the doc at the top of this file, then here you can introduce `x` and `x_prime`, etc

plonky3/src/circuit_builder.rs
```diff
@@ -232,12 +240,34 @@ impl<'a, T: FieldElement, AB: AirBuilderWithPublicValues<F = Val>> Air<AB> for P
...
+                    local[self.analyzed.commitment_count() + 3 * index + 2],
+                );
+
+                //set incrementors for each column to be row_id and decrement each row
+                let incrementor_next = next[self.analyzed.commitment_count() + 3 * index];
+
+                let mut when_first_row = builder.when_first_row();
+                when_first_row.assert_eq(
+                    incrementor,
+                    AB::Expr::from(cast_to_goldilocks(GoldilocksField::from(*row_id as u32))),
```
Schaeff, line 259:
```suggestion
                    cast_to_goldilocks(GoldilocksField::from(*row_id as u32)).into(),
```

Maybe the `into` is not needed even?

plonky3/src/circuit_builder.rs
```diff
@@ -232,12 +240,34 @@ impl<'a, T: FieldElement, AB: AirBuilderWithPublicValues<F = Val>> Air<AB> for P
...
-                let selector = local[self.analyzed.commitment_count() + index];
+            |(index, ((_, col_id, row_id), public_value))| {
+                let (incrementor, inverse, selector) = (
+                    local[self.analyzed.commitment_count() + 3 * index],
+                    local[self.analyzed.commitment_count() + 3 * index + 1],
+                    local[self.analyzed.commitment_count() + 3 * index + 2],
+                );
+
+                //set incrementors for each column to be row_id and decrement each row
+                let incrementor_next = next[self.analyzed.commitment_count() + 3 * index];
```
Schaeff, line 254: we can merge these

PR #1485, opened 2024-06-28
Extract ByteShift machine

Schaeff:
~Fails in witgen~
- Extract the fixed table in Shift to its own machine
- Adjust optimiser to fix witgen


pilopt/src/lib.rs
```diff
@@ -346,6 +347,26 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
     }
 }
 
+/// Removes lookup and permutation selectors which are equal to 1
+/// TODO: refactor `SelectedExpression` to not use an optional selector
```
pacheco, line 351: Maybe its better to fix witgen here? Selectors of `1` should be equivalent to no selector right? (i.e., consider every row for the lookup).
It shouldn't be the case that the optimized pil passes witgen but the unoptimized doesn't...

Schaeff: I see your point but this is not the case in the codebase now, with the optimizer off the following pipeline tests fai:

```
    block_to_block
    book::modules
    book::operations_and_links
    pass_range_constraints
    permutation_to_block_to_block
    side_effects
    single_operation
    sqrt
    sqrt_asm
```

I'll check if the witgen fix is manageable

Schaeff: I think we should rely on the optimiser for now. If we want witgen to work on unoptimized code we should add tests for that. I don't think it's a priority now.

leonardoalt (approved):

PR #1487, opened 2024-06-28
Extract Byte2 machine

Schaeff:
requires #1513


importer/src/path_canonicalizer.rs
```diff
@@ -140,7 +140,32 @@ impl<'a> Folder for Canonicalizer<'a> {
                         canonicalize_inside_expression(e, &self.path, self.paths);
                     }
                 }
-                _ => {}
+                MachineStatement::InstructionDeclaration(_, _, i) => {
```
chriseth, line 143: Can we implement `Children<E>` for Instruction?

std/machines/arith.asm
```diff
@@ -10,6 +10,19 @@ use std::convert::expr;
 use std::prover::eval;
 use std::prover::Query;
 
+// A machine to check that a field element represents two bytes. It uses an exhaustive lookup table.
```
chriseth, line 13:
```suggestion
/// A machine to check that a field element represents two bytes. It uses an exhaustive lookup table.
```

leonardoalt (commented):
Could go to std instead?

Schaeff:
> Could go to std instead?

done

georgwiese (approved):
LGTM

PR #1489, opened 2024-06-28
Trait parsing

gzanitti:
This PR splits from the main Trait implementation PR #1450 to simplify the review process.

It includes only the parsing of the traits (not impls) and some functionality necessary for the code to compile.


analysis/src/machine_check.rs
```diff
@@ -333,6 +333,12 @@ impl TypeChecker {
                                 Item::TypeDeclaration(enum_decl),
                             );
                         }
+                        asm::SymbolValue::TraitDeclaration(trait_decl) => {
```
chriseth, line 336: I was just wondering if we should combine TypeDeclaration and TraitDeclaration, but I realized that traits are not types, they cannot be used where types are expected - if you want to do that, you have to introduce a type var and bind it to the trait. Rust does it the same, it just allows traits to be used as a shortcut for the "correct" way.

ast/src/analyzed/display.rs
```diff
@@ -49,6 +49,9 @@ impl<T: Display> Display for Analyzed<T> {
                         if matches!(
                             definition,
                             Some(FunctionValueDefinition::TypeConstructor(_, _))
+                        ) || matches!(
+                            definition,
+                            Some(FunctionValueDefinition::TraitFunction(_, _))
                         ) {
                             // These are printed as part of the enum.
```
chriseth, line 56:
```suggestion
                            // These are printed as part of the enum / trait.
```

ast/src/parsed/display.rs
```diff
@@ -559,13 +561,63 @@ impl Display for FunctionDefinition {
...
+impl Display for TraitDeclaration<Expression> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        let type_vars = if self.type_vars.is_empty() {
+            Default::default()
+        } else {
+            format!("<{}>", self.type_vars)
+        };
+        write!(
+            f,
+            "trait {name} {type_vars} {{\n{functions}}}",
```
chriseth, line 580: This produces two spaces if there are no type vars.

chriseth: Ah wait, there should be no space after name in any case, right?

ast/src/parsed/display.rs
```diff
@@ -559,13 +561,63 @@ impl Display for FunctionDefinition {
...
+                self.functions.iter().map(|m| format!("{m},\n")).format(""),
+                1
+            )
+        )
+    }
+}
+
+impl Display for TraitFunction<Expression> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        write!(f, "{}: {}", self.name, self._type)
```
chriseth, line 611: I think `ty` is better than `_type`

ast/src/parsed/mod.rs
```diff
@@ -120,6 +124,8 @@ impl PilStatement {
     /// and their category.
     /// For an enum, returns the name of the enum and all the variants, where the first
     /// component is the name of the enum and the second the name of the variant.
+    /// Similarly, for a trait implementation, returns the name of the trait and all the functions,
```
chriseth, line 127:
```suggestion
    /// Similarly, for a trait declaration, returns the name of the trait and all the functions,
```

ast/src/analyzed/mod.rs
```diff
@@ -422,6 +424,16 @@ pub fn type_from_definition(
             FunctionValueDefinition::TypeConstructor(enum_decl, variant) => {
                 Some(variant.constructor_type(enum_decl))
             }
+            FunctionValueDefinition::TraitDeclaration(_) => {
+                panic!("Requested type of trait declaration.")
+            }
+            FunctionValueDefinition::TraitFunction(trait_decl, trait_func) => {
```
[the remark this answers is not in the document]
chriseth: Yeah, now that I thought about it a bit, I think that trait functions should be very similar to type constructors, so I think this is correct here.

ast/src/parsed/display.rs
```diff
@@ -559,13 +561,63 @@ impl Display for FunctionDefinition {
...
+            name = self.name,
+            functions = indent(
+                self.functions.iter().map(|m| format!("{m},\n")).format(""),
+                1
+            )
+        )
+    }
+}
+
+impl Display for TraitDeclaration {
```
chriseth, line 590: Why the second implementation? Can't you implement it for `<E: Display>`?

ast/src/parsed/display.rs
```diff
@@ -559,13 +561,63 @@ impl Display for FunctionDefinition {
...
+    }
+}
+
+impl Display for TraitFunction<Expression> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        write!(f, "{}: {}", self.name, self._type)
+    }
+}
+
+impl Display for TraitFunction {
```
chriseth, line 615: Same question here, why not implement for `<E: Display>`?

ast/src/parsed/mod.rs
```diff
@@ -46,6 +48,7 @@ impl SymbolCategory {
                 // Type constructors can also satisfy requests for values.
                 request == SymbolCategory::TypeConstructor || request == SymbolCategory::Value
             }
+            SymbolCategory::TraitDeclaration => request == SymbolCategory::Type,
```
chriseth, line 51: I think we need a new category called `Trait`. Let's consider `Into<A, B>` as a trait.

We cannot declare variables of that type. If you e.g. want a function parameter of that type you need
```
let<A, B, Into<A, B>> f: A -> B = |a| Into::into(a);
```

So you can "use" `Into` in an expression, but only as a path to get to one of the functions (`Into::into`).

The only point where you can use a trait (apart from an impl for that trait) is in the list of trait bounds:
`<A, B, Into<A, B>>`.

So I think it is fair to create a new category for that.

chriseth: Ah OK, sorry, you do have a new symbol category, but I think it should only be compatible with itself.

ast/src/parsed/mod.rs
```diff
@@ -306,6 +317,52 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct TraitDeclaration<E = u64> {
+    pub name: String,
+    pub type_vars: TypeBounds,
```
chriseth, line 323: I'm wondering, I think we will really just allow a list of type variables for a trait declaration, so maybe just a vector of strings?

chriseth: If it's easier to use TypeBounds then it's also fine!

ast/src/parsed/mod.rs
```diff
@@ -306,6 +317,52 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
...
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct TraitFunction<E = u64> {
+    pub name: String,
+    pub _type: Type<E>,
+}
+
+impl<R> Children<Expression<R>> for TraitFunction<u64> {
```
chriseth, line 357: For the non-u64 version, the type can contain expression children, so I think you should call that function for that case.

importer/src/path_canonicalizer.rs
```diff
@@ -870,6 +887,25 @@ fn check_type(
...
+    check_path(location.clone().join(symbol_path), state)?;
+    trait_decl.functions.iter().try_fold(
+        BTreeSet::default(),
+        |mut acc, TraitFunction { name, .. }| {
+            acc.insert(name.clone()).then_some(acc).ok_or(format!(
+                "Duplicate method `{name}` defined in trait `{location}`"
+            ))
+        },
+    )?;
+
```
chriseth, line 924: Why don't you call `check_type` on the types?

importer/src/path_canonicalizer.rs
```diff
@@ -870,6 +887,25 @@ fn check_type(
     })
 }
 
+fn check_trait_declaration(
+    location: &AbsoluteSymbolPath,
+    trait_decl: &TraitDeclaration<Expression>,
+    state: &mut State<'_>,
+) -> Result<(), String> {
+    let symbol_path = SymbolPath::from_identifier(trait_decl.name.clone());
+    check_path(location.clone().join(symbol_path), state)?;
```
chriseth, line 915: Are you sure this is correct? We don't do this for enum.

parser/src/lib.rs
```diff
@@ -532,4 +533,52 @@ namespace N(2);
...
+    }"#;
+
+        let printed = format!("{}", parse(Some("input"), input).unwrap_err_to_stderr());
+        assert_eq!(expected.trim(), printed.trim());
+    }
+
+    #[test]
+    fn parse_trait_multi_params() {
+        let input = r#"
+    trait Add<T: Add, Q> {
```
chriseth, line 568: I think we shouldn't allow bounds on trait type variables for now.

parser/src/lib.rs
```diff
@@ -532,4 +533,52 @@ namespace N(2);
...
+
+    #[test]
+    fn parse_trait_multi_params2() {
+        let input = r#"
+    trait Iterator<S, I> {
+        next: S -> (S, Option<I>),
+    }"#;
+
+        let expected = r#"
+    trait Iterator <S, I> {
```
chriseth, line 589: should be
```suggestion
    trait Iterator<S, I> {
```

pil-analyzer/src/statement_processor.rs
```diff
@@ -457,6 +465,37 @@ where
             .collect();
         }
 
+        if let Some(FunctionDefinition::TraitDeclaration(trait_decl)) = value {
```
chriseth, line 460:
```suggestion
        } else if let Some(FunctionDefinition::TraitDeclaration(trait_decl)) = value {
```

pil-analyzer/tests/parse_display.rs
```diff
@@ -721,3 +721,19 @@ namespace T(8);
     assert_eq!(analyzed.degree, Some(8));
     assert_eq!(expected, analyzed.to_string());
 }
+
+#[test]
```
chriseth, line 720: Please add a test that shows that types are properly processed here, e.g. a trait that uses an array type of fixed length that is computed by some expression.

chriseth (changes requested):
Looks good, just some small changes left.

gzanitti:
> Looks good, just some small changes left.

Thank you for your comments @chriseth !


I've been thinking, and maybe we have to find another way to handle some definitions like enums, structs, traits, etc. As things stand today and considering that both enums and traits insert definitions, this code would not be possible:


```
trait Add<T, Q> {
    add: T, T -> Q,
}
enum Add {
    Add,
}
```


is not the clearest example in the world, but the idea is that both Add collide. And it gets more complex when we add Structs and their fields. Maybe it's easier to modify it now that we only have enums, what do you think?

pil-analyzer/tests/parse_display.rs
```diff
@@ -719,16 +719,57 @@ namespace T(8);
...
 ";
 
     let analyzed = analyze_string::<GoldilocksField>(input);
     assert_eq!(expected, analyzed.to_string())
 }
+
+#[test]
+fn array_type_trait() {
+    let input = "trait ArraySum<T> {
+        array_sum: T[5] -> T,
```
chriseth, line 739: That's not what I meant. It should be something like `4 + 1`, something that is not just a number.

ast/src/parsed/mod.rs
```diff
@@ -320,7 +320,7 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct TraitDeclaration<E = u64> {
     pub name: String,
-    pub type_vars: TypeBounds,
+    pub type_vars: Option<Vec<String>>,
```
chriseth, line 323: I'm thinking, does it make sense to have no type variables? Then we can have exactly one implementation of the trait.

So it sounds like it's better to have them non-optional and non-empty (already ensured by the parser).

ast/src/parsed/mod.rs
```diff
@@ -330,15 +330,6 @@ impl TraitDeclaration<u64> {
...
-        Box::new(empty())
-    }
-    fn children_mut(&mut self) -> Box<dyn Iterator<Item = &mut Expression<R>> + '_> {
-        Box::new(empty())
-    }
-}
-
 impl<R> Children<Expression<R>> for TraitDeclaration<Expression<R>> {
     fn children(&self) -> Box<dyn Iterator<Item = &Expression<R>> + '_> {
         Box::new(empty())
```
chriseth, line 335: This is still missing: You have to iterate through the functions

chriseth:
the parser still allows the trait var list to be empty

ast/src/parsed/mod.rs
```diff
@@ -138,6 +144,9 @@ impl PilStatement {
                         .map(move |v| (name, Some(&v.name), SymbolCategory::TypeConstructor)),
                 ),
             ),
+            PilStatement::TraitDeclaration(_, TraitDeclaration { name, .. }) => {
+                Box::new(once((name, None, SymbolCategory::TraitDeclaration)))
+            }
```
chriseth, line 149: Shouldn't this also return the trait functions?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -229,6 +229,7 @@ impl PILAnalyzer {
             .iter_mut()
             .filter(|(_name, (_symbol, value))| {
                 !matches!(value, Some(FunctionValueDefinition::TypeDeclaration(_)))
+                    && !matches!(value, Some(FunctionValueDefinition::TraitDeclaration(_)))
```
chriseth, line 232: You could actually use the or feature in the pattern itself

chriseth:
could you also add some tests for the asm handling?

ast/src/parsed/mod.rs
gzanitti: I don't think so, since both, `TraitImplementation` and `TraitDeclaration`, define the same `Definitions`, if functions are also included there would be collisions.

chriseth: I mean for enum it does...

gzanitti: Should be fixed :)

ast/src/parsed/mod.rs
```diff
@@ -140,9 +140,18 @@ impl PilStatement {
...
+            PilStatement::TraitDeclaration(
+                _,
+                TraitDeclaration {
+                    name, functions, ..
+                },
+            ) => Box::new(
+                once((name, None, SymbolCategory::TraitDeclaration)).chain(
+                    functions
+                        .iter()
+                        .map(move |f| (name, Some(&f.name), SymbolCategory::TypeConstructor)),
```
chriseth, line 152: I think `Value` should be more suitable. The main difference between TypeConstructor and Value is that TypeConstructor can be used inside patterns - but trait functions should not be usable inside patterns.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -348,6 +348,26 @@ impl PILAnalyzer {
                 vec![]
             }
             PilStatement::Include(_, _) => unreachable!(),
+            PilStatement::TraitDeclaration(_, _) => {
+                let names = statement
+                    .symbol_definition_names_and_contained()
+                    .filter_map(|(name, sub_name, symbol_category)| match sub_name {
+                        None => Some((self.driver().resolve_decl(name), symbol_category)),
+                        Some(_) => None,
```
chriseth, line 356: Why are you filtering out the trait functions here (i.e. why are you not using the generic impl below)? Again: If it's not done for enums, why should it be done for trait functions?

ast/src/parsed/mod.rs
gzanitti: Oh, I see, thanks!

test_data/asm/trait_parsing.asm
```diff
@@ -0,0 +1,13 @@
+machine Trait with degree: 8 {
+    reg pc[@pc];
+    reg X[<=];
+    reg A;
+
+    trait ArraySum<T> {
+        array_sum: T[4 + 1] -> T,
```
chriseth, line 7: To check that the path canonicalizer is correctly implemented, could you add a generic enum type and use it here like for example `Option<T>`?

test_data/asm/trait_parsing.asm
```diff
@@ -0,0 +1,13 @@
+machine Trait with degree: 8 {
+    reg pc[@pc];
+    reg X[<=];
+    reg A;
+
+    trait ArraySum<T> {
```
chriseth, line 6: Oh you don't really need machines here for a test of the asm part. I would actually consider not allowing trait declarations inside machines if that makes it easier somehow (maybe it will have to later).

pil-analyzer/src/pil_analyzer.rs
chriseth: I think we shouldn't even make the implementation visible. You will always call the trait function and you will specify the type params. It is up to the evaluator to find the correct implementation (and of course the type checker).

Also there will be multiple implementations, if you add them to the namespace, they will clash.

test_data/asm/trait_parsing.asm
```diff
@@ -0,0 +1,16 @@
+mod types {
+    enum DoubleOpt<T> {
+        None,
+        Some(T, T)
+    }
+
+    trait ArraySum<T, DoubleOpt<T> > {
+        array_sum: T[4 + 1] -> DoubleOpt,
```
chriseth, line 8: this is what I meant:
```suggestion
    trait ArraySum<T> {
        array_sum: T[4 + 1] -> DoubleOpt<T>,
```

And if you want to be even more thorough, you could check that when this has been exported to pil (if there is a set of suitable tests for that), it says `types::DoubleOpt<T>`

pil-analyzer/src/pil_analyzer.rs
gzanitti: Perfect. I had a solution in my other PR to avoid collisions between implementations, but surely now it will be solved in a cleaner way. Thanks and sorry for the time lost.

chriseth (approved):

PR #1490, opened 2024-06-28
Impl parsing

gzanitti:
This PR splits from the main Trait implementation PR https://github.com/powdr-labs/powdr/pull/1450 to simplify the review process.

It includes only the parsing of the impls (trait parsing PR here: #1489 ) and some functionality necessary for the code to compile.


ast/src/parsed/mod.rs
```diff
@@ -140,6 +144,18 @@ impl PilStatement {
                         .map(move |v| (name, Some(&v.name), SymbolCategory::TypeConstructor)),
                 ),
             ),
+            PilStatement::TraitImplementation(
```
chriseth, line 147: No, I think the trait implementation should not be accessible at all. It is only accessible through the trait declaration.

parser/src/powdr.lalrpop
```diff
@@ -33,6 +33,7 @@ ModuleStatement: ModuleStatement = {
             name: <>.name.clone(),
             value: SymbolValue::TypeDeclaration(<>),
         }),
+    <ImplDeclaration> => ModuleStatement::TraitImplementation(<>),
```
chriseth, line 36: should we also call it TraitImplementation?

parser/src/powdr.lalrpop
```diff
@@ -681,6 +683,25 @@ PatternList: Vec<Pattern> = {
 
 // ---------------------------- Trait/Impl Declaration -----------------------------
 
+GenericTraitName: (String, Option<TraitScheme<Expression>>) = {
+    <name:Identifier> => (name, None),
```
chriseth, line 687: If we require traits to have at least one type var, then it doesn't make sense to parse impls without at least type arguments for the trait (it is legal to not declare any type params like in `impl Into<StructA, StructB> { ... }`

chriseth: Oh and `name` should actually be a Path here, because it is a reference and not a new declaration.

parser/src/powdr.lalrpop
```diff
@@ -681,6 +683,25 @@ PatternList: Vec<Pattern> = {
...
+    <vars:("<" <TypeVarBounds> ">")?> <name:Identifier> <types:("<" <TypeTermList> ">")> =>
+        (name, Some(TraitScheme{ vars: vars.unwrap_or_default(), types }))
+}
+
+ImplDeclaration: TraitImplementation<Expression> = {
+    "impl" <type_scheme: GenericTraitName> "{" <functions:ImplFunctions> "}" => TraitImplementation { name: type_scheme.0, type_scheme: type_scheme.1, functions }
+}
+
+ImplFunctions: Vec<NamedExpression<Expression>> = {
```
chriseth, line 696: We could use the same grammatic element (`{`/`}`-separated list of fields with values) for struct expressions

ast/src/parsed/mod.rs
```diff
@@ -320,6 +338,32 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
...
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct TraitImplementation<Expr> {
+    pub name: String,
+    pub type_scheme: Option<TraitScheme<Expr>>,
+    pub functions: Vec<NamedExpression<Expr>>,
+}
+
+impl<R> Children<Expression<R>> for TraitImplementation<Expression<R>> {
+    fn children(&self) -> Box<dyn Iterator<Item = &Expression<R>> + '_> {
+        Box::new(self.functions.iter().flat_map(|m| m.body.children()))
```
chriseth, line 350: `type_scheme` has an `Expr` parameter as well, don't forget that!

chriseth: By the way, this might get a bit complicated later on, because in expression processor change the types from `Type<Expression>` to `Type<u64>`, so we probably need a new type for trait impls after that.

The alternative is to parse it as `Type<u64>` right away - which is probably good enough since we cannot handle statically-sized array types in the type checker any way.

I did the same here: https://github.com/powdr-labs/powdr/pull/1545/files#diff-786a78af64899fc3240c68b7f68cb48a1f9e5772564adbe2cf3d53e08720a902R737

ast/src/parsed/mod.rs
```diff
@@ -320,6 +338,32 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct TraitImplementation<Expr> {
+    pub name: String,
+    pub type_scheme: Option<TraitScheme<Expr>>,
```
chriseth, line 344: I don't think it should be optional.

ast/src/parsed/mod.rs
gzanitti: Yes, in the giant PR on Traits I had to create both versions as you mentione here. I definitely think it's better to improve the types in the grammar rather than duplicate TraitImplementation so I'm basing this one on the types in your PR to avoid duplication.

asm-to-pil/src/lib.rs
```diff
@@ -23,6 +23,9 @@ pub fn compile<T: FieldElement>(file: AnalysisASMFile) -> AnalysisASMFile {
                         }
                         Item::Expression(e) => Item::Expression(e),
                         Item::TypeDeclaration(enum_decl) => Item::TypeDeclaration(enum_decl),
+                        Item::TraitImplementation(trait_impl) => {
+                            Item::TraitImplementation(trait_impl)
+                        }
                         Item::TraitDeclaration(trait_decl) => Item::TraitDeclaration(trait_decl),
```
chriseth, line 29: Can you use something like
```suggestion
                        Item::Expression(_) | Item::TypeDeclaration(_) | ... => m
```
?

chriseth: Maybe we could also just match on `Machine` and only modify the machine in that case, and pass the rest un-modified.

chriseth: Could also use the `machines_mut` function

ast/src/parsed/display.rs
```diff
@@ -621,6 +623,38 @@ impl<E: Display> EnumDeclaration<E> {
...
+        let trait_vars = if self.type_scheme.types.is_empty() {
+            Default::default()
+        } else {
+            let formatted_elements: Vec<String> = self
+                .type_scheme
+                .types
+                .iter()
+                .map(|t| format!("{t}"))
+                .collect();
+            format!("<{}>", formatted_elements.join(", "))
```
chriseth, line 643: Why not (using itertools)
```suggestion
            format!("<{}>", self
                .type_scheme
                .types
                .format(", "))
```

ast/src/parsed/mod.rs
```diff
@@ -123,6 +125,8 @@ impl PilStatement {
     /// and their category.
     /// For an enum, returns the name of the enum and all the variants, where the first
     /// component is the name of the enum and the second the name of the variant.
+    /// Similarly, for a trait implementation, returns the name of the trait and all the functions,
+    /// where the first component is the name of the trait and the rest the name of the functions.
```
chriseth, line 129:
```suggestion
```

ast/src/parsed/types.rs
```diff
@@ -405,6 +405,12 @@ impl From<Type> for TypeScheme {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct TraitScheme {
```
chriseth, line 409: I'm wondering (really wondering, not suggesting to do this) if it would be simpler to use a regular TypeScheme whose type is a tuple instead.

importer/src/powdr_std.rs
```diff
@@ -97,13 +97,17 @@ impl Folder for StdAdder {
...
+                ModuleStatement::TraitImplementation(trait_impl) => {
+                    self.fold_trait_implementation(trait_impl).map(From::from)
+                }
             })
             .collect::<Result<Vec<_>, _>>()?;
 
         // Check whether the module already has a definition for `std`
         // (E.g. the main module)
         let has_std = statements.iter().any(|s| match s {
             ModuleStatement::SymbolDefinition(d) => d.name == "std",
```
chriseth, line 109: Would it work to use `defined_names` here?

parser/src/lib.rs
```diff
@@ -481,6 +481,22 @@ namespace N(2);
...
+        let input = r#"
+    impl<T> Iterator<ArrayIterator<T>, T> {
+        next: |it| if it.pos >= 7 { (it, none) } else { some((increment(it), it.arr[it.pos])) },
+    }"#;
+
+        let expected = r#"
+    impl<T> Iterator<ArrayIterator<T>, T> {
+        next: (|it| if it.pos >= 7 { (it, none) } else { some((increment(it), it.arr[it.pos])) }),
+    }"#;
+
```
chriseth, line 495: Can you add another one with `impl<A, B>`? Also do we already support bounds here like `impl<A: Eq, B: Add>`?

parser/src/powdr.lalrpop
```diff
@@ -694,6 +696,24 @@ PatternList: Vec<Pattern> = {
...
+TraitImplementation: TraitImplementation<Expression> = {
+    "impl" <type_scheme: GenericTraitName> "{" <functions:ImplFunctions> "}" => TraitImplementation { name: type_scheme.0, type_scheme: type_scheme.1, functions }
+}
+
+ImplFunctions: Vec<NamedExpression<Expression>> = {
+    => vec![],
+    <mut list:( <ImplFunction> "," )*> <end:ImplFunction> ","  => { list.push(end); list }
+}
+
+ImplFunction: NamedExpression<Expression> = {
```
chriseth, line 713: Why not also call this NamedExpression?

parser/src/powdr.lalrpop
```diff
@@ -694,6 +696,24 @@ PatternList: Vec<Pattern> = {
...
+        (name, TraitScheme{ vars, types })
+}
+
+TraitImplementation: TraitImplementation<Expression> = {
+    "impl" <type_scheme: GenericTraitName> "{" <functions:ImplFunctions> "}" => TraitImplementation { name: type_scheme.0, type_scheme: type_scheme.1, functions }
+}
+
+ImplFunctions: Vec<NamedExpression<Expression>> = {
+    => vec![],
+    <mut list:( <ImplFunction> "," )*> <end:ImplFunction> ","  => { list.push(end); list }
```
chriseth, line 710: Can we make the last `,` optional?
```suggestion
    <mut list:( <ImplFunction> "," )*> <end:ImplFunction> ","?  => { list.push(end); list }
```

parser/src/powdr.lalrpop
```diff
@@ -694,6 +696,24 @@ PatternList: Vec<Pattern> = {
 
 // ---------------------------- Trait/Impl Declaration -----------------------------
 
+GenericTraitName: (SymbolPath, TraitScheme) = {
```
chriseth, line 699: I would expect this section to start with TraitDeclaration and GenericTraitName to go after TraitImplementation

ast/src/parsed/types.rs
gzanitti: I didn't really think about this solution, but it looks really more elegant than the current option. I'll take a look at it

parser/src/lib.rs
gzanitti: Sure. No, we don't support bounds at the moment.

importer/src/powdr_std.rs
chriseth: Isn't there a function on ModuleStatement that returns the names defined by the statement? if yes, I think it would be cleaner to use that function instead of relying on an "internal" property of ModuleStatement::SymbolDefinition.

ast/src/parsed/display.rs
```diff
@@ -631,21 +631,25 @@ impl<E: Display> Display for TraitImplementation<E> {
             format!("<{}>", self.type_scheme.vars)
         };
 
-        let trait_vars = if self.type_scheme.types.is_empty() {
-            Default::default()
-        } else {
-            format!("<{}>", self.type_scheme.types.iter().format(", "))
-        };
+        if let Type::Tuple(TupleType { items }) = &self.type_scheme.ty {
```
chriseth, line 634:
```suggestion
        let Type::Tuple(TupleType { items }) = &self.type_scheme.ty else { panic!("Type from trait scheme is not a tuple.") }
```

ast/src/parsed/asm.rs
```diff
@@ -44,6 +44,15 @@ pub enum ModuleStatement {
     TraitImplementation(TraitImplementation<Expression>),
 }
 
+impl ModuleStatement {
+    pub fn defined_name(&self) -> impl Iterator<Item = &String> {
+        match self {
+            ModuleStatement::SymbolDefinition(d) => once(&d.name),
+            ModuleStatement::TraitImplementation(t) => once(t.name.name()),
```
chriseth, line 51: It doesn't define any name.

ast/src/parsed/asm.rs
```diff
@@ -45,10 +45,10 @@ pub enum ModuleStatement {
 }
 
 impl ModuleStatement {
-    pub fn defined_names(&self) -> impl Iterator<Item = &String> {
+    pub fn defined_names(&self) -> Box<dyn Iterator<Item = &String> + '_> {
```
chriseth, line 48: For now, it could also just return `Option<&string>`.

chriseth (approved):

PR #1513, opened 2024-07-01
Visit instructions and links in canonicalizer

Schaeff:


importer/src/path_canonicalizer.rs
```diff
@@ -255,8 +255,6 @@ fn canonicalize_inside_expression(
                 // If resolving the reference fails, we assume it is a local variable that has been checked below.
                 if let Some(n) = paths.get(&path.clone().join(reference.path.clone())) {
                     *reference = n.relative_to(&Default::default()).into();
-                } else {
-                    assert!(reference.path.try_to_identifier().is_some());
```
Schaeff, line 248: @chriseth we talked about the real issue here being that the first step of canonicalization did not run. It is not, the first step does run. Also, I checked with `instr incr_a { A = utils::super::utils::incremented(A) }` and it gets resolved correctly (with this assertion removed).

Schaeff: looking into this

importer/src/path_canonicalizer.rs
```diff
@@ -628,22 +639,17 @@ fn check_machine(
...
-                for link_decl in &instr.links {
-                    check_expression(&module_location, &link_decl.flag, state, &local_variables)?;
-                    link_decl
-                        .link
-                        .params
-                        .inputs_and_outputs()
-                        .try_for_each(|e| {
-                            check_expression(&module_location, e, state, &local_variables)
-                        })?;
+                // Add the instruction parameters, ignore collisions.
```
chriseth, line 648: Shouldn't we only add the parameters that are actual parameters that introduce a new name i.e. only the typed ones?

Schaeff: Good point, done

importer/test_data/instruction.expected.asm
```diff
@@ -0,0 +1,19 @@
...
+    }
+}
+machine Id {
+    operation id<0> x, y;
+        pol commit x;
+        pol commit y;
+        x = y;
+}
+machine Main {
+    ::Id id;
```
chriseth, line 13: why does this add a double colon here? It doesn't add it at other places

chriseth (approved):

PR #1515, opened 2024-07-01
Change column map to work efficiently with slices of column IDs.

chriseth:


executor/src/witgen/rows.rs
```diff
@@ -458,6 +487,7 @@ impl<'row, 'a, T: FieldElement> RowPair<'row, 'a, T> {
     pub fn get_value(&self, poly: &AlgebraicReference) -> Option<T> {
         let row = self.get_row(poly.next);
         if self.unknown_strategy == UnknownStrategy::Zero {
+            // TODO this should panic if the column is out of range.
```
chriseth, line 490: Ah no, I still wanted to look into which calls to get_value should panic and which should return  None.

Issue #1016, opened 2024-02-05
PIL: Structs, traits and enums


chriseth:
Implement the last missing part of a fully-fledged functional language.

```rust
enum Option<T> {
  None,
  Some(T)
}

// This could be implicit like in Rust or not.
use Option::None;
use Option::Some;

trait Iterator<S, I> {
  next: S -> (S, Option<I>),
}

trait Default<S> {
  default: -> S,
}

impl<T> Default<T[]> {
  default: || [],
}

trait Add<T> {
  add: T, T -> T
}

struct ArrayIterator<T> {
  arr: T[],
  pos: int
}

let<T> increment: ArrayIterator<T> -> ArrayIterator<T> = |it| ArrayIterator{ arr: it.arr, pos: it.pos + 1 };

impl<T> Iterator<ArrayIterator<T>, T> {
  next: |it| if it.pos >= array::len(it.arr) { (it, none) } else { some((increment(it), it.arr[it.pos])) },
}

trait Into<A, B> {
  into: A -> B,
}

impl<T> Into<T[], ArrayIterator<T>> {
  into: |arr| ArrayIterator{ arr, pos: 0 },
}

let<T: Default + Add, I: Iterator<I, T>>  sum_iter: I -> T = |it| match Iterator::next(it) {
  (it, Option::None) => Default::default(),
  (it, Option::Some(v)) => sum_iter(it) + v,
};

let sum = |x| sum_iter(Into::into(x));
```

The idea with traits is that you can define a set of functions that use any number of type variables. Then you can implement functions of the trait on an arbitrary specialization of these type variables. Impls can themselves be generic again. We do not (yet?) have any special support for "self' types.

You can call trait functions only by specifying the trait (``Into::into()``). You can but do not have to provide the parameters for the generic type arguments (``Into::<int[], ArrayIterator<int>>::into()``).

You can use traits as bounds on generic type variables. If the trait has exactly one generic parameter, there is a shorthand where `T: TraitName` stands for `T: TraitName<T>`. The trait bound requires that there is an implementation for the type variable.

I'm not sure about which syntax to use in an `impl`. We could use a struct constructor syntax (makes sense because we have to mention each trait item exactly once, just like for struct members and because the type is dictated from outside) or the `let` syntax.

PR #1516, opened 2024-07-01
Structs in PIL

gzanitti:
This PR implements structs based on the issue #1016.

~~To avoid ambiguity with the grammar, for creating an instance of a struct and for accessing its fields, a temporary variation of the grammar is used which should be discussed.~~

~~Currently, an instance of struct `X {x: int, y: int}` is created using `X with {x: 1, y:0}`.~~


importer/src/path_canonicalizer.rs
```diff
@@ -197,9 +213,11 @@ fn free_inputs_in_expression<'a>(
         Expression::LambdaExpression(_, _) => todo!(),
         Expression::ArrayLiteral(_, _) => todo!(),
         Expression::IndexAccess(_, _) => todo!(),
+        Expression::FieldAccess(_, _) => todo!(),
```
chriseth, line 216: Can you split field access out into its own PR?

importer/src/path_canonicalizer.rs
```diff
@@ -852,6 +882,28 @@ fn check_type_declaration(
...
+    struct_decl: &StructDeclaration<Expression>,
+    state: &mut State<'_>,
+) -> Result<(), Error> {
+    struct_decl
+        .fields
+        .iter()
+        .try_fold(BTreeSet::default(), |mut acc, (name, _)| {
+            acc.insert(name.clone())
+                .then_some(acc)
+                .ok_or(format!("Duplicate variant `{name}` in enum `{location}`"))
```
chriseth, line 896: uses enum wording

parser/src/powdr.lalrpop
```diff
@@ -622,6 +633,22 @@ IfExpression: Box<Expression> = {
         <end:@R> => ctx.to_expr_with_source_ref(IfExpression{condition, body, else_body}, start, end),
 }
 
+StructExpression: Box<Expression> = {
+    <start:@L> <name:Identifier> "with" "{" <fields:StructExpressionList> "}" <end:@R> => ctx.to_expr_with_source_ref(StructExpression{name, fields}, start, end),
```
chriseth, line 637: What is the obstacle that forces us to use `with`?

parser/src/powdr.lalrpop
```diff
@@ -622,6 +633,22 @@ IfExpression: Box<Expression> = {
         <end:@R> => ctx.to_expr_with_source_ref(IfExpression{condition, body, else_body}, start, end),
 }
 
+StructExpression: Box<Expression> = {
```
chriseth, line 636: StructConstructorCall?

parser/src/powdr.lalrpop
```diff
@@ -622,6 +633,22 @@ IfExpression: Box<Expression> = {
...
+    <start:@L> <name:Identifier> "with" "{" <fields:StructExpressionList> "}" <end:@R> => ctx.to_expr_with_source_ref(StructExpression{name, fields}, start, end),
+}
+
+StructExpressionList: BTreeMap<String, Box<Expression>> = {
```
chriseth, line 640: I wonder if we should use `vec` here instead. If we re-format, BTreeMap sorts the fields.
I think as long as we use BTreeMap for the Values in the evaluator, it should not be bad WRT performance, what do you think?

parser/src/powdr.lalrpop
```diff
@@ -669,6 +697,26 @@ EnumPattern: Pattern = {
     <start:@L> <n:SymbolPath> <items:( "(" <PatternList> ")" )?> <end:@R> => Pattern::Enum(ctx.source_ref(start, end), n, items)
 }
 
+// Temporary using SymbolPath{} as an empty struct pattern to avoid ambiguity with EnumPattern
+StructPattern: Pattern = {
```
chriseth, line 701: I think we could even split out the patterns into their own PR

parser/src/test_utils.rs
```diff
@@ -124,12 +125,16 @@ impl<R> ClearSourceRefs for Expression<R> {
...
             | Expression::BinaryOperation(_, _)
             | Expression::IndexAccess(_, _)
+            | Expression::FieldAccess(_, _)
             | Expression::FunctionCall(_, _)
             | Expression::FreeInput(_, _)
             | Expression::IfExpression(_, _) => {}
             Expression::BlockExpression(_, b) => b.clear_source_refs(),
             Expression::MatchExpression(_, m) => m.clear_source_refs(),
             Expression::LambdaExpression(_, l) => l.clear_source_refs(),
+            Expression::StructExpression(_, f) => {
```
chriseth, line 135: This is only needed for stuff that has patterns. "Ordinary" expressions are already handled in line 116/117

pil-analyzer/src/statement_processor.rs
```diff
@@ -428,18 +445,56 @@ where
...
             return iter::once(PILItem::Definition(
                 symbol,
-                Some(FunctionValueDefinition::TypeDeclaration(enum_decl.clone())),
+                Some(FunctionValueDefinition::TypeDeclaration(
+                    TypeDeclarationAnalyzed::Enum(enum_decl.clone()),
+                )),
             ))
             .chain(var_items)
             .collect();
+        } else if let Some(FunctionDefinition::TypeDeclaration(TypeDeclarationParsed::Struct(
```
chriseth, line 462: I think at this point, we should really split out the structs and enums into functions. This function is getting too long. Also isn't there any code that can be shared?

parser/src/powdr.lalrpop
gzanitti: The parser cannot distinguish between Matchs and Structs when finding the first {. And if we try to modify MatchExpression in any way, then Structs and IfExpression become ambiguous.


I tried some variations to try to fix it but I couldn't, so I decided to stop wasting time and use a temporary patch.

pil-analyzer/src/statement_processor.rs
```diff
@@ -428,18 +445,56 @@ where
...
+            struct_decl,
+        ))) = value
+        {
+            assert_eq!(symbol_kind, SymbolKind::Other());
+            let struct_decl = self.process_struct_declaration(struct_decl);
+            let shared_struct_decl = Arc::new(struct_decl.clone());
+            let field_items = struct_decl.fields.keys().map(|field_name| {
```
chriseth, line 469: I don't think the field items should be accessible through the type. Are they in rust?

pil-analyzer/src/type_inference.rs
```diff
@@ -662,6 +675,13 @@ impl TypeChecker {
                 self.local_var_types.truncate(original_var_count);
                 result?
             }
+            Expression::StructExpression(_, struct_expr) => {
+                for expr in struct_expr.fields.values_mut() {
+                    self.infer_type_of_expression(expr)?;
```
chriseth, line 680: Shouldn't we have an expected type here?

parser/src/powdr.lalrpop
chriseth: This very weird because `if` and `match` use keywords in front of the identifiers...

Do you know if `S({...})` works?

parser/src/powdr.lalrpop
chriseth: ah sorry, if rust calls it structexpression then it's of course fine!

parser/src/powdr.lalrpop
gzanitti: > This very weird because `if` and `match` use keywords in front of the identifiers...
> 
> Do you know if `S({...})` works?

Yes, it works. In fact, anything between the `Identifier` and `{` of the struct works, because it helps disambiguate with the MatchExpression.

I think the problem comes from the `<scrutinee:BoxedExpression>` of the match, which can also be a Struct, so the parser can't tell if the first `{` is a Struct in the scrutinee or the beginning of the match branches.

We stick with S({...}) ?

parser/src/powdr.lalrpop
gzanitti: I don't think having a vector here will create a problem. In fact, I went for a BTreeMap mainly because it seemed the most "ergonomic" option to store expressions associated to strings. I'll change it :)

parser/src/powdr.lalrpop
chriseth: Yes, let's use `S({...})` for now. The rust grammar distinguishes between `ExpressionWithBlock` and `ExpressionWithoutBlock`: https://doc.rust-lang.org/reference/expressions.html

And match is in a different category than struct. Maybe that would help?

chriseth: Ah no, haha and that is interesting: https://doc.rust-lang.org/reference/expressions/match-expr.html

the scrutinee in the rust grammar excludes structs :)

chriseth:
It would be nice to fix struct expression parsing - we can do this in a separate PR that potentially also just parses it into nothing (so that we don't need all the other ast types).

chriseth:
Ah sorry, you already made a PR for that. Will take a look!

gzanitti:
> Ah sorry, you already made a PR for that. Will take a look!

Sorry, it's my fault, I should have specified the PR order somewhere. 
So. this is the base PR. Then, I will recommend the follwoing order:

- #1591
- #1586 
- #1585

ast/src/analyzed/mod.rs
```diff
@@ -468,9 +468,16 @@ pub fn type_from_definition(
...
-            FunctionValueDefinition::TypeConstructor(enum_decl, variant) => {
+            FunctionValueDefinition::TypeConstructor(TypeConstructor::Enum(enum_decl, variant)) => {
                 Some(variant.constructor_type(enum_decl))
             }
+            FunctionValueDefinition::TypeConstructor(TypeConstructor::Struct(
+                struct_decl,
+                field,
+            )) => Some(TypeScheme {
+                vars: struct_decl.type_vars.clone(),
+                ty: field.1.clone(),
```
chriseth, line 479: This doesn't look right.

importer/src/path_canonicalizer.rs
```diff
@@ -925,6 +954,28 @@ fn check_type_declaration(
...
+    struct_decl: &StructDeclaration<Expression>,
+    state: &mut State<'_>,
+) -> Result<(), Error> {
+    struct_decl
+        .fields
+        .iter()
+        .try_fold(BTreeSet::default(), |mut acc, (name, _)| {
+            acc.insert(name.clone())
+                .then_some(acc)
+                .ok_or(format!("Duplicate field `{name}` in struct `{location}`"))
```
chriseth, line 968: Are you doing the same in statement processor? I think it should be done there as well.

ast/src/parsed/display.rs
```diff
@@ -68,6 +68,15 @@ impl Display for SymbolDefinition {
     }
 }
 
+impl Display for asm::TypeDeclaration {
```
chriseth, line 71: Did you try using `derive_more(Display)` for this enum?

ast/src/parsed/display.rs
```diff
@@ -681,6 +685,55 @@ impl<E: Display> Display for EnumVariant<E> {
     }
 }
 
+impl<E: Display> Display for StructDeclaration<E> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        let name = self.name.to_string();
```
chriseth, line 690: It's way more efficient to not turn this into a string first.

ast/src/parsed/display.rs
```diff
@@ -681,6 +685,55 @@ impl<E: Display> Display for EnumVariant<E> {
...
+        let type_vars = if self.type_vars.is_empty() {
```
chriseth, line 691: Maybe we could create a helper for this, I have seen it a lot. The same code is also inside `TypeScheme::type_vars_to_string`

ast/src/parsed/display.rs
```diff
@@ -681,6 +685,55 @@ impl<E: Display> Display for EnumVariant<E> {
...
+        } else {
+            format!("<{}>", self.type_vars)
+        };
+        write!(
+            f,
+            "struct {name}{type_vars} {{\n{}}}",
+            indent(
+                self.fields
+                    .iter()
+                    .map(|v| format!("{}: {},\n", v.0, v.1))
```
chriseth, line 702:
```suggestion
                    .map(|(name, ty)| format!("{name}: {ty},\n"))
```

chriseth: Didn't we want to re-use some struct from the trait definition here?

ast/src/parsed/display.rs
```diff
@@ -681,6 +685,55 @@ impl<E: Display> Display for EnumVariant<E> {
...
+impl<E: Display> Display for StructExpression<E> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        write!(
+            f,
+            "{}{}",
+            self.name,
+            if self.fields.is_empty() {
+                "".to_string()
+            } else {
+                format!(
```
chriseth, line 719:
```suggestion
        write!(f, "{}", self.name)?;
        if !self.fields.is_empty() {
            write!(f,
```

chriseth (commented):
Only very partial review.

ast/src/analyzed/mod.rs
```diff
@@ -621,12 +628,56 @@ pub enum SymbolKind {
...
+            TypeDeclaration::Enum(enum_decl) => enum_decl.children_mut(),
+            TypeDeclaration::Struct(struct_decl) => struct_decl.children_mut(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub enum TypeConstructor {
+    Enum(Arc<EnumDeclaration>, EnumVariant),
+    Struct(Arc<StructDeclaration>, (String, Type)),
```
chriseth, line 662: A type constructor for struct and a struct expression are the same thing. Is this the case here or do you make a distinction? I have the impression that we don't need struct type constructors but I might be wrong.

gzanitti: I used it mainly in the Statement processor when processing the Struct declaration. `TypeDeclaration` declares the new type of the struct and `TypeConstructor`, constructed the types of the internal functions (similar to enums, as you mentioned), using the type declarations of each one of the functions inside the declaration. 

For example: `struct Dot { x: int, y: int }`
would give us a `TypeDeclaration` Dot and two `TypeConstructor` Dot::x and Dot::y.

On the other hand, the Struct expression is used at the time of instantiation, for example `let f: int -> Dot = |i| Dot{a: 0, b: i};`

is this wrong?

My biggest question here is: somehow I have to process the internal types so that they are available later, but if they are not TypeConstructor, what should they be?

ast/src/parsed/display.rs
gzanitti: `TraitFunction` would be the ideal option, so I think we should rename them: https://github.com/powdr-labs/powdr/pull/1816

ast/src/object/mod.rs
```diff
@@ -50,10 +50,28 @@ pub struct PILGraph {
 
 #[derive(Clone)]
 pub enum TypeOrExpression {
```
chriseth, line 52: TypeDeclarationOrExpression?

ast/src/object/mod.rs
```diff
@@ -50,10 +50,28 @@ pub struct PILGraph {
...
-    Type(EnumDeclaration<Expression>),
+    Type(Type),
     Expression(TypedExpression),
 }
 
+#[derive(Clone)]
+pub enum Type {
```
chriseth, line 58: Please find a better name or see if we can re-use some other enum.

parser/src/lib.rs
```diff
@@ -394,6 +394,44 @@ namespace N(2);
         assert_eq!(input.trim(), printed.trim());
     }
 
+    #[test]
+    fn struct_decls() {
+        let input = r#"
+namespace N(2);
+    struct X {
```
chriseth, line 401: make it generic?

chriseth: and self-referential

pil-analyzer/src/condenser.rs
```diff
@@ -1101,6 +1101,32 @@ fn try_value_to_expression<T: FieldElement>(value: &Value<'_, T>) -> Result<Expr
...
+                            Err(e) => return Err(e),
+                        };
+                        Ok(NamedExpression {
+                            name: name.to_string(),
+                            body: Box::new(body),
+                        })
+                    })
+                    .collect::<Result<Vec<_>, _>>()?,
+            }
+            .into()
```
chriseth, line 1127:
```suggestion
            let name = Reference::Poly(PolynomialReference {
                name: name.to_string(),
                // We do not know the type args here.
                type_args: None,
            });
            let fields = fields
                    .iter()
                    .map(|(name, value)| {
                        let body = match try_value_to_expression(value) {
                            Ok(expr) => expr,
                            Err(e) => return Err(e),
                        };
                        Ok(NamedExpression {
                            name: name.to_string(),
                            body: Box::new(body),
                        })
                    })
                    .collect::<Result<Vec<_>, _>>()?;

            StructExpression { name, fields }.into()
```

pil-analyzer/src/evaluator.rs
```diff
@@ -976,8 +1005,36 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
...
+                StructExpression {
+                    name: reference,
+                    fields,
+                },
+            ) => {
+                let mut exp_fields = Vec::new();
+                for NamedExpression { name, body } in fields.iter() {
+                    self.expand(body)?;
+
+                    let value = self.value_stack.pop().ok_or(EvalError::SymbolNotFound(
```
chriseth, line 1019: should panic

pil-analyzer/src/evaluator.rs
```diff
@@ -976,8 +1005,36 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
...
-                unimplemented!("Struct expressions are not yet supported.")
+            Expression::StructExpression(
+                _,
+                StructExpression {
+                    name: reference,
+                    fields,
+                },
+            ) => {
+                let mut exp_fields = Vec::new();
+                for NamedExpression { name, body } in fields.iter() {
```
chriseth, line 1016: Please test for the fields to be specified in a different order.

chriseth, line 1016:
```suggestion
                for NamedExpression { name, body } in &fields {
```

chriseth, line 1016: Consider changing `body` to `value` inside `NamedExpression` (if it makes sense in all cases)

pil-analyzer/src/evaluator.rs
```diff
@@ -976,8 +1005,36 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
...
+
+                let reference = match reference {
+                    Reference::Poly(poly) => poly.name.as_str(),
+                    _ => unreachable!(),
+                };
+
+                self.value_stack.push(
+                    Value::Struct(StructValue {
+                        name: reference,
+                        fields: exp_fields,
```
chriseth, line 1034: Could be a HashMap instead of a vector of pairs.

pil-analyzer/src/evaluator.rs
```diff
@@ -1145,7 +1202,9 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
                 let body = if *condition { body } else { else_body };
                 return self.expand(body);
             }
-
+            Expression::StructExpression(_, StructExpression { name: _, fields: _ }) => {
```
chriseth, line 1205:
```suggestion
            Expression::StructExpression(_, _) => {
```

pil-analyzer/src/evaluator.rs
```diff
@@ -976,8 +1005,36 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
...
+            Expression::StructExpression(
+                _,
+                StructExpression {
+                    name: reference,
+                    fields,
+                },
+            ) => {
+                let mut exp_fields = Vec::new();
+                for NamedExpression { name, body } in fields.iter() {
+                    self.expand(body)?;
```
chriseth, line 1017: Don't call expend directly on them, but put an Expand on the op stack instead.

You can expand the first item directly if you like.

pil-analyzer/src/expression_processor.rs
```diff
@@ -189,7 +190,34 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
...
+                    .map(|args| args.into_iter().map(|t| self.process_type(t)).collect());
+
+                Expression::StructExpression(
+                    src,
+                    StructExpression {
+                        name: Reference::Poly(PolynomialReference {
+                            name: self.driver.resolve_type_ref(&reference.path),
+                            type_args,
+                        }),
+                        fields: fields
```
chriseth, line 211: I think we should check here that all the fields exist and that they are also all mentioned.

pil-analyzer/src/evaluator.rs
gzanitti: For now it's used only in Traits and Structs. Since they are NamedExpressions it seems to me that the most accurate would be `expr`.

What do you think?

pil-analyzer/src/expression_processor.rs
gzanitti: The biggest problem with this is that at this point we don't have the struct declaration. So this check is done during type checking (which was also improved).

ast/src/object/mod.rs
gzanitti: In fact, this enum doesn't seem to be used anymore. It seems to have been abandoned after the recent refactor.

pil-analyzer/src/expression_processor.rs
chriseth: The AnalysisDriver knows about all the symbols, so we could do it at this point. Not saying we should if it's too difficult.

pil-analyzer/src/evaluator.rs
chriseth: Yes, makes sense for `NamedExpression`  to use `name` and `expr` :)

pil-analyzer/src/evaluator.rs
```diff
@@ -1145,7 +1182,30 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
...
             }
+            Expression::StructExpression(_, StructExpression { name, fields }) => {
+                let new_fields = fields
+                    .iter()
+                    .rev()
+                    .map(|named_expr| {
+                        let value = self
+                            .value_stack
+                            .pop()
+                            .unwrap_or_else(|| panic!("Symbol {name} not found"));
```
chriseth, line 1193: This doesn't sound like the right error message.

pil-analyzer/src/evaluator.rs
```diff
@@ -1145,7 +1182,30 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
...
+                        (named_expr.name.as_str(), value)
+                    })
+                    .collect::<HashMap<_, _>>();
+
+                let v_name = match name {
+                    Reference::Poly(poly) => &poly.name,
+                    _ => unreachable!(),
+                };
```
chriseth, line 1201:
```suggestion
                let Reference::Poly(poly) = name else { unreachable!(); }
```

pil-analyzer/src/evaluator.rs
```diff
@@ -1145,7 +1182,30 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
                 let body = if *condition { body } else { else_body };
                 return self.expand(body);
             }
+            Expression::StructExpression(_, StructExpression { name, fields }) => {
+                let new_fields = fields
```
chriseth, line 1186: Why not just `fields`? In which way are they new?

gzanitti:
I started the simplification of this PR:
1) Definitions and statement processor: #1906 
2) Field checking before type checking: #1907

gzanitti:
This has been simplified into the following PRs:
1) Definitions and statement processor: #1906 
2) Field checking before type checking: #1907
3) Type inference: #1910 
4) Condenser & Evaluator: #1911

PR #1520, opened 2024-07-02
profiler fix for callgrind showing duplicate functions

pacheco:
use function id mapping in callgrind file to properly merge costs for each function


lvella (approved):
I am not sure what is happening here, but the code looks pretty.

PR #1530, opened 2024-07-03
Fix printing non-function fixed columns.

chriseth:


pil-analyzer/tests/parse_display.rs
```diff
@@ -728,3 +728,13 @@ namespace N(16);
     let formatted = analyze_string::<GoldilocksField>(input).to_string();
     assert_eq!(formatted, input);
 }
+
+#[test]
+fn reparse_non_function_fixed_cols() {
+    let input = r#"namespace X(16);
+    let A: int -> int = (|i| i);
```
leonardoalt, line 735: do we still need these parens?

leonardoalt (approved):

Parasos (commented):
fix_printing_non_function_fixed_cols

pil-analyzer/tests/parse_display.rs
chriseth: Yes, they are still printed in this case, don't really know what causes it.

PR #1531, opened 2024-07-03
Fix plonky3 empty test on release builds

chriseth:


plonky3/src/prover/mod.rs
```diff
@@ -134,6 +134,9 @@ mod tests {
 
     #[test]
     #[should_panic = "assertion failed: width >= 1"]
+    // The above failure is from a debug assertion, so we can only run this
+    // test if debug assertions are activated.
+    #[cfg(debug_assertions)]
```
pacheco, line 139: should the debug assertion be a real assertion?

chriseth: it's not in code we control

chriseth: `/plonky3-2a8c63bcd5ef83b0/7e4f7af/matrix/src/dense.rs:31`

Schaeff (approved):

PR #1545, opened 2024-07-08
Fixed cols in functions

chriseth:


pipeline/tests/asm.rs
```diff
@@ -790,3 +791,34 @@ fn trait_parsing() {
...
+            } = 0
+        };
+        machine Main {
+            col witness w;
+            f();
+        }
+        "#;
+    let output = asm_string_to_pil::<GoldilocksField>(input).to_string();
+    let expected = "    2 = 0;\n";
+    assert_eq!(output, expected);
```
leonardoalt, line 823: Maybe this test should be moved to one of the compiler crates instead of Pipeline which does end-to-end stuff?

pil-analyzer/tests/types.rs
```diff
@@ -560,3 +560,30 @@ fn empty_conditional() {
...
+    let input = "
+    let<T> f: T -> T[] = |i| [i];
+    let<Q> x: Q -> Q[][] = |i| {
+        let y: Q[] = f(i);
+        [y]
+    };
+    ";
+    type_check(input, &[]);
+}
+// TODO also check type vars used in types and also check this all for asm code.
```
leonardoalt, line 589: this PR or future?

pil-analyzer/src/type_inference.rs
```diff
@@ -641,21 +641,33 @@ impl TypeChecker {
...
-                                Type::Expr
+                            let value_type = match (ty, value) {
+                                (Some(ty), Some(value)) => {
+                                    self.process_concrete_symbol(ty.clone(), value)?;
+                                    ty.clone()
+                                }
+                                (None, Some(value)) => self.infer_type_of_expression(value)?,
+                                (Some(ty), None) => {
+                                    if *ty != Type::Col {
+                                        // TODO better source ref
```
leonardoalt, line 662: here or after?

pil-analyzer/src/type_inference.rs
```diff
@@ -641,21 +641,33 @@ impl TypeChecker {
...
+                                        return Err(source_ref.with_error("Let-declared variables without value must have type 'col'.".to_string()));
```
leonardoalt, line 663: is there a test that reaches this?

pil-analyzer/src/condenser.rs
```diff
@@ -309,13 +310,37 @@ impl<'a, T: FieldElement> SymbolLookup<'a, T> for Condenser<'a, T> {
...
+            {
+                if !lambda.outer_var_references.is_empty() {
+                    return Err(EvalError::TypeError(format!("Lambda expression for fixed column {name} must not reference outer variables.")))
+                }
+                Ok(FunctionValueDefinition::Expression(TypedExpression {
+                    e: Expression::LambdaExpression(source.clone(), (*lambda).clone()),
+                    type_scheme: None,
+                }))
+            } else {
+                Err(EvalError::TypeError(format!("Only lambda expressions are allowed for dynamically-created fixed columns. Got {v}.")))
```
leonardoalt, line 340: Is there a test that reaches this?

chriseth: I don't think so, it should not be able to pass the type checker.

pil-analyzer/src/type_inference.rs
chriseth: after

pil-analyzer/src/condenser.rs
chriseth: Added a test.

chriseth: Should I change this here to panic?

pil-analyzer/src/type_inference.rs
chriseth: Added a test.

pipeline/tests/asm.rs
chriseth: We don't have a crate that spans asm and pil, so pipeline is the correct crate IMO

pil-analyzer/src/condenser.rs
leonardoalt: Err sounds better

chriseth:
It seems that the type inference algorithm does not assign types to local variables (i.e. it does not modify the ast there). I think it should. Will add that.

pil-analyzer/src/type_inference.rs
leonardoalt: where? I don't see a diff from my latest refiew

leonardoalt: ah now

leonardoalt (approved):

PR #1546, opened 2024-07-08
Constr functions expected type moved to builtins

gzanitti:
This small PR pulls the definition of the ExpectedType `constr_function_statement_type` from pil_analyzer and moves it into type builtins.

### DIFF
--- a/pil-analyzer/src/type_builtins.rs
+++ b/pil-analyzer/src/type_builtins.rs
@@ -89,6 +89,7 @@ lazy_static! {
     .into_iter()
     .map(|(op, (vars, ty))| (op, parse_type_scheme(vars, ty)))
     .collect();
+
     static ref CONSTR_FUNCTION_STATEMENT_TYPE: ExpectedType = ExpectedType {
         ty: Type::NamedType(SymbolPath::from_str("std::prelude::Constr").unwrap(), None),
         allow_array: true,
--- a/pil-analyzer/src/type_builtins.rs
+++ b/pil-analyzer/src/type_builtins.rs
@@ -1,12 +1,16 @@
 use std::collections::HashMap;
 
-use powdr_ast::{
-    parsed::types::{ArrayType, Type, TypeScheme},
-    parsed::{BinaryOperator, UnaryOperator},
+use powdr_ast::parsed::{
+    asm::SymbolPath,
+    types::{ArrayType, Type, TypeScheme},
+    BinaryOperator, UnaryOperator,
 };
 use powdr_parser::parse_type_scheme;
+use std::str::FromStr;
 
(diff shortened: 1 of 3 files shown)


pil-analyzer/src/type_inference.rs
```diff
@@ -671,7 +668,7 @@ impl<'a> TypeChecker<'a> {
     /// Returns the type expected at statement level, given the current function context.
     fn statement_type(&self) -> ExpectedType {
         if self.lambda_kind == FunctionKind::Constr {
-            self.constr_function_statement_type.clone()
+            constr_function_statement_type()
```
gzanitti, line 671: If we allow this expected_type to be a builtin instead of coming from the typechecker itself, `self.constr_function_statement_type` is never read and can be deleted along with its lifetime `'a`


Of course, the price is that expected_type is no longer parameterised in the type_checker. What do you think?

chriseth: I mean yeah, that was the idea. It doesn't need to be parameterized because it will always be the same type. The main idea of moving it out was that the type checker should remain "pure" and all the builtins and the specific stuff should be somewhere else.

gzanitti: Perfect. Then it's ready for review :)

pil-analyzer/src/type_builtins.rs
```diff
@@ -85,6 +89,11 @@ lazy_static! {
     .into_iter()
     .map(|(op, (vars, ty))| (op, parse_type_scheme(vars, ty)))
     .collect();
+    static ref CONSTR_FUNCTION_STATEMENT_TYPE: ExpectedType = ExpectedType {
```
chriseth, line 92:
```suggestion

    static ref CONSTR_FUNCTION_STATEMENT_TYPE: ExpectedType = ExpectedType {
```

pil-analyzer/src/type_builtins.rs
```diff
@@ -99,6 +108,10 @@ pub fn unary_operator_scheme(op: UnaryOperator) -> TypeScheme {
     UNARY_OPERATOR_SCHEMES[&op].clone()
 }
 
+pub fn constr_function_statement_type() -> ExpectedType {
```
chriseth, line 111:
```suggestion
/// Returns the type allowed at statement level in `constr` functions.
pub fn constr_function_statement_type() -> ExpectedType {
```

pil-analyzer/src/type_builtins.rs
gzanitti: Had to rollback this change to avoid an error with `cargo fmt`

chriseth (approved):

PR #1548, opened 2024-07-08
Move counters inside condenser.

chriseth:


pil-analyzer/src/statement_processor.rs
```diff
@@ -56,6 +56,27 @@ impl Default for Counters {
...
+        let mut counters = Self::default();
+        if let Some(id) = identity {
+            counters.identity_counter = id + 1;
+        }
+        if let Some(id) = public {
+            counters.public_counter = id + 1;
+        }
+        for (kind, id) in symbols {
+            let counter = counters.symbol_counters.get_mut(&kind).unwrap();
+            *counter = std::cmp::max(*counter, id + 1);
```
chriseth, line 75: Ah sorry, this is actually wrong. Will fix.

pil-analyzer/src/condenser.rs
```diff
@@ -162,77 +161,31 @@ pub struct Condenser<'a, T> {
...
-            .values()
-            .filter_map(|(sym, _)| match sym.kind {
-                SymbolKind::Poly(PolynomialType::Committed) => {
-                    Some(sym.id + sym.length.unwrap_or(1))
-                }
-                _ => None,
-            })
-            .max()
-            .unwrap_or_default();
+        let counters = Counters::with_existing(symbols.values().map(|(sym, _)| sym), None, None);
```
lvella, line 175: use `.keys()` instead of `.values().map(|(sym, _)| sym)`

lvella (changes requested):

pil-analyzer/src/condenser.rs
chriseth: Unfortunately, not. The values are pairs. It would have worked to replace `items().map(...`, but not `values().map(...`

lvella: Ah, sorry. I may be still sleepy.

lvella (approved):

Issue #1550, opened 2024-07-08
Type inference ignores generic arguments 


chriseth:
The following code errs with "Unable to derive concrete type for reference to generic symbol cols.f", even though they were explicitly set in the call:

```rust
mod cols {
    enum Generic<T> {
        A(T),
        B,
    }
    let<T> f: -> Generic<T> = || {
            Generic::B
    };
}


machine Empty {
    let x;
    let fi: col = |i| match cols::f::<int>() {
        cols::Generic::A(_) => 7,
        cols::Generic::B => 9,
    };
    x = fi;

}
```

chriseth:
The issue seems to be that the generic type args (the `<int>` in `cols::f::<int>()`) is not preserved when translating asm to pil.

PR #1551, opened 2024-07-09
Type inference ignores generic arguments

gzanitti:
This PR solves issue #1550

### DIFF
--- a/pipeline/tests/asm.rs
+++ b/pipeline/tests/asm.rs
@@ -731,3 +731,9 @@ fn connect_no_witgen() {
     let i = [];
     verify_asm(f, slice_to_vec(&i));
 }
+
+#[test]
+fn generics_preservation() {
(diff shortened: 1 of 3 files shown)


pipeline/tests/asm.rs
```diff
@@ -731,3 +731,9 @@ fn connect_no_witgen() {
     let i = [];
     verify_asm(f, slice_to_vec(&i));
 }
+
+#[test]
+fn generics_preservation() {
+    let f = "asm/generics_preservation.asm";
+    verify_asm(f, Default::default());
```
gzanitti, line 738: Is there anything else to take out of here? Or maybe I should move the test somewhere else?

chriseth: I think we don't really have good tests for assembly apart from the pipeline tests. You can see if you can find a better place also for future tests

chriseth (approved):

PR #1552, opened 2024-07-09
Fail gracefully on different degrees in plonky3

Schaeff:
We had missed some backends in the original PR


leonardoalt:
Looks like this PR has no diff anymore? Closing for now

PR #1553, opened 2024-07-09
Generic type for Trait functions

gzanitti:


parser/src/powdr.lalrpop
```diff
@@ -691,11 +691,11 @@ TraitVars: Vec<String> = {
 
 TraitFunctions: Vec<TraitFunction<Expression>> = {
     => vec![],
-    <mut list:( <TraitFunction> "," )*> <end:TraitFunction> ","  => { list.push(end); list }
+    <mut list:( <TraitFunction> ";" )*> <end:TraitFunction> ";"  => { list.push(end); list }
 }
 
 TraitFunction: TraitFunction<Expression> = {
-    <name:Identifier> ":" <params:TypeTermList> "->" <value:TypeTermBox> => TraitFunction { name, ty: Type::Function(FunctionType{params, value}) }
+    <name:Identifier> ":" <ty:Type> => TraitFunction { name, ty }
```
chriseth, line 698: Do you know where the parsing ambiguity comes from?

I think we could change this to TypeTerm here and fix it, but then we would need to use parentheses for functions which is not ideal.

Would it be an option to have to lines here, one for an explicit function as before and one for TypeTerm?

chriseth: On the other hand, it's not too bad to use `;`

chriseth: Ah ok I think it's because `a: x, y, z -> w` would need too much look-ahead to find out the meaning of the `,` after the `x`. Although it's really not too much look-ahead, it's just a single token...

gzanitti: > Ah ok I think it's because a: x, y, z -> w would need too much look-ahead to find out the meaning of the , after the x. Although it's really not too much look-ahead, it's just a single token...

Yes, exactly, by using commas at the end of functions, the parser cannot differentiate when it's another type and when it's the end of the function.


> Would it be an option to have to lines here, one for an explicit function as before and one for TypeTerm?

Not really, because it would be equivalent to simply "opening" `Type` in its two possible variants and replacing them.


> On the other hand, it's not too bad to use `;`

And it's like in Rust :1st_place_medal:  :stuck_out_tongue_closed_eyes:

chriseth (approved):

parser/src/powdr.lalrpop
chriseth: I mean rust traits use `fn` functions.

I would really like traits to be similar to structs, and I think the two alternatives would work: Note that the second is not Type but TypeTerm, so if you want to use that to represent a function type you would have to put it in parentheses. But you could still use
```
trait X<T> {
  some_const: T,
  some_fun: T -> T,
}
```

chriseth:
Yeah ok, I understand that it won't work.

So I think we have to make a decision:
1. require parentheses for all functions
2. use `;`
3. only allow function types

I'm actually tending towards 3 to be honest (i.e. ignore this PR). We can still use constants by using a `-> T` type.

gzanitti:
Perfect, we'll use option three then. I will close this PR.
Thanks Chris!

PR #1554, opened 2024-07-10
Store references to captured variables.

chriseth:


pil-analyzer/tests/processor.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+#[test]
+fn determine_outer_var_refs() {
+    let input = "
+        let f: int, int -> int = |i, _| i;
+        let g: int, int -> int = |_, i| i;
+        let h: int -> (int -> (int, int)) = |i| |j| (f(i, j), g(i, j));
+        let k: int, int -> (int -> (int, int)) = |k, i| |j| (f(i, j), g(i, j));
+        ";
```
pacheco, line 40: i'm not sure i get what "outer vars" are capturing: aren't `f` and `g` captured by `h`?
why is `outer_vars_of_lambda(h)` empty?

chriseth: There are two different kinds of references in PIL (also in rust): local variables and globally accessible symbols.

Local variables have a scope and they can only be accessed after their declaration.
Global accessible symbols can be accessed "above" their declaration and they can also be recursive.

Let-declarations at the namespace scope in pil (or module scope in asm) are globally accessible symbols, while variables declared inside a function or more generally a `{`/`}`-block are local variables.

Another main difference between the two is that references to local variables always depend on the environment, so their value can be different depending on the "execution path" taken to get to the reference.

The function `h` for example returns a closure that references `i`. And that closure is a different thing depending on where I call `g(8)` or `g(2)`.

Long story short: The outer variable references only care about references to local variables because those are the ones relevant when capturing the environment.

Oh and another thing: This feature is used because I want to be able to declare fixed columns (and also define them as functions) dynamically inside a function. Since the condenser evaluates these functions and then "returns" pil code that also works in isolation, we need to make sure that all references still "work" in this "returned PIL". References to globally available symbols are not a problem, because they just need to be absolute instead of relative paths. For local variable references this doesn't work because they cannot be turned into absolute paths.

pacheco: nice, makes sense!

pacheco (approved):

PR #1555, opened 2024-07-10
Extract ROM machine

Schaeff:
This PR continues the task to replace lookups by links. The advantages of this approach are:
- with a link, the two machines can have different degrees and be proven separately
- with a lookup, only a monolithic proof works
- some backends such as Plonky3 do not support lookups
- this is backwards compatible, since in the monolithic setting, links are turned into lookups

In the compiler, we currently reduce each machine to another machine which has only pil code and links. To this end, in the case of virtual machines, we encode the program in fixed columns. This change introduces a separate machine to store the ROM. Therefore, each VM gets turned into not one, but two machines:

```
machine MyVM {
}
```

becomes

```
machine MyVM {
   MyVMROM _rom;
}

machine MyVMROM {
}
```

We introduce a new name alongside the original name, which pollutes the module. When raised, it was decided that we should not currently allow defining the ROM machine *inside* the VM.

A better long term solution would be to have a generic `ROM` machine in the stdlib which can be instantiated with the fixed columns which encode the program, using them in the `get_line` operation. There are a few missing pieces in the asm language to enable that.


airgen/src/lib.rs
```diff
@@ -26,13 +26,14 @@ const MAIN_FUNCTION: &str = "main";
 pub fn compile(input: AnalysisASMFile) -> PILGraph {
     let main_location = Location::main();
 
-    let non_std_machines = input
+    let neither_std_nor_rom_machines = input
         .machines()
         .filter(|(k, _)| k.parts().next() != Some("std"))
+        .filter(|(k, _)| !k.parts().last().unwrap().contains("ROM"))
```
chriseth, line 32: Who supplies ROM and what makes a machine a ROM machine?

I still don't understand the purpose of this PR but shouldn't the detection based on semantics instead of the name?

chriseth:
Can you please again list why we want to have this?

airgen/src/lib.rs
Schaeff: > Who supplies ROM and what makes a machine a ROM machine?

The ROM machine is generated internally from the machine's program. I agree the detection can be better here.

Schaeff:
@chriseth sure

The reasoning is that we want to generate a different proof for the VM and for the ROM. The way we do this is replacing the ROM lookup by a link to a ROM submachine.

chriseth:
Ok, that's a good reason, so you want to turn the lookup inside the machine into a cross-machine lookup? Do you want to do the same for all other lookups like fixed lookups for range constraints? I hope we can undo that compiler complexity soon...

Schaeff:
Yes, I already did it for a few lookups and I have a few more coming in another PR. I updated the description of this PR with what I think would be a way to simplify this, in general I suspect it will become easier once machines act more like structs.

airgen/src/lib.rs
Schaeff: I tried to come up with something more clever but I think this is fine. This would only fail if the user's main machine ends with "ROM".. I don't think it makes a big difference to base it on the inside of the machine, for example having a single operation named `get_line`.

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -43,11 +56,64 @@ pub enum LiteralKind {
...
+    mut pil: Vec<PilStatement>,
+    mut line_lookup: impl Iterator<Item = &'a str>,
+) -> Machine {
+    Machine {
+        operation_id: Some(ROM_OPERATION_ID.into()),
+        latch: Some(ROM_LATCH.into()),
+        pil: {
+            pil.extend([
+                parse_pil_statement(&format!("pol fixed {ROM_OPERATION_ID} = [0]*;")),
+                parse_pil_statement(&format!("pol fixed {ROM_LATCH} = [1]*;")),
```
leonardoalt, line 74:
```suggestion
                parse_pil_statement(&format!("let {ROM_OPERATION_ID} = [0]*;")),
                parse_pil_statement(&format!("let {ROM_LATCH} = [1]*;")),
```
?

Schaeff: reverted

leonardoalt (approved):

PR #1565, opened 2024-07-11
ArrayExpression implementation in PIL

Schaeff:
I implemented the `col fixed a = [1, 2, 3]* + [4, 5]` way to define fixed columns, in pil.
This was mostly an occasion for me to write some pil, but it might also be useful for us, since this implementation does not rely on the degree being available at analysis time, which is a blocker in #1558.

We could turn the above example into `expand(ArrayTerm::Repeat[1, 2, 3], ArrayTerm::Once([4, 5]))` somewhere in the compiler and remove the Rust implementation.

Note that I had to disable the optimizer for the tests to properly detect bugs, I'm not sure how to do it otherwise?

Thoughts welcome!


test_data/pil/expand_fixed.pil
```diff
@@ -0,0 +1,172 @@
...
+    let degree: -> int = [];
+
+namespace std::utils;
+    /// Evaluates to folder(...folder(folder(initial, f(0)), f(1)) ..., f(length - 1)),
+    /// i.e. calls f(0), f(1), ..., f(length - 1) and combines the results
+    /// using the function `folder`, starting with the value `initial`.
+    ///
+    /// See `sum` for an example use.
+    let<T1, T2>
+        fold: int, (int -> T1), T2, (T2, T1 -> T2) -> T2 = |length, f, initial, folder|
```
chriseth, line 30: I would propose to turn this test into an `.asm` test so that you don't have to copy the std

test_data/pil/expand_fixed.pil
```diff
@@ -0,0 +1,172 @@
...
+    /// TODO: Assert that lengths are equal when expressions are supported.
+    let<T1, T2, T3> zip: T1[], T2[], (T1, T2 -> T3) -> T3[] = |array1, array2, fn| new(len(array1), |i| fn(array1[i], array2[i]));
+
+namespace Main(8);
+
+    // A ArrayTerm in an `ArrayExpression`
+    enum ArrayTerm {
+        Repeat(int[]),
+        Once(int[])
+    }
```
chriseth, line 82:
```suggestion
    enum ArrayTerm {
        Repeat(ArrayTerm),
        Concatenate(ArrayTerm, ArrayTerm),
        Literal(int[]),
    }
```
?

test_data/pil/expand_fixed.pil
```diff
@@ -0,0 +1,172 @@
...
+            std::prover::degree() - total_size_of_non_repeated
+        }
+    };
+
+    /// repeat `repeated` until we reach a length of `total_len`, truncating
+    let expand_to_len = |repeated, total_len|
+        std::array::new(total_len, |j| repeated[j % std::array::len(repeated)]);
+
+    // expand an `ArrayExpression` into a single complete array
+    let expand: ArrayTerm[] -> int[] = |terms| {
```
chriseth, line 114: I think this should return an `int -> int` function instead. Explicitly creating the big arrays in the pil evaluator should be quite expensive.

test_data/pil/expand_fixed.pil
```diff
@@ -0,0 +1,172 @@
...
+    // returns the total size of the repeated array in this `ArrayExpression`
+    let solve: ArrayTerm[] -> int = |terms| {
+        let (_, total_size_of_non_repeated) = std::array::fold(terms, (std::prelude::false, 0), |(found_repeated, len), ArrayTerm| {
+            match ArrayTerm {
+                // we can have `[]*` as many times as we want
+                ArrayTerm::Repeat([]) => (found_repeated, len),
+                ArrayTerm::Repeat(a) => {
+                    if found_repeated {
+                        std::check::panic("unsolvable because two terms are repeated");
+                        (found_repeated, 42)
```
chriseth, line 93: I think this should work because panic returns `!`
```suggestion
                        std::check::panic("unsolvable because two terms are repeated")
```

test_data/pil/expand_fixed.pil
```diff
@@ -0,0 +1,172 @@
...
+                    } else {
+                        (std::prelude::true, len)
+                    }
+                },
+                ArrayTerm::Once(a) => (found_repeated, len + std::array::len(a))
+            }
+        });
+        if total_size_of_non_repeated > std::prover::degree() {
+            std::check::panic("non repeated array terms do not fit in degree");
```
chriseth, line 102: Could also use `std::assert(total... > degree, || "non repeated ...")`

test_data/asm/expand_fixed.asm
```diff
@@ -0,0 +1,101 @@
...
+                if found_repeated {
+                    std::check::panic("unsolvable because two terms are repeated")
+                } else {
+                    (true, len)
+                }
+            },
+            ArrayTerm::Once(a) => (found_repeated, len + std::array::len(a))
+        }
+    });
+    //std::check::assert(total_size_of_non_repeated <= degree(), || "non repeated array terms do not fit in degree");
```
Schaeff, line 26: @chriseth if I uncomment this line, I get `Tried to add constraints outside of statement context`

test_data/pil/expand_fixed.pil
Schaeff: I can change that but is necessary? I don't think the parser lets you create `Repeat(x)` where `x` is not `int[]`

test_data/asm/expand_fixed.asm
```diff
@@ -0,0 +1,101 @@
...
+let repeat: int[] -> ArrayTerm = |a| ArrayTerm::Repeat(a);
+
+let once: int[] -> ArrayTerm = |a| ArrayTerm::Once(a);
+
+// test utility
+let assert_eq: expr, expr -> () = constr |left, right| {
+    let diff;
+    diff = left - right;
+    diff = 0;
+};
```
chriseth, line 65: Would change this so that the check does not rely on the witness generator:
```suggestion
let assert_eq: (int -> int), int[] -> () = |fun, expected| {
    let _ = std::array::new(degree(), |i| assert_eq(fun(i), expected[i]))
};
```

test_data/asm/expand_fixed.asm
chriseth: I think this is a bug in the evaluator: If a statement evaluates to `()`, it should not add a constraint.

chriseth: Here is the fix: https://github.com/powdr-labs/powdr/pull/1568

Schaeff:
Closing in favor of #1623

PR #1576, opened 2024-07-15
fix riscv tests verification when using a composite backend

leonardoalt:
Needed for registers in memory. This will make some RISCV tests slower, but we won't use the normal EStarkDump backend anymore for such tests but rather composites backend only which should make them faster again.


georgwiese (approved):
Cool!

PR #1586, opened 2024-07-16
Field access in PIL

gzanitti:
This RP adds field access to PIL. At the moment, they are only valid for accessing fields within structs.
It goes on top of PR https://github.com/powdr-labs/powdr/pull/1516 which implements structs, so that one should be reviewed first.


leonardoalt:
Closing for now.

PR #1591, opened 2024-07-18
Parameterised expressions for Structs

gzanitti:
This PR 'parameterise' Expressions to be able to choose when they can be converted into a StructExpression and when they cannot. This modification allows us to overcome ambiguities with Structs and to adopt the standard Structs syntax:

```
struct Point{ x: int, y: int }
let p: Point = Point{ x: 1, y: 2 };
```

It goes on top of PR https://github.com/powdr-labs/powdr/pull/1516 which implements structs, so that one should be reviewed first.


chriseth:
Would it be a lot of work to put this PR first?

It looks quite complicated, but I also don't see a better option.

We want both `if a { ... }` and `if x >= 0 { ... }` to work. So the condition should support both Terms and more complex expressions.

gzanitti:
> We want both `if a { ... }` and `if x >= 0 { ... }` to work. So the condition should support both Terms and more complex expressions.

That shouldn't be a problem. The modification, the only thing it doesn't allow in the if, is something like
`if X{a:1, b:2}`, which is invalid anyway.

Closing this in favor of PR #1627, which includes the modification but not the structs.

PR #1594, opened 2024-07-22
Extract range checks from riscv into std machines

Schaeff:
Also, allow vms to have links outside instructions.

PR #1596, opened 2024-07-23
reduce the degree of compiled asm to pil so it works with P3

leonardoalt:
Depends on https://github.com/powdr-labs/powdr/pull/1594


georgwiese (approved):
Nice! I think in the future, these types of things should be automatic, but for now LGTM

leonardoalt:
yea agree

georgwiese (dismissed):

georgwiese (dismissed):

PR #1600, opened 2024-07-24
fix sub implementation in link_merging.asm test

onurinanc:
The previous version of link_merging.asm is not working correctly since the `operation sub<1>` implementation is wrong. With this PR, `operation sub<1>` implementation is fixed.


test_data/asm/permutations/link_merging.asm
```diff
@@ -29,7 +29,7 @@ machine Main {
 
     // these are merged into 1 link
     instr add X, Y -> Z link => Z = submachine.add(X, Y);
-    instr sub_with_add X, Y -> Z link => Y = submachine.add(X, Z);
+    instr sub_with_add X, Y -> Z link => X = submachine.add(Y, Z);
```
leonardoalt, line 32:
```suggestion
    instr sub_with_add X, Y -> Z link => Z = submachine.add(X, Y);
```
Shouldn't it be this?

onurinanc: `instr sub_with_add X, Y -> Z` means that we want X - Y = Z (X = Y + Z). So, shouldn't we use `link => X = add(Y, Z)`

onurinanc: It is not `instr add X, Y -> Z`, it is ìnstr sub_with_add`

leonardoalt: ah right, I missed the sub part, agree with your PR

leonardoalt (dismissed):

leonardoalt:
looks like some tests need updating though

leonardoalt (approved):

PR #1601, opened 2024-07-24
Traits overlapping

gzanitti:


pil-analyzer/src/pil_analyzer.rs
```diff
@@ -71,6 +75,7 @@ struct PILAnalyzer {
     symbol_counters: Option<Counters>,
     /// Symbols from the core that were added automatically but will not be printed.
     auto_added_symbols: HashSet<String>,
+    implementations: HashMap<String, Vec<(SourceRef, TraitImplementation<Expression>)>>,
```
chriseth, line 78: Please document.

Also maybe we could move the source ref into the implementation struct?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -144,8 +149,21 @@ impl PILAnalyzer {
 
         for PILFile(file) in files {
             self.current_namespace = Default::default();
-            for statement in file {
-                self.handle_statement(statement);
+            for ref statement in file {
+                if let PilStatement::TraitImplementation(sr, trait_impl) = statement {
```
chriseth, line 153: Why do you need to do it this early? This way we don't even see what is coming later on (we cannot guarantee that trait implementations are always after trait declarations).

I would have expected it to be right before type checking, or is there a reason to do it during statement processing?

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,99 @@
...
+                .as_ref()
+                .unwrap_or_else(|| panic!("Trait definition for {absolute_name} not found"));
+
+            let trait_decl = match trait_decl {
+                FunctionValueDefinition::TraitDeclaration(trait_decl) => trait_decl,
+                _ => unreachable!("Invalid trait declaration"),
+            };
+
+            if types1.len() != trait_decl.type_vars.len() {
+                panic!(
```
chriseth, line 45: I think we should try to return the error in new code.

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,99 @@
...
+                        types1.len(),
+                    ))
+                );
+            }
+
+            for (sr2, impl2) in implementations
+                .iter()
+                .skip(i + 1)
+                .filter_map(|stmt2| match stmt2 {
+                    (sr2, impl_) if impl_.name == impl1.name => Some((sr2, impl_)),
```
chriseth, line 60: shouldn't they all implement the same trait?

pil-analyzer/tests/types.rs
```diff
@@ -561,6 +561,122 @@ fn empty_conditional() {
...
+    let input = "
+    namespace std::convert(4);
+        let fe = || fe();
+    namespace F(4);
+    
+        trait Add<T> {
+            add: T, T -> T,
+        }
+
+        impl<T> Add<fe> {
```
chriseth, line 653: Shouldn't this be `impl Add<fe>`? It should be an error if the impl introduces a type variable and does not use it.

pil-analyzer/tests/types.rs
```diff
@@ -614,3 +730,50 @@ fn new_fixed_column_wrong_type() {
...
+        not: T -> T,
+    }
+    
+    impl<T> Not<Bool> {
+        not: |b| match b {
+            Bool::True => Bool::False,
+            Bool::False => Bool::True,
+        },
+    }
+    let b: Bool = Not::not(Bool::True);
```
chriseth, line 749:
```suggestion
    let b = Not::not(Bool::True);
```
Does this work?

pil-analyzer/tests/types.rs
```diff
@@ -614,3 +730,50 @@ fn new_fixed_column_wrong_type() {
...
+        },
+    }
+
+    impl<T, U> Convert<V2, V1> {
+        convert: |x| match x {
+            V2::B => V1::A,
+        },
+    }
+
+    let r1: V2 = Convert::convert(V1::A);
```
chriseth, line 775: Same here, would be more interesting without explicit type.

pil-analyzer/src/pil_analyzer.rs
gzanitti: > Why do you need to do it this early? This way we don't even see what is coming later on (we cannot guarantee that trait implementations are always after trait declarations).

Since this step is just to collect the implementations, I thought this would be the right location. In fact, this is one of the few steps where we can go over all the statements and keep the implementations before they are 'lost'.

The only thing that bothers me is having to call StatementProcessor, to transform the parsed::Expression into analyzed::Expression, but I don't think I have another option.


> I would have expected it to be right before type checking, or is there a reason to do it during statement processing?

The rest of the unification happens before the type check.

pil-analyzer/tests/types.rs
gzanitti: Yes, this one works (probably because there is only one impl). 
The one that doesn't work is the following case, `Convert`.

pil-analyzer/tests/types.rs
gzanitti: Why? shouldn't it be possible to derive the type of each let?

pil-analyzer/src/traits_processor.rs
gzanitti: Except for the saving of implementations, everything happens just before type checking:

https://github.com/powdr-labs/powdr/blob/1b8d5585957c88b137aa120d9de7cedb761a2c0e/pil-analyzer/src/pil_analyzer.rs#L55

pil-analyzer/tests/types.rs
gzanitti: Added a check about this: https://github.com/powdr-labs/powdr/blob/b175ff3eb91a6ac3f9ad4719e54716a427fa1471/pil-analyzer/src/traits_processor.rs#L66

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -144,8 +152,21 @@ impl PILAnalyzer {
 
         for PILFile(file) in files {
             self.current_namespace = Default::default();
-            for statement in file {
-                self.handle_statement(statement);
+            for ref statement in file {
+                if let PilStatement::TraitImplementation(_, trait_impl) = statement {
```
chriseth, line 156: I don't see why trait implementations cannot be handled in the same way as other pil statements:

StatementProcessor::handle_statement gets another section for trait implementations and returns a new kind of PILItem. There you also have access to the AnalysisDriver and the expression processor to turn parsed types into analyzed types.

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+    analyzed::{Expression, FunctionValueDefinition, Symbol},
+    parsed::{
+        types::{TupleType, Type},
+        TraitImplementation,
+    },
+};
+
+use crate::{pil_analyzer::Driver, type_unifier::Unifier, AnalysisDriver};
+
+/// Checks for overlapping trait implementations in the current `PILAnalyzer` instance.
```
chriseth, line 13: Please explain in the comment what overlapping means. This is much more important to explain here than how it does it (we can look at the code to check how it is done).

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Validates the trait implementation definitions in the given `implementations` map against the trait
+/// declarations in the `definitions` map.
+fn validate_impl_definitions(
+    implementations: &HashMap<String, Vec<TraitImplementation<Expression>>>,
+    definitions: &HashMap<String, (Symbol, Option<FunctionValueDefinition>)>,
+    driver: Driver,
+) {
+    for impls in implementations.values() {
+        for trait_impl in impls.iter() {
```
chriseth, line 36:
```suggestion
fn validate_impl_definitions(
    implementations: impl Iterator<Item = &TraitImplementation<Expression>>,
) {
    for trait_impl in implementations {
```

Maybe even just pass a single implementation?

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+            let absolute_name = driver.resolve_decl(trait_impl.name.name());
+
+            let trait_decl = definitions
+                .get(&absolute_name)
+                .unwrap_or_else(|| panic!("Trait {absolute_name} not found"))
+                .1
+                .as_ref()
+                .unwrap_or_else(|| panic!("Trait definition for {absolute_name} not found"));
+
+            let trait_decl = match trait_decl {
```
chriseth, line 49: Maybe rather use `if let ... else panic!()`?

pil-analyzer/src/traits_processor.rs
chriseth: What I meant was: If the driver is used to resolve names properly at processing time, the errors above will not be user errors but assertion failures. So we can keep them as panics. But the error here that the number of trait type parameters does not match should be a Result::Err.

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+                        trait_decl.type_vars.len(),
+                        types.len(),
+                    ))
+                );
+            }
+
+            for var in trait_impl.type_scheme.vars.vars() {
+                if !types
+                    .iter()
+                    .any(|t| matches!(t, Type::NamedType(v, _) if v.name() == var))
```
chriseth, line 69: I think you should compare `types.iter().map(|t| t.contained_type_vars())` to the type vars in the type scheme and check for equality. It might help to keep the tuple type around, then you don't need to iterate over `types`

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+                }
+            }
+        }
+    }
+}
+
+/// Ensures that there are no overlapping trait implementations in the given `implementations` map.
+///
+/// This function iterates through all the trait implementations comparing them with each other and checks that:
+/// - The number of type variables between implementations is the same.
```
chriseth, line 86:
```suggestion
```

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+/// - There are no traits with overlapping type variables and the same name between the implementations.
+fn ensure_unique_impls(
+    implementations: &HashMap<String, Vec<TraitImplementation<Expression>>>,
+    driver: Driver,
+) {
+    for implementations in implementations.values() {
```
chriseth, line 92: I think the interface could be nice if this loop is outside the function

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+        for (i, impl1) in implementations.iter().enumerate() {
+            for impl2 in implementations.iter().skip(i + 1) {
+                let Type::Tuple(TupleType { items: types1 }) = &impl1.type_scheme.ty else {
+                    panic!("Type from trait scheme is not a tuple.")
+                };
+                let Type::Tuple(TupleType { items: types2 }) = &impl2.type_scheme.ty else {
+                    panic!("Type from trait scheme is not a tuple.")
+                };
+
+                if types1.len() != types2.len() {
```
chriseth, line 102: Isn't this checked in the other function?

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+                            "Impl types have different lengths: {} with {} vs {} with {}",
+                            driver.resolve_decl(impl1.name.name()),
+                            types1.len(),
+                            driver.resolve_decl(impl2.name.name()),
+                            types2.len()
+                        ))
+                    );
+                }
+
+                unify_traits_types(impl1.type_scheme.ty.clone(), impl2.type_scheme.ty.clone())
```
chriseth, line 115: I think you need to be careful with the type variables here: If impl1 and impl2 use the same name for a type variable, the unifier will assume the type variables to be the same. You essentially need what TypeChecker::instantiate_scheme.

Maybe it helps to extract the type var counter and the instantiate_scheme function into a new component that handles type variables.

Although maybe I'm wrong and it doesn't matter, at least I cannot think of a way to exploit this.

pil-analyzer/tests/types.rs
```diff
@@ -561,6 +561,136 @@ fn empty_conditional() {
...
+#[should_panic = "Impls for Add: Types (int, fe) and (int, fe) overlap"]
+fn duplicated_trait() {
+    let input = "
+    trait Add<T, Q> {
+        add: T, T -> Q,
+    }
+    impl Add<int, fe> {
+        add: |a, b| a + b,
+    }
+    impl Add<int, fe> {
```
chriseth, line 599: Can you add another test where this second implementation is generic, e.g. `impl<T> Add<T, fe>{}`?

pil-analyzer/tests/types.rs
```diff
@@ -561,6 +561,136 @@ fn empty_conditional() {
...
+
+#[test]
+#[should_panic = "Impls for Add: Types (int, fe) and (int, fe) overlap"]
+fn duplicated_trait() {
+    let input = "
+    trait Add<T, Q> {
+        add: T, T -> Q,
+    }
+    impl Add<int, fe> {
+        add: |a, b| a + b,
```
chriseth, line 597: If we ever swap type checking and overlap checking around, this test will not test the overlapping any more. Can you change this so that everything is correct except the overlapping?

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+            let trait_decl = match trait_decl {
+                FunctionValueDefinition::TraitDeclaration(trait_decl) => trait_decl,
+                _ => unreachable!("Invalid trait declaration"),
+            };
+
+            if types.len() != trait_decl.type_vars.len() {
+                panic!(
+                    "{}",
+                    trait_impl.source_ref.with_error(format!(
+                        "Trait {} has {} type vars, but implementation has {}",
```
chriseth, line 58: It's more like type parameters, right?

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,133 @@
...
+
+            for var in trait_impl.type_scheme.vars.vars() {
+                if !types
+                    .iter()
+                    .any(|t| matches!(t, Type::NamedType(v, _) if v.name() == var))
+                {
+                    panic!(
+                        "{}",
+                        trait_impl.source_ref.with_error(format!(
+                            "Impl {absolute_name} has a type var {var} that is not defined in the type tuple",
```
chriseth, line 74:
```suggestion
                            "Impl {absolute_name} introduces a type variable {var} that is not used",
```

pil-analyzer/tests/types.rs
```diff
@@ -561,6 +561,136 @@ fn empty_conditional() {
...
+    }
+    impl<Q> Add<int> {
+        add: |a, b| a + b,
+    }
+    ";
+    type_check(input, &[]);
+}
+
+#[test]
+fn impl_type_resolution() {
```
chriseth, line 657: Shouldn't this be part of the other PR?

pil-analyzer/tests/types.rs
```diff
@@ -561,6 +561,136 @@ fn empty_conditional() {
...
+            add: |a, b| a + b,
+        }
+
+        let r: int = Add::add(3, 4);
+    ";
+    type_check(input, &[("F.r", "", "int")]);
+}
+
+#[test]
+fn trait_multi_generics() {
```
chriseth, line 681: Same here, this test is only useful once we resolve impls.

pil-analyzer/tests/types.rs
```diff
@@ -614,3 +744,50 @@ fn new_fixed_column_wrong_type() {
     "#;
     type_check(input, &[]);
 }
+
+#[test]
+fn trait_with_user_defined_enum() {
```
chriseth, line 749: These two tests also belong to the other PR

pil-analyzer/src/traits_processor.rs
gzanitti: You were right, but in order to isolate that functionality I had to create this:

https://github.com/powdr-labs/powdr/blob/2bba7283dcc9ec8e85c5622f065db8d52be17c48/pil-analyzer/src/traits_processor.rs#L142

that receives a unifier (if we want to check bounds) and that should be shared by the new traits step and the type checker (I assume replacing last_type_var by something like type_var_manager in the type checker, not yet implemented) to be able to keep track of the amount of variables created.

But it results in a rather annoying integration between the parts. Especially having to include another unifier (which would give us two different unifiers between the TypeChecker and this new manager).

chriseth: We could also move the functionality inside the unifier. It kind of makes sense, since all type variables that are used inside the unifier must have been created via the `new_type_var` function - otherwise we get unexpected name clashes.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -50,10 +51,16 @@ pub fn analyze_string<T: FieldElement>(contents: &str) -> Analyzed<T> {
...
+    let mut unifier = Unifier::new();
+
     analyzer.process(files);
     analyzer.side_effect_check();
-    check_traits_overlap(&mut analyzer.implementations, &analyzer.definitions);
-    analyzer.type_check();
+    check_traits_overlap(
+        &mut analyzer.implementations,
+        &analyzer.definitions,
+        &mut unifier,
```
chriseth, line 61: Are you sure it needs to be the same instance?

pil-analyzer/src/type_inference.rs
```diff
@@ -962,12 +953,11 @@ impl TypeChecker {
     }
 
     fn new_type_var_name(&mut self) -> String {
```
chriseth, line 955: Maybe we don't really need this function any more (same for the other)

pil-analyzer/src/type_unifier.rs
```diff
@@ -137,4 +150,31 @@ impl Unifier {
         self.substitutions.insert(type_var, ty);
         Ok(())
     }
+
+    pub fn instantiate_scheme(&mut self, scheme: TypeScheme) -> (Type, Vec<Type>) {
```
chriseth, line 154: Can you move the pub functions to the top?

parser/src/powdr.lalrpop
```diff
@@ -726,9 +726,9 @@ NamedExpression: NamedExpression<Expression> = {
     <name:Identifier> ":" <body:BoxedExpression> => NamedExpression { name, body }
 }
 
-GenericTraitName: (SymbolPath, TypeScheme) = {
-    <vars:("<" <TypeVarBounds> ">")> <name:SymbolPath> <items:("<" <TypeTermList<ArrayLengthNumber>> ">")> =>
-        (name, TypeScheme{ vars, ty: Type::Tuple(TupleType{items}) })
+GenericTraitName: (String, TypeScheme) = {
```
chriseth, line 729: I think this should stay symbolpath

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -49,8 +50,10 @@ pub fn analyze_string<T: FieldElement>(contents: &str) -> Analyzed<T> {
 
 fn analyze<T: FieldElement>(files: Vec<PILFile>) -> Analyzed<T> {
     let mut analyzer = PILAnalyzer::new();
+
     analyzer.process(files);
     analyzer.side_effect_check();
+    check_traits_overlap(&mut analyzer.implementations, &analyzer.definitions);
```
chriseth, line 56: To keep the style, it might make sense to add a `check_traits_overlap` function to Analyzer which then calls the free function

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -71,6 +74,8 @@ struct PILAnalyzer {
     symbol_counters: Option<Counters>,
     /// Symbols from the core that were added automatically but will not be printed.
     auto_added_symbols: HashSet<String>,
+    /// Implementations found, organized according to their associated trait name.
```
chriseth, line 77:
```suggestion
    /// Trait implementations found, organized according to their associated trait name.
```

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -436,7 +446,7 @@ impl PILAnalyzer {
 }
 
 #[derive(Clone, Copy)]
-struct Driver<'a>(&'a PILAnalyzer);
+pub struct Driver<'a>(&'a PILAnalyzer);
```
chriseth, line 449:
```suggestion
struct Driver<'a>(&'a PILAnalyzer);
```

pil-analyzer/src/statement_processor.rs
```diff
@@ -647,4 +653,29 @@ where
             functions,
         }
     }
+
+    pub fn process_trait_implementation(
```
chriseth, line 657:
```suggestion
    fn process_trait_implementation(
```

pil-analyzer/src/statement_processor.rs
```diff
@@ -647,4 +653,29 @@ where
...
+        &self,
+        trait_impl: parsed::TraitImplementation<parsed::Expression>,
+    ) -> TraitImplementation<Expression> {
+        let type_vars = trait_impl.type_scheme.vars.vars().collect();
+        let functions = trait_impl
+            .functions
+            .iter()
```
chriseth, line 664:
```suggestion
            .into_iter()
```
(then you don't need to clone the name I think)

pil-analyzer/src/statement_processor.rs
```diff
@@ -647,4 +653,29 @@ where
...
+                name: named.name.clone(),
+                body: Box::new(
+                    self.expression_processor(&type_vars)
+                        .process_expression(named.body.as_ref().clone()),
+                ),
+            })
+            .collect();
+
+        TraitImplementation {
+            name: self.driver.resolve_decl(&trait_impl.name),
```
chriseth, line 675: I think this should use `resolve_ref` with TraitDeclaration as symbol category.

chriseth: Also maybe we could rename `name` to `trait`? But not sure if that makes it easier to understand.

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,139 @@
...
+    },
+};
+
+use crate::type_unifier::Unifier;
+
+/// Checks for overlapping trait implementations in the current `PILAnalyzer` instance.
+///
+/// This method iterates through all the trait implementations.
+/// For each implementation, it checks that there are no traits with overlapping type vars and the same name between them.
+/// Overlapping occurs when there are two implementations of the same trait whose trait types can be unified.
```
chriseth, line 17: ... If two trait implementation are overlapping, there is the possibility for a trait function call to be ambiguous even though all types are fully concrete.

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+fn unify_traits_types(ty1: TypeScheme, ty2: TypeScheme) -> Result<(), String> {
+    let mut unifier = Unifier::new();
+    let instantiated_ty1 = unifier.instantiate_scheme(ty1).0;
+    let instantiated_ty2 = unifier.instantiate_scheme(ty2).0;
+
+    match unifier.unify_types(instantiated_ty1.clone(), instantiated_ty2.clone()) {
+        Ok(_) => Err(format!(
+            "Types {instantiated_ty1} and {instantiated_ty2} overlap"
```
chriseth, line 135: We should improve this error: First of all, we should use the non-instantiated type schemes (which use type variables familiar to the user), and then we should apply the substitutions to the type scheme to give an example type that applies to  both implementations.

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,139 @@
...
+
+use crate::type_unifier::Unifier;
+
+/// Checks for overlapping trait implementations in the current `PILAnalyzer` instance.
+///
+/// This method iterates through all the trait implementations.
+/// For each implementation, it checks that there are no traits with overlapping type vars and the same name between them.
+/// Overlapping occurs when there are two implementations of the same trait whose trait types can be unified.
+///
+/// It also checks that the number of type variables in the implementation matches
```
chriseth, line 19: Then the function should be renamed to a more generic name

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,139 @@
...
+            .unwrap_or_else(|| panic!("Trait {trait_name} not found"))
+            .1
+            .as_ref()
+            .unwrap_or_else(|| panic!("Trait definition for {trait_name} not found"));
+
+        let trait_decl = if let FunctionValueDefinition::TraitDeclaration(trait_decl) = trait_decl {
+            trait_decl
+        } else {
+            panic!("Invalid trait declaration");
+        };
```
chriseth, line 39:
```suggestion
        if let FunctionValueDefinition::TraitDeclaration(trait_decl) = trait_decl else {
            panic!("Invalid trait declaration");
        };
```

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,139 @@
...
+///
+/// It also checks that the number of type variables in the implementation matches
+/// the number of type variables in the corresponding trait declaration.
+pub fn check_traits_overlap(
+    implementations: &mut HashMap<String, Vec<TraitImplementation<Expression>>>,
+    definitions: &HashMap<String, (Symbol, Option<FunctionValueDefinition>)>,
+) {
+    for trait_impls in implementations.values_mut() {
+        // All the impls in trait_impls are of the same trait declaration.
+        let trait_name = trait_impls[0].name.clone();
```
chriseth, line 27: Isn't the name also the key? so you could use `implementations.items_mut()` or something

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,139 @@
...
+///
+/// This function iterates through all the trait implementations comparing them with each other and ensure that
+/// there are no traits with overlapping type variables.
+fn ensure_unique_impls(implementations: &mut [TraitImplementation<Expression>]) {
+    for i in 0..implementations.len() {
+        let type_vars: HashSet<_> = implementations[i].type_scheme.vars.vars().collect();
+        implementations[i]
+            .type_scheme
+            .ty
+            .map_to_type_vars(&type_vars);
```
chriseth, line 105: This should be done already at this stage.

leonardoalt:
Closing for now.

PR #1602, opened 2024-07-24
Executor generating witnesses for all columns

pacheco:
Hand-written witness generation for the RISCV `large_field` executor.
Generates `main` and default submachine columns (no "coprocessors" like poseidon).


leonardoalt:
We can also wait for the new implementations of poseidon and arith that use memory which will simplify their implementation here too

cli-rs/src/main.rs
```diff
@@ -113,7 +114,8 @@ enum Commands {
         #[arg(default_value_t = false)]
         continuations: bool,
     },
-    /// Executes a powdr-asm file with given inputs.
+    /// Execute a powdr-asm file with given inputs.
+    /// Does not generate a witness.
     Execute {
```
pacheco, line 119: now there are two separate commands: `execute` for running the riscv executor in "fast" non witgen mode, and `witgen` which runs in "trace" mode and generates a valid witness

cli-rs/src/main.rs
```diff
@@ -134,15 +136,46 @@ enum Commands {
...
+        #[arg(long)]
+        #[arg(default_value_t = false)]
+        generate_flamegraph: bool,
+
+        /// Generate callgrind file of the execution ("[file].callgrind")
+        #[arg(long)]
+        #[arg(default_value_t = false)]
+        generate_callgrind: bool,
+    },
+    /// Execute and generate a valid witness for a powdr-asm file with the given inputs.
```
leonardoalt, line 149:
```suggestion
    /// Execute and generate a valid witness for a RISCV powdr-asm file with the given inputs.
```

cli-rs/src/main.rs
```diff
@@ -325,34 +417,67 @@ fn execute<F: FieldElement>(
...
+            &[],
+            None,
+            profiling,
+        );
+
+        let duration = start.elapsed();
+        log::info!("Executor done in: {:?}", duration);
+        log::info!("Execution trace length: {}", execution.main_trace_len);
+
+        let pil = pipeline.compute_optimized_pil().unwrap();
```
leonardoalt, line 446: Isn't this the same as line 427?

riscv/src/large_field/code_gen.rs
```diff
@@ -323,8 +323,13 @@ fn preamble(field: KnownField, runtime: &Runtime, with_bootloader: bool) -> Stri
     
     // Jump to the address in register X and store the return program counter in register W.
     instr jump_dyn X, W
-        link ~> pc' = regs.mload(X, STEP)
-        link ~> regs.mstore(W, STEP, pc + 1);
+        link ~> tmp1_col = regs.mload(X, STEP)
+        link ~> regs.mstore(W, STEP, pc + 1)
+    {
+        // TODO: using a tmp col here avoids an extra selector column, because
+        // links with next references on LHS can't currenlty be merged with other links
```
leonardoalt, line 330:
```suggestion
        // links with next references on LHS can't currently be merged with other links
```

leonardoalt (commented):
Partial review, everything outside of riscv-executor.

riscv/tests/common/mod.rs
```diff
@@ -28,25 +28,36 @@ pub fn verify_riscv_asm_string<T: FieldElement, S: serde::Serialize + Send + Syn
...
-            Default::default(),
+            None,
+            None,
         );
-        let pipeline_gl: Pipeline<GoldilocksField> =
-            unsafe { std::mem::transmute(pipeline.clone()) };
-        run_pilcom_with_backend_variant(pipeline_gl, BackendVariant::Composite).unwrap();
+        pipeline.rollback_from_witness();
+        let executor_trace: Vec<_> = execution.trace.into_iter().collect();
+        let pipeline = pipeline.add_external_witness_values(executor_trace);
```
leonardoalt, line 58: Should we add a check somewhere in the tests that all cols were computed? Or does that not make sense atm?

riscv-executor/src/lib.rs
```diff
@@ -61,6 +68,26 @@ impl<F: FieldElement> Elem<F> {
...
+    pub fn from_u32_as_fe(value: u32) -> Self {
+        Self::Field(F::from(value))
+    }
+
+    pub fn from_i64_as_fe(value: i64) -> Self {
+        Self::Field(F::from(value))
+    }
+
+    pub fn from_u64_as_fe(value: u64) -> Self {
+        Self::Field(F::from(value))
```
leonardoalt, line 80: is this checked or unchecked?

riscv-executor/src/lib.rs
```diff
@@ -272,13 +374,18 @@ mod builder {
...
 
     use crate::{
-        Elem, ExecMode, ExecutionTrace, MemOperation, MemOperationKind, MemoryState, RegWrite,
-        RegisterMemory, RegisterMemoryState, PC_INITIAL_VAL,
+        BinaryMachine, Elem, ExecMode, Execution, ExecutionTrace, MemOperation, MemOperationKind,
+        MemoryMachine, MemoryState, RegWrite, RegisterMemory, ShiftMachine, SplitGlMachine,
+        Submachine, SubmachineBoxed, PC_INITIAL_VAL,
     };
 
+    // TODO: where do we get this from? I don't think we want to depend on the linker crate just for this
```
leonardoalt, line 382: It should be in the parsed asm somewhere. Both min and max are defined explicitly in the current impl of the riscv gl machine.

riscv-executor/src/lib.rs
```diff
@@ -272,13 +374,18 @@ mod builder {
...
+    const MIN_DEGREE: u32 = 1 << 5;
+
     fn register_names(main: &Machine) -> Vec<&str> {
         main.registers
             .iter()
+            //.map(|statement| &statement.name[..])
```
leonardoalt, line 388: leftover?

riscv-executor/src/lib.rs
```diff
@@ -422,7 +555,8 @@ mod builder {
         ///
         /// to set the PC, use set_pc() instead of this
         pub(crate) fn set_reg(&mut self, idx: &str, value: impl Into<Elem<F>>) {
-            self.set_reg_impl(idx, value.into())
+            let v = value.into();
+            self.set_reg_impl(idx, v)
```
leonardoalt, line 559: why the change?

riscv-executor/src/lib.rs
```diff
@@ -433,7 +567,7 @@ mod builder {
 
         /// raw set next value of register by register index instead of name
         fn set_reg_idx(&mut self, idx: u16, value: Elem<F>) {
-            // Record register write in trace.
+            // Record register write in trace. Only for non-assignment registers.
```
leonardoalt, line 570: I guess this only happens for the pc now then?

riscv-executor/src/lib.rs
```diff
@@ -510,8 +697,76 @@ mod builder {
             self.reg_mem.second_last = self.reg_mem.last.clone();
         }
 
-        pub fn finish(self) -> (ExecutionTrace<F>, MemoryState, RegisterMemoryState<F>) {
-            (self.trace, self.mem, self.reg_mem.for_bootloader())
+        pub fn finish(mut self) -> Execution<F> {
+            if let ExecMode::Fast = self.mode {
+                return Execution {
+                    // TODO: in FAST mode we don't extend the trace to next power of two. What should this value be in TRACE/FAST?
+                    main_trace_len: self.len(),
```
leonardoalt, line 704: Answering the comment: I guess this value is fine?

riscv-executor/src/lib.rs
```diff
@@ -510,8 +697,76 @@ mod builder {
...
+
+            // add submachine traces to main trace
+            for (name, mut machine) in self.submachines {
+                if name.is_empty() {
+                    // ignore empty machines
+                    continue;
+                }
+                machine.final_row_override();
+                // extend with dummy blocks up to the required machine degree
+                let machine_degree = std::cmp::max(machine.len().next_power_of_two(), MIN_DEGREE);
```
leonardoalt, line 730: should use the MIN_DEGREE of the machine here, not  Main's

riscv-executor/src/lib.rs
```diff
@@ -510,8 +697,76 @@ mod builder {
...
+                    machine.push_dummy_block();
+                }
+                for (col_name, col) in machine.take_cols() {
+                    assert!(self.trace.cols.insert(col_name, col).is_none());
+                }
+            }
+
+            // add regs memory trace
+            let regs_degree =
+                std::cmp::max(self.regs_machine.len().next_power_of_two(), MIN_DEGREE);
```
leonardoalt, line 741: same here

riscv-executor/src/lib.rs
```diff
@@ -645,59 +904,202 @@ struct Executor<'a, 'b, F: FieldElement> {
     label_map: HashMap<&'a str, Elem<F>>,
     inputs: &'b Callback<'b, F>,
     bootloader_inputs: Vec<Elem<F>>,
-    _stdout: io::Stdout,
+    fixed: Arc<Vec<(String, VariablySizedColumn<F>)>>,
+    // program columns: maps fixed cols to respective witness cols
```
leonardoalt, line 908: maps fixed to witness? wdym?

riscv-executor/src/lib.rs
```diff
@@ -645,59 +904,202 @@ struct Executor<'a, 'b, F: FieldElement> {
...
+            .map(|(_, v)| v.get_uniquely_sized().expect("not uniquely sized!"))
+    }
+
+    fn sink_id(&self) -> u32 {
+        // sink_id is the length of the ROM. we find it by looking at the line instr__loop activates
+        for (i, &val) in self
+            .get_fixed("main__rom::p_instr__loop")
+            .unwrap()
+            .iter()
+            .enumerate()
```
leonardoalt, line 949:
```suggestion
            .position(|&v| v.is_one())
```

riscv-executor/src/lib.rs
```diff
@@ -645,59 +904,202 @@ struct Executor<'a, 'b, F: FieldElement> {
...
+                let val = Elem::Field(
+                    *self
+                        .get_fixed(fixed)
+                        .unwrap_or(&Vec::new())
+                        .get(pc as usize)
+                        .unwrap_or(&F::zero()),
+                );
+                self.proc.set_col(col, val);
+            }
+            // always 2 because RISCV only uses constrainted submachines
```
leonardoalt, line 990: Don't really get what this means

riscv-executor/src/lib.rs
```diff
@@ -645,59 +904,202 @@ struct Executor<'a, 'b, F: FieldElement> {
...
+        set_col!(W, get_col!(W_const));
+        self.proc
+            .set_col(&format!("main::instr_{name}"), Elem::from_u32_as_fe(1));
+
+        let r = match name {
             "set_reg" => {
                 let addr = args[0].u();
-                self.proc.set_reg_mem(addr, args[1]);
+                let val = args[1];
+                self.reg_write(0, addr, val, 3);
```
leonardoalt, line 1027: 3?

riscv-executor/src/lib.rs
```diff
@@ -645,59 +904,202 @@ struct Executor<'a, 'b, F: FieldElement> {
...
+                if !get_col!(Y_read_free).is_zero() {
+                    set_col!(Y_free_value, val);
+                }
 
                 Vec::new()
             }
             "get_reg" => {
                 let addr = args[0].u();
-                let val = self.proc.get_reg_mem(addr);
+                let val = self.reg_read(0, addr, 0);
```
leonardoalt, line 1039: The first and last arguments are kinda confusing overall

riscv-executor/src/lib.rs
```diff
@@ -645,59 +904,202 @@ struct Executor<'a, 'b, F: FieldElement> {
...
+                let value = self.reg_read(2, read_reg3, 2);
 
                 let addr = addr1.bin() - addr2.bin() + offset;
                 let addr = addr as u32;
                 assert_eq!(addr % 4, 0);
                 self.proc.set_mem(addr, value.u());
 
+                // Calculate the address with wrapping
+                let addr_full = addr1.bin() - addr2.bin() + offset;
+                // let addr_wrapped = (addr_full % (2_u64.pow(32) as i64)) as u32;
```
leonardoalt, line 1074: leftover?

riscv-executor/src/lib.rs
```diff
@@ -707,27 +1109,51 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
+                self.proc.memory_machine.read(
+                    self.step + 1,
+                    addr as u32 & 0xfffffffc,
+                    val.into(),
+                    0,
+                );
 
                 Vec::new()
             }
+            // TODO: update to witness generation for continuations
```
leonardoalt, line 1139: what's missing here for continuations?

leonardoalt (commented):
Done.

cli-rs/src/main.rs
pacheco: yep, removing

riscv-executor/src/lib.rs
pacheco: this one is actually not used, i'll remove it... and I think the other one can also be removed in favor of the 32 bit ones

riscv-executor/src/lib.rs
pacheco: prob. leftover from some "println debugging", ill revert

riscv-executor/src/lib.rs
pacheco: there are also the two `query_arg` registers

riscv-executor/src/lib.rs
pacheco: changed this so it returns the "actual execution" trace len, before extending to the power of two (which is what the continuations code expects)

riscv-executor/src/lib.rs
pacheco: maps the "rom" `p_foo` fixed columns to their respective `foo` witness column

riscv-executor/src/lib.rs
pacheco: we're only ever executing inside `main` function in our "virtual" asm machine. If we were to call functions in some other "virtual machine" (non block/constrained machines), we'd get different operation id's here afaiu... but I will remove the comment as it's kind of obscure and unlikely to ever be an issue

riscv-executor/src/lib.rs
pacheco: yep, that's the selector assigned to the link doing this "reg_memory.write"

riscv-executor/src/lib.rs
pacheco: idk what to do about it. These are "STEP offset" and the selector assigned to the link. I think the step is not a problem, comes directly from the asm (STEP, STEP+1, etc). The selector OTH, you can figure out by looking at the optimized PIL

riscv-executor/src/lib.rs
pacheco: mmm no, i think it should do the modulo to match the asm, ill uncomment

riscv-executor/src/lib.rs
```diff
@@ -740,31 +1166,43 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
-                self.proc.set_reg_mem(write_reg, next_pc.into());
+                self.reg_write(0, write_reg, next_pc.into(), 3);
 
                 self.proc.set_pc(addr);
 
+                set_col!(tmp1_col, addr);
+
                 Vec::new()
             }
+            // TODO: update to witness generation for continuations
```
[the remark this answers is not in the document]
pacheco: same as above

riscv-executor/src/lib.rs
```diff
@@ -1308,16 +2077,100 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
+    pub register_memory: RegisterMemoryState<F>,
+}
+
+#[derive(Clone, Copy)]
+enum ExecMode {
+    Fast,
+    Trace,
+}
+
+/// Execute a Powdr/RISCV assembly program, without generating a witness.
```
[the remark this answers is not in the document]
pacheco: it does calculate the values (didn't want to make that part even more complicated), but doesn't really save values into columns or submachines

riscv-executor/src/lib.rs
```diff
@@ -1308,16 +2077,100 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
+        bootloader_inputs,
+        max_steps_to_execute.unwrap_or(usize::MAX),
+        ExecMode::Trace,
+        profiling,
+    )
+}
+
+/// FIXME: copied from `riscv/runtime.rs` instead of adding dependency.
+/// Helper function for register names used in submachine instruction params.
+fn register_by_idx(idx: usize) -> String {
```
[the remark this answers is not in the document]
pacheco: kind of yes, for the coprocessors that don't use memory yet

riscv-executor/src/lib.rs
```diff
@@ -1426,55 +2360,83 @@ pub fn execute_ast<F: FieldElement>(
...
+        e.proc.set_col("main::_operation_id", sink_id.into());
+
+        // loop
+        e.proc.set_col("main::pc_update", sink_id.into());
+        e.proc.set_pc(sink_id.into());
+        assert!(e.proc.advance().is_none());
+        e.proc.push_row();
+        e.set_program_columns(sink_id);
+        e.proc.set_col("main::pc_update", sink_id.into());
+        e.proc.set_col("main::_operation_id", sink_id.into());
```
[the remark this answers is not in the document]
pacheco: We're populating the final 3 rows of the trace, these are the values they take here. This last "loop" row is then going to be repeated until the next power of two

riscv-executor/src/memory.rs
```diff
@@ -0,0 +1,186 @@
...
+
+use crate::Elem;
+
+#[derive(Debug, Eq, PartialEq)]
+/// Order of fields matter: will be ordered by addr then step.
+struct Op<F: FieldElement> {
+    addr: u32,
+    step: u32,
+    value: Elem<F>,
+    write: u32,
```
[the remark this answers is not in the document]
pacheco: mmm don't remember why, i don't think there's a reason it can't be a bool, ill change it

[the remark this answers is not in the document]
pacheco: i think it was just convenient to convert to a FE by using `.into()`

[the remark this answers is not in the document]
pacheco: in exactly one line, so kinda pointless :P

riscv-executor/src/submachines.rs
```diff
@@ -0,0 +1,533 @@
+use super::decompose_lower32;
+use super::Elem;
+use powdr_number::{FieldElement, LargeInt};
+
+use std::collections::HashMap;
+
+/// Each submachine kind (i.e., binary, shift) must implement this trait
```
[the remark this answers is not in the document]
pacheco: cause it's calculated differently than the other submachines, its that "double sorted witness" logic, where we generate the witness only at the end, after we can order all operations

riscv-executor/src/submachines.rs
```diff
@@ -0,0 +1,533 @@
...
+
+    fn final_row_override(&mut self) {
+        self.trace.final_row_override()
+    }
+
+    fn push_dummy_block(&mut self) {
+        self.trace.push_dummy_block(M::BLOCK_SIZE, M::SELECTORS);
+    }
+
+    fn take_cols(&mut self) -> Vec<(String, Vec<Elem<F>>)> {
```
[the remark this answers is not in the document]
pacheco: that won't stop it from being called twice... rn it already clears the trace when called

riscv/tests/common/mod.rs
pacheco: idk if it makes sense to break tests for that rn, as we're not using it yet. It will only break if proving fails or if its partial and witgen can't finish it.

riscv-executor/src/submachines.rs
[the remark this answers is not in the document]
leonardoalt: ah right

riscv-executor/src/submachines.rs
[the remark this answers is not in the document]
leonardoalt: ok

riscv/tests/common/mod.rs
leonardoalt: Ok. At some point (next PRs I guess) it would be good to get quick feedback from CI on whether all columns are there. Maybe some specific tests.

riscv-executor/src/lib.rs
leonardoalt: ok

riscv-executor/src/lib.rs
leonardoalt: ah yea

riscv-executor/src/lib.rs
leonardoalt: yea the step is fine, the selector is more obscure. I wonder if just shuffling the args to put both at the end would help, but probably not worth it

riscv-executor/src/lib.rs
leonardoalt: I think the instructions are correct actually. Ah I guess maybe missing the columns?

riscv-executor/src/lib.rs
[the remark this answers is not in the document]
leonardoalt: ah I see

riscv-executor/src/lib.rs
```diff
@@ -379,13 +382,23 @@ mod builder {
         Submachine, SubmachineBoxed, PC_INITIAL_VAL,
     };
 
-    // TODO: where do we get this from? I don't think we want to depend on the linker crate just for this
-    const MIN_DEGREE: u32 = 1 << 5;
+    fn namespace_degree_range<F: FieldElement>(
```
leonardoalt, line 385: Wouldn't it be easier to read this from the PILGraph or some higher level data structure in Pipeline that could be passed here? Or is this safe enough?

riscv-executor/src/lib.rs
```diff
@@ -379,13 +382,23 @@ mod builder {
...
+        opt_pil: &Analyzed<F>,
+        namespace: &str,
+    ) -> DegreeRange {
+        opt_pil
+            .committed_polys_in_source_order()
+            .find(|(s, _)| s.absolute_name.contains(&format!("{namespace}::")))
+            .and_then(|(s, _)| s.degree)
+            .unwrap_or(DegreeRange {
+                min: 0,
+                max: u32::MAX as u64,
```
leonardoalt, line 395: we know that the range must be known for every single machine in the RISCV case, so I think it'd be safer to panic instead of using [0, MAX]

riscv-executor/src/lib.rs
```diff
@@ -1065,26 +1117,24 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
+                    Elem::Field(F::zero())
+                };
+                let addr = addr as u32;
 
+                self.proc.set_mem(addr, value.u());
                 self.proc
                     .memory_machine
                     .write(self.step + 3, addr, value, 1);
 
+                set_col!(wrap_bit, wrap_bit);
```
leonardoalt, line 1136: wrap_bit twice?

riscv-executor/src/memory.rs
```diff
@@ -8,20 +8,25 @@ struct Op<F: FieldElement> {
     addr: u32,
     step: u32,
     value: Elem<F>,
-    write: u32,
+    write: bool,
+    // each machine that's called via permutation has selector array, with one entry per incoming permutation.
```
leonardoalt, line 12:
```suggestion
    // each machine that's called via permutation has a selector array, with one entry per incoming permutation.
```

riscv-executor/src/lib.rs
pacheco: It should be the same in the end i think? Columns in a namespace should have the same degree no?

leonardoalt: yea I just meant the way to retrieve the numbers

riscv-executor/src/lib.rs
pacheco: one is the name of the column (used by the macro), the other is the local variable... it does look funny, so ill drop the local variable and build the value in place here

riscv-executor/src/lib.rs
pacheco: here we're already dealing in namespaces/columns and we already have the optimized PIL, seemed like the easiest way

leonardoalt: ok

riscv-executor/src/lib.rs
leonardoalt: ah I see

leonardoalt (approved):
lgtm

PR #1606, opened 2024-07-25
Make main instantiate all submachines and pass them as arguments

leonardoalt:


Schaeff (dismissed):

leonardoalt:
fixed tests

riscv/src/runtime.rs
```diff
@@ -356,11 +440,13 @@ impl Runtime {
         self
     }
 
+    #[allow(clippy::too_many_arguments)]
```
chriseth, line 443: hm....

chriseth (approved):
It's technically ok but really really ugly. From an architectural standpoint, we should never need to pass internals to machines like that.

Is there a plan to improve this somehow later on? Maybe by passing a generic object where we can request a certain machine type instance from?

leonardoalt:
I'm not sure I follow. Are you talking about the assembly part or the Rust part?

chriseth:
Both, the whole concept of passing a machine to another machine because we know that it needs it, but in the end, it should be that machine's concern what internal machines it needs.

leonardoalt:
> it should be that machine's concern what internal machines it needs.

It still is, as expressed in the machine's arguments. The point here is that this actually is an optimization that can be done in the language: before this PR, multiple machines that require the same submachine were instantiating them multiple times. With this PR only the top-level machine instantiates the shared ones and passes them around for the ones that need them. I'd say that's the opposite of ugly, and turned out quite nice.

leonardoalt:
The part that I *do* think is ugly is the way we create the RISCV machine powdr code in Rust with a bunch of string concatenations as opposite to files containing templates and rendering. If the whole thing was already expressed in terms of powdr-asm, it would be much clearer which machines are being instantiated by the top-level RISCV main machine and what's passed to what machine.

chriseth (approved):

Schaeff:
> Both, the whole concept of passing a machine to another machine because we know that it needs it, but in the end, it should be that machine's concern what internal machines it needs.

I have a plan for this. I think this is fine for now because we currently don't instantiate machines explicitly, so declaring and instantiating is the same. Eventually, I think we should have something like:

```
// declaration
machine Main {
   arith: Arith,
}

machine Arith {
}

// instantiation, where we pass instances
let arith = Arith {};
let main = Main { arith };
```

PR #1609, opened 2024-07-25
Set hint

chriseth:


pil-analyzer/src/type_builtins.rs
```diff
@@ -48,6 +48,7 @@ lazy_static! {
         ("std::field::modulus", ("", "-> int")),
         ("std::prelude::challenge", ("", "int, int -> expr")),
         ("std::prover::degree", ("", "-> int")),
+        ("std::prover::add_hint", ("", "expr, (int -> fe) ->")),
```
chriseth, line 51: Should we call it `add_hint` or `set_hint`? We probably won't support more than one hint...

pacheco: `set_hint` makes sense then

pil-analyzer/tests/side_effects.rs
```diff
@@ -100,3 +100,31 @@ fn fixed_with_constr_type() {
...
+
+#[test]
+fn set_hint() {
+    let input = r#"
+    namespace std::prover;
+        let set_hint = 8;
+        enum Query { Hint(fe), None, }
+    namespace N(16);
+        let x;
+        std::prover::set_hint(x, query |i| std::prover::Query::Hint(1));
```
pacheco, line 112: is `set_hint` just a different way of doing `col ... query` or does it do/allow different things?

pil-analyzer/tests/condenser.rs
```diff
@@ -243,3 +243,124 @@ fn new_fixed_column_as_closure() {
...
+        enum Query { Hint(fe), None, }
+    namespace N(16);
+        let x;
+        std::prover::set_hint(x, query |_, _| std::prover::Query::Hint(1));
+    "#;
+    analyze_string::<GoldilocksField>(input);
+}
+
+#[test]
+#[should_panic = "Array elements are not supported for std::prover::set_hint (called on N.x[0])."]
```
pacheco, line 295: will it be supported eventually?

pil-analyzer/tests/condenser.rs
```diff
@@ -243,3 +243,124 @@ fn new_fixed_column_as_closure() {
...
+        let set_hint = 8;
+        let eval = 8;
+        enum Query { Hint(fe), None, }
+    namespace N(16);
+        let x;
+        let y;
+        std::prover::set_hint(y, |i| std::prover::Query::Hint(std::prover::eval(x)));
+        {
+            let z;
+            std::prover::set_hint(z, query |_| std::prover::Query::Hint(1));
```
pacheco, line 260: can you `set_hint` a column from an outer scope? what about from a function? should those be tested?

pil-analyzer/tests/condenser.rs
```diff
@@ -243,3 +243,124 @@ fn new_fixed_column_as_closure() {
...
+#[should_panic = "Expected type: int -> std::prover::Query"]
+fn set_hint_invalid_function() {
+    let input = r#"
+    namespace std::prover;
+        let set_hint = 8;
+        let eval = 8;
+        enum Query { Hint(fe), None, }
+    namespace N(16);
+        let x;
+        std::prover::set_hint(x, query |_, _| std::prover::Query::Hint(1));
```
pacheco, line 289: should we test passing a "non-query" lambda?

pil-analyzer/src/condenser.rs
```diff
@@ -513,3 +586,48 @@ fn to_expr<T: Clone>(value: &Value<'_, T>) -> AlgebraicExpression<T> {
...
 }
+
+/// Turns a value of function type (i.e. a closure) into a FunctionValueDefinition.
+/// Does not allow captured variables.
+fn closure_to_function<T: Clone + Display>(
+    source: &SourceRef,
+    value: &Value<'_, T>,
+    expected_kind: FunctionKind,
+) -> Result<FunctionValueDefinition, EvalError> {
+    if let Value::Closure(evaluator::Closure {
```
pacheco, line 597: nitpick, `let-else` might read a bit better here

pil-analyzer/src/condenser.rs
```diff
@@ -164,10 +183,12 @@ pub struct Condenser<'a, T> {
...
-    /// The generated columns since the last extraction.
-    new_columns: Vec<(Symbol, Option<FunctionValueDefinition>)>,
-    /// The names of all new olumns ever generated, to avoid duplicates.
-    all_new_names: HashSet<String>,
+    /// The generated columns since the last extraction in creation order.
+    new_columns: Vec<Symbol>,
+    /// The hints and fixed column definitions added since the last extraction.
+    new_column_values: HashMap<String, FunctionValueDefinition>,
+    /// The names of all new columns ever generated, to avoid duplicates.
+    new_symbols: HashSet<String>,
```
pacheco, line 191: maybe `all_symbols`?

pil-analyzer/tests/side_effects.rs
chriseth: the `col ... query` syntax is only valid in pil at (outer) statement level. In the long run, we want to get rid of that and only have the syntax that is also available inside functions. So essentially, `set_hint` allows you to set the query function for a witness column from inside a function.

pil-analyzer/tests/condenser.rs
chriseth: function or block should no make a difference, but I can add some more tests where the column is "somewhere else"

pil-analyzer/tests/condenser.rs
chriseth: Only once we remove the old syntax, because you would have to set a hint for the whole array and we don't have syntax for that now.

pacheco (approved):

PR #1610, opened 2024-07-25
Plonky public values constrained by fixed columns

topanisto:
Addressing #1607 -- changing the implementation for constraining public values to reflect changes from #1543, as well as a slight refactoring of PowdrCircuit struct to include publics.

### DIFF
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -322,6 +322,30 @@ impl<T> Analyzed<T> {
             .filter_map(|(_poly, definition)| definition.as_mut())
             .for_each(|definition| definition.post_visit_expressions_mut(f))
     }
+
+    /// Retrieves (col_name, col_idx, offset) of each public witness in the trace.
+    pub fn get_publics(&self) -> Vec<(String, usize, usize)> {
+        let mut publics = self
+            .public_declarations
+            .values()
+            .map(|public_declaration| {
+                let witness_name = public_declaration.referenced_poly_name();
+                let witness_column = {
+                    let base = public_declaration.polynomial.poly_id.unwrap().id as usize;
+                    match public_declaration.array_index {
+                        Some(array_idx) => base + array_idx,
+                        None => base,
+                    }
+                };
+                let witness_offset = public_declaration.index as usize;
+                (witness_name, witness_column, witness_offset)
+            })
+            .collect::<Vec<_>>();
+
+        // Sort, so that the order is deterministic
+        publics.sort();
+        publics
+    }
 }
 
 impl<T: FieldElement> Analyzed<T> {
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -329,8 +329,8 @@ impl<T> Analyzed<T> {
             .public_declarations
             .values()
             .map(|public_declaration| {
-                let public_name = public_declaration.referenced_poly_name();
-                let public_col = {
+                let column_name = public_declaration.referenced_poly_name();
+                let column_idx = {
                     let base = public_declaration.polynomial.poly_id.unwrap().id as usize;
                     match public_declaration.array_index {
                         Some(array_idx) => base + array_idx,
@@ -338,7 +338,7 @@ impl<T> Analyzed<T> {
                     }
                 };
                 let row_offset = public_declaration.index as usize;
-                (public_name, public_col, row_offset)
+                (column_name, column_idx, row_offset)
             })
             .collect::<Vec<_>>();
 
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -329,16 +329,16 @@ impl<T> Analyzed<T> {
             .public_declarations
             .values()
             .map(|public_declaration| {
-                let witness_name = public_declaration.referenced_poly_name();
-                let witness_column = {
+                let public_name = public_declaration.referenced_poly_name();
+                let public_col = {
                     let base = public_declaration.polynomial.poly_id.unwrap().id as usize;
                     match public_declaration.array_index {
                         Some(array_idx) => base + array_idx,
                         None => base,
                     }
                 };
-                let witness_offset = public_declaration.index as usize;
-                (witness_name, witness_column, witness_offset)
+                let row_offset = public_declaration.index as usize;
+                (public_name, public_col, row_offset)
             })
             .collect::<Vec<_>>();
 
--- a/plonky3/src/circuit_builder.rs
+++ b/plonky3/src/circuit_builder.rs
@@ -28,6 +28,8 @@ pub(crate) struct PowdrCircuit<'a, T> {
     witness: Option<&'a [(String, Vec<T>)]>,
     /// Callback to augment the witness in the later stages
     _witgen_callback: Option<WitgenCallback<T>>,
+    /// Identity of all public cells
+    publics: Vec<(String, usize, usize)>,
     /// The matrix of preprocessed values, used in debug mode to check the constraints before proving
     #[cfg(debug_assertions)]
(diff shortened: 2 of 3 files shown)


plonky3/src/circuit_builder.rs
```diff
@@ -283,42 +255,24 @@ impl<'a, T: FieldElement, AB: AirBuilderWithPublicValues<F = Val> + PairBuilder>
...
-                builder.assert_eq(s, AB::Expr::one() - inv_decr * decr);
-                builder.assert_zero(s * decr); //constraining is_zero
+        let pi_deref = pi_moved.clone();
+        let fixed_local = fixed.row_slice(0);
+        let public_offset = self.analyzed.constant_count();
+
+        self.publics.iter().zip(pi_moved).enumerate().for_each(
+            |(index, ((_, col_id, _), public_value))| {
+                let selector = fixed_local[public_offset + index];
+                builder.assert_bool(selector);
```
Schaeff, line 271: If the selector is a fixed column we do not need to constrain it inside the circuit

plonky3/src/stark.rs
```diff
@@ -91,9 +91,18 @@ impl<T: FieldElement> Plonky3Prover<T> {
         // get fixed columns
         let fixed = &self.fixed;
 
-        if fixed.is_empty() {
-            return;
```
Schaeff, line 95: This still needs to happen in the case where we have no fixed columns and no public inputs. In that case we don't need keys.

plonky3/src/stark.rs
```diff
@@ -104,8 +113,22 @@ impl<T: FieldElement> Plonky3Prover<T> {
...
             self.analyzed.degree() as usize,
         );
-        // get the preprocessed matrix
-        let matrix = self.get_preprocessed_matrix();
+        // write fixed into matrix row by row
+        let matrix = RowMajorMatrix::new(
+            (0..self.analyzed.degree())
+                .flat_map(|i| {
+                    fixed
+                        .iter()
```
Schaeff, line 121: We can chain publics here to avoid repeating casting to gl

ast/src/analyzed/mod.rs
```diff
@@ -322,6 +322,30 @@ impl<T> Analyzed<T> {
...
             .for_each(|definition| definition.post_visit_expressions_mut(f))
     }
+
+    /// Retrieves (col_name, col_idx, offset) of each public witness in the trace.
+    pub fn get_publics(&self) -> Vec<(String, usize, usize)> {
+        let mut publics = self
+            .public_declarations
+            .values()
+            .map(|public_declaration| {
+                let witness_name = public_declaration.referenced_poly_name();
```
Schaeff, line 332: This does not have to be a witness, public inputs can reference any column (even if it doesn't make sense in practice to expose a cell of a fixed column)

plonky3/src/circuit_builder.rs
```diff
@@ -46,6 +28,8 @@ pub(crate) struct PowdrCircuit<'a, T> {
     witness: Option<&'a [(String, Vec<T>)]>,
     /// Callback to augment the witness in the later stages
     _witgen_callback: Option<WitgenCallback<T>>,
+    /// Identity of all public cells
```
Schaeff, line 31: Do we need this field? Can't it be computed from the other fields and therefore be a method?

plonky3/src/circuit_builder.rs
```diff
@@ -225,13 +172,23 @@ impl<'a, T: FieldElement> PowdrCircuit<'a, T> {
                     }
                 }
             }
-            AlgebraicExpression::PublicReference(_) => unimplemented!(
-                "public references are not supported inside algebraic expressions in plonky3"
-            ),
+            AlgebraicExpression::PublicReference(id) => {
+                assert!(
```
Schaeff, line 176: This could be more concise by using `Iter::find`

plonky3/src/circuit_builder.rs
```diff
@@ -257,13 +215,27 @@ impl<'a, T: FieldElement> PowdrCircuit<'a, T> {
     }
 }
 
+/// An extension of [Air] allowing access to the number of fixed columns
```
Schaeff, line 218: I think we already have this in BaseAir in our fork

Schaeff: I think it's called `preprocessed_width`

plonky3/src/circuit_builder.rs
```diff
@@ -257,13 +215,27 @@ impl<'a, T: FieldElement> PowdrCircuit<'a, T> {
...
+
 impl<'a, T: FieldElement> BaseAir<Val> for PowdrCircuit<'a, T> {
     fn width(&self) -> usize {
-        self.analyzed.commitment_count() + 3 * self.analyzed.publics_count()
+        self.analyzed.commitment_count()
     }
 
     fn preprocessed_width(&self) -> usize {
-        self.analyzed.constant_count()
+        self.analyzed.constant_count() + self.analyzed.publics_count()
```
Schaeff, line 238: Oh, here!

plonky3/src/circuit_builder.rs
topanisto: Oops sorry this was leftover from the old fork

plonky3/src/circuit_builder.rs
topanisto: Clippy recommended to use `any` is that equivalent?

Schaeff: Here you assert any, then find the position, then return the element at that position. The three can be done with a single call to find.

plonky3/src/circuit_builder.rs
```diff
@@ -193,11 +135,12 @@ impl<'a, T: FieldElement> PowdrCircuit<'a, T> {
     }
 
     /// Conversion to plonky3 expression
-    fn to_plonky3_expr<AB: AirBuilder<F = Val>>(
+    fn to_plonky3_expr<AB: AirBuilder<F = Val> + AirBuilderWithPublicValues>(
         &self,
         e: &AlgebraicExpression<T>,
         main: &AB::M,
         fixed: &AB::M,
+        publics: &Vec<<AB as AirBuilderWithPublicValues>::PublicVar>,
```
Schaeff, line 143: It would be clearer for this to be `&HashMap<String, <AB as AirBuilderWithPublicValues>::PublicVar>`. Basically a map of the values of the public inputs, indexed by name.

plonky3/src/circuit_builder.rs
```diff
@@ -225,13 +168,21 @@ impl<'a, T: FieldElement> PowdrCircuit<'a, T> {
                     }
                 }
             }
-            AlgebraicExpression::PublicReference(_) => unimplemented!(
-                "public references are not supported inside algebraic expressions in plonky3"
-            ),
+            AlgebraicExpression::PublicReference(id) => {
```
Schaeff, line 171: Then here we simply return `publics[&id]`

plonky3/src/circuit_builder.rs
```diff
@@ -283,42 +237,24 @@ impl<'a, T: FieldElement, AB: AirBuilderWithPublicValues<F = Val> + PairBuilder>
...
         assert_eq!(publics.len(), pi.len());
 
+        let local = main.row_slice(0);
+
         // public constraints
         let pi_moved = pi.to_vec();
-        let (local, next) = (main.row_slice(0), main.row_slice(1));
-
-        let public_offset = self.analyzed.commitment_count();
+        let pi_deref = pi_moved.clone();
```
Schaeff, line 247: And we can create the hashmap a single time here.

ast/src/analyzed/mod.rs
Schaeff: probably `column` makes more sense in this block

Schaeff (approved):

PR #1613, opened 2024-07-25
Even more quick tests

chriseth:


test_data/std/binary_test.asm
```diff
@@ -1,7 +1,7 @@
 use std::machines::binary::Binary;
 
 
-machine Main with degree: 262144 {
+machine Main with degree: 196608 {
```
leonardoalt, line 4: I'm not sure this makes a difference?

leonardoalt (dismissed):

test_data/std/binary_test.asm
chriseth: ah wait, does this even work if it's not a power of two?

chriseth: looks like it does, even in the backends?

leonardoalt: It gets padded to the next power of 2

leonardoalt (approved):

test_data/std/binary_test.asm
chriseth: Then it's a nice test...

PR #1614, opened 2024-07-25
Fix clippy with disabled halo2

chriseth:


chriseth:
That this has not been detected shows that our setup is broken. We always compile everything with halo2.

As far as I understand features, they are always "add-ons", which means when a crate is requested twice, once without the feature and once with the feature, the request with the feature wins.

Since our "powdr" crate requests the pipeline with halo2, it is always activated in all tests and all binaries.

leonardoalt (approved):

PR #1617, opened 2024-07-25
Create separate runner for examples.

chriseth:


leonardoalt (dismissed):
Dont they need to also be removed from somewhere then?

chriseth:
It is removed, the diff is a bit weird.

leonardoalt (approved):

PR #1622, opened 2024-07-26
Mark slow tests

chriseth:
This marks more tests as ignored because they are slow.

It also adds a list of "nightly tests", which are ignored tests that are explicitly excluded in the regular PR runs (but executed in the nightly run that runs all tests).


.github/workflows/pr-tests.yml
```diff
@@ -202,13 +209,10 @@ jobs:
     - uses: taiki-e/install-action@nextest
     - name: Run slow tests
       # Number threads is set to 1 because the runner does not have enough memory for more.
+      # TODO is that true?
+      # TODO exclue the nightly tests
```
leonardoalt, line 213: wen

leonardoalt, line 213:
```suggestion
      # TODO exclude the nightly tests
```

.github/workflows/pr-tests.yml
```diff
@@ -202,13 +209,10 @@ jobs:
...
       run: |
-        if [[ "${{ matrix.test }}" == "subset1" ]]; then
-          TESTS="test(=arith_test)"
-        elif [[ "${{ matrix.test }}" == "subset2" ]]; then
-          TESTS="test(=keccak) | test(=vec_median) | test(=instruction_tests::addi) | test(=many_chunks) | test(=sum_serde)"
-        fi
-        cargo nextest run --archive-file tests.tar.zst --verbose --run-ignored=ignored-only --no-capture -E "$TESTS"
+        cargo nextest run --archive-file tests.tar.zst --verbose --run-ignored=ignored-only --no-capture --partition count:"${{ matrix.test }}"/8
```
leonardoalt, line 215: this doesn't look equivalent, what's the new behavior?

chriseth:
It would be nice if we could just have two levels of "ignore": One "slow" and one "nightly".

chriseth:
It's weird that nextest ran 5 tests in the first bucket but only 1 test in the last buckets. As per the documentation, the counting is performed after the filter is applied, but maybe "ignored" is not such a filter?

Let's try if the division is more even with hashed partitions.

.github/workflows/pr-tests.yml
```diff
@@ -203,12 +210,8 @@ jobs:
     - name: Run slow tests
       # Number threads is set to 1 because the runner does not have enough memory for more.
```
chriseth, line 211: Where does it actually limit the number of tests? We should try if it works with two threads.

chriseth:
Nice! The slow tests all passed within 13 minutes, using two parallel tests per machine.

I think this is a much cleaner solution than we previously had because we can just mark tests as "slow" if they are too slow for the regular (local!) run and then they will be placed in the matrix run.

I will change the quick tests again to not have that much parallelism.

leonardoalt (approved):

.github/workflows/pr-tests.yml
pacheco: if i remember correctly, in nextest, enabling `no-capture` sets running threads to 1

chriseth: Yes, that's the answer.

PR #1624, opened 2024-07-26
Implement LogUp via Bus

onurinanc:
LogUp implementation using Bus

As @georgwiese suggested in https://github.com/powdr-labs/powdr/issues/1573#issuecomment-2250875201

- That makes the machine detection simpler, because we wouldn't have constraints that depend on both the LHS and RHS of a lookup.
- It's what we'll actually need for VADCOP, so maybe we don't even have to support std::protocols::lookup as a first step.


georgwiese (commented):
LGTM, but I think to make the CI pass this branch needs to be merged with `main` and and the `bus_lookup_via_challenges` needs to be added to [this blacklist](https://github.com/powdr-labs/powdr/blob/d63525fb1611da80613621cd053764dadb1f426f/pipeline/tests/powdr_std.rs#L379) to fix the failing test.

georgwiese (approved):
Nice!

PR #1625, opened 2024-07-26
Traits unification

gzanitti:


pil-analyzer/src/evaluator.rs
```diff
@@ -452,6 +465,13 @@ impl<'a> Definitions<'a> {
                         Value::TypeConstructor(&variant.name).into()
                     }
                 }
+                Some(FunctionValueDefinition::TraitFunction(trait_decl, function)) => {
+                    let impls = implementations.get(&trait_decl.name).ok_or_else(|| {
```
chriseth, line 469: I think it should be possible to resolve the exact implementation at type checking time and not during evaluation. Once we change internal operators to be traits, it would be way too slow to do this for every single addition.

My plan was to store this association somewhere in the AST (next to the generic type args), and just don't print it.

One difficulty is that the exact type might also depend on the values for type variables, so we might have to store a map.

chriseth:
I spent some more thoughts on this: Essentially we need to do what a compiler would do: After type checking is done, we "walk" the call tree starting from the outer statements (we ignore function definitions). Whenever we see a function reference (not call), we follow that reference. At some point, we will see references to generic functions. Since we start at the outermost code, the reference needs fully specific type arguments. We follow that reference combined with the type arguments. So our stack for the depth-first search does not use references but references plus a list of types. Whenever we enter a generic function, we have a list of type arguments, so all type variables can be fully substituted during this search.

And that is the key point: When we encounter a trait function, we have fully specific types. This means we can collect all references to trait functions combined with the specific types. Then we just need to resolve the impls once and store them at the references in the ast as a mapping from list of types to impl function.

gzanitti:
> I spent some more thoughts on this: Essentially we need to do what a compiler would do: After type checking is done, we "walk" the call tree starting from the outer statements (we ignore function definitions). Whenever we see a function reference (not call), we follow that reference. At some point, we will see references to generic functions. Since we start at the outermost code, the reference needs fully specific type arguments. We follow that reference combined with the type arguments. So our stack for the depth-first search does not use references but references plus a list of types. Whenever we enter a generic function, we have a list of type arguments, so all type variables can be fully substituted during this search.
> 
> And that is the key point: When we encounter a trait function, we have fully specific types. This means we can collect all references to trait functions combined with the specific types. Then we just need to resolve the impls once and store them at the references in the ast as a mapping from list of types to impl function.

This makes a lot of sense. However, I was reviewing the code and I have questions about how to walk the `call tree starting from the outer statements` after the type check. At this point the files have been processed and the closest thing to the code we have is `source_order`, but it is not what we are looking for. What am I missing?

pil-analyzer/src/type_unifier.rs
```diff
@@ -10,9 +13,19 @@ pub struct Unifier {
     type_var_bounds: HashMap<String, HashSet<String>>,
     /// Substitutions for type variables
     substitutions: HashMap<String, Type>,
+    /// Last used type variable index.
+    last_type_var: usize,
 }
 
 impl Unifier {
+    pub fn new() -> Self {
```
chriseth, line 21: This could also be Default::default

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,166 @@
...
+            .filter(|(_, (_, def))| {
+                def.is_some()
+                    && !matches!(
+                        def.as_ref().unwrap(),
+                        FunctionValueDefinition::TraitFunction(_, _)
+                    )
+            })
+            .map(|(name, _)| name.clone())
+            .collect::<Vec<_>>();
+        for name in keys {
```
chriseth, line 47: I think we should also filter out every definition that has type variables

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    pub fn traits_resolution(&mut self) {
+        let keys = self
+            .definitions
+            .iter()
+            .filter(|(_, (_, def))| {
+                def.is_some()
+                    && !matches!(
+                        def.as_ref().unwrap(),
+                        FunctionValueDefinition::TraitFunction(_, _)
```
chriseth, line 42:
```suggestion
                        def,
                        Some(FunctionValueDefinition::TraitFunction(_, _))
```

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.stack.push(current.to_string());
+
+        let next_name = {
+            let current_def = &self.definitions.get(current).unwrap().1;
+            match current_def {
+                Some(FunctionValueDefinition::Expression(TypedExpression {
+                    e: Expression::FunctionCall(_, FunctionCall { function, .. }),
+                    ..
+                })) => match function.as_ref() {
+                    Expression::Reference(
```
chriseth, line 65: You can use the following here:
```suggestion
                    e,
                    ..
                })) => e.all_children().filter_map(|e| match e { Expression::Reference(_, Reference::Poly(PolynomialReference{name, type_args: Some(types)})) => Some((name.clone(), types.clone()), _=> None })
```

chriseth: We can also take a look at this together at some point. If this is too tricky, you can also just do fully resolved trait function calls first and we do this one as another PR.

ast/src/analyzed/mod.rs
```diff
@@ -1325,6 +1325,8 @@ pub struct PolynomialReference {
     /// The type arguments if the symbol is generic.
     /// Guaranteed to be Some(_) after type checking is completed.
     pub type_args: Option<Vec<Type>>,
+    ///
+    pub resolved_impl: Option<Box<Expression>>,
```
chriseth, line 1329: If this is an expression, we need to clone it, but we also don't want to have lifetime parameters in this struct. Another option would be to store an index into the implementation vector.

pil-analyzer/src/traits_processor.rs
gzanitti: > You can use the following here:

But, with this change, wouldn't we be losing the "reference path" to the trait function?

My original idea was to jump from reference to reference waiting for the last reference to jump to a TraitFunction (otherwise rollback and continue searching in other branches) and there have all the necessary type_args.

With this change we would be keeping all the references related to the initial expression, those that go to a TraitFunction and those that don't, in pre-order, which I understand would make more difficult the idea of "finding the reference path".

But maybe I misunderstood the original idea?

pil-analyzer/src/traits_processor.rs
gzanitti: Shouldn't it be the other way around? Shouldn't it filter out all those that don't have type vars?

gzanitti:
Hey @chriseth , I've been thinking and reviewing some examples. Isn't it a problem that we try to match implementations after type checking?

For example, this code:
```
trait Add<T> {
    add: T, T -> T,
}
impl Add<int> {
    add: |a, b| a + b,
}

let r = Add::add(3, 4);
```

will fail type checking because it's unable to derive `r`'s type. It works if we change the last line to `let r: int = ...`, but I understand that this is something we should be able to solve internally.

gzanitti:
> Hey @chriseth , I've been thinking and reviewing some examples. Isn't it a problem that we try to match implementations after type checking?
> 
> For example, this code:
> 
> ```
> trait Add<T> {
>     add: T, T -> T,
> }
> impl Add<int> {
>     add: |a, b| a + b,
> }
> 
> let r = Add::add(3, 4);
> ```
> 
> will fail type checking because it's unable to derive `r`'s type. It works if we change the last line to `let r: int = ...`, but I understand that this is something we should be able to solve internally.

Adding a non-related question: I was finishing with the identities and at the time of making a test I realized that I am not sure how we plan to use traits in these cases.

Could you give me a brief example?
Thanks!

chriseth:
This specific case is fine, because the trait function has the signature `T, T -> T`, so we know that `3`, `4` and `r` need to be of the same type. I think it's OK to require specifying at least one of the types, and probably `add(3, 4)` is not a very good example. More realistic is something like:

```
let w; // witness column
w * (w - 1) = 0; // constrain it to 0 or 1
```

In the future, the operators will be trait functions, so it will look like this:
```
Equals::equals(Mul::mul(w, Sub::sub(w, 1)), 0);
```

Both `mul` and `sub` have the signature `T, T -> T` and `equals` is `expr, expr -> Constr` (I don't think it makes sense to make it a trait).
We know that `w` is an `expr` (reference to a witness column has type `expr`). So the only sub-expressions that do not have a type yet are `1` and `0` which both must be `expr`.

Now the more complicated case is where we have something like `Into::into: A -> B`, where there is no relation between `A` and `B`. I think in this case, we really need both types to be known, and I think it's the same in Rust.

pil-analyzer/src/evaluator.rs
```diff
@@ -138,6 +138,7 @@ pub enum Value<'a, T> {
     Enum(&'a str, Option<Vec<Arc<Self>>>),
     BuiltinFunction(BuiltinFunction),
     Expression(AlgebraicExpression<T>),
+    TraitFunction(Closure<'a, T>),
```
chriseth, line 141: I don't see a difference to Closure in your code - is this variant needed?

pil-analyzer/src/evaluator.rs
```diff
@@ -821,7 +827,42 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
                         }
                         ta
                     });
-                    self.symbols.lookup(&poly.name, &type_args)?
+
+                    let impl_pos = type_args.as_ref().and_then(|type_args| {
+                        let key = type_args.iter().format(",").to_string();
```
chriseth, line 832: Formatting types to string is very expensive. Can you use `Vec<Type>` as a key instead (might have to use BTreeMap, which is fine)?

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,254 @@
...
+            let identity = &mut self.identities[index];
+            TraitsProcessor::update_references_in_identity(
+                identity,
+                &trait_name,
+                &resolved_impl_pos,
+            );
+        }
+    }
+
+    fn collect_refs_in_identity(
```
chriseth, line 73: We do something similar for the type checker in pil_analyzer.rs - maybe we can share the routine and change the interface of the trait processor to just get a vector of expressions to process?

pil-analyzer/tests/types.rs
```diff
@@ -614,3 +666,124 @@ fn new_fixed_column_wrong_type() {
     "#;
     type_check(input, &[]);
 }
+
+#[test]
+fn impl_type_resolution() {
```
chriseth, line 671: Does this test add anything that `defined_trait` does not test?

pil-analyzer/tests/types.rs
```diff
@@ -561,6 +561,58 @@ fn empty_conditional() {
...
+fn defined_trait_generic() {
+    let input = "
+    namespace std::convert(4);
+        let fe = || fe();
+    namespace F(4);
+        trait Add<T, Q> {
+            add: T, T -> Q,
+        }
+        impl<T> Add<T, fe> {
+            add: |a, b| std::convert::fe(a + b),
```
chriseth, line 588: This test is interesting because it should actually fail - but it is not related to trait unification / impl resolution:

You cannot call `std::convert::fe` on any type `T`. It has type constraints and these need to be matched to the ones in the `impl` here.

ast/src/analyzed/mod.rs
```diff
@@ -43,6 +44,7 @@ pub struct Analyzed<T> {
     pub source_order: Vec<StatementIdentifier>,
     /// Symbols from the core that were added automatically but will not be printed.
     pub auto_added_symbols: HashSet<String>,
+    pub implementations: HashMap<String, Vec<TraitImplementation<Expression>>>,
```
chriseth, line 47: Please use `trait_implementations` or `trait_impls`, I think just 'implementations' is too generic. Also please document.

chriseth: Oh and maybe move it up below definitions?

pil-analyzer/src/evaluator.rs
```diff
@@ -837,20 +858,33 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
                                 })
                                 .unwrap();
 
-                            // let Expression::LambdaExpression(_, body) = &expr.as_ref() else {
```
chriseth: I think it would be nice to move all these things into the trait (i.e. just into the lookup function, then we don't need a separate function for trait impls). Maybe the only thing the evaulator should do is substitute the type args. Please check if it is better to change the signature of the lookup function to take a poly reference instead of just the name.

pil-analyzer/src/evaluator.rs
```diff
@@ -821,7 +836,57 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
...
+                                            let key = ty.to_string();
+                                            self.type_args
+                                                .get(&key)
+                                                .map(|value| (key, value.clone()))
+                                        })
+                                        .collect::<HashMap<_, _>>()
+                                })
+                                .unwrap();
+
+                            let poly_name = poly.name.clone().replace('.', "::");
```
chriseth, line 861: This is needed because traits are formatted with `.` since they (usually) just have two components while the trait function uses `::` with its three components, right?

Since we do not maintain compatibility with polygon-pil any more, it's probably time to remove the `.` in the Display function.

Can you do this in a separate (independent) PR?

pil-analyzer/src/evaluator.rs
```diff
@@ -821,7 +836,57 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
...
+                                        unreachable!()
+                                    };
+                                    let closure = Closure {
+                                        lambda,
+                                        environment: vec![],
+                                        type_args: local_type_args,
+                                    };
+                                    Value::Closure(closure).into()
+                                }
+                                None => Err(EvalError::SymbolNotFound(format!(
```
chriseth, line 883: Can this be a panic?

pil-analyzer/src/evaluator.rs
```diff
@@ -821,7 +836,57 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
                         }
                         ta
                     });
-                    self.symbols.lookup(&poly.name, &type_args)?
+
+                    let impl_pos = type_args
```
chriseth, line 840: Can this be None at this stage?

pil-analyzer/src/evaluator.rs
```diff
@@ -880,9 +880,9 @@ impl<'a, 'b, T: FieldElement, S: SymbolLookup<'a, T>> Evaluator<'a, 'b, T, S> {
                                     };
                                     Value::Closure(closure).into()
                                 }
-                                None => Err(EvalError::SymbolNotFound(format!(
-                                    "Function {fn_name} not found in trait {trait_name}"
-                                )))?,
+                                None => {
+                                    panic!("Function {fn_name} not found in trait {trait_name}")
```
chriseth, line 884: I think an 'unwrap' is fine here.

pil-analyzer/src/evaluator.rs
gzanitti: Here: https://github.com/powdr-labs/powdr/pull/1694

ast/src/analyzed/mod.rs
```diff
@@ -1271,6 +1271,17 @@ pub struct PolynomialReference {
     pub resolved_impls: BTreeMap<Vec<Type>, usize>,
```
chriseth, line 1271: Can you add documentation in the line above?

pil-analyzer/src/evaluator.rs
```diff
@@ -470,41 +471,54 @@ impl<'a> Definitions<'a> {
 impl<'a, T: FieldElement> SymbolLookup<'a, T> for Definitions<'a> {
     fn lookup(
         &mut self,
-        name: &str,
+        poly: &PolynomialReference,
         type_args: &Option<Vec<Type>>,
     ) -> Result<Arc<Value<'a, T>>, EvalError> {
-        Self::lookup_with_symbols(self.0, name, type_args, self)
+        let impl_pos = type_args
```
chriseth, line 477: Don't you have to modify lookup_with_symbols?

pil-analyzer/src/evaluator.rs
```diff
@@ -470,41 +471,54 @@ impl<'a> Definitions<'a> {
...
+            .as_ref()
+            .and_then(|type_args| poly.resolved_impls.get(type_args).as_ref().copied());
+
+        match impl_pos {
+            Some(index) => {
+                let parts = poly.name.split("::").collect::<Vec<_>>();
```
chriseth, line 483: If a search for `::` is still needed after #1694 then please store the trait name inside the poly reference. At least I think it should be faster that way.

pil-analyzer/src/evaluator.rs
```diff
@@ -470,41 +471,54 @@ impl<'a> Definitions<'a> {
...
+                let fn_name = parts.last().unwrap().to_string();
+                let trait_name = parts[..parts.len() - 1].join("."); // TODO GZ Replace '.' by '::' after we merge #1694
+
+                let impls = match self.1.get(&trait_name) {
+                    Some(impls) => impls,
+                    None => Err(EvalError::SymbolNotFound(format!(
```
chriseth, line 489: I think you can use `self.1.get().unwrap()`, since the type checker (or some earlier component) guarantees that the trait is available, or does it not?

pil-analyzer/src/evaluator.rs
```diff
@@ -1355,10 +1321,10 @@ mod test {
         .to_string()
     }
 
-    pub fn evaluate_function<T: FieldElement>(input: &str, function: &str) -> T {
+    pub fn evaluate_function<T: FieldElement>(input: &str, poly_fn: &PolynomialReference) -> T {
```
chriseth, line 1324: I think this can keep its signature (it's a helper function) or do you need the functionality to call type-arg-qualified names?

ast/src/analyzed/mod.rs
```diff
@@ -1267,8 +1267,10 @@ pub struct PolynomialReference {
     /// The type arguments if the symbol is generic.
     /// Guaranteed to be Some(_) after type checking is completed.
     pub type_args: Option<Vec<Type>>,
-    ///
+    /// Keep track of valid implementations for this reference, indexed by their type arguments.
```
chriseth, line 1270:
```suggestion
    /// Store implementations for this reference (if it is a trait function), for each type it has been referenced somewhere in the code.
```

ast/src/analyzed/mod.rs
```diff
@@ -1267,8 +1267,10 @@ pub struct PolynomialReference {
...
     pub resolved_impls: BTreeMap<Vec<Type>, usize>,
+    /// If resolved implementation is matched with this reference, this is the name of the implemented trait.
```
chriseth, line 1272: Maybe we should combine this with the one above in a struct? It is only relevant for trait function references and we could abstract that away.

pil-analyzer/src/evaluator.rs
```diff
@@ -445,24 +452,48 @@ impl<'a> Definitions<'a> {
...
-                })) => {
-                    let type_args = type_arg_mapping(type_scheme, type_args);
-                    evaluate_generic(value, &type_args, symbols)?
+            let impl_pos = type_args
+                .as_ref()
+                .and_then(|type_args| poly.resolved_impls.get(type_args).as_ref().copied());
+
+            match impl_pos {
+                Some(index) => {
+                    let fn_name = poly.name.rsplit("::").next().unwrap().to_string();
```
chriseth, line 461: Looks like we also need to store the function name.

chriseth: At best we could store a reference to the function directly (i.e. no impl index, no function name), do you think this can be done? We probably need a lifetime parameter and that could be quite a big change we don't want to do.

pil-analyzer/src/evaluator.rs
```diff
@@ -445,24 +452,48 @@ impl<'a> Definitions<'a> {
...
+
+                    let trait_name = poly.trait_name.as_ref().unwrap();
+                    let impls = trait_impls.get(trait_name).unwrap();
+                    let impl_ = &impls[*index];
+
+                    let func = impl_.function_by_name(&fn_name).unwrap();
```
chriseth, line 467: Why not
```suggestion
                    let func = trait_impls.get(trait_name).unwrap()[*index].function_by_name(&fn_name).unwrap();
```
?

pil-analyzer/src/evaluator.rs
chriseth: Oh and it probably doesn't even work to assign that reference since we need a mut ref to the ast to do that but we store a ref to one of its children inside it. :(

chriseth: Another solution could be to sort the functions in the same way as we have them in the trait (and then only store an index and not a function name). Then the re-printing might swap them around but maybe that's ok?

pil-analyzer/src/traits_processor.rs
```diff
@@ -30,13 +30,18 @@ impl<'a> TraitsProcessor<'a> {
                 continue;
             };
 
-            let Some((trait_name, resolved_impl_pos)) =
+            let Some((full_name, resolved_impl_pos)) =
                 self.resolve_trait_function((name.to_string(), types), &traits_functions_defs)
```
chriseth, line 34: Why don't you use a &str here for the name?

pil-analyzer/src/statement_processor.rs
```diff
@@ -642,4 +650,62 @@ where
...
+                    .split('.')
+                    .map(|p| {
+                        if p == "super" {
+                            Part::Super
+                        } else {
+                            Part::Named(p.to_string())
+                        }
+                    })
+                    .collect::<Vec<_>>(),
+            ),
```
chriseth, line 700:
```suggestion
            name: SymbolPath::from_str(resolved_name),
```

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -317,9 +322,54 @@ impl PILAnalyzer {
         }
     }
 
+    pub fn traits_resolution(&mut self) {
```
chriseth, line 325: Why is it public?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -317,9 +322,54 @@ impl PILAnalyzer {
...
+                _ => expr.children_mut().flat_map(collect_references).collect(),
+            }
+        }
+
+        let traits_functions_defs: HashMap<String, FunctionValueDefinition> = self
+            .definitions
+            .iter()
+            .filter_map(|(name, (_, def))| match def {
+                Some(FunctionValueDefinition::TraitFunction(_, _)) => {
+                    Some((name.clone(), def.clone().unwrap()))
```
chriseth, line 338: Please don't clone, especially not the def

chriseth: Actually why do you need this hashmap any way? Can't you just use `self.definitions`?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -317,9 +322,54 @@ impl PILAnalyzer {
         }
     }
 
+    pub fn traits_resolution(&mut self) {
+        fn collect_references(expr: &mut Expression) -> Vec<&mut Reference> {
+            match expr {
+                Expression::Reference(_, ref mut r @ Reference::Poly(_)) => vec![r],
+                _ => expr.children_mut().flat_map(collect_references).collect(),
```
chriseth, line 329: You can introduce `all_children_mut` and avoid the recursive call.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -317,9 +322,54 @@ impl PILAnalyzer {
...
+                identity
+                    .left
+                    .selector
+                    .iter_mut()
+                    .chain(once(identity.left.expressions.as_mut()))
+                    .chain(identity.right.selector.iter_mut())
+                    .chain(once(identity.right.expressions.as_mut()))
+                    .flat_map(collect_references)
+            }))
+            .collect();
```
chriseth, line 363: Does this need to be a vector? Can't we pass the iterator?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -317,9 +322,54 @@ impl PILAnalyzer {
...
+
+        TraitsProcessor::new(&self.implementations)
```
chriseth, line 365: If we don't do anything with the TraitsProcessor object, maybe it would be better to define a public function in the traits_processor.rs and call that here without creating an object (and making the TraitsProcessor private)

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -317,9 +322,54 @@ impl PILAnalyzer {
...
+        let mut references: Vec<&mut Reference> = self
+            .definitions
+            .iter_mut()
+            .flat_map(|(_, (_, def))| match def {
+                Some(FunctionValueDefinition::Expression(TypedExpression { e: expr, .. })) => {
+                    collect_references(expr)
+                }
+                _ => Vec::new(),
+            })
+            .chain(self.identities.iter_mut().flat_map(|identity| {
```
chriseth, line 353: children_mut would also help here

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,103 @@
+use core::panic;
```
chriseth, line 1: I think we should not call this `_processor`, because statement processor and expression processor both translate from parsed expressions to analyzed expressions but this one does not.

pil-analyzer/src/pil_analyzer.rs
chriseth: Also if you create the iterator here, why not call trait resolution only on the element?

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+fn update_reference(
+    ref_name: &str,
+    c: &mut Reference,
+    resolved_impl_pos: &BTreeMap<Vec<Type>, usize>,
+    trait_name: String,
+) {
+    if let Reference::Poly(PolynomialReference {
```
chriseth, line 86: Why don't you pass a PolynomialReference to this function already?

pil-analyzer/src/traits_processor.rs
```diff
@@ -0,0 +1,103 @@
...
+        if let Some(impls) = self.trait_impls.get(&trait_decl.name) {
+            for (index, impl_) in impls.iter().enumerate() {
+                let Some(_) = impl_.function_by_name(&trait_fn.name) else {
+                    panic!(
+                        "Could not find function {} for {}",
+                        trait_fn.name, trait_decl.name
+                    );
+                };
+
+                resolved_impl_pos.insert(collected_ref.1.clone(), index);
```
chriseth, line 69: I don't understand how this works. Don't you have to check if colelcted_ref.1 unifies with impl_.type_scheme?

ast/src/analyzed/visitor.rs
```diff
@@ -40,4 +40,13 @@ impl<T> AllChildren<AlgebraicExpression<T>> for AlgebraicExpression<T> {
     fn all_children(&self) -> Box<dyn Iterator<Item = &AlgebraicExpression<T>> + '_> {
         Box::new(iter::once(self).chain(self.children().flat_map(|e| e.all_children())))
     }
+
+    fn all_children_mut(&mut self) -> Box<dyn Iterator<Item = &mut AlgebraicExpression<T>> + '_> {
```
chriseth, line 44: This PR is already huge, please try to factor out this change (and mabye other more or less unrelated ones)

ast/src/analyzed/visitor.rs
```diff
@@ -40,4 +40,13 @@ impl<T> AllChildren<AlgebraicExpression<T>> for AlgebraicExpression<T> {
...
+        let self_iter = iter::once(self as *mut _);
+        let children_iter = self
+            .children_mut()
+            .flat_map(|child| child.all_children_mut());
+        let self_iter = self_iter.map(|ptr| unsafe { &mut *ptr });
```
chriseth, line 49: It doesn't work without unsafe?

chriseth: If it doesn't work, there is also pre_visit_expressions_mut

chriseth:
I'm not sure if this is a problem, but I could imagine it is difficult to both hold a reference to definitions to actually do the trait resolution and then also a mutable reference to definitions to update the references.

What about splitting this up into a two-step process: First iterate through all references collecting trait references plus type, then resolve all the references (this also has the benefit that you don't need to do this multiple times if you have the same reference to a trait multiple times). And then with just a mutable ref to the definitions / references, store all the information you found.

chriseth:
Thinking about it, this might also just be a different way to do it: Store a map of trait plus type to impl just as a separate structure and don't store it in the AST. This has the benefit that we don't store stuff in the AST that we don't print, it's only relevant for the evaluator and we don't store the resolved impls multiple times if they are used multiple times.
What do you think?

gzanitti:
> Thinking about it, this might also just be a different way to do it: Store a map of trait plus type to impl just as a separate structure and don't store it in the AST. This has the benefit that we don't store stuff in the AST that we don't print, it's only relevant for the evaluator and we don't store the resolved impls multiple times if they are used multiple times. What do you think? Oh and doesn't that also allow us to store references to impl functions?

I ended up going for this option. I think now the code in general is clearer and cleaner. Also, I reverted all the changes in the lookups to go back to using `&str` instead of `PolynomialReference`, because it is no longer necessary. That should simplify the PR a lot.

ast/src/analyzed/mod.rs
```diff
@@ -1268,6 +1269,16 @@ pub struct PolynomialReference {
     pub type_args: Option<Vec<Type>>,
 }
 
+impl PolynomialReference {
```
chriseth, line 1272: I think this is no longer needed.

ast/src/parsed/mod.rs
```diff
@@ -336,10 +336,20 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct TraitImplementation<Expr> {
     pub name: SymbolPath,
+    pub source_ref: SourceRef,
     pub type_scheme: TypeScheme,
     pub functions: Vec<NamedExpression<Expr>>,
```
chriseth, line 341: What I meant was to use `Arc` here as well, so that we don't need to clone the expression.

chriseth: And then we can make solved_impls "point" at the expression directly and don't need to search for the function or index-access the vector.

ast/src/parsed/mod.rs
```diff
@@ -336,8 +338,35 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
...
+    }
+
+    pub fn function_bodies_mut(&mut self) -> TraitBodyIterMut<R> {
+        TraitBodyIterMut {
+            functions: self.functions.iter_mut(),
+        }
+    }
+}
+
+pub struct TraitBodyIterMut<'a, R> {
```
gzanitti, line 358: Iterator to be able to obtain mutable references from `Arc<Expr>`. As far as I understand`Arc::get_mut` should not give us problems.

importer/src/path_canonicalizer.rs
```diff
@@ -116,8 +116,8 @@ impl<'a> Folder for Canonicalizer<'a> {
                         .map(|value| value.map(|value| SymbolDefinition { name, value }.into()))
                     }
                     ModuleStatement::TraitImplementation(mut trait_impl) => {
-                        for f in &mut trait_impl.functions {
-                            canonicalize_inside_expression(&mut f.body, &self.path, self.paths)
+                        for f in trait_impl.function_bodies_mut() {
```
gzanitti, line 119: This is the only other use of Arc's mutable iterator.

parser/src/powdr.lalrpop
```diff
@@ -744,21 +754,22 @@ TraitFunction: TraitFunction<Expression> = {
...
+
+NamedArcExpressions: Vec<NamedExpression<Arc<Expression>>> = {
     => vec![],
-    <mut list:( <NamedExpression> "," )*> <end:NamedExpression> ","?  => { list.push(end); list }
+    <mut list:( <NamedArcExpression> "," )*> <end:NamedArcExpression> ","?  => { list.push(end); list }
 }
 
-NamedExpression: NamedExpression<Expression> = {
-    <name:Identifier> ":" <body:BoxedExpression> => NamedExpression { name, body }
+NamedArcExpression: NamedExpression<Arc<Expression>> = {
```
gzanitti, line 766: I had to duplicate the `NamedExpressions` because they are also used in the structs, where Arc is not necessary (it uses Box). I don't think it's worth the cost to unify them.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,131 @@
...
+                    return Some(((name, type_args), Arc::clone(&expr.body)));
+                }
+            }
+        }
+
+        None
+    }
+}
+
+pub fn traits_resolution(
```
chriseth, line 125: I personally prefer it for public functions to be at the top.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,131 @@
...
+        types::{TupleType, Type},
+        visitor::AllChildren,
+        SelectedExpressions, TraitImplementation,
+    },
+};
+use std::{collections::HashMap, sync::Arc};
+
+type SolvedImpl = ((String, Vec<Type>), Arc<Expression>);
+
+pub struct TraitsResolver<'a> {
```
chriseth, line 16: Does this need to be public?

pil-analyzer/src/traits_resolver.rs
chriseth: And please also document it.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -71,6 +74,12 @@ struct PILAnalyzer {
     symbol_counters: Option<Counters>,
     /// Symbols from the core that were added automatically but will not be printed.
     auto_added_symbols: HashSet<String>,
+    /// Implementations found, organized according to their associated trait name.
+    implementations: HashMap<String, Vec<TraitImplementation<Expression>>>,
+    /// A map between the name and type_Args of the caller reference
+    /// and the expression to be called.
+    /// Empty until resolve_trait_impls() is called.
+    solved_impls: HashMap<String, HashMap<Vec<Type>, Arc<Expression>>>,
```
chriseth, line 82: Could also be a local variable inside `analyze()` that is then passed to the `condense` function.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,131 @@
...
+    ) -> Self {
+        Self {
+            definitions,
+            identities,
+        }
+    }
+
+    pub fn resolve_traits(
+        &'a self,
+        trait_impls: &HashMap<String, Vec<TraitImplementation<Expression>>>,
```
chriseth, line 34: Why is this not part of the state?
Also I wonder if the code would be cleaner by moving some of the logic from this function to pil_analyzer. The trait resolver does not really need to know how to extract expressions from the various kinds of function value definitions.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,131 @@
...
+                    result
+                        .entry(key.0)
+                        .or_insert_with(HashMap::new)
+                        .insert(key.1, value);
+                }
+            }
+        }
+
+        for identity in self.identities.iter() {
+            for selector in identity
```
chriseth, line 71: Doesn't identity implement `all_children()`?

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,131 @@
...
+                result
+                    .entry(key.0)
+                    .or_insert_with(HashMap::new)
+                    .insert(key.1, value);
+            }
+        }
+
+        result
+    }
+    fn resolve_reference_trait(
```
chriseth, line 100:
```suggestion
    fn resolve_trait_function_reference(
```

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,131 @@
...
+        &self,
+        reference: &PolynomialReference,
+        trait_impls: &HashMap<String, Vec<TraitImplementation<Expression>>>,
+    ) -> Option<SolvedImpl> {
+        let name = reference.name.clone();
```
chriseth, line 105: Do you need the `clone`? Can't you split `&str`?

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,131 @@
...
+        let (trait_decl_name, trait_fn_name) = name.rsplit_once("::")?;
+        if let Some(impls) = trait_impls.get(trait_decl_name) {
```
chriseth, line 107: Please add a test where you call a trait function that has zero implementations.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,131 @@
...
+            for impl_ in impls.iter() {
+                let Type::Tuple(TupleType { ref items }) = impl_.type_scheme.ty else {
+                    unreachable!()
+                };
+                let type_args = reference.type_args.as_ref().unwrap().to_vec();
+
+                let expr = impl_.function_by_name(trait_fn_name).unwrap();
+                if type_args == *items {
```
chriseth, line 115: You need to call unify or are you saying this is fine because trait impls are not generic yet? If yes, please add a todo

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,131 @@
...
+        trait_impls: &HashMap<String, Vec<TraitImplementation<Expression>>>,
+    ) -> Option<SolvedImpl> {
+        let name = reference.name.clone();
+        let (trait_decl_name, trait_fn_name) = name.rsplit_once("::")?;
+        if let Some(impls) = trait_impls.get(trait_decl_name) {
+            for impl_ in impls.iter() {
+                let Type::Tuple(TupleType { ref items }) = impl_.type_scheme.ty else {
+                    unreachable!()
+                };
+                let type_args = reference.type_args.as_ref().unwrap().to_vec();
```
chriseth, line 112: Would make sense to do that only once outside the loop.

ast/src/parsed/mod.rs
chriseth: Do you actually need a separate struct? `self.functions.iter_mut().map(|named_expr| Arc::get_mut(&mut named_expr.body).unwrap())` does not work?

pil-analyzer/src/evaluator.rs
```diff
@@ -440,24 +450,43 @@ impl<'a> Definitions<'a> {
...
-                Some(FunctionValueDefinition::Expression(TypedExpression {
-                    e: value,
-                    type_scheme,
-                })) => {
-                    let type_args = type_arg_mapping(type_scheme, type_args);
-                    evaluate_generic(value, &type_args, symbols)?
+            let impl_ = match type_args {
+                Some(type_args) => solved_impls
+                    .get(&name)
+                    .and_then(|inner_map| inner_map.get(type_args)),
```
chriseth, line 456: I think you need to first call `type_arg_mapping(type_scheme, type_args)` and then use the result here.

chriseth: Please also add a test to check that, i.e. call a trait function from a generic function and then at the very outside, call that generic function with two types.

pil-analyzer/src/untyped_evaluator.rs
```diff
@@ -19,6 +19,7 @@ pub fn evaluate_expression_to_int(
     evaluator::evaluate_expression::<GoldilocksField>(
         &ExpressionProcessor::new(driver, &Default::default()).process_expression(expr),
         driver.definitions(),
+        driver.solved_impls(),
```
chriseth, line 22: At this point, we have not even run the type checker, so the solved_impls will always me empty.

I think you can remove the function and just pass Default::default() here.

importer/src/path_canonicalizer.rs
```diff
@@ -116,7 +117,11 @@ impl<'a> Folder for Canonicalizer<'a> {
                         .map(|value| value.map(|value| SymbolDefinition { name, value }.into()))
                     }
                     ModuleStatement::TraitImplementation(mut trait_impl) => {
-                        for f in trait_impl.function_bodies_mut() {
+                        for f in trait_impl
```
chriseth, line 120: `children_mut()`?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -326,15 +323,52 @@ impl PILAnalyzer {
...
+                    if let Expression::Reference(
+                        _,
+                        Reference::Poly(
+                            reference @ PolynomialReference {
+                                type_args: Some(_), ..
+                            },
+                        ),
+                    ) = expr
+                    {
+                        Some(reference.clone())
```
chriseth, line 341: Why do you need to clone here? Would it work for the `resolve_references` function to return an iterator (and not converting the iterator to a vector at all)?

pil-analyzer/src/traits_resolver.rs
```diff
@@ -1,131 +1,85 @@
...
+
+impl TraitsResolver {
+    pub fn new() -> Self {
+        Self {}
     }
 
     pub fn resolve_traits(
-        &'a self,
+        &self,
+        references: Vec<PolynomialReference>,
```
chriseth, line 35: As far as I understand it, this function can fail. Not in the case where the reference cannot be uniquely resolved (this should be handled generically by checking that each pair of impls does not unify and is done before this step), but in the case where there is no impl for the type args. Because of that, I think we should have source references here and return a vector of errors (there is no reason to fail on the first error).

chriseth: But this can be a new PR.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -1,131 +1,85 @@
...
         if let Some(impls) = trait_impls.get(trait_decl_name) {
+            let type_args = reference.type_args.as_ref().unwrap().to_vec();
             for impl_ in impls.iter() {
-                let Type::Tuple(TupleType { ref items }) = impl_.type_scheme.ty else {
+                let Type::Tuple(TupleType { items: _ }) = impl_.type_scheme.ty else {
                     unreachable!()
                 };
-                let type_args = reference.type_args.as_ref().unwrap().to_vec();
+
+                let tuple_args = Type::Tuple(TupleType {
```
chriseth, line 66: Can you assert that `tuple_args.is_concrete_type()`? At some point we need to implement this for generic types, but we can do it later.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -1,131 +1,85 @@
...
-                        .entry(key.0)
-                        .or_insert_with(HashMap::new)
-                        .insert(key.1, value);
-                }
-            }
-            for (key, value) in resolve_references(identity.left.expressions.as_ref()) {
-                result
-                    .entry(key.0)
+        for ref_poly in references {
+            if let Some(solved) = self.resolve_trait_function_reference(&ref_poly, trait_impls) {
```
chriseth, line 41: If this function retuns None, it's an error

pil-analyzer/src/traits_resolver.rs
```diff
@@ -1,131 +1,85 @@
...
-        definitions: &'a HashMap<String, (Symbol, Option<FunctionValueDefinition>)>,
-        identities: &'a Vec<Identity<SelectedExpressions<Expression>>>,
-    ) -> Self {
-        Self {
-            definitions,
-            identities,
-        }
+type SolvedImpl = ((String, Vec<Type>), Arc<Expression>);
+
+struct TraitsResolver {}
```
chriseth, line 26: Again, this is my taste, but maybe it would be nice to store the solved_impl map and the `trait_impls` in the struct and then call a function on this struct on a single reference at a time and then have a function to extract the solved_impl map (taking `self` by value). This is a rather flexible interface and we can then iterate over all expressions in the pil_analyzer. Also it is easier to handle the errors singe we can just return a simple Result here (because resolving the reference either fails or not).

pil-analyzer/src/traits_resolver.rs
gzanitti: This type of error is caught during #1601

pil-analyzer/src/traits_resolver.rs
gzanitti: The problem is that there is not always an implementation for every reference. Since we take every reference without distinction, they must `fail` in some silent way.

pil-analyzer/src/traits_resolver.rs
chriseth: Are you sure? I think the overlapping can only detect if there are multiple matching implementations.

pil-analyzer/src/traits_resolver.rs
chriseth: Sure, and I think this is the main place where this can be detected, so it should not fail silently, but maybe I'm wrong.

ast/src/parsed/mod.rs
```diff
@@ -395,6 +395,15 @@ pub struct TraitFunction<E = u64> {
     pub ty: Type<E>,
 }
 
+impl TraitFunction<u64> {
+    pub fn type_scheme(&self, vars: TypeBounds) -> TypeScheme<u64> {
```
chriseth, line 399: Wasn't there a struct where you also have a reference to the trait (so you could get the vars from there)?
If not I'm sorry, this way the function is not too useful...

pil-analyzer/src/evaluator.rs
```diff
@@ -1818,4 +1813,31 @@ mod test {
...
         assert_eq!(parse_and_evaluate_symbol(input, "F::r"), "5".to_string());
     }
+
+    #[test]
+    fn traits_generic_function() {
+        let input = "
+        namespace std::convert(4);
+            let fe = || fe();
+        namespace F(4);
+            trait Add<T> {
```
chriseth, line 1823: To improve this test, you could rename the type parameter from `T` to `Q`. If it's not renamed, it is the same as in `generic_add`, and the test might not detect if we are doing both substitutions correctly.

pil-analyzer/src/evaluator.rs
```diff
@@ -1818,4 +1813,31 @@ mod test {
...
+                add: T, T -> T,
+            }
+            impl Add<int> {
+                add: |a, b| a + b,
+            }
+            impl Add<fe> {
+                add: |a, b| a + b,
```
chriseth, line 1830: Can you maybe use `a + 2*b` or something? The evaluator does not take the determined types into account, but only the runtime types - so even if we pick the wrong `add` function here in the evaluator, the result would still be correct.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -9,54 +9,44 @@ use std::{collections::HashMap, sync::Arc};
...
-    trait_impls: &HashMap<String, Vec<TraitImplementation<Expression>>>,
-) -> HashMap<String, HashMap<Vec<Type>, Arc<Expression>>> {
-    TraitsResolver::new().resolve_traits(references, trait_impls)
-}
-
 type SolvedImpl = ((String, Vec<Type>), Arc<Expression>);
 
-struct TraitsResolver {}
+pub struct TraitsResolver {
+    trait_impls: HashMap<String, Vec<TraitImplementation<Expression>>>,
```
chriseth, line 15: Is there a reason this can not be a reference?

pil-analyzer/src/traits_resolver.rs
```diff
@@ -9,54 +9,44 @@ use std::{collections::HashMap, sync::Arc};
...
-                solved_impls
+    pub fn resolve_trait(&mut self, ref_poly: &PolynomialReference) -> Result<(), String> {
+        match self.resolve_trait_function_reference(ref_poly) {
+            Some(((key, type_args), expr)) => {
+                self.solved_impls
                     .entry(key)
-                    .or_insert_with(HashMap::new)
-                    .insert(type_args, solved.1);
+                    .or_default()
+                    .insert(type_args, expr);
```
chriseth, line 33: It would be nice if we could avoid running the unifier and all that in case we already have an entry in the hash maps here.

pil-analyzer/src/traits_resolver.rs
gzanitti: Yes, but since overlapping happens before type checking, it is responsible for catching most of these errors. 

Anyway, if you consider that it is also necessary here, I can add a test for this case and the necessary checks. No problem.

pil-analyzer/src/traits_resolver.rs
gzanitti: The problem I see is that when we receive a reference that has no implementation, let's say any variable `F::x`, it will never find an implementation and will always fail.

pil-analyzer/src/condenser.rs
```diff
@@ -178,6 +180,8 @@ pub struct Condenser<'a, T> {
     degree: Option<DegreeRange>,
     /// All the definitions from the PIL file.
     symbols: &'a HashMap<String, (Symbol, Option<FunctionValueDefinition>)>,
+    /// All the trait implementations from the PIL file.
```
chriseth, line 183:
```suggestion
    /// Pointers to expressions for all referenced trait implementations and the concrete types.
```

pil-analyzer/src/evaluator.rs
```diff
@@ -461,6 +471,20 @@ impl<'a> Definitions<'a> {
                         Value::TypeConstructor(&variant.name).into()
                     }
                 }
+                Some(FunctionValueDefinition::TraitFunction(_, _)) => {
+                    let type_arg = type_args.clone().unwrap();
```
chriseth, line 475: Why not `.as_ref().unwrap()`?

chriseth: I think it would be nice to have an assertion that `type_arg` is concrete.

chriseth: although actually I think this should always be the case here, right? So we could put it at the start of the function.

chriseth: I mean we could do that at the call site, but I yeah, probably it's fine. Then let's not do that at all.

gzanitti: There is already one, but inside the traits_resolver, before the unifier.
https://github.com/powdr-labs/powdr/blob/0af8e32a68aca886660f8d421305a2f938f8888c/pil-analyzer/src/traits_resolver.rs#L68

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -72,6 +75,8 @@ struct PILAnalyzer {
     symbol_counters: Option<Counters>,
     /// Symbols from the core that were added automatically but will not be printed.
     auto_added_symbols: HashSet<String>,
+    /// Implementations found, organized according to their associated trait name.
```
chriseth, line 78:
```suggestion
    /// All trait implementations found, organized according to their associated trait name.
```

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -318,9 +323,48 @@ impl PILAnalyzer {
         }
     }
 
-    pub fn condense<T: FieldElement>(self) -> Analyzed<T> {
+    fn resolve_trait_impls(&mut self) -> HashMap<String, HashMap<Vec<Type>, Arc<Expression>>> {
```
chriseth, line 326:
```suggestion
    /// Creates and returns a map for every referenced trait and every concrete type to the
    /// corresponding trait implementation function.
    fn resolve_trait_impls(&mut self) -> HashMap<String, HashMap<Vec<Type>, Arc<Expression>>> {
```

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -318,9 +323,48 @@ impl PILAnalyzer {
...
+        };
+
+        for (_, (_, def)) in self.definitions.iter() {
+            if let Some(FunctionValueDefinition::Expression(TypedExpression { e: expr, .. })) = def
+            {
+                resolve_references(expr);
+            }
+        }
+
+        for identity in self.identities.iter() {
```
chriseth, line 352: Do you need to call `iter()`? Doesn't `in &self.identities {` work as well?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -318,9 +323,48 @@ impl PILAnalyzer {
...
+                        },
+                    ),
+                ) = expr
+                {
+                    let _ = trait_solver.resolve_trait(reference);
+                }
+            });
+        };
+
+        for (_, (_, def)) in self.definitions.iter() {
```
chriseth, line 345:
```suggestion
        for (_, def) in self.definitions.values() {
```

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -318,9 +323,48 @@ impl PILAnalyzer {
...
+            if let Some(FunctionValueDefinition::Expression(TypedExpression { e: expr, .. })) = def
```
chriseth, line 348: Actually for Arrays you need to do the same. Essentially the same as in infer_type, just that we don't care about the type of the expression.

chriseth: Oh and also `self.implementations` - and please add a test for that: a trait impl function that calls another trait function.

pil-analyzer/src/statement_processor.rs
```diff
@@ -643,4 +651,50 @@ where
             functions,
         }
     }
+
+    fn process_trait_implementation(
+        &self,
+        trait_impl: parsed::TraitImplementation<parsed::Expression>,
+    ) -> TraitImplementation<Expression> {
+        let type_vars = trait_impl.type_scheme.vars.vars().collect();
```
chriseth, line 659: Since we don't support generic impls, we should add an `unimplemented` assert here that `type_vars` should be empty.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,82 @@
...
+        TraitImplementation,
+    },
+};
+use std::{collections::HashMap, sync::Arc};
+
+use crate::type_unifier::Unifier;
+
+type SolvedImpl = ((String, Vec<Type>), Arc<Expression>);
+
+pub struct TraitsResolver<'a> {
```
chriseth, line 14: Please document.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,82 @@
...
+
+impl<'a> TraitsResolver<'a> {
+    pub fn new(trait_impls: &'a HashMap<String, Vec<TraitImplementation<Expression>>>) -> Self {
+        Self {
+            trait_impls,
+            solved_impls: HashMap::new(),
+        }
+    }
+
+    pub fn resolve_trait(&mut self, ref_poly: &PolynomialReference) -> Result<(), String> {
```
chriseth, line 27: We could clarify a bit better what it does by calling this `resolve_trait_function_reference`.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,82 @@
...
+        if let Some(inner_map) = self.solved_impls.get(&ref_poly.name) {
+            if let Some(type_args) = &ref_poly.type_args {
```
chriseth, line 29: If ref_poly.type_args is None, we can return right away at the start of the function because it's not a trait function reference.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    fn resolve_trait_function_reference(
+        &self,
+        reference: &PolynomialReference,
+    ) -> Option<SolvedImpl> {
+        let (trait_decl_name, trait_fn_name) = reference.name.rsplit_once("::")?;
+        if let Some(impls) = self.trait_impls.get(trait_decl_name) {
+            let type_args = reference.type_args.as_ref().unwrap().to_vec();
+            for impl_ in impls.iter() {
+                let Type::Tuple(TupleType { items: _ }) = impl_.type_scheme.ty else {
```
chriseth, line 60: I don't think this is needed any more.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,82 @@
...
+                    unreachable!()
+                };
+
+                let tuple_args = Type::Tuple(TupleType {
```
chriseth, line 64: This can all move outside of the loop.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -0,0 +1,82 @@
...
+                    items: type_args.clone(),
+                });
+
+                assert!(tuple_args.is_concrete_type());
+
+                let mut unifier: Unifier = Default::default();
+
+                let res = unifier.unify_types(tuple_args.clone(), impl_.type_scheme.ty.clone());
+                if let Ok(()) = res {
```
chriseth, line 73:
```suggestion
                if res.is_ok() {
```
Didn't clippy complain here?

pil-analyzer/tests/types.rs
```diff
@@ -561,6 +561,61 @@ fn empty_conditional() {
...
+    impl Add<fe> {
+        add: |a, b| a + b,
+    }
+    let r: fe = Add::add(3, 4);
+    ";
+    type_check(input, &[("r", "", "fe")]);
+}
+
+#[test]
+#[should_panic = "Unable to derive concrete type for literal 2."]
```
chriseth, line 579: Since we don't support generic impls, could you move the test to the generic impl PR?

pil-analyzer/tests/types.rs
```diff
@@ -561,6 +561,61 @@ fn empty_conditional() {
...
+            add: |a, b| std::convert::fe(a + b),
+        }
+
+        let r: fe = Add::add(1,2);
+    ";
+    type_check(input, &[]);
+}
+
+#[test]
+fn impl_combined_test() {
```
chriseth, line 598: Same here.

pil-analyzer/src/statement_processor.rs
```diff
@@ -643,4 +651,50 @@ where
...
+        let functions = trait_impl
+            .functions
+            .into_iter()
+            .map(|named| NamedExpression {
+                name: named.name,
+                body: Arc::new(
+                    self.expression_processor(&type_vars)
+                        .process_expression(named.body.as_ref().clone()),
```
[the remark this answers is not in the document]
gzanitti: Because `process_expression` needs an `Expression`, and `named.body` has the `Arc` on it.

pil-analyzer/tests/types.rs
```diff
@@ -614,3 +669,88 @@ fn new_fixed_column_wrong_type() {
...
+    let input = "
+    trait ToTuple<S, I> {
+        get: S -> (S, I),
+    }
+    impl ToTuple<int, (int, int)> {
+        get: |n| (n, (1, n+2)),
+    }
+    let r: (int, (int, int)) = ToTuple::get(3);
+    ";
+    type_check(input, &[("r", "", "(int, (int, int))")]);
```
chriseth, line 708: It is useless to check for the type (i.e. to list `r` in the array you pass to `type_check`), since you already fully specified the type. You can just call the function with an empty array.

pil-analyzer/tests/types.rs
```diff
@@ -614,3 +669,88 @@ fn new_fixed_column_wrong_type() {
...
+            V1::A => V2::B,
+        },
+    }
+    impl Convert<V2, V1> {
+        convert: |x| match x {
+            V2::B => V1::A,
+        },
+    }
+
+    let r1: V2 = Convert::convert(V1::A);
```
chriseth, line 752: I think to test anything here, you should remove the types of r1 and r2.

chriseth: But then the type checker will complain because it cannot assign types, and it's correct.

chriseth: So I think the test is good, but what we need is a negative test as well: Call `Convert::convert(false)`.

pil-analyzer/src/traits_resolver.rs
gzanitti: not at all

pil-analyzer/src/traits_resolver.rs
```diff
@@ -11,6 +11,9 @@ use crate::type_unifier::Unifier;
 
 type SolvedImpl = ((String, Vec<Type>), Arc<Expression>);
 
+/// TraitsResolver implements a trait resolver for polynomial references.
+/// It manages trait implementations and provides functionality to resolve
+/// trait function references based on given polynomial references.
```
chriseth, line 16:
```suggestion
/// For each reference to a trait function with type arguments, it finds the corresponding
/// trait implementation and stores this association in a map that is returned.
```

pil-analyzer/src/traits_resolver.rs
```diff
@@ -24,16 +27,21 @@ impl<'a> TraitsResolver<'a> {
         }
     }
 
-    pub fn resolve_trait(&mut self, ref_poly: &PolynomialReference) -> Result<(), String> {
+    /// Resolves a trait function reference for a given polynomial reference.
+    /// If successful, it caches the resolved implementation for future use.
```
chriseth, line 31:
```suggestion
    /// If successful, it stores the resolved implementation to be returned via `solved_impls()`.
```

pil-analyzer/src/traits_resolver.rs
```diff
@@ -24,16 +27,21 @@ impl<'a> TraitsResolver<'a> {
...
+    pub fn resolve_trait_function_reference(
+        &mut self,
+        ref_poly: &PolynomialReference,
+    ) -> Result<(), String> {
         if let Some(inner_map) = self.solved_impls.get(&ref_poly.name) {
```
chriseth, line 36:
```suggestion
        if ref_poly.type_args.is_none() {
            // not a trait function reference
            return Ok(());
        }
        if let Some(inner_map) = self.solved_impls.get(&ref_poly.name) {
```
and then we can just use `unwrap()` on the type args in all the other functions.

pil-analyzer/src/evaluator.rs
```diff
@@ -1733,4 +1760,131 @@ mod test {
...
+            };
+
+            let z: int = y(2);
+            let r: fe = x(2);
+            ";
+
+        assert_eq!(parse_and_evaluate_symbol(input, "F::r"), "11".to_string());
+        assert_eq!(parse_and_evaluate_symbol(input, "F::z"), "6".to_string());
+    }
 }
```
chriseth, line 1890: Please add a test where you call a trait function inside a trait impl, because I think there is still something missing in the pil analyzer regarding that.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -318,9 +323,55 @@ impl PILAnalyzer {
...
+                _ => {}
+            }
+        }
+
+        for identity in &self.identities {
+            for expr in identity.all_children() {
+                resolve_references(expr);
+            }
+        }
+
```
chriseth, line 364: I think you also need to resolve inside self.implementations.

pil-analyzer/src/statement_processor.rs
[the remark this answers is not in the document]
chriseth: Did you try it?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -323,6 +341,61 @@ impl PILAnalyzer {
         }
     }
 
+    fn specialize_trait_type<T>(
```
chriseth, line 344: Maybe a good candidate for a function of TraitImplementation?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -323,6 +341,61 @@ impl PILAnalyzer {
...
+            .collect();
+
+        let trait_fn = trait_decl
+            .function_by_name(&named_expr.name)
+            .expect("Function not found in trait declaration");
+
+        Self::replace_type_vars(&trait_fn.ty, &type_var_mapping)
+    }
+
+    fn replace_type_vars(ty: &Type, type_var_mapping: &HashMap<String, Type>) -> Type {
```
chriseth, line 371: Isn't there  a function on Type for that purpose?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -231,6 +232,22 @@ impl PILAnalyzer {
         // We filter out enum type declarations (the constructor functions have been added
         // by the statement processor already).
         // For Arrays, we also collect the inner expressions and expect them to be field elements.
+
+        for (name, trait_impls) in self.implementations.iter_mut() {
```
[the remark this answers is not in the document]
gzanitti: I don't think so, the `iter_mut` of the `let definitions...` doesn't let me reuse `self.definitions`.

chriseth:
Just saw that side_effect_check also needs to take the impls into account. Maybe there is some code to be shared between side_effect_check and the trait checks?

pil-analyzer/src/traits_resolver.rs
```diff
@@ -83,3 +83,32 @@ impl<'a> TraitsResolver<'a> {
...
+        .iter()
+        .cloned()
+        .zip(items.iter().cloned())
+        .collect();
+
+    let trait_fn: &mut TraitFunction = trait_decl
+        .function_by_name_mut(&named_expr.name)
+        .expect("Function not found in trait declaration");
+
+    trait_fn.ty.substitute_type_vars(&type_var_mapping);
```
chriseth, line 111: Ah, now I see! Does it even have a type? Shouldn't we just take the type from the trait and then let the type inferenec algorithm do its job?

chriseth: I.e. just return `trait_type.clone().substitute_type_vars(&type_var_mapping)`?

chriseth: And then I think the type inference algorithm will store the type anyway at the end.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -228,22 +228,24 @@ impl PILAnalyzer {
...
+        // Collect all definitions and traits implementations with their types and expressions.
         // We filter out enum type declarations (the constructor functions have been added
         // by the statement processor already).
         // For Arrays, we also collect the inner expressions and expect them to be field elements.
 
         for (name, trait_impls) in self.implementations.iter_mut() {
             let (_, def) = self
                 .definitions
-                .get(name)
+                .get_mut(name)
```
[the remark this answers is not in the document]
gzanitti: The generic type that I am going to “specialize” using the implementation, is stored in the trait declaration, that's why the mutability.

ast/src/parsed/mod.rs
```diff
@@ -362,15 +362,7 @@ impl<R> TraitImplementation<Expression<R>> {
...
-    pub fn specialize_trait_type(
-        &self,
-        def: &mut Option<FunctionValueDefinition>,
-        expr_name: &str,
-    ) -> Type {
-        let Some(FunctionValueDefinition::TraitDeclaration(trait_decl)) = def else {
-            panic!("Expected trait declaration");
-        };
-
+    pub fn specialize_trait_type(&self, trait_decl: &TraitDeclaration, fn_name: &str) -> Type {
```
chriseth, line 365: Nice! And now I think we could rename this to `type_of_function`.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -220,7 +220,9 @@ impl PILAnalyzer {
             for impl_ in impls {
```
chriseth, line 220: Is this inside the loop over the definitions?

chriseth (approved):

PR #1627, opened 2024-07-29
Parameterized expressions

gzanitti:
This PR 'parameterise' Expressions to be able to choose when they can be converted into a StructExpression and when they cannot. This modification allows us to overcome ambiguities with Structs and to adopt the standard Structs syntax:
```
struct Point{ x: int, y: int }
let p: Point = Point{ x: 1, y: 2 };
```

This PR replace #1591 to be able to include this modifications before merging structs.


parser/src/powdr.lalrpop
```diff
@@ -445,25 +445,25 @@ LabelStatement: FunctionStatement = {
...
-ExpressionList: Vec<Expression> = {
+ExpressionList<StructOption>: Vec<Expression> = {
     => vec![],
-    <mut list:( <Expression> "," )*> <end:Expression>  => { list.push(end); list }
+    <mut list:( <Expression<StructOption>> "," )*> <end:Expression<StructOption>>  => { list.push(end); list }
 }
 
-Expression: Expression = {
-    BoxedExpression => *<>,
+Expression<StructOption>: Expression = {
```
chriseth, line 453: Maybe add a comment explaining StructOption?

parser/src/powdr.lalrpop
```diff
@@ -619,35 +633,35 @@ GenericReference: NamespacedPolynomialReference = {
     <path:GenericSymbolPath> => NamespacedPolynomialReference{path: path.0, type_args: path.1},
 }
 
-MatchExpression: Box<Expression> = {
-    <start:@L> "match" <scrutinee:BoxedExpression> "{" <arms:MatchArms> "}" <end:@R> => ctx.to_expr_with_source_ref(MatchExpression{scrutinee, arms}, start, end),
+MatchExpression<StructOption>: Box<Expression> = {
+    <start:@L> "match" <scrutinee:BoxedExpression<TermWithoutStruct>> "{" <arms:MatchArms<StructOption>> "}" <end:@R> => ctx.to_expr_with_source_ref(MatchExpression{scrutinee, arms}, start, end),
```
chriseth, line 637: Do we actually need to "forward" the StructOption here to MatchArms (and many other places)? Isn't it only relevant for the Term?

parser/src/powdr.lalrpop
```diff
@@ -555,58 +555,72 @@ ProductOp: BinaryOperator = {
...
 
 PowOp: BinaryOperator = {
     "**" => BinaryOperator::Pow,
 }
 
-Unary: Box<Expression> = {
-    <start:@L> <op:PrefixUnaryOp> <expr:PostfixUnary> <end:@R> => ctx.to_expr_with_source_ref(UnaryOperation{op, expr}, start, end),
-    PostfixUnary,
+Unary<StructOption>: Box<Expression> = {
+    <start:@L> <op:PrefixUnaryOp> <expr:PostfixUnary<StructOption>> <end:@R> => ctx.to_expr_with_source_ref(UnaryOperation{op, expr}, start, end),
```
chriseth, line 568: I think even here we should unconditionally allow structs. I think otherwise the following (maybe weird but still) example would not parse:
```
match -MyStruct{a: 1} {
...
}
```

chriseth: Oh, I see, that's true! But it should be possible to do this in at least some nested expressions like MatchArm, shouldn't it? Or is the problem there that we don't know where the arm body begins?

chriseth: Ok, but at least inside `[]` and `()` it should be possible to allow structs.

chriseth: So I think that ExpressionList can be defined without a parameter...

parser/src/powdr.lalrpop
```diff
@@ -555,58 +555,72 @@ ProductOp: BinaryOperator = {
...
     <start:@L> <value:Number> <end:@R> => ctx.to_expr_with_source_ref(Number {value: value.into(), type_: None}, start, end),
     <start:@L> <s:StringLiteral> <end:@R> => Box::new(Expression::String(ctx.source_ref(start, end), s)),
-    MatchExpression,
-    IfExpression,
-    BlockExpression,
-    <start:@L> "[" <items:ExpressionList> "]" <end:@R> => ctx.to_expr_with_source_ref(ArrayLiteral{items}, start, end),
-    <start:@L> "(" <head:Expression> "," <tail:ExpressionList> ")" <end:@R> => { let mut list = vec![head]; list.extend(tail); Box::new(Expression::Tuple(ctx.source_ref(start, end), list)) },
+    MatchExpression<StructOption>,
+    IfExpression<StructOption>,
+    BlockExpression<StructOption>,
```
chriseth, line 610: I think that this as also not needed, is it?

parser/src/powdr.lalrpop
```diff
@@ -555,58 +555,72 @@ ProductOp: BinaryOperator = {
...
-Term: Box<Expression> = {
-    <start:@L> <i:IndexAccess> <end:@R> => ctx.to_expr_with_source_ref(i, start, end),
-    <start:@L> <f:FunctionCall> <end:@R> => ctx.to_expr_with_source_ref(f, start, end),
+Term<StructOption>: Box<Expression> = {
+    StructOption,
+}
+
+TermWithStruct: Box<Expression> = {
+    TermCommon<TermWithStruct>,
+    //StructExpression<TermWithStruct>,
```
chriseth, line 592: In order to test it out, you could already parse struct expressions and then just turn them into something like an empty tuple value.

parser/src/powdr.lalrpop
gzanitti: You are right, several `StructOption` were not necessary. I simplified the grammar by removing all these cases and leaving `Expression<TermWithoutStruct>` on the edge cases (match, if, and some other error-producing rules like Linkdeclaration). All other cases are now Expression<TermWithStruct>.

parser/src/powdr.lalrpop
```diff
@@ -444,10 +444,14 @@ LabelStatement: FunctionStatement = {
 }
 
 // ---------------------------- Expressions -----------------------------
+// Expressions are parameterized with StructOption to be able to define when we want to allow them to become structs.
```
chriseth, line 447:
```suggestion
// Expressions are parameterized with StructOption because in some cases, we want to disallow the parsing of struct expressions.
```

parser/src/powdr.lalrpop
```diff
@@ -72,7 +72,7 @@ Part: Part = {
 // Same as SymbolPath plus we allow "::<...>" at the end.
 GenericSymbolPath: (SymbolPath, Option<Vec<Type<Expression>>>) = {
     // If we "inline" SymbolPath here, we get an ambiguity error.
-    <abs:"::"?> <parts:( <Part> "::" )*> <end:Part> <types:("::" "<" <TypeTermList<Expression>> ">")?>  => (
+    <abs:"::"?> <parts:( <Part> "::" )*> <end:Part> <types:("::" "<" <TypeTermList<Expression<TermWithStruct>>> ">")?>  => (
```
chriseth, line 75: Does it make sense to define `Expression` as `Expression<TermWithStruct>` and use some other name for the one that takes the parameter? Like ExpressionRestricted or something?

parser/src/powdr.lalrpop
```diff
@@ -595,18 +613,18 @@ Term: Box<Expression> = {
     IfExpression,
     BlockExpression,
     <start:@L> "[" <items:ExpressionList> "]" <end:@R> => ctx.to_expr_with_source_ref(ArrayLiteral{items}, start, end),
-    <start:@L> "(" <head:Expression> "," <tail:ExpressionList> ")" <end:@R> => { let mut list = vec![head]; list.extend(tail); Box::new(Expression::Tuple(ctx.source_ref(start, end), list)) },
+    <start:@L> "(" <head:Expression<StructOption>> "," <tail:ExpressionList> ")" <end:@R> => { let mut list = vec![head]; list.extend(tail); Box::new(Expression::Tuple(ctx.source_ref(start, end), list)) },
```
chriseth, line 616: Shouldn't this also allow structs?

parser/src/powdr.lalrpop
```diff
@@ -595,18 +613,18 @@ Term: Box<Expression> = {
...
     <start:@L> "(" ")" <end:@R> => Box::new(Expression::Tuple(ctx.source_ref(start, end), vec![])),
-    "(" <BoxedExpression> ")",
-    <start:@L> "${" <b:BoxedExpression> "}" <end:@R> => Box::new(Expression::FreeInput(ctx.source_ref(start, end), b))
+    "(" <BoxedExpression<StructOption>> ")",
```
chriseth, line 618: Same in these two lines?

parser/src/powdr.lalrpop
```diff
@@ -629,25 +647,42 @@ MatchArms: Vec<MatchArm> = {
...
         <body:BlockExpression>
         "else"
         <else_body:BlockExpression>
         <end:@R> => ctx.to_expr_with_source_ref(IfExpression{condition, body, else_body}, start, end),
 }
 
-
 BlockExpression: Box<Expression> = {
-    <start:@L> "{" <statements:StatementInsideBlock*> <expr:BoxedExpression?> "}" <end:@R> => ctx.to_expr_with_source_ref(BlockExpression{statements, expr}, start, end),
+    <start:@L> "{" <statements:StatementInsideBlock *> <expr:BoxedExpression<TermWithStruct>?> "}" <end:@R> => ctx.to_expr_with_source_ref(BlockExpression{statements, expr}, start, end),
```
chriseth, line 662: Why does it not work to allow structs here?

parser/src/powdr.lalrpop
gzanitti: Can't have both `Expression` as a rename and `Expression<StructOption>` as a Macro. The only way would be something like `RestrictedExpression` and `FullExpression`, what do you think?

parser/src/powdr.lalrpop
gzanitti: It does. `<TermWithStruct>` means that Structs are possible.

parser/src/powdr.lalrpop
```diff
@@ -446,26 +446,37 @@ LabelStatement: FunctionStatement = {
...
-BoxedExpression: Box<Expression> = {
-    LambdaExpression,
+MacroBoxedExpression<StructOption>: Box<Expression> = {
+    LambdaExpression<StructOption>,
 }
 
-LambdaExpression: Box<Expression> = {
-    <start:@L> <kind:FunctionKind> "||" <body:BoxedExpression> <end:@R>
+LambdaExpression<StructOption>: Box<Expression> = {
+    <start:@L> <kind:FunctionKind> "||" <body:MacroBoxedExpression<StructOption>> <end:@R>
```
chriseth, line 475: Does this not allow generic expressions (because there was a `||` before)?

parser/src/powdr.lalrpop
```diff
@@ -446,26 +446,37 @@ LabelStatement: FunctionStatement = {
...
         => ctx.to_expr_with_source_ref(LambdaExpression{kind, params: vec![], body, outer_var_references: Default::default()}, start, end),
-    <start:@L> <kind:FunctionKind> "|" <params:ParameterList> "|" <body:BoxedExpression> <end:@R>
+    <start:@L> <kind:FunctionKind> "|" <params:ParameterList> "|" <body:MacroBoxedExpression<StructOption>> <end:@R>
```
chriseth, line 477: Same here.

chriseth (approved):

PR #1640, opened 2024-07-31
Enable the linker to output dynamic namespaces

Schaeff:
What we can have after this PR:
- only dynamic namespaces
- dynamic main, others dynamic or static

What we still cannot have:
- static main and another machines dynamic

The change to `_linker_first_step` can be reverted after #1565

### DIFF
--- a/book/src/architecture/linker.md
+++ b/book/src/architecture/linker.md
@@ -4,13 +4,13 @@ A linker is used to turn an [AIR tree](./compiler.md#air-generation) into a sing
 The linking process operates in the following way:
 
 1. Create an empty PIL file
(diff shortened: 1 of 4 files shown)


chriseth (approved):

georgwiese (approved):
Nice!

I tried to change `test_data/asm/vm_to_block_different_length.asm` to set the size of the main machine to `None`, and it kind of works, with some caveats:
- Still a few cases of fixed columns being declared using the array syntax.
- We don't currently set the size of the ROM
- For VMs, witness generation still picks the largest available size. It should instead downsize to the next power of 2 as soon as it sees a loop.

PR #1641, opened 2024-07-31
fake std support

lvella:
We don't plan to actually support all the features of `std`, but this will allow Powdr to build and execute `std` crates that limits themselves to the features we do support (which, as of this PR, is reading and writing to specific file descriptors). Only works via ELF path.


lvella:
I changed the default rust target from `riscv32imac-unknown-none-elf` to `riscv32im-risc0-zkvm-elf` as there isn't any significant drawback in doing so, and it allows `std` to be used.

The binaries are bigger because it lacks C extension, and you can't use inline assembly to issue atomic instructions, as it lacks A extension, but that is all.

.github/workflows/nightly-tests.yml
```diff
@@ -68,8 +68,6 @@ jobs:
       run: rustup toolchain install 1.77-x86_64-unknown-linux-gnu
     - name: Install nightly
       run: rustup toolchain install nightly-2024-08-01-x86_64-unknown-linux-gnu
-    - name: Install riscv target
-      run: rustup target add riscv32imac-unknown-none-elf --toolchain nightly-2024-08-01-x86_64-unknown-linux-gnu
```
lvella, line 72: We never needed this in the first place, because we have always used `-Zbuild-std=core,alloc`. We might want it if we drop support for ASM path, so that `-Zbuild-std=core,alloc` won't be needed anymore.

leonardoalt:
So `std` is the default for guests now, and the user has to opt-in `no-std`?

cli-rs/src/main.rs
```diff
@@ -41,8 +41,8 @@ struct Cli {
 
 #[derive(Subcommand)]
 enum Commands {
-    /// Compiles (no-std) rust code to riscv assembly.
-    /// Needs `rustup target add riscv32imac-unknown-none-elf`.
+    /// Compiles rust code to riscv assembly.
```
leonardoalt, line 44: actually this option ultimately compiles it to powdr-asm, no?

riscv-runtime/rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
 channel = "nightly-2024-08-01"
-targets = ["riscv32imac-unknown-none-elf"]
+targets = ["riscv32imac-unknown-none-elf", "riscv32im-risc0-zkvm-elf"]
```
leonardoalt, line 3: is this used only by tooling to ensure both are installed? or does the crate get compiled to both in a `cargo build`?

riscv-runtime/src/std_support.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use powdr_riscv_syscalls::Syscall;
+
+use crate::io::write_slice;
+
+#[no_mangle]
+extern "C" fn sys_rand(buf: *mut u32, words: usize) {
+    let slice = unsafe { slice::from_raw_parts_mut(buf, words) };
+
+    // I promise I rolled a fair die and this is the result.
```
leonardoalt, line 20: maybe replace this line with an explanation, similar to what you wrote in the chat?

riscv/tests/riscv_data/std_hello_world/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[dependencies]
+powdr-riscv-syscalls = { path = "../../../../riscv-syscalls" }
+powdr-riscv-runtime = { path = "../../../../riscv-runtime", features = ["std"]}
+
+[profile.release]
+panic = 'abort'
+
+[profile.dev]
+panic = 'abort'
```
leonardoalt, line 14: Are these needed or are you just testing it?

riscv/tests/riscv_data/std_hello_world/README.md
```diff
@@ -0,0 +1,35 @@
+My best shot at compiling this so far is:
```
leonardoalt, line 1: Do we still need this file?

riscv/tests/riscv.rs
```diff
@@ -420,13 +428,15 @@ fn verify_riscv_crate_with_data<S: serde::Serialize + Send + Sync + 'static>(
...
+    verify_riscv_crate_impl(case, inputs, runtime, true, true, Some(data))
 }
 
-fn verify_riscv_crate_from_both_paths<S: serde::Serialize + Send + Sync + 'static>(
+fn verify_riscv_crate_impl<S: serde::Serialize + Send + Sync + 'static>(
     case: &str,
     inputs: Vec<GoldilocksField>,
     runtime: &Runtime,
+    via_elf: bool,
+    via_asm: bool,
```
leonardoalt, line 439: Boolean parameters are discouraged, an enum might be clearer.

leonardoalt (commented):
Awesome!

riscv-runtime/rust-toolchain.toml
lvella: The targets does not affect `cargo build`, only the channel (toolchain) does. In fact, I don't know how the targets are used here. I thought it would be in `rust-analyzer`, but apparently not.

cli-rs/src/main.rs
lvella: Yes. Fixed.

riscv-runtime/src/std_support.rs
lvella: Done.

riscv/tests/riscv_data/std_hello_world/Cargo.toml
lvella: This is either needed here or at the `RUSTFLAGS` during build. Maybe it is better to put it there, together with the rest of the crazy flags we use.

riscv/tests/riscv.rs
lvella: Enum of one, another or both? The values are independent, I think bool is better here.

riscv/tests/riscv_data/std_hello_world/README.md
lvella: No, but it would make an interesting comment in the function that generates the compilation command.

riscv/tests/riscv_data/std_hello_world/Cargo.toml
leonardoalt: Yea agree, otherwise the user has to do it themselves

riscv/tests/riscv.rs
leonardoalt: yea ok

leonardoalt (approved):

PR #1623, opened 2024-07-26
Make `ArrayLiteralExpression` sugar for PIL implementation

Schaeff:

PR #1648, opened 2024-08-02
Improve evaluator

chriseth:
Extracted from @Schaeff `s https://github.com/powdr-labs/powdr/pull/1623

### DIFF
--- a/executor/src/constant_evaluator/mod.rs
+++ b/executor/src/constant_evaluator/mod.rs
@@ -94,17 +94,16 @@ fn generate_values<T: FieldElement>(
             } else {
                 e
             };
+            let fun = evaluator::evaluate(e, &mut symbols.clone()).unwrap();
             (0..degree)
                 .into_par_iter()
                 .map(|i| {
-                    let mut symbols = symbols.clone();
-                    let fun = evaluator::evaluate(e, &mut symbols).unwrap();
                     evaluator::evaluate_function_call(
-                        fun,
+                        fun.clone(),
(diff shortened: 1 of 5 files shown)


executor/src/constant_evaluator/mod.rs
```diff
@@ -94,17 +94,16 @@ fn generate_values<T: FieldElement>(
...
             };
+            let fun = evaluator::evaluate(e, &mut symbols.clone()).unwrap();
             (0..degree)
                 .into_par_iter()
                 .map(|i| {
-                    let mut symbols = symbols.clone();
-                    let fun = evaluator::evaluate(e, &mut symbols).unwrap();
                     evaluator::evaluate_function_call(
-                        fun,
+                        fun.clone(),
```
chriseth, line 102: This is an `Arc` so it should be relatively cheap.

chriseth: If we want to optimize this further, we could avoid thread synchronization by having one copy per thread, but I guess we should rather compile to rust instead...

pil-analyzer/src/condenser.rs
```diff
@@ -313,18 +313,23 @@ impl<'a, T: FieldElement> SymbolLookup<'a, T> for Condenser<'a, T> {
     fn lookup(
         &mut self,
         name: &'a str,
-        type_args: Option<Vec<Type>>,
+        type_args: &Option<Vec<Type>>,
```
chriseth, line 316: This avoids a clone in the case the cache entry exists already.

leonardoalt (approved):

PR #1657, opened 2024-08-06
Detect submachine not being a machine

Schaeff:
The way `machine_check` works now is pretty basic: it keeps most of the parsed AST, except for machines where it applies some basic checks and returns a stricter type. In particular, it does not have memory or look-ahead, so that we cannot check, for example, that a submachine is actually a machine.

This PR introduces a flow based on paths, so that at any point in the analysis, we can jump to another place in the parsed AST and check it.

I'd be interested to hear your opinion @gzanitti as you've been more active in this part of the codebase :)

### DIFF
--- a/analysis/src/machine_check.rs
+++ b/analysis/src/machine_check.rs
@@ -1,9 +1,6 @@
 #![deny(clippy::print_stdout)]
 
-use std::{
-    cell::RefCell,
-    collections::{BTreeMap, HashSet},
-};
+use std::{cell::RefCell, collections::BTreeMap};
 
 use powdr_ast::{
     asm_analysis::{
@@ -30,34 +27,48 @@ pub fn check(file: ASMProgram) -> Result<AnalysisASMFile, Vec<String>> {
     let checker = TypeChecker::new(file.main);
     checker.check_path(&ctx)?;
     Ok(AnalysisASMFile {
-        items: checker.analyzed.into_inner(),
+        items: checker
+            .analyzed
+            .into_inner()
+            .into_iter()
+            .filter_map(|(k, v)| match v {
+                Some(ItemOrModule::Item(i)) => Some((k, i)),
+                Some(ItemOrModule::Module) => None,
+                None => unreachable!(),
+            })
+            .collect(),
     })
+}
+
+enum ItemOrModule {
+    Item(Item),
+    Module,
 }
 
 struct TypeChecker {
     /// The parsed AST as a single [SymbolValue::Module]
     ast: SymbolValue,
-    /// The already checked paths
-    checked_paths: RefCell<HashSet<AbsoluteSymbolPath>>,
     /// The checked items
     /// We use [RefCell] because we want to mutate this while borrowing from `parsed`
-    analyzed: RefCell<BTreeMap<AbsoluteSymbolPath, Item>>,
+    analyzed: RefCell<BTreeMap<AbsoluteSymbolPath, Option<ItemOrModule>>>,
 }
 
 impl TypeChecker {
     fn new(root: ASMModule) -> Self {
         Self {
             ast: SymbolValue::Module(Module::Local(root)),
-            checked_paths: Default::default(),
             analyzed: Default::default(),
         }
     }
 
     fn check_path(&self, ctx: &AbsoluteSymbolPath) -> Result<(), Vec<String>> {
         // check if the path is already checked
-        if self.checked_paths.borrow().contains(ctx) {
+        if self.analyzed.borrow().contains_key(ctx) {
             return Ok(());
         }
+
+        // insert `None` in the map to keep track of circular dependencies
+        self.analyzed.borrow_mut().entry(ctx.clone()).or_default();
 
         let symbol = ctx.parts().fold(&self.ast, |acc, part| match acc {
             SymbolValue::Module(Module::Local(m)) => m
@@ -71,18 +82,14 @@ impl TypeChecker {
             s => s,
         });
 
-        let res = match symbol {
+        match symbol {
             SymbolValue::Module(Module::Local(module)) => self.check_module(module, ctx),
             SymbolValue::Machine(machine) => self.check_machine_type(machine.clone(), ctx),
             SymbolValue::Expression(e) => self.check_expression(e.clone(), ctx),
             SymbolValue::TypeDeclaration(d) => self.check_type_declaration(d.clone(), ctx),
             SymbolValue::TraitDeclaration(d) => self.check_trait_declaration(d.clone(), ctx),
             _ => unreachable!(),
-        };
-
-        self.checked_paths.borrow_mut().insert(ctx.clone());
-
-        res
+        }
     }
 
     fn check_machine_type(
@@ -361,6 +368,8 @@ impl TypeChecker {
             }
         }
 
+        self.insert_module(ctx);
+
         if !errors.is_empty() {
             Err(errors)
         } else {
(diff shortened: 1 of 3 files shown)


analysis/src/machine_check.rs
```diff
@@ -280,84 +334,69 @@ impl TypeChecker {
...
-                        }
+                ModuleStatement::SymbolDefinition(SymbolDefinition { name, .. }) => {
+                    if let Err(e) = self.check_path(&ctx.with_part(name)) {
+                        errors.extend(e);
                     }
                 }
                 ModuleStatement::TraitImplementation(trait_impl) => {
-                    res.insert(ctx.clone(), Item::TraitImplementation(trait_impl));
+                    // comment by Thibaut: here I'm confused why we add an implementation to the items, as it is not bound to a name (here it's bound to the name of the module). Shouldn't we rather, for example, have a single `Item::Trait` and store both the declaration and the implementations there?
+                    self.insert_item(ctx, Item::TraitImplementation(trait_impl.clone()));
```
gzanitti, line 358: I think the main difference is that implementations do not define symbols, so we keep them separate and at module level.

analysis/src/machine_check.rs
```diff
@@ -280,84 +334,69 @@ impl TypeChecker {
...
+        ctx: &AbsoluteSymbolPath,
+    ) -> Result<(), Vec<String>> {
+        // TODO: actually check the type declaration?
+        self.insert_item(ctx, Item::TypeDeclaration(enum_decl));
+        Ok(())
+    }
+
+    fn check_trait_declaration(
+        &self,
+        enum_decl: TraitDeclaration<parsed::Expression>,
```
gzanitti, line 392: This should be `trait_decl`

gzanitti (changes requested):
Looks good to me. Just a minor fix.

analysis/src/machine_check.rs
Schaeff: What if we have two implementations in the same module, doesn't it lead to a collision of the keys?

gzanitti: Yes, that's right, it will. I'm not an expert on the ASM part, but let me check how we can solve it.

Schaeff: Ok cool, it can be a different PR. With the changes I introduce in this PR, when we see `impl Foo for Bar` it would be straightforward to add it to the `Foo` `Item::Trait` or the `Bar` `Item::?`, so that we can remove the `TraitImplementation` item entirely.

chriseth: I think I would prefer to keep them separate. Currently it makes no difference in which module the impl is, but we might make that significant (like in rust). If we re-format the asm, it will be weird that the impl suddenly moves to the module of the trait.

gzanitti: I have discussed it with Chris and we believe that `ASMAnalysisFile` should include a way to handle anonymous items, but it will be included in another PR

chriseth: Also note that impls in powdr are really anonymous. There is no `impl A for B`, there is just `impl TraitName<type1, type2>{...}`. So the only sensible place to put the impls (apart from the trait) is the module where they are written.

So we could change ASMAnalysisFile to contain namedItems (the current items) and statements or directives, which are a map from String (the module name) to a vector of such statements.

Schaeff: > Also note that impls in powdr are really anonymous. There is no `impl A for B`, there is just `impl TraitName<type1, type2>{...}`. So the only sensible place to put the impls (apart from the trait) is the module where they are written.

Oh I see! However how about my other idea to put the implementations in the `Item::Trait`? Is it important where an implementation is written?

> So we could change ASMAnalysisFile to contain namedItems (the current items) and statements or directives, which are a map from String (the module name) to a vector of such statements.

Then this wouldn't need to change

No strong opinion on this of course, just asking out of curiosity!

Schaeff: Ah is it because trait implementations are just forwarded and end up in a PIL namespace?

Schaeff:
@gzanitti can we merge?

analysis/src/machine_check.rs
chriseth: If we re-print the file, I would find it weird to move the implementation to the place where the trait was defined.

Also I could imagine other items at the module level that do not have a name, but only side-effects.

analysis/src/machine_check.rs
```diff
@@ -13,31 +16,80 @@ use powdr_ast::{
...
     })
 }
 
-#[derive(Default)]
-struct TypeChecker {}
+struct TypeChecker {
+    /// The parsed AST as a single [SymbolValue::Module]
+    parsed: SymbolValue,
+    /// The already checked paths
+    checked: RefCell<HashSet<AbsoluteSymbolPath>>,
```
chriseth, line 41:
```suggestion
    checked_paths: RefCell<HashSet<AbsoluteSymbolPath>>,
```
?

analysis/src/machine_check.rs
```diff
@@ -13,31 +16,80 @@ use powdr_ast::{
...
-        items: machines.into_iter().collect(),
+        items: checker.analyzed.into_inner(),
     })
 }
 
-#[derive(Default)]
-struct TypeChecker {}
+struct TypeChecker {
+    /// The parsed AST as a single [SymbolValue::Module]
+    parsed: SymbolValue,
```
chriseth, line 39:
```suggestion
    ast: SymbolValue,
```
?

analysis/src/machine_check.rs
```diff
@@ -13,31 +16,80 @@ use powdr_ast::{
...
+    /// The already checked paths
+    checked: RefCell<HashSet<AbsoluteSymbolPath>>,
+    /// The checked items
+    /// We use [RefCell] because we want to mutate this while borrowing from `parsed`
+    analyzed: RefCell<BTreeMap<AbsoluteSymbolPath, Item>>,
```
chriseth, line 44: What is the difference between checked and the keys of analyzed?

analysis/src/machine_check.rs
```diff
@@ -13,31 +16,80 @@ use powdr_ast::{
...
+        }
+    }
+
+    fn check_path(&self, ctx: &AbsoluteSymbolPath) -> Result<(), Vec<String>> {
+        // check if the path is already checked
+        if self.checked.borrow().contains(ctx) {
+            return Ok(());
+        }
+
+        let symbol = ctx.parts().fold(&self.parsed, |acc, part| match acc {
```
chriseth, line 62: This is quite expensive if there are many statements in the module. Why don't we have a map from names to items at this point? This is done after the path canonicalizer, right?

chriseth: Ah ok machine_check is the component that creates that map. Would it make sense to create the map first instead of iterating through the ast all the time here?

analysis/src/machine_check.rs
```diff
@@ -391,6 +431,54 @@ impl TypeChecker {
             links: instruction.links,
         })
     }
+
+    /// Get the type of a path. If required, check the path first.
+    fn type_of(&self, ty: &AbsoluteSymbolPath) -> Result<Type, Vec<String>> {
```
chriseth, line 436:
```suggestion
    fn type_of(&self, path: &AbsoluteSymbolPath) -> Result<Type, Vec<String>> {
```

analysis/src/machine_check.rs
```diff
@@ -13,31 +16,80 @@ use powdr_ast::{
...
+        let res = match symbol {
+            SymbolValue::Module(Module::Local(module)) => self.check_module(module, ctx),
+            SymbolValue::Machine(machine) => self.check_machine_type(machine.clone(), ctx),
+            SymbolValue::Expression(e) => self.check_expression(e.clone(), ctx),
+            SymbolValue::TypeDeclaration(d) => self.check_type_declaration(d.clone(), ctx),
+            SymbolValue::TraitDeclaration(d) => self.check_trait_declaration(d.clone(), ctx),
+            _ => unreachable!(),
+        };
+
+        self.checked.borrow_mut().insert(ctx.clone());
```
chriseth, line 83: If an item references itself, and we only insert it into checked here, I think we might get infinite recursion.

analysis/src/machine_check.rs
Schaeff: Good point. It seems like the path canonicalizer should be the one responsible for creating the map (as it's the one making sure no key is duplicated). Let me try to implement that in a separate PR before this.

Schaeff: Done in #1725. If it's just performance, we could merge this first and the other one later, up to you.

analysis/src/machine_check.rs
Schaeff: Good point but this is already the case in airgen currently (not infinite recursion but infinite loop) for a machine which has an instance of itself as a submachine. Can you think of other examples which this PR would break? If not, I think we should have a separate PR.

analysis/src/machine_check.rs
Schaeff: Solved in #1725 by having a separate section for implementations (can eventually be extended to other non-declaration statement)

analysis/src/machine_check.rs
Schaeff: done

analysis/src/machine_check.rs
```diff
@@ -30,36 +27,50 @@ pub fn check(file: ASMProgram) -> Result<AnalysisASMFile, Vec<String>> {
...
+}
+
 struct TypeChecker {
     /// The parsed AST as a single [SymbolValue::Module]
-    parsed: SymbolValue,
-    /// The already checked paths
-    checked: RefCell<HashSet<AbsoluteSymbolPath>>,
+    ast: SymbolValue,
     /// The checked items
     /// We use [RefCell] because we want to mutate this while borrowing from `parsed`
```
chriseth, line 52:
```suggestion
    /// We use [RefCell] because we want to mutate this while borrowing from `ast`
```

analysis/src/machine_check.rs
```diff
@@ -30,36 +27,50 @@ pub fn check(file: ASMProgram) -> Result<AnalysisASMFile, Vec<String>> {
...
-    analyzed: RefCell<BTreeMap<AbsoluteSymbolPath, Item>>,
+    analyzed: RefCell<BTreeMap<AbsoluteSymbolPath, Option<ItemOrModule>>>,
```
chriseth, line 53: Nice!

Schaeff:
Closing this for now as it's too outdated but will probably recycle it into a new version

PR #1659, opened 2024-08-06
Decrease degree of constraints in poseidon_gl_memory

leonardoalt:
Depends on https://github.com/powdr-labs/powdr/pull/1599

Needed to be used with Plonky3.

### DIFF
--- a/std/machines/hash/poseidon_gl_memory.asm
+++ b/std/machines/hash/poseidon_gl_memory.asm
@@ -156,10 +156,17 @@ machine PoseidonGLMemory(mem: Memory, split_gl: SplitGL) with
     let a = array::zip(state, C, |state, C| state + C);
 
     // Compute S-Boxes (x^7)
-    let x2: inter[STATE_SIZE] = array::map(a, |a| a * a);
-    let x4: inter[STATE_SIZE] = array::map(x2, |x2| x2 * x2);
-    let x6: inter[STATE_SIZE] = array::zip(x4, x2, |x4, x2| x4 * x2);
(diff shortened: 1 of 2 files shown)


std/machines/hash/poseidon_gl_memory.asm
```diff
@@ -156,10 +156,17 @@ machine PoseidonGLMemory(mem: Memory, split_gl: SplitGL) with
     let a = array::zip(state, C, |state, C| state + C);
 
     // Compute S-Boxes (x^7)
-    let x2: inter[STATE_SIZE] = array::map(a, |a| a * a);
-    let x4: inter[STATE_SIZE] = array::map(x2, |x2| x2 * x2);
-    let x6: inter[STATE_SIZE] = array::zip(x4, x2, |x4, x2| x4 * x2);
-    let x7: inter[STATE_SIZE] = array::zip(x6, a, |x6, a| x6 * a);
+    col witness x2[STATE_SIZE];
+    array::zip(x2, array::map(a, |a| a * a), |x2, a_a| x2 = a_a);
```
pacheco, line 160: can't these be just `array::map(x2, |x2| x2 = a*a);`?

pacheco (commented):
Should the other `poseidon_gl.asm` file also be changed?

std/machines/hash/poseidon_gl_memory.asm
pacheco: hmmm doesn't work

leonardoalt:
> Should the other `poseidon_gl.asm` file also be changed?

changed it as well. I left poseidon_bn254 as is because we're not gonna use that with plonky3 anyway

georgwiese (approved):
LGTM

PR #1682, opened 2024-08-13
Baby bear field

leonardoalt:


Schaeff:
We should extract adding support for the field to a different PR which should be straightforward to merge

Schaeff:
Seems like it's [a95c12f](https://github.com/powdr-labs/powdr/pull/1682/commits/a95c12fba195438af8836be56a9c012b771d4a67) and [eacf9da](https://github.com/powdr-labs/powdr/pull/1682/commits/eacf9da8606dc010529ed2417ef155bdf40ea67a)

leonardoalt:
that's what I'm trying in this PR, it only

- adds the field
- adds some tests
- fixes some tests

but CI doesn't like it.

But ok, I'll try to make it even smaller and only add the field

number/src/baby_bear.rs
```diff
@@ -0,0 +1,467 @@
...
+)]
+pub struct BabyBearField(BabyBear);
+
+const P: u32 = 0x78000001;
+const MONTY_BITS: u32 = 32;
+const MONTY_ZERO: u32 = to_monty(0);
+const MONTY_ONE: u32 = to_monty(1);
+
+const fn to_monty(x: u32) -> u32 {
+    (((x as u64) << MONTY_BITS) % P as u64) as u32
```
chriseth, line 43: Didn't we ditch montgomery? ;)

leonardoalt: This is all copied from Plonky3, our type is just a wrapper over their type. I had to copy this directly here because `const BabyBear::new()` is private within the `p3_field` crate, as well as this function, and I needed it for the `ConstZero` and `ConstOne` traits we need to implement here.

number/src/baby_bear.rs
```diff
@@ -0,0 +1,467 @@
...
+    fn is_in_lower_half(&self) -> bool {
+        self.to_canonical_u32() <= (Self::ORDER - 1) / 2
+    }
+
+    fn known_field() -> Option<crate::KnownField> {
+        Some(KnownField::BabyBearField)
+    }
+
+    fn try_into_i32(&self) -> Option<i32> {
+        Some(self.to_canonical_u32() as i32)
```
chriseth: I don't think this is the right implementation. It should return `-1` for `bb(0) - bb(1)`.

leonardoalt: btw these files were git's merge mess, they were already merged in https://github.com/powdr-labs/powdr/pull/1696

I'll fix it in another PR

onurinanc (commented):
Both Binary8 and Binary16 look good to me! Since 2^27 * 15 + 1 < 2^32, we need those two implementation right? (It might be out of concept but, In Goldilocks, we also perform with 32-bit registers and don't need an extra implementation. Is it because the RISCV operates on 32-bits?

leonardoalt:
> Both Binary8 and Binary16 look good to me! Since 2^27 * 15 + 1 < 2^32, we need those two implementation right? (It might be out of concept but, In Goldilocks, we also perform with 32-bit registers and don't need an extra implementation. Is it because the RISCV operates on 32-bits?

Exactly, the purpose of all these machines is to be used by our RISCV machine which operates on 32 bit registers and memory. On GL we can support 1 value per FE, but here I've added two variants of BabyBear  impls which use 2 FE per RV value (as 16 bit limbs) or one FE per byte.

onurinanc (approved):

PR #1684, opened 2024-08-14
Typevar count & schema instantiation to Unifier

gzanitti:
Shifts the typevars counting and schema instantiation logic from TypeChecker to Unifier. 
This change simplifies schema instantiation so that the implementation of traits is cleaner.


pil-analyzer/src/type_inference.rs
```diff
@@ -932,21 +929,7 @@ impl TypeChecker {
     /// Returns the new type and a vector of the type variables used for those
     /// declared in the scheme.
     fn instantiate_scheme(&mut self, scheme: TypeScheme) -> (Type, Vec<Type>) {
```
chriseth, line 931: I think this can also fully move into the unifier.

chriseth (approved):

PR #1693, opened 2024-08-15
BabyBear with only Plonky3 changes

topanisto:
Adding BabyBear params to Plonky3 crate.


plonky3/src/stark.rs
```diff
@@ -126,11 +126,11 @@ impl<T: FieldElement> Plonky3Prover<T> {
         }
 
         // get the config
-        let config = get_config();
+        let config = get_config_goldilocks();
```
Schaeff, line 129: Since we are now making this generic over the field `F`, this cannot refer to a particular field (goldilocks) but rather should call a method of the `FieldElementMap` trait. Why suggests moving `get_config` to the trait.

plonky3/src/params.rs
```diff
@@ -1,98 +1,19 @@
...
-use p3_goldilocks::MdsMatrixGoldilocks;
-use p3_merkle_tree::FieldMerkleTreeMmcs;
-use p3_poseidon::Poseidon;
-use p3_symmetric::{PaddingFreeSponge, TruncatedPermutation};
-use p3_uni_stark::StarkConfig;
+pub trait FieldElementMap: FieldElement {
+    type Config: StarkGenericConfig + Sync;
 
-use rand::{distributions::Standard, Rng, SeedableRng};
+    fn to_p3_field(&self) -> Plonky3Field<Self>;
```
Schaeff, line 14:
```suggestion
    fn into_p3_field(self) -> Plonky3Field<Self>;
```

It makes sense to move `self` in this case

leonardoalt:
fixed conflicts

leonardoalt:
running p3 bb composite on two tests

plonky3/src/baby_bear.rs
```diff
@@ -0,0 +1,109 @@
...
+            .take(ROUNDS_P)
+            .collect(),
+        DiffusionMatrixBabyBear::default()
+    );
+}
+
+impl FieldElementMap for BabyBearField {
+    type Config = StarkConfig<MyPcs, FriChallenge, FriChallenger>;
+    fn to_p3_field(&self) -> Plonky3Field<Self> {
+        BabyBear::from_canonical_u32(self.to_integer().try_into_u32().unwrap())
```
Schaeff, line 80: This does not need to go through integers, as our BabyBearField type wraps a p3 babybear element. Opened #1761

plonky3/src/circuit_builder.rs
```diff
@@ -6,24 +6,22 @@
...
-};
+use std::collections::{BTreeMap, HashSet};
 
+use crate::params::{Commitment, FieldElementMap, Plonky3Field, ProverData};
 use p3_air::{Air, AirBuilder, AirBuilderWithPublicValues, BaseAir, PairBuilder};
-use p3_field::AbstractField;
-use p3_goldilocks::Goldilocks;
+// use p3_baby_bear::BabyBear;
+// use p3_field::{extension::ComplexExtendable, AbstractField, PrimeField};
+// use p3_goldilocks::Goldilocks;
```
Schaeff, line 15: Remove

plonky3/src/circuit_builder.rs
```diff
@@ -6,24 +6,22 @@
...
 use p3_matrix::{dense::RowMajorMatrix, Matrix};
+// use p3_mds::MdsPermutation;
```
Schaeff, line 17: Remove

Schaeff (changes requested):

plonky3/src/baby_bear.rs
leonardoalt: fixed in the latest commit

Schaeff:
I think `into_inner` would be cleaner on the BabyBearField struct than `pub` but not a huge deal

Schaeff (approved):

PR #1694, opened 2024-08-15
Depreciation of `.` by `::` in paths

gzanitti:


airgen/src/lib.rs
```diff
@@ -341,7 +341,7 @@ impl<'a> ASMPILConverter<'a> {
             .find(|o| o.name == callable)
             .unwrap_or_else(|| {
                 panic!(
-                    "function/operation not found: {}.{}",
+                    "function/operation not found: {}::{}",
```
chriseth, line 344: Please double-check, but I think this should stay as `.`, because it accesses the member of an instance (a value) and not a type or module element.

ast/src/parsed/asm.rs
```diff
@@ -162,11 +162,9 @@ impl SymbolPath {
         self
     }
 
-    /// Formats the path and uses `.` as separator if
-    /// there are at most two components.
+    /// Formats the path using `::` as separator
```
chriseth, line 165: I think this can just be the Display trait now.

ast/src/parsed/asm.rs
```diff
@@ -362,11 +356,9 @@ impl AbsoluteSymbolPath {
         Self { parts }
     }
 
-    /// Formats the path without leading `::` and uses `.` as separator if
-    /// there are at most two components.
+    /// Formats the path without leading `::` and uses `::` as separator
     pub fn to_dotted_string(&self) -> String {
-        let separator = if self.parts.len() <= 2 { "." } else { "::" };
-        self.parts.join(separator)
+        self.parts.join("::")
```
chriseth, line 361: Same here, should just be the Display trait.

analysis/README.md
```diff
@@ -33,9 +33,9 @@ machine Main with degree: 16 {
     reg Y[<=];
     reg A;
 
-    instr identity X -> Y = sub.identity;
-    instr one -> Y = sub.one;
-    instr nothing = sub.nothing;
+    instr identity X -> Y = sub::identity;
```
chriseth, line 36: Same here: `sub` is a value, so `sub.identity` is a member access and not a path access.

ast/src/analyzed/display.rs
```diff
@@ -454,16 +454,7 @@ impl Display for PolynomialReference {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
         if let Some(type_args) = &self.type_args {
```
chriseth, line 455: Can now be simplified to:
```
write!(f, "{}", self.name)?;
if has non-empty type args {
  write!(f, "::{}", format_type_args(...))?;
}
Ok(())
```

Schaeff:
There are at least two `.asm` files which can be simplified with this, since they currently use a deeper module tree to avoid `.`-separated paths:
`test_data/asm/vm_args_relative_path.asm` and `importer/test_data/instruction.asm`

analysis/README.md
gzanitti: Oh, I see, I thought we stopped using `.` in general. I'll fix it. Thanks!

ast/src/parsed/display.rs
```diff
@@ -192,7 +192,7 @@ impl Display for CallableRef {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
         write!(
             f,
-            "{}{}.{}({})",
+            "{}{}::{}({})",
```
chriseth, line 195: I think this is also a member access.

backend/src/composite/split.rs
```diff
@@ -63,7 +63,7 @@ pub(crate) fn machine_witness_columns<F: FieldElement>(
                 panic!("Machine {machine_name} has witness columns of different sizes")
             }
         });
-    let dummy_column_name = format!("{machine_name}.{DUMMY_COLUMN_NAME}");
+    let dummy_column_name = format!("{machine_name}::{DUMMY_COLUMN_NAME}");
```
chriseth, line 66: Also a value.

backend/src/estark/json_exporter/mod.rs
```diff
@@ -197,7 +197,7 @@ fn fixup_name(name: &str) -> String {
     if name.contains('.') {
```
chriseth, line 197: This cannot be the case any more, but I think the part you changed should remain as it is, shouldn't it? It replaces the last `::` by a `.`.

parser/src/lib.rs
```diff
@@ -485,12 +485,12 @@ namespace N(2);
     fn parse_impl() {
         let input = r#"
     impl<T> Iterator<ArrayIterator<T>, T> {
-        next: |it| if it.pos >= 7 { (it, none) } else { some((increment(it), it.arr[it.pos])) },
```
chriseth, line 488: Is this merged to main? It uses member access notation we don't support yet. Also I think this should stay as `.`.

backend/src/estark/json_exporter/mod.rs
gzanitti: It was not being replaced correctly. With the original code, `a::b::c` becomes `a::b.:c`, with this change, it now returns `a::b.c` which I understand is the correct version.

parser/src/lib.rs
gzanitti: Yes, but it was just a test on the parser. Anyway, I fixed it.

pipeline/tests/asm.rs
```diff
@@ -146,19 +146,19 @@ fn block_to_block_with_bus_monolithic() {
...
-    //   The shared challenges are used during witness generation.
-    // - `CompositeBackend::verify` simply verifies each machine proof independently, using the local
-    //   challenges. As a result, the challenges during verification differ and the constraints are
-    //   not satisfied.
-    let f = "asm/block_to_block_with_bus.asm";
-    let pipeline = make_simple_prepared_pipeline(f);
-    test_halo2_with_backend_variant(pipeline.clone(), BackendVariant::Composite);
-}
+// #[test]
+// #[should_panic = "called `Result::unwrap()` on an `Err` value: [\"Circuit was not satisfied\"]"]
```
gzanitti, line 150: I'm not sure why, but I had to comment this because the panic is no longer reached.

Schaeff: I had the same issue on my branch and it seems like pulling main solved it

backend/src/composite/split.rs
gzanitti: But isn't it forming the column namespace?

pipeline/tests/asm.rs
gzanitti: Thanks!

backend/src/composite/split.rs
chriseth: ah right, machine name is a namespace here not a powdr machine name!

ast/src/parsed/asm.rs
```diff
@@ -362,11 +351,9 @@ impl AbsoluteSymbolPath {
         Self { parts }
     }
 
-    /// Formats the path without leading `::` and uses `.` as separator if
-    /// there are at most two components.
-    pub fn to_dotted_string(&self) -> String {
-        let separator = if self.parts.len() <= 2 { "." } else { "::" };
-        self.parts.join(separator)
+    // Returns the path as a string without the leading "::".
+    pub fn to_relative_string(&self) -> String {
```
chriseth, line 355: I'm not 100% convinced by this. What is the string relative to?
In some parts of the cod, we use `.relative_to(&Default::default())` and then format the result. This is the same as this function.

What if we implement `.relative_to_root()`, which is much simpler than `relative_to`. It would return a SymbolPath which can then be formatted. If you implement that, you should also change all other places where we use `relative_to(&Default::default())`.

ast/src/parsed/asm.rs
```diff
@@ -362,11 +351,9 @@ impl AbsoluteSymbolPath {
         Self { parts }
     }
 
-    /// Formats the path without leading `::` and uses `.` as separator if
-    /// there are at most two components.
-    pub fn to_dotted_string(&self) -> String {
-        let separator = if self.parts.len() <= 2 { "." } else { "::" };
-        self.parts.join(separator)
+    // Returns the path as a string without the leading "::".
```
chriseth, line 354: triple slash

chriseth (approved):

PR #1696, opened 2024-08-16
baby bear field only

leonardoalt:
Most minimal attempt at adding BabyBear, only adds the field and field tests, but no powdr tests using the field.

### DIFF
--- a/number/src/baby_bear.rs
+++ b/number/src/baby_bear.rs
@@ -32,7 +32,7 @@ use p3_field::{AbstractField, Field, PrimeField32};
     Deserialize,
     derive_more::Display,
 )]
-pub struct BabyBearField(pub BabyBear);
+pub struct BabyBearField(BabyBear);
 
 const P: u32 = 0x78000001;
 const MONTY_BITS: u32 = 32;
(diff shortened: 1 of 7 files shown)


number/src/baby_bear.rs
```diff
@@ -0,0 +1,467 @@
...
+    Eq,
+    Hash,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Serialize,
+    Deserialize,
+    derive_more::Display,
+)]
+pub struct BabyBearField(pub BabyBear);
```
Schaeff, line 35:
```suggestion
pub struct BabyBearField(BabyBear);
```

Schaeff (changes requested):

number/src/baby_bear.rs
leonardoalt: done

Schaeff (approved):

number/src/baby_bear.rs
```diff
@@ -0,0 +1,467 @@
...
+        let modulus = Self::modulus();
+        if n >= modulus.to_arbitrary_integer() {
+            Err(format!("Decimal number \"{s}\" too large for field."))
+        } else {
+            Ok(n.into())
+        }
+    }
+}
+
+impl Neg for BabyBearField {
```
chriseth, line 214: Can use derive_more for all of these (see goldilocks for an example)

Issue #1699, opened 2024-08-18
Explicit instantiation of machines


Schaeff:
Currently we interpret the machine type called `Main` as the entry point. We also allow reusing instances by passing them as arguments to submachines.

So for example:

```rust
use machines::{Foo, Bar};

machine Main {
    foo Foo;
    bar Bar(foo);
}
```

instead, switch to explicit instantiation by expecting:

```
use machines::{Foo, Bar};

machine Main {
   Foo foo;
   Bar bar;
}

let foo = Foo {};
let bar = Bar { foo };
let main = Main { foo, bar };
```

and interpret the machine instance named `main` as the entry point.
for backward compatibility, if none is found, add a default which just instantiates everything as above.

It would help a lot to rely on the parser change for structs, and intercept struct expressions which correspond to machines in machine check. If we don't find a machine, leave it as a struct and it will be checked later in PIL.

PR #1705, opened 2024-08-19
Explicit machine instantiation

Schaeff:
Partly addresses #1699 by using explicit instantiation behind the scenes.


airgen/Cargo.toml
```diff
@@ -11,10 +11,14 @@ repository = { workspace = true }
 powdr-ast.workspace = true
 powdr-number.workspace = true
 powdr-analysis.workspace = true
-
 log = "0.4.17"
-
 itertools = "0.13"
+topo_sort = "0.4.0"
```
chriseth, line 16: The source repository was alive for three weeks, has a single contributor and uses unsafe features. Zero changes, pull requests or issue answers for three years.

chriseth: call_graph.rs contains a topo sort alg, but uses recursion.

Schaeff: fair, I'll make the `call_graph` one generic and use it here as well

Schaeff: done

airgen/src/lib.rs
```diff
@@ -20,71 +21,205 @@ use itertools::Itertools;
...
+        .iter()
+        .zip(args)
+        .map(|(param, argument)| {
+            let ty = AbsoluteSymbolPath::default().join(param.ty.clone().unwrap());
+            match input.items.get(&ty) {
+                Some(Item::Machine(_)) => MachineInstance {
+                    ty,
+                    value: MachineInstanceExpression::Reference(argument.clone()),
+                },
+                _ => unimplemented!(),
```
pacheco, line 120: is this unreachable instead because of a previous check (i don't recall)? this is a reference to a non-existent machine type right?

airgen/src/lib.rs
```diff
@@ -20,71 +21,205 @@ use itertools::Itertools;
...
+#[derive(Default, Debug)]
+struct Instances {
+    abs_to_loc: BTreeMap<AbsoluteSymbolPath, Location>,
+    map: BTreeMap<Location, Instance>,
+}
+
+impl Instances {
+    /// Trim to remove instances unreachable from a given Location
+    /// Returns the instance map
+    /// untested!
```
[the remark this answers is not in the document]
Schaeff: I removed this in the end, it can be done somewhere else

utils/src/lib.rs
```diff
@@ -0,0 +1,92 @@
+use std::{
+    collections::{HashMap, HashSet},
+    hash::Hash,
+};
+
+/// Take a set of nodes and a function returning the dependencies of any node and return a topological order where any node always comes after its dependencies
+/// Circular dependencies are returned in an arbitrary order
+pub fn topo_sort<'a, E, F, I>(nodes: I, get_dependencies: F) -> Vec<&'a E>
+where
+    F: Fn(&E) -> HashSet<&'a E>,
```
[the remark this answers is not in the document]
Schaeff: It does simplify this PR :)

airgen/src/lib.rs
```diff
@@ -20,71 +21,205 @@ use itertools::Itertools;
...
+    fn trim(self, from: &Location) -> BTreeMap<Location, Instance> {
+        let mut reached: BTreeSet<Location> = Default::default();
+        let mut queue = vec![from];
+        let mut map = self.map;
+        while let Some(location) = queue.pop() {
+            // `location` is reached
+            reached.insert(location.clone());
```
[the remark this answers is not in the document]
Schaeff: removed

airgen/src/lib.rs
```diff
@@ -20,71 +21,205 @@ use itertools::Itertools;
...
+struct Instances {
+    abs_to_loc: BTreeMap<AbsoluteSymbolPath, Location>,
+    map: BTreeMap<Location, Instance>,
+}
+
+impl Instances {
+    /// Trim to remove instances unreachable from a given Location
+    /// Returns the instance map
+    /// untested!
+    fn trim(self, from: &Location) -> BTreeMap<Location, Instance> {
```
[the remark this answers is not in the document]
Schaeff: removed

airgen/src/lib.rs
Schaeff: Do we expect any other types of parameters here? My understanding is that we only ever get machines as arguments.
Now assuming we get a machine, I'm not sure we currently enforce that this machine must be defined somewhere. This is something we need to do in `machine_check`.

I'm removing this check here entirely, as we make this assumption in a many places in this code, for example in `fold_instance`.

PR #1706, opened 2024-08-19
Backend artifact

pacheco:
Allow the pipeline to setup and cache the backend before calls to prove/verify.
`Backend` objects are not cloneable, so upon cloning the pipeline, the instance is cleared (but will be regenerated when needed).


georgwiese (approved):
Nice!

Issue #1691, opened 2024-08-15
Implement extension field of degree 4 (Fp4) for BabyBear Field


onurinanc:
To make the permutation and lookup arguments implemented in PIL sound enough for the BabyBear Field, we need to implement Fp4 for the BabyBear field

PR #1707, opened 2024-08-20
Implement Fp4 for BabyBear Field in PIL

onurinanc:
Solves the issue in #1691

You can run the test using `--field bb`


georgwiese (commented):
Cool! I think similar to `fp2`, we could do this in a way that it's not field-specific!

std/math/fp4_bb.asm
```diff
@@ -0,0 +1,200 @@
...
+
+/// Converts a base field element to the extension field
+let<T: FromLiteral> from_base: T -> Fp4<T> = |x| Fp4::Fp4(x, 0, 0, 0);
+
+/// Addition for extension field
+let<T: Add> add_ext: Fp4<T>, Fp4<T> -> Fp4<T> = |a, b| match (a, b) {
+    (Fp4::Fp4(a0, a1, a2, a3), Fp4::Fp4(b0, b1, b2, b3)) => Fp4::Fp4(
+        a0 + b0,
+        a1 + b1,
+        a2 + b2,
```
ShuangWu121, line 30: just a question, how do we make these addition operation "+" calculated in the BabyBear field? I mean, the result of the sums like a0+b0 ...  are modules babybear prime?

onurinanc: When I run the program with this command `cargo run -r -bin powdr pil test_data/std/fp4_bb.asm --field bb -f`, I can specify the field element using `--field` flag. (we currently have bb, gl, bn254). So, the operations are done in the bb

std/math/fp4_bb.asm
```diff
@@ -0,0 +1,203 @@
...
+use std::convert::expr;
+use std::prover::eval;
+
+/// BETA = 11
+/// NBETA = 2013265910
+
+let BABYBEAR_PRIME = 0x78000001;
+
+/// An element of the extension field over the BabyBear Field
+/// relative to the irreducible polynomial X^4 + 11,
```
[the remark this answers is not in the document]
onurinanc: I double-checked using Sage, and both $X^4 + 11$ and $X^4 - 11$ works for all the fields that we currently support.

I think we can keep the same instead of using $X^4 - 11$, cause there will be no efficiency difference, we can keep $X^4 = -11$ instead of $X^4 = 11$ since in the `inv_ext` we're doing a similar approach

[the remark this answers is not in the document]
onurinanc: But, we have `2013265910`make the things not-field specific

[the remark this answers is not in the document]
onurinanc: Okay, I'm going to try to make it field-specific rn

pipeline/tests/powdr_std.rs
```diff
@@ -388,13 +388,14 @@ mod reparse {
...
     /// run on Goldilocks.
-    const BLACKLIST: [&str; 6] = [
+    const BLACKLIST: [&str; 7] = [
         "std/bus_permutation_via_challenges.asm",
         "std/permutation_via_challenges.asm",
         "std/lookup_via_challenges.asm",
         "std/poseidon_bn254_test.asm",
         "std/split_bn254_test.asm",
         "std/bus_lookup_via_challenges.asm",
+        "std/fp4.asm",
```
georgwiese, line 398: I don't think this is necessary!

test_data/std/fp4.asm
```diff
@@ -0,0 +1,8 @@
+use std::math::fp4::test;
+
+machine Main {
+    test::add();
+    test::sub();
+    test::mul();
+    test::inverse()[0];
+}
```
georgwiese, line 8: Instead of this, we should have the analog of [this](https://github.com/powdr-labs/powdr/blob/3d4fedfcc47f30a85b6b4ef8f9dfeeaf004c1e3e/pipeline/tests/powdr_std.rs#L318), no? Then we can test that it works on all 3 fields as well.

std/math/fp2.asm
```diff
@@ -5,13 +5,24 @@ use std::check::panic;
...
+/// BN254 = 21888242871839275222246405745257275088548364400416034343698204186575808495617
+/// GL = 0xffffffff00000001
+/// F = GF(GL)
+/// R.<x> = PolynomialRing(F)
+/// f = x^2 - 7
+/// f.is_irreducible()
+
 /// An element of the extension field over the implied base field (which has to be either
 /// the Goldilocks or the BN254 field) relative to the irreducible polynomial X^2 - 7,
+/// (This irreducible polynomial also works for Mersenne31)
```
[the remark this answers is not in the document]
onurinanc: Cool! Opened the issue for the new PR: https://github.com/powdr-labs/powdr/issues/1714

std/math/fp4.asm
```diff
@@ -0,0 +1,197 @@
...
+};
+
+/// Inversion for an Fp4 element
+// The inverse of (a0, a1, a2, a3) is a point (b0, b1, b2, b3) such that:
+/// (a0 + a1 * x + a2 * x^2 + a3 * x^3) (b0 + b1 * x + b2 * x^2 + b3 * x^3) = 1 (mod x^4 + 11)
+/// Multiplying out and plugging in x^4 = -11 yields the following result
+/// a[0] * b0 + BETA * a[2] * b2
+/// + (-a[1] * b0 + NBETA * a[3] * b2) * X
+/// + (-a[0] * b2 + a[2] * b0) * X^2
+/// + (a[1] * b2 - a[3] * b0) * X^3
```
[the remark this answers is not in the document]
onurinanc: We're not directly solving a system of linear equations, instead, we're using a little trick as in the complex numbers to find the inverse more easily.

Let's say we want to find the inverse  (let's call it z)
$z = 1/a$
Multiply lhs and rhs part of the equation with a' = $a_0$ - $a_1 * x$ + $a_2 * x^2$ - $a_3*x^3$ (negating the 1st and 

Now, we have:
$z = a' / (a' * a)$

The multiplication is $b = (a' * a)$. In this calculation, we are just going to have b0 and b2 of b, which are the following:

`let b0 = a0 * a0 - 11 * (a1 * (a3 + a3) - a2 * a2);`
`let b2 = a0 * (a2 + a2) - a1 * a1 - 11 * (a3 * a3);`

Now, we have:

$z = a' / b$ and mutliply each side with b' = $b_0$ - $b_2^2$ * $x^2$

Now, we have:

$z = a' * b'/ b * b'$

Let's call $c = b * b'$ = $b_0^2$ + $b_2^2 * BETA$

Now, we have:

$z = a' * b' * inverse(c)$

After calculating this, we have z is equal to:

$(a_0 * b_0 - 11 * a_2 * b_2) + (-1 * a_1 * b_0 + 11 * a_3 * b_2) * x + (-1 * a_0 * b_2 + a_2 * b_0) * x^2 + (a_1 * b_2 - a_3 * b_0) * x^3$

pipeline/tests/powdr_std.rs
onurinanc: I got a reparsing error in the tests, and added this one

test_data/std/fp4.asm
onurinanc: Also, fixed the documentation

pipeline/tests/powdr_std.rs
Schaeff: What was the error?

std/math/fp4.asm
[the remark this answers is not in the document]
georgwiese: OK, cool! Didn't know of this, but seems to work and we're testing it (and it's only used in hints, not to generate constraints).

pipeline/tests/powdr_std.rs
onurinanc: - I removed this one from the blacklist, and it seems it's not necessary
- I double-checked and it is a std::reparse error (something was wrong related to addition, i supposed it is about the extension field) but, not related to this blacklist.

georgwiese (approved):
Cool!

PR #1711, opened 2024-08-21
Set hint to existing witness in an array

qwang98:
See witgen output in arith_test.txt, which has lots of `?` values, so I'm not sure if it worked as intended.


chriseth:
This should be included in https://github.com/powdr-labs/powdr/pull/1645 already.

chriseth:
Closing because this does not add anything new.

PR #1716, opened 2024-08-22
Baby Bear params for benchmarking

topanisto:
Stark and params in the Plonky3 backend for Baby Bear after trouble with implementing generic types.


Schaeff:
Closing as we figured out a way with generic types

PR #1717, opened 2024-08-22
Prover functions

chriseth:


executor/src/witgen/processor.rs
```diff
@@ -109,6 +114,25 @@ impl<'a, 'b, 'c, T: FieldElement, Q: QueryCallback<T>> Processor<'a, 'b, 'c, T,
...
             .collect();
+        // TODO we should create an error if a prover function cannot be mapped
+        // to any processor. So we should probably do this in the machine xtractor.
+        let prover_functions: Vec<_> = fixed_data
+            .analyzed
+            .prover_functions
+            .iter()
+            .filter_map(|f| {
+                f.all_children()
+                    .filter_map(|e| match e {
```
chriseth, line 125: Use the ReferencedSymbols trait instead

test_data/asm/sqrt.asm
```diff
@@ -28,8 +28,9 @@ machine Sqrt with
             sqrt_rec((y + x / y) / 2, x)
         };
 
-    col witness y(i) query Query::Hint(sqrt_hint(std::prover::eval(x)));
-    
+    col witness y;
+    query |i| std::prover::provide_value(y, i, sqrt_hint(std::prover::eval(x)));
```
chriseth, line 32: Somehow this eval call fails even though the one above succeeds...

pil-analyzer/tests/condenser.rs
```diff
@@ -471,3 +484,158 @@ fn intermediate_arr_wrong_length() {
...
+    let provide_value = 9;
+namespace N(16);
+    col witness x;
+    col witness y;
+    N::x = N::y;
+    (query |i| {
+        std::prover::provide_value(N::x, i, std::convert::fe::<int>(i % 2));
+        std::prover::provide_value(N::y, i, std::convert::fe::<int>(i % 2));
+
+    });
```
qwang98, line 586: I appears to me that `set_hint` and `provide_value` works differently in that `set_hint` "captures" the context and moves it into the lambda whereas `provide_value` directly uses context without moving it into the lambda. Is it a deliberate design difference? Are we removing `set_hint` eventually?

pil-analyzer/tests/condenser.rs
```diff
@@ -471,3 +484,158 @@ fn intermediate_arr_wrong_length() {
...
+        let y = N::y;
+        (query |i| {
+            std::prover::provide_value(x, i, std::convert::fe::<int>(i % 2));
+            std::prover::provide_value(y, i, std::convert::fe::<int>(i % 2));
+
+        })
+    };
+"#;
+    assert_eq!(analyzed.to_string(), expected);
+}
```
qwang98, line 641: What does "dynamic" mean here? What if there are `x` and `y` already created outside `gen()`, will `gen()` dynamically generated two different column names?

pil-analyzer/tests/condenser.rs
chriseth: Yes, the idea is to replace `set_hint` by query functions that can provide hints to multiple columns at the same time. This is useful since often, multiple columns depend on the same intermediate value being computed. Without that, we would have to re-compute the intermediate value multiple times.

pil-analyzer/tests/condenser.rs
chriseth: Dynamic means that in line 600, a column with the proposed name `x` is created each time the funciton `gen` is called. The name is disambiguated by adding a suitable prefix if the name is already taken.

pil-analyzer/tests/condenser.rs
qwang98: Makes sense thanks :)

pil-analyzer/tests/condenser.rs
qwang98: Alrighty. Thanks

std/machines/arith.asm
```diff
@@ -132,66 +132,38 @@ machine Arith with
...
-        _ => Query::None
+    col witness y1[16];
+    col witness x2[16];
+    query |i| {
+        match is_ec_operation() {
+            0 => {
+                match std::prover::try_eval(y1[0]) {
+                    Option::Some(_) => {
+                        // y1 is an input, in this case we do not need a hint.
+                    },
```
qwang98, line 143: This match block is new to this PR. Did you add this because when `y1` is already provided, `provide_value` will error out? When the operation is `affine_256`, `is_ec_operation = 0` but `y1` is provided as an input, so these hints are only used for `mod_256`.

test_data/asm/sqrt.asm
qwang98, line 32: Are both syntaxes available in the future, i.e. defining `y(i)` as a fixed column via query function vs declaring `y` as a column and `provide_value` to it?

Can you do `col fixed y` and `query |i| provide_value(i)`, i.e. does `provide_value` work for fixed columns as well?

pil-analyzer/tests/condenser.rs
```diff
@@ -597,3 +597,85 @@ namespace N(16);
...
+namespace std::prover;
+    let provide_value = 9;
+namespace N(16);
+    col witness x;
+    col witness y;
+    N::x = N::y;
+    query |i| {
+        std::prover::provide_value(N::x, i, std::convert::fe::<int>(i % 2));
+        std::prover::provide_value(N::y, i, std::convert::fe::<int>(i % 2));
+    };
```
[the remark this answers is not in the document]
chriseth: It tests that query functions are valid at statement level and actually get processed through the condenser and re-printed.

std/machines/arith.asm
chriseth: Yes, that is the reason. This is just a first step, the full change of arith.asm is in https://github.com/powdr-labs/powdr/pull/1760

test_data/asm/sqrt.asm
chriseth: I wanted to deprecate the old syntax because it is weird and requires the std::prelude::Query enum. Calling provide_value for fixed columns should result in a runtime error.

executor/src/witgen/machines/machine_extractor.rs
```diff
@@ -77,26 +101,45 @@ pub fn split_out_machines<'a, T: FieldElement>(
             .collect::<BTreeMap<_, _>>();
         assert!(connecting_identities.contains_key(&id.id));
 
+        let prover_functions = prover_functions
+            .iter()
+            .copied()
+            .enumerate()
```
Schaeff, line 107: unused

executor/src/witgen/machines/machine_extractor.rs
```diff
@@ -77,26 +101,45 @@ pub fn split_out_machines<'a, T: FieldElement>(
...
+                .format("\n"),
+            prover_functions
+                .iter()
+                .map(|(_, pf)| format!("{pf}"))
+                .format("\n")
         );
 
+        for (i, pf) in &prover_functions {
+            if !extracted_prover_functions.insert(*i) {
+                log::warn!("Prover function was assigned to multiple machines:\n{pf}");
```
Schaeff, line 139: Does this happen when a prover function references witnesses of two different machines?

chriseth: Yes

executor/src/witgen/query_processor.rs
```diff
@@ -30,6 +32,44 @@ impl<'a, 'b, T: FieldElement, QueryCallback: super::QueryCallback<T>>
...
+            rows.current_row_index,
+        ))))];
+
+        let mut symbols = Symbols {
+            fixed_data: self.fixed_data,
+            rows,
+            size: self.size,
+            updates: Constraints::new(),
+        };
+        let res = match evaluator::evaluate(fun, &mut symbols)
```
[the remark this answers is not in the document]
chriseth: This is not really expensive, it essentially creates a Value::Closure that references the body of the lambda, not much more.

pil-analyzer/src/type_inference.rs
```diff
@@ -43,6 +43,8 @@ pub struct ExpectedType {
     pub allow_array: bool,
     /// If true, the empty tuple is also allowed.
     pub allow_empty: bool,
+    /// If "int -> ()" is allowed.
+    pub allow_int_to_empty_fun: bool,
```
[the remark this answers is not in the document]
Schaeff: @chriseth there is still this outdated comment ☝️

Schaeff (approved):

PR #1718, opened 2024-08-23
Remove reset instruction

Schaeff:
Remove the `_reset` instruction in the dispatcher.
When turning asm into pil, we have a dispatcher which first resets all registers and jumps to the correct part of the ROM. As a result, the write registers are unconstrained on the first row.

After this PR, registers are reset after calling `return`, just like for the `pc`. This removes the need for the `_reset` instruction, as long as we constrain the write registers to be 0 on the first row.

Here are the tradeoffs of this change:
- pros
   - simplify the dispatcher, as now the pc and write registers are treated more similarly and we remove an instruction
   - fully constrain the write registers, including on the first row (note: this could also be achieved by constraining them to an arbitrary value on the first row)
   - save a main column `instr__reset`
   - save a ROM column `p_instr__reset` (negligible)
   - save a ROM row (negligible)
- cons
   - adds a witness column per write register: because of the initialisation to 0 on the first row, the register update is now higher degree. Maybe this could be an intermediate column?


Schaeff:
I think the benchmark alert is a false positive because #2264 is out of sync with main. The regression was introduced in https://github.com/powdr-labs/powdr/pull/2242/commits/8dd0a00c1fd15c5748a759ed08385f9557feff02

leonardoalt (dismissed):
lgtm, but maybe you want someone else to also take a look?

leonardoalt:
has a bunch of conflicts now =/

lvella:
Did you base this PR off some other PR that was merged? Then you need to merge `main` back into the branch of that base PR, then merge that branch into the branch of this PR, and all the conflicts and spurious commits will magically disappear...

leonardoalt:
@Schaeff do we still want this soon? Maybe?

leonardoalt:
Closing for now, branch still exists.

PR #1723, opened 2024-08-26
Remove poly_id property.

chriseth:


backend/src/estark/json_exporter/mod.rs
```diff
@@ -65,12 +65,12 @@ pub fn export<T: FieldElement>(analyzed: &Analyzed<T>) -> PIL {
             }
             StatementIdentifier::PublicDeclaration(name) => {
                 let pub_def = &analyzed.public_declarations[name];
-                let pub_ref = &pub_def.polynomial;
-                let poly_id = pub_ref.poly_id.unwrap();
+                let symbol = &analyzed.definitions[&pub_def.polynomial.name].0;
+                assert_eq!(symbol.kind, SymbolKind::Poly(PolynomialType::Committed));
```
Schaeff, line 69: In practice I agree but in theory does a public value have to expose a committed polynomial? Fixed polynomials seem fine too. Or do we have a cleaner error constraining this somewhere else?

pilopt/src/lib.rs
```diff
@@ -441,55 +437,59 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
+            Reference::Poly(PolynomialReference { name, type_args: _ }),
         ) = e
         {
-            if let Some(value) = substitutions.get(poly_id) {
+            if let Some(value) = substitutions_by_name.get(name) {
                 *e = Number {
-                    value: value.clone(),
-                    type_: Some(Type::Fe),
+                    value: (*value).clone(),
+                    type_: Some(Type::Expr),
```
Schaeff, line 475: Why did this type change?

pilopt/src/lib.rs
```diff
@@ -441,55 +437,59 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
+                *e = AlgebraicExpression::Number(T::checked_from((*value).clone()).unwrap());
             }
         }
     });
 }
 
 fn constrained_to_constant<T: FieldElement>(
     expr: &AlgebraicExpression<T>,
-) -> Option<(PolyID, BigUint)> {
+) -> Option<(String, PolyID, BigUint)> {
```
Schaeff, line 492: It was not clear to me that this is still a key and a value.
```suggestion
) -> Option<((String, PolyID), BigUint)> {
```
Or introduce a `SubstitutionKey` struct

pilopt/src/lib.rs
chriseth: A test failure made me change this. I think it was always wrong but just not detected.

PR #1725, opened 2024-08-26
Better parsed AST

Schaeff:
Changes:
- Introduce two versions of parsed ast
   - a `non_unique` one where duplicate names in modules are allowed. It is output by the parser.
   - a `unique` one where duplicate names are not allowed. It is output by the importer. 
- Store trait implementations separately from symbols

The first item solves [the issue](https://github.com/powdr-labs/powdr/pull/1657#discussion_r1731010638) in #1657 where we do a lot of linear searches inside modules, since the `unique` version uses a map instead of a vector.

The second item solves the [existing issue](https://github.com/powdr-labs/powdr/pull/1657#discussion_r1720013776) where trait implementations are treated as symbol declarations.

### DIFF
--- a/ast/src/parsed/asm/generic.rs
+++ b/ast/src/parsed/asm/generic.rs
@@ -14,59 +14,82 @@ use serde::{Deserialize, Serialize};
 
 use crate::parsed::{BinaryOperation, BinaryOperator};
 
-use super::{
+use crate::parsed::{
     visitor::Children, EnumDeclaration, EnumVariant, Expression, PilStatement, SourceReference,
     TraitDeclaration, TraitImplementation, TypedExpression,
 };
 
 #[derive(Default, Clone, Debug, PartialEq, Eq)]
-pub struct ASMProgram {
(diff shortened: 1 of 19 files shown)


ast/src/parsed/asm/generic.rs
```diff
@@ -14,59 +14,85 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, PartialEq, Eq)]
+pub struct SymbolDefinitionMut<'a, S> {
+    pub name: &'a str,
+    pub value: &'a mut SymbolValue<S>,
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, From)]
-pub enum ModuleStatement {
-    SymbolDefinition(SymbolDefinition),
+pub enum ModuleStatement<S> {
```
Schaeff, line 65: This is only used as a helper in the parser, see `ASMModule::new`

Schaeff:
#1736

PR #1731, opened 2024-08-28
Unifying implementation of fake random number generator.

lvella:
There are two interfaces we must provide: the one from `getrandom` crate and the one used internally by `std`.

### DIFF
--- a/riscv-runtime/src/entropy_source.rs
+++ b/riscv-runtime/src/entropy_source.rs
@@ -14,5 +14,10 @@ pub(crate) fn getrandom(s: &mut [u8]) {
 
 #[cfg(not(feature = "allow_fake_rand"))]
 pub(crate) fn getrandom(_: &mut [u8]) {
-    panic!("there is no real entropy source in Powdr");
+    panic!(
+        r#"There is no real entropy source in Powdr.
+You may enable, at your own risk, the "allow_fake_rand" feature of
+"powdr-riscv-runtime" crate to get a deterministic value instead
+of this panic."#
+    );
 }
--- a/riscv-runtime/src/entropy_source.rs
+++ b/riscv-runtime/src/entropy_source.rs
@@ -1,0 +1,12 @@
+/// This is a placeholder to pretend to provide a random number generator, for
+/// places like the hash function of HashMap who needs something.
+///
+/// This could be improved by using the rand_chacha crate, but I think it is
+/// worse, as it will just mask the fact that we are not providing a real
+/// entropy source.
+///
+/// TODO: figure how to be truly random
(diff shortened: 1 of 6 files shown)


riscv-runtime/src/std_support.rs
```diff
@@ -14,30 +14,36 @@ use powdr_riscv_syscalls::Syscall;
...
 }
-
 register_custom_getrandom!(powdr_getrandom);
 
+/// This is a placeholder to pretend to provide a random number generator, for
+/// places like the hash function of HashMap who needs something.
+///
+/// This could be improved by using the rand_chacha crate, but I think it is
+/// worse, as it will just mask the fact that we are not providing a real
+/// entropy source.
```
pacheco, line 39: does it make sense to borrow from risc0's approach, with the disclaimer + feature flag or panic otherwise?

riscv-runtime/src/std_support.rs
```diff
@@ -14,30 +14,36 @@ use powdr_riscv_syscalls::Syscall;
...
     }
 }
 
-fn powdr_getrandom(s: &mut [u8]) -> Result<(), Error> {
-    // TODO: getrandom expects this u8 slice as input, so we're not using sys_rand.
-    // Both should probably be fixed/unified later once we have a proper RNG.
-    const VALUE: u8 = 3;
-    s.iter_mut().for_each(|v| *v = VALUE);
-
+/// The de-factor standard rust interface to low level random number generation.
```
pacheco, line 27: de facto
but i think this comment belongs to `sys_rand`?

lvella: No, sys_rand is risc0 invention.

getrandom is the defacto standard, it is even used inside std in some targets

riscv-runtime/src/std_support.rs
lvella: yes, let me do that

riscv-runtime/src/entropy_source.rs
```diff
@@ -0,0 +1,18 @@
...
+/// TODO: figure how to be truly random
+#[cfg(feature = "allow_fake_rand")]
+pub(crate) fn getrandom(s: &mut [u8]) {
+    const VALUE: u8 = 3;
+    s.iter_mut().for_each(|v| *v = VALUE);
+}
+
+#[cfg(not(feature = "allow_fake_rand"))]
+pub(crate) fn getrandom(_: &mut [u8]) {
+    panic!("there is no real entropy source in Powdr");
```
pacheco, line 17: point to fake rand feature in the msg?

lvella: Done.

pacheco (approved):

Issue #1728, opened 2024-08-28
Lambda expression formatting


chriseth:
The formatting for lambda expressions still has some flaws:

1. There are always parentheses around lambda expressions, even though they are not always needed (they are needed in some cases for example `|i| |x| x + i` vs `|i| (|x| x) + i`

2. For block expressions that do not have an expression, the formatter adds an empty line (instead of just omitting it):
```rust
    let set_hint: expr -> () = (constr |c| {
        std::prelude::set_hint(c, (query |_| std::prelude::Query::Hint(8)));

    });
```

PR #1733, opened 2024-08-29
LambdaExpression formatting

gzanitti:
This PR fixes issue #1728.
1) Removes the blank space that was printed even if the body of the blocks was empty.
2) Adds the Precedence trait to LambdaExpressions to correctly handle the use of parentheses.

### DIFF
--- a/ast/src/parsed/display.rs
+++ b/ast/src/parsed/display.rs
@@ -748,18 +748,22 @@ impl Display for NamespacedPolynomialReference {
     }
 }
 
-impl<E: Display> Display for LambdaExpression<E> {
+impl<E> Display for LambdaExpression<E>
+where
+    E: Display + Precedence,
+{
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
-        write!(
-            f,
-            "({}|{}| {})",
-            match self.kind {
-                FunctionKind::Pure => "".into(),
-                _ => format!("{} ", &self.kind),
-            },
-            format_list(&self.params),
-            self.body
-        )
+        let prefix = match self.kind {
+            FunctionKind::Pure => "".into(),
+            _ => format!("{} ", &self.kind),
+        };
+        let params = format_list(&self.params);
+
+        if self.body.precedence() < self.precedence() {
+            write!(f, "{}|{}| {}", prefix, params, self.body)
+        } else {
+            write!(f, "{}|{}| ({})", prefix, params, self.body)
+        }
     }
 }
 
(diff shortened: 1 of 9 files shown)


ast/src/parsed/mod.rs
```diff
@@ -1003,11 +1003,18 @@ impl Precedence for BinaryOperator {
     }
 }
 
+impl<E> Precedence for LambdaExpression<E> {
+    fn precedence(&self) -> Option<ExpressionPrecedence> {
+        Some(13)
```
gzanitti, line 1008: Former highest precedence + 1

ast/src/parsed/display.rs
```diff
@@ -910,8 +914,9 @@ impl<E: Display> Display for BlockExpression<E> {
             write_items_indented(f, &self.statements)?;
             if let Some(expr) = &self.expr {
                 write_indented_by(f, expr, 1)?;
+                writeln!(f)?;
             }
-            write!(f, "\n}}")
+            write!(f, "}}")
```
gzanitti, line 919: This is the only change regarding (1) white space in empty blocks.

ast/src/parsed/display.rs
```diff
@@ -1195,6 +1200,10 @@ mod tests {
                 ("let x = (1 + (|i| i + 2));", "let x = 1 + (|i| i + 2);"),
                 // Index access
                 ("(|i| i)[j];", "(|i| i)[j];"),
+                // Lambda expression
+                ("|i| (|x| x) + i;", "|i| (|x| x) + i;"),
+                ("|i| (|x| x + i);", "|i| (|x| x + i);"),
+                ("|i| |x| x + i;", "|i| (|x| x + i);"),
```
chriseth, line 1206: Can you add another test set (with all combinations of parentheses) where the lambda expression is inside a binary operator, like `|i| i + |x| x`?

ast/src/parsed/display.rs
```diff
@@ -1200,10 +1200,27 @@ mod tests {
...
+        }
+
+        #[test]
+        fn lambda_parentheses() {
+            let test_cases: Vec<TestCase> = vec![
+                ("1 + (|x| x)(2);", "1 + (|x| x)(2);"),
                 ("|i| (|x| x) + i;", "|i| (|x| x) + i;"),
                 ("|i| (|x| x + i);", "|i| (|x| x + i);"),
                 ("|i| |x| x + i;", "|i| (|x| x + i);"),
+                ("(|i| i) + (|x| x);", "(|i| i) + (|x| x);"),
```
chriseth: Interestingly, we would actually not need the second pair of parentheses, i.e. the following would also be ok: `(|i| i) + |x| x`, but this is good enough and maybe even better because less ambiguous.

chriseth (approved):

PR #1734, opened 2024-08-29
run pr-tests on self hosted runner

pacheco:


leonardoalt:
shouldn't we rather run the build in the self hosted runner and the pr tests on GH?

pacheco:
Should already be what this is doing (pr-tests just refers to the name of the action), in any case its simple to change as we want.

leonardoalt:
should we merge this soon?

leonardoalt (approved):

PR #1736, opened 2024-08-30
Refactor asm ast

Schaeff:


airgen/src/lib.rs
```diff
@@ -178,14 +174,25 @@ fn resolve_submachine_arg(
...
+            (
+                module_path,
+                module
+                    .into_inner()
+                    .1
+                    .into_iter()
+                    .filter(|s| {
+                        matches!(
+                            s,
+                            PilStatement::EnumDeclaration(..) | PilStatement::LetStatement(..)
```
chriseth, line 191: I already envision spending quite some time trying to figure out why something is missing from the output. Is there a safer way to do this check so that we add variants here if the code is extended?

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -1365,51 +1358,4 @@ machine Main {
 ";
         parse_analyze_and_compile::<GoldilocksField>(asm);
     }
-
-    #[test]
-    fn trait_parsing() {
```
chriseth, line 1370: Was this removed in a merge?

ast/src/asm_analysis/mod.rs
```diff
@@ -808,28 +784,97 @@ pub struct Rom {
...
+            module.machines.into_iter().map(move |(name, machine)| {
+                let mut machine_path = module_path.clone();
+                machine_path.push(name.to_string());
+                (machine_path, machine)
+            })
         })
     }
+
+    pub fn get_machine(&self, ty: &AbsoluteSymbolPath) -> Option<&Machine> {
+        let mut path = ty.clone();
```
chriseth, line 821: I the clone needed here?

ast/src/object/mod.rs
```diff
@@ -41,7 +41,7 @@ pub struct PILGraph {
     pub main: Machine,
     pub entry_points: Vec<Operation>,
     pub objects: BTreeMap<Location, Object>,
-    pub definitions: BTreeMap<AbsoluteSymbolPath, TypeOrExpression>,
+    pub definitions: BTreeMap<AbsoluteSymbolPath, Vec<PilStatement>>,
```
chriseth, line 44: Is the key here the module name?

ast/src/parsed/asm.rs
```diff
@@ -29,26 +29,19 @@ pub struct ASMModule {
...
 
 impl ModuleStatement {
-    pub fn defined_names(&self) -> Option<&String> {
+    pub fn defined_names(&self) -> Box<dyn Iterator<Item = &String> + '_> {
         match self {
-            ModuleStatement::SymbolDefinition(d) => Some(&d.name),
-            ModuleStatement::TraitImplementation(_) => None,
+            ModuleStatement::SymbolDefinition(d) => Box::new(once(&d.name)),
+            ModuleStatement::PilStatement(s) => {
+                Box::new(s.symbol_definition_names().map(|(name, _)| name))
```
chriseth, line 43: I think we should call `symbol_definition_names_and_contained` here, but I'm not sure. `.._and_contained` also returns the variants in an enum for example.

ast/src/parsed/asm.rs
```diff
@@ -97,7 +81,7 @@ pub enum SymbolValueRef<'a> {
     /// A module definition
     Module(ModuleRef<'a>),
     /// A generic symbol / function.
-    Expression(&'a TypedExpression),
+    Expression(&'a Option<Expression>, &'a Option<TypeScheme<Expression>>),
```
chriseth, line 84: Why not keep TypedExpression? Lifetime issues?

importer/src/path_canonicalizer.rs
```diff
@@ -263,6 +232,45 @@ fn free_inputs_in_expression_mut<'a>(
...
+            for variant in &mut enum_decl.variants {
+                if let Some(fields) = &mut variant.fields {
+                    for field in fields {
+                        canonicalize_inside_type(field, &type_vars, path, paths);
+                    }
+                }
+            }
+        }
+        PilStatement::TraitImplementation(_, trait_impl) => {
+            for f in &mut trait_impl.functions {
```
chriseth, line 260: Should also call `canonicalize_inside_type_scheme` for `trait_impl.type_scheme`.

asm-to-pil/src/vm_to_constrained.rs
chriseth: OK, fine to remove, there is `test_data/asm/trait_parsing.asm` with the same content.

ast/src/object/mod.rs
Schaeff: yes, comment added

chriseth: Maybe we should also rename this to `statements`?

ast/src/asm_analysis/mod.rs
Schaeff: Yes since modules are indexed by path and machines by name. To avoid this clone, we'd need slices of `AbsoluteSymbolPath` which is not implemented in this PR.

ast/src/parsed/asm.rs
Schaeff: So... if we have
```
enum Foo {
    Bar
}
```
Then the defined names should be `[Foo, Bar]`? Or `["Foo", "Foo::Bar"]`?

ast/src/parsed/asm.rs
Schaeff: Because it's constructed from a reference over `PilStatement::LetStatement` which is
```
    LetStatement(
        SourceRef,
        String,
        Option<TypeScheme<Expression>>,
        Option<Expression>,
    ),
```
My understanding is that `SymbolValueRef::Expression` also covers statements like `let a: Type;` which seem to be allowed. `TypedExpression` would not cover that.

importer/src/path_canonicalizer.rs
Schaeff: Extracted to https://github.com/powdr-labs/powdr/pull/1750/files because I'm not sure

ast/src/object/mod.rs
Schaeff: done

airgen/src/lib.rs
Schaeff: using an exhaustive match instead

ast/src/parsed/asm.rs
chriseth: `[("Foo", None), ("Foo", Some("Bar"))]`

airgen/src/lib.rs
```diff
@@ -178,14 +174,35 @@ fn resolve_submachine_arg(
...
+                        PilStatement::PublicDeclaration(..) => false,
+                        PilStatement::PolynomialConstantDeclaration(..) => false,
+                        PilStatement::PolynomialConstantDefinition(..) => false,
+                        PilStatement::PolynomialCommitDeclaration(..) => false,
+                        PilStatement::PlookupIdentity(..) => false,
+                        PilStatement::PermutationIdentity(..) => false,
+                        PilStatement::ConnectIdentity(..) => false,
+                        PilStatement::TraitImplementation(..) => false,
+                        PilStatement::TraitDeclaration(..) => false,
+                        PilStatement::Expression(..) => false,
```
chriseth, line 202:
```suggestion
                        PilStatement::Include(..) |
                        PilStatement::Namespace(..) |
                        PilStatement::PolynomialDefinition(..) |
                        PilStatement::PublicDeclaration(..) |
                        PilStatement::PolynomialConstantDeclaration(..) |
                        PilStatement::PolynomialConstantDefinition(..) |
                        PilStatement::PolynomialCommitDeclaration(..) |
                        PilStatement::PlookupIdentity(..) |
                        PilStatement::PermutationIdentity(..) |
                        PilStatement::ConnectIdentity(..) |
                        PilStatement::TraitImplementation(..) |
                        PilStatement::TraitDeclaration(..) |
                        PilStatement::Expression(..) => false,
```

ast/src/parsed/asm.rs
Schaeff: I mean, what should this function return? Currently it returns all names defined in a given statement, as strings. Should the signature then change?

chriseth: At least before your PR this was essentially only used by the call to local_names() inside check_machine to see if there is a name clash, so I think it's not needed to return the nested names.

Schaeff: It's used only in the importer to prevent duplicates, and to detect the standard library. I think we do not need the nested names for that. Then the current code is correct?

importer/src/path_canonicalizer.rs
```diff
@@ -364,7 +365,7 @@ fn canonicalize_inside_type(
     }
 
     for tne in ty.children_mut() {
-        canonicalize_inside_expression(tne, path, paths);
+        canonicalize_inside_type(tne, type_vars, path, paths);
```
Schaeff, line 368: Is it correct to forward the `type_vars` here?

importer/src/path_canonicalizer.rs
Schaeff: Following discussion, removed the commit and instead used generics on the existing data structures

importer/src/path_canonicalizer.rs
Schaeff: Ok no this is wrong

importer/src/path_canonicalizer.rs
```diff
@@ -351,21 +354,21 @@ fn canonicalize_inside_type_scheme(
...
-    ty: &mut Type<Expression>,
+fn canonicalize_inside_type<E>(
+    ty: &mut Type<E>,
     type_vars: &HashSet<&String>,
     path: &AbsoluteSymbolPath,
     paths: &'_ PathMap,
-) {
+) where
+    Type<E>: ExpressionVisitable<Expression>,
+{
```
Schaeff, line 364: My understanding is that these are required because of the following implementation:

```rust
impl<Expr: ExpressionVisitable<Expr>, C: Children<Expr>> ExpressionVisitable<Expr> for C
```

`Type` implements `Children<Type>`, so if we set `C` and `Expr` to `Type`, we have a recursive trait dependency: `Type` implements `ExpressionVisitable<Type>` if it implements `ExpressionVisitable<Type>`.

This is fine as long as we are not calling methods from `ExpressionVisitable<Expr>` with `Expr` potentially being `Type`.

Below, when calling `pre_visit_expressions_mut`, even if we pass a closure which takes an `Expression` as input, the compiler still tries to set `Expr` to `Type`, which triggers the recursive dependency. Note that this still happens even though we pass a closure which takes `Expression` as input (and therefore excludes `ExpressionVisitable<Type>` which would require a closure taking a `Type`), so it seems like the compiler is not using this knowledge.

Now the confusing part: adding the where clause somehow tells the compiler not to try to apply `ExpressionVisitable<Type>`, even though it is implemented!

ast/src/parsed/types.rs
```diff
@@ -136,6 +136,62 @@ impl<E> Type<E> {
...
+                    .chain(base.contained_expressions()),
+            ),
+            t => Box::new(t.children().flat_map(|t| t.contained_expressions())),
+        }
+    }
+}
+
+/// A trait to operate the possible types for the array type lengths
+pub trait ArrayLength: std::fmt::Display + std::fmt::Debug {
+    fn try_to_expression_mut(&mut self) -> Option<&mut Expression>;
```
chriseth, line 170: try_as_expression_mut?

ast/src/parsed/types.rs
```diff
@@ -136,6 +136,62 @@ impl<E> Type<E> {
...
+                    .into_iter()
+                    .chain(base.contained_expressions()),
+            ),
+            t => Box::new(t.children().flat_map(|t| t.contained_expressions())),
+        }
+    }
+}
+
+/// A trait to operate the possible types for the array type lengths
+pub trait ArrayLength: std::fmt::Display + std::fmt::Debug {
```
chriseth, line 169: ExpressionInArrayLength?

ast/src/parsed/types.rs
Schaeff: done

ast/src/parsed/types.rs
Schaeff: done

PR #1744, opened 2024-09-02
Make csv export work with multiple fixed column sizes

Schaeff:


pipeline/src/pipeline.rs
```diff
@@ -549,9 +550,32 @@ impl<T: FieldElement> Pipeline<T> {
...
+                let witness_sizes: HashMap<&str, u64> = witness
+                    .iter()
+                    .map(|(name, values)| {
+                        let namespace = name.split("::").next().unwrap();
+                        (namespace, values.len() as u64)
+                    })
+                    .fold(Default::default(), |mut map, (name, size)| {
+                        map.entry(name).or_insert(size);
+                        map
+                    });
```
chriseth, line 563: Isn't this just
```suggestion
                    .collect();
```
?

chriseth (dismissed):

pipeline/src/pipeline.rs
```diff
@@ -549,9 +550,30 @@ impl<T: FieldElement> Pipeline<T> {
 
         if self.arguments.export_witness_csv {
             if let Some(path) = self.path_if_should_write(|name| format!("{name}_columns.csv"))? {
-                // TODO: Handle multiple sizes
-                let fixed = get_uniquely_sized_cloned(fixed).unwrap();
-                let columns = fixed.iter().chain(witness.iter()).collect::<Vec<_>>();
+                // get the column size for each namespace. This assumes all witness columns of the same namespace have the same size.
```
ShuangWu121, line 553: I'm still getting familiar with the codebase, so I have a few questions :) :
-Is my understanding correct that this code is related to VadCop, which allows different "machines" (namespaces) to have different column sizes? 
-Does namespace refer to "machines," where each "machine" has columns of the same size?

Thank you for your help in clarifying these points :)!

ShuangWu121 (approved):

pipeline/src/pipeline.rs
Schaeff: Good questions, your understanding is correct.
In PIL, each machine instance is represented as a AIR/namespace, within which all columns have the same length.
Since dynamic VadCop, we only know the size of each AIR after witgen, so this PR uses that to determine the correct fixed columns to export to csv.

Today we can think of namespaces as machine instances, but machines are an asm concept which disappears in PIL, and once we have proof objects there's likely going to be a way to put many machine instance inside the same AIR.

ShuangWu121: Got it. Thanks for clarification!

PR #1746, opened 2024-09-02
Remove query functions from asm processing

chriseth:


executor/src/witgen/mod.rs
```diff
@@ -47,6 +47,10 @@ mod vm_processor;
 
 static OUTER_CODE_NAME: &str = "witgen (outer code)";
 
+// TODO change this so that it has functions
```
chriseth, line 50: Created issue here: https://github.com/powdr-labs/powdr/issues/1789

lvella (changes requested):
It seems the `Option` is not necessary and using the `&mut` field "just works" (apparently rebind is automatic).

lvella (approved):

std/prover.asm
```diff
@@ -14,6 +14,32 @@ let try_eval: expr -> Option<fe> = [];
...
+let get_input_from_channel: int, int -> fe = [];
+
+/// Outputs a byte to a file descriptor.
+let output_byte: int, int -> () = [];
+
+let handle_query: expr, int, std::prelude::Query -> () = query |column, row, v| match v {
+    Query::Hint(h) => provide_if_unknown(column, row, || h),
+    Query::Input(i) => provide_if_unknown(column, row, || get_input(i)),
+    Query::DataIdentifier(i, j) => provide_if_unknown(column, row, || get_input_from_channel(i, j)),
+    Query::Output(fd, b) => provide_if_unknown(column, row, || { output_byte(fd, b); 0 }),
```
chriseth, line 39: The bug was that the system actually expects that we set the value to zero, and that was also the former behaviour.

lvella (approved):

PR #1750, opened 2024-09-03
Use expression as type param in trait implementation

Schaeff:
https://github.com/powdr-labs/powdr/pull/1736#discussion_r1740736729

### DIFF
--- a/ast/src/parsed/mod.rs
+++ b/ast/src/parsed/mod.rs
@@ -349,7 +349,7 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct TraitImplementation<Expr> {
     pub name: SymbolPath,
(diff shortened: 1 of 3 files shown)


gzanitti (approved):

ast/src/parsed/mod.rs
```diff
@@ -349,7 +349,7 @@ impl<R> Children<Expression<R>> for EnumVariant<Expression<R>> {
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct TraitImplementation<Expr> {
     pub name: SymbolPath,
-    pub type_scheme: TypeScheme,
+    pub type_scheme: TypeScheme<Expr>,
```
chriseth, line 352: I don't think this is correct. As with TypedExpression, we need the flexibility for `functions` and `type_scheme` to use different type var values, because they are used differently. For type, it is u64 after analysis, for expressions, it is analyzed::Expression.

Maybe we should talk about this in person, it's a bit tricky.

PR #1768, opened 2024-09-09
Turn machine statements into block statements.

chriseth:


gzanitti (approved):

lvella (commented):
Why in some `let` you set the type to `col`, and others you leave the type unspecified?

Should I use just `let` instead of `col witness`?

chriseth:
This is just a partial PR, I didn't want to make it gigantic. In the future, please use `let` instead of `col witness`.

chriseth:
Ah forget it.

PR #1770, opened 2024-09-09
Lookups as Binary ops

gzanitti:


parser/src/lib.rs
```diff
@@ -210,44 +209,6 @@ mod test {
         );
     }
 
-    #[test]
```
chriseth, line 213: This is just parsed as an operator now, nothing special.

chriseth (approved):

PR #1778, opened 2024-09-10
Div and mul for 32 bit numbers for baby bear

qwang98:
Two operations: 
- `mul` takes two 32 bit-numbers, multiply them, and outputs two 32 bit-numbers, which are the upper and lower halves of the product
- `div` takes two 32 bit-numbers, which are the dividend and divisor, respectively, and outputs two 32 bit-numbers, which are the quotient and the remainder, respectively

Warning:
- Although we do provide hints to calculate a unique remainder that's less than the divisor, this is not constrained and thus witness for `div` is not unique and not sound.

Testing:
- I created new test vectors but should be quite comprehensive (uses all variables).

Implementation:
- Uses the affine equation `x1 * y1 + x2 = y2 * 2**32 + y3`.
- `x2` is constrained to 0 for `mul`.
- `y2` is constrained to 0 for `div`.
- This approach has the disadvantage of using five sets of columns, i.e. `x1`, `y1`, `x2`, `y2`, and `y3`, instead of just four sets. However, five sets of columns has the advantage of using roughly half of the number of constraints compared to if we were to have four sets of columns. This is because `mul` and `div` can be both defined by affine with five sets of columns, whereas defining `mul` and `div` separately in four sets of columns requires using the variables for different purposes depending on the operations, and thus more constraints.
- Tl;dr: it's a trade off between one more variable (4 more columns) and fewer constraints.


lvella:
I believe for other fields, division is unique and sound. The unsoundness is just for babybear?

qwang98:
> I believe for other fields, division is unique and sound. The unsoundness is just for babybear?

As discussed with @lvella, the RISCV machine constrains that the remainder is less than the divisor in `div`: https://github.com/powdr-labs/powdr/blob/main/riscv/src/code_gen.rs#L626-L649

We should do the same in the main machine using BabyBear. @leonardoalt

test_data/std/arith_bb_test_16.asm
```diff
@@ -0,0 +1,59 @@
+use std::machines::arith_bb::Arith16;
```
onurinanc, line 1: naming here

std/machines/arith_bb.asm
```diff
@@ -0,0 +1,197 @@
...
+    // Allow this machine to be connected via a permutation
+    call_selectors: sel,
+{
+    Byte byte;
+    
+    // The operation ID will be bit-decomposed to yield selEq[], controlling which equations are activated.
+    col witness operation_id;
+
+    // operation_id has to be either mul or div.
+    operation_id * (1 - operation_id) = 0;
```
onurinanc, line 28: You can also use std::utils::force_bool(operation_id); here

lvella (dismissed):
I don't think this should be merged without the uniqueness constraints.

test_data/std/arith_bb_test_16.asm
qwang98: Agree

std/machines/arith_bb.asm
qwang98: Agree

qwang98:
> I don't think this should be merged without the uniqueness constraints.

Will do :)

qwang98:
> I don't think this should be merged without the uniqueness constraints.



> I don't think this should be merged without the uniqueness constraints.

I think this has been implemented in the main machine that `remainder < divisor`. Same for other RISCV specifications, such as setting the quotient to `0xffffffff` when the divisor is zero: https://github.com/powdr-labs/powdr/pull/1790/files#diff-edc6d3bea1d3d72d7d7a853447050294a9a2ff3a0189e2948eb0d6f34c062bc0R731-R732

Therefore, I think this no longer needs to be added in this sub machine.

std/machines/arith_bb.asm
qwang98: Done.

std/machines/mod.asm
```diff
@@ -1,4 +1,5 @@
 mod arith;
+mod arith_bb;
```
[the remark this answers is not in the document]
qwang98: Done

leonardoalt:
> We should do the same in the main machine using BabyBear? @leonardoalt

Yes this is now done in the RISCV machine for BB.

std/machines/arith16.asm
```diff
@@ -0,0 +1,191 @@
...
+    col witness operation_id;
+
+    // operation_id has to be either mul or div.
+    force_bool(operation_id);
+
+    // Computes x1 * y1 + x2, where all inputs / outputs are 32-bit words (represented as 16-bit limbs in little-endian order).
+    // More precisely, affine_256(x1, y1, x2) = (y2, y3), where x1 * y1 + x2 = 2**16 * y2 + y3
+
+    // x1 * y1 = y2 * 2**16 + y3
+    operation mul<0> x1c[0], x1c[1], y1c[0], y1c[1] -> y2c[0], y2c[1], y3c[0], y3c[1];
```
leonardoalt, line 33: small detail, but can we switch to big endian here? I think it's more natural to represent numbers (known width) and the other machines are also written like that.

So it would be

```
mul A_high, A_low, B_high, B_low, -> result_high_high, result_high_low, result_low_high, result_low_low
div A_high, A_low, B_high, B_low, -> quot_high, quot_low, rem_high, rem_low
```

std/machines/arith16.asm
```diff
@@ -0,0 +1,191 @@
...
+
+// Arithmetic machine, ported mainly from Polygon: https://github.com/0xPolygonHermez/zkevm-proverjs/blob/main/pil/arith.pil
+// This machine supports eq0, which is the affine equation. Currently we only expose operations for mul and div.
+machine Arith16 with
+    latch: CLK8_7,
+    operation_id: operation_id,
+    // Allow this machine to be connected via a permutation
+    call_selectors: sel,
+{
+    Byte byte;
```
leonardoalt, line 22:
```suggestion
machine Arith16(byte: Byte) with
    latch: CLK8_7,
    operation_id: operation_id,
    // Allow this machine to be connected via a permutation
    call_selectors: sel,
{
```

leonardoalt (commented):
Nice! Just a few small comments. I used this machine in the RISCV BB PR and it seems to work well.

std/machines/arith16.asm
qwang98: Done!

std/machines/arith16.asm
qwang98: Done

leonardoalt (approved):

PR #1782, opened 2024-09-10
Remove lookups from PilStatement

gzanitti:


ast/src/parsed/display.rs
```diff
@@ -142,17 +142,13 @@ impl Display for InstructionBody {
 }
 
 fn format_instruction_statement(stmt: &PilStatement) -> String {
```
chriseth, line 144: Can you create an issue to simplify the caller code and data structures so that we can have `stmt: &Expression` here?

linker/src/lib.rs
```diff
@@ -191,43 +195,60 @@ fn process_link(link: Link) -> PilStatement {
...
+        let permutation_expr = Expression::BinaryOperation(
+            SourceRef::unknown(),
+            BinaryOperation {
+                left: Box::new(lhs),
+                right: Box::new(rhs),
+                op: BinaryOperator::Is,
+            },
+        );
+
+        PilStatement::Expression(SourceRef::unknown(), permutation_expr)
```
chriseth, line 233:
```suggestion
        ).into()
```

chriseth: Does that work?

linker/src/lib.rs
```diff
@@ -236,29 +257,43 @@ fn process_link(link: Link) -> PilStatement {
...
+        let identity_expr = Expression::BinaryOperation(
+            SourceRef::unknown(),
+            BinaryOperation {
+                left: Box::new(lhs),
+                right: Box::new(rhs),
+                op: BinaryOperator::In,
+            },
+        );
+
+        PilStatement::Expression(SourceRef::unknown(), identity_expr)
```
chriseth, line 296: Could also use `into` here.

linker/src/lib.rs
```diff
@@ -236,29 +257,43 @@ fn process_link(link: Link) -> PilStatement {
...
+                                )
+                            }))
+                            .collect(),
+                    }
+                    .into(),
+                ),
+            },
+        );
+
+        let identity_expr = Expression::BinaryOperation(
```
chriseth, line 287: Maybe in general, it might be useful to create helper functions to create lookups. Maybe in the same file as the `namespaced_reference` function?

linker/src/lib.rs
gzanitti: No, it doesn't

chriseth (approved):

PR #1784, opened 2024-09-11
BabyBear shift machine

qwang98:
All tests passed. :)

Operations:
- shl<0> A1, A2, B -> C1, C2
- shr<1> A1, A2, B -> C1, C2
- `A1` `A2` are 16 bit limbs of 32 bit `A` in little-endian order. Likewise for `C1` and `C2`

Implementation:
- We adopted a similar implementation to our prior shift machine, which decomposes `A` to 4 bytes and looks up each byte to a lookup table of `[A_byte, B (shift amount), block row, operation id]`, so the size of the lookup table is `[256, 32, 4, 2] = 65536`. Each row looks up to the resulting byte after the bit shifting, and the results are added together to obtain `C`.
- In our design, instead of looking up to 32-bit `C` column, we are looking up to two 16-bit `C1` and `C2` columns. Overall, there are more witness columns due to decomposing to 16-bit limbs the in the main shift machine and one more fixed column in the lookup table, but the same number of lookups performed.

Future optimization:
- There's ample ground for "reshaping" the main machine to have more columns but fewer rows, and do more lookups in each row so that we are not just processing one byte in each row. In the most aggressive case, we can even process everything in the same row.
- For example, processing two bytes in one row (instead of one byte) should have half the number of rows but less than twice the number of columns, and therefore fewer cells in total. This should be good for provers with time linear to the number of cells.


std/machines/shift_bb.asm
```diff
@@ -0,0 +1,70 @@
...
+use std::convert::int;
+
+// Shift for single bytes using an exhaustive table
+// TODO this way, we cannot prove anything that shifts by more than 31 bits.
+machine ByteShiftBB with
+    latch: latch,
+    operation_id: operation_id,
+    degree: 65536
+{
+    // P_C1 and P_C2 are both 16 bit limbs of P_C, where P_C1 is the less significant limb.
```
chriseth, line 12: Don't we usually have big endian ordering?

std/machines/shift_bb.asm
```diff
@@ -0,0 +1,70 @@
+use std::utils::unchanged_until;
+use std::utils::cross_product;
+use std::convert::int;
+
+// Shift for single bytes using an exhaustive table
```
chriseth, line 5:
```suggestion
// Shift for single bytes using an exhaustive table, returning two 8-bit values as result.
```

std/machines/shift_bb.asm
```diff
@@ -0,0 +1,70 @@
...
+{
+    // P_C1 and P_C2 are both 16 bit limbs of P_C, where P_C1 is the less significant limb.
+    operation run<0> P_operation, P_A, P_B, P_ROW -> P_C1, P_C2;
+
+    col fixed latch = [1]*;
+    col fixed operation_id = [0]*;
+
+    let bit_counts = [256, 32, 4, 2];
+    let min_degree = std::array::product(bit_counts);
+    std::check::assert(std::prover::min_degree() >= std::array::product(bit_counts), || "The shift machine needs at least 65536 rows to work.");
```
chriseth, line 20: Could also add an assertion for the min field modulus this  machine works with.

std/machines/shift_bb.asm
```diff
@@ -0,0 +1,70 @@
...
+    col fixed P_C2(i) { (c(i) >> 16) & 0xffff };
+}
+
+machine ShiftBB(byte_shift: ByteShiftBB) with
+    latch: latch,
+    operation_id: operation_id,
+    // Allow this machine to be connected via a permutation
+    call_selectors: sel,
+{
+    operation shl<0> A1, A2, B -> C1, C2;
```
chriseth, line 44: could call this `Ahi` and `Alow`, to make endianness clear

std/machines/shift_bb.asm
qwang98: We had little endian in binary bb machine, but as just discussed with Leo both should be good.

std/machines/shift_bb.asm
qwang98: Agree

std/machines/shift_bb.asm
```diff
@@ -0,0 +1,70 @@
+use std::utils::unchanged_until;
+use std::utils::cross_product;
+use std::convert::int;
+
+// Shift for single bytes using an exhaustive table
+// TODO this way, we cannot prove anything that shifts by more than 31 bits.
+machine ByteShiftBB with
```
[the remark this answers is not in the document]
qwang98: Agree

std/machines/shift_bb.asm
```diff
@@ -0,0 +1,70 @@
...
+    col fixed A1_SWITCH = [1, 0, 0, 1]*;
+
+    col witness A_byte;
+    col witness C_part1, C_part2;
+
+    col witness A1, A2;
+    col witness B;
+    col witness C1, C2;
+
+    A1' = A1 * (1 - latch) + A_byte * FACTOR * A1_SWITCH;
```
chriseth, line 63: Can we reduce the degree here by creating columns for `FACTOR * A1_SWITCH` and `FACTOR * (1 - A1_SWITCH)` instead of `FACTOR` and `A1_SWITCH`?

qwang98: Agree

std/machines/shift_bb.asm
qwang98: Agree

qwang98:
@chriseth All comments fixed. Ready for a final review. :)

chriseth (approved):

PR #1787, opened 2024-09-11
Removing useless steps from cache building.

lvella:


.github/workflows/build-cache.yml
```diff
@@ -19,16 +19,8 @@ jobs:
       run: rustup toolchain install 1.77-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain 1.77-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain 1.77-x86_64-unknown-linux-gnu
     - name: Install EStarkPolygon prover dependencies
       run: sudo apt-get install -y nlohmann-json3-dev libpqxx-dev nasm
-    - name: Lint
-      run: cargo clippy --all --all-targets --all-features --profile pr-tests -- -D warnings
-    - name: Lint
-      run: cargo clippy --all --all-targets --no-default-features --profile pr-tests -- -D warnings
```
chriseth, line 25: Are you sure we don't need any of the artifacts created in this step?

lvella: No, I am not.

In fact, I think the step that is more likely to be relevant for caching is the "Check without Halo2", not the lints. But the uncertainty makes this PR not worth it.

PR #1788, opened 2024-09-11
add generic fp2 for all the field elements (bb, gl, bn254, m31)

onurinanc:


std/math/fp2.asm
```diff
@@ -20,9 +20,8 @@ use std::prover::eval;
 /// f = x^2 - 7
```
lvella, line 20: Outdated comment. I guess the X^2 - 7 is not irreducible in BabyBear?

std/math/fp2.asm
```diff
@@ -166,7 +165,7 @@ mod test {
 
         // Subtract arbitrary elements
         let _ = test_sub(Fp2::Fp2(123, 1234), Fp2::Fp2(567, 5678), Fp2::Fp2(123 - 567, 1234 - 5678));
-        test_sub(Fp2::Fp2(-1, -1), Fp2::Fp2(0x100000000, 1), Fp2::Fp2(-0x100000000 - 1, -2))
+        test_sub(Fp2::Fp2(-1, -1), Fp2::Fp2(2013265920, 1), Fp2::Fp2(-2013265920 - 1, -2))
```
lvella, line 168: Personal preference, but I like hex more here...

lvella (commented):
What is fp2?

std/math/fp2.asm
```diff
@@ -20,9 +20,8 @@ use std::prover::eval;
 /// f = x^2 - 7
 /// f.is_irreducible()
 
-/// An element of the extension field over the implied base field (which has to be either
-/// the Goldilocks or the BN254 field) relative to the irreducible polynomial X^2 - 7,
-/// (This irreducible polynomial also works for Mersenne31)
+/// An element of the extension field over the implied base field (which has to be one
+/// of the field elements: Goldilocks, BN254, BabyBear, M31) relative to the irreducible polynomial X^2 - 11,
```
chriseth, line 24: Do we have a test somewhere that verifies that it is really irreducible on all the fields we care about?

chriseth: We should somehow check that `Fp2` can only be constructed for the known fields. But that would mean that this check has to be run every time we call some function here...

lvella: Maybe the method returning the irreducible polynomial should be available in the field's trait, so it could be made to work with any field.

chriseth: The problem is that in pil, the field is not really a type. We could use the `known_field` function, but then we would have to do it at every multiplication.

lvella: My idea was to introduce a `get_f2_irreducible()` function, but yeah, it would have to be called everywhere it is needed.

std/math/fp2.asm
onurinanc: - fp2 is a quadratic tower extension of a field element. We use the fp2 extension field element (instead of ff.asm) in permutation/lookup arguments in PIL to increase the security of bits. For example, if we use Goldilocks field (2^64) on permutation arguments in PIL, we could have 64-18 = 46 bits of estimated security for the 2^18 number of rows. Using fp2 = (fp, fp) basically. We know have 64+64-18 = 110 bits of estimated security

- Yes, x^2 - 7 is not irreducible, we should've have x^2 - 11, thanks for the catch

std/math/fp2.asm
onurinanc: Same, i'll change it to the hex value

std/math/fp2.asm
onurinanc: We have the sage test in here as a comment: https://github.com/powdr-labs/powdr/blob/b2a06c2b03997f77b04ae2b0e41f9bcd996f83e3/std/math/fp2.asm#L13-L21 (x^2 - 7 is wrong, should be x^2 - 11). I would like to add the code inside the following gist as a comment: https://gist.github.com/onurinanc/ccb37ae5dde012dc8bb67617a0f89376

onurinanc: If we had a constructor for Fp2 we could do the check when creating a new fp2 instance, but we are using Fp2::Fp2()

lvella: Well, this PR doesn't solve the problem, but it hasn't created it, either, so I am approving it.

lvella (approved):

PR #1796, opened 2024-09-12
Create is_first inside the lookup.

chriseth:
Since we can now create constant columns inside `constr` functions, we don't need to pass `is_first` any more.

### DIFF
--- a/std/protocols/lookup.asm
+++ b/std/protocols/lookup.asm
@@ -53,14 +53,13 @@ let compute_next_z: Fp2<expr>, Fp2<expr>, Fp2<expr>, Constr, expr -> fe[] = quer
     
 /// Adds constraints that enforce that rhs is the lookup for lhs
 /// Arguments:
-/// - is_first: A column that is 1 for the first row and 0 for the rest
 /// - acc: A phase-2 witness column to be used as the accumulator. If 2 are provided, computations
 ///        are done on the F_{p^2} extension field.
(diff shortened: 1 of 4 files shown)


Schaeff (approved):

std/protocols/lookup.asm
```diff
@@ -93,12 +92,12 @@ let lookup: expr, expr[], Fp2<expr>, Fp2<expr>, Constr, expr -> Constr[] = |is_f
...
-        // (because of wrapping, the acc[0] and acc[N] are the same)
-        is_first * acc_1 = 0,
-        is_first * acc_2 = 0
-    ] + constrain_eq_ext(update_expr, from_base(0))
+    let (acc_1, acc_2) = unpack_ext(acc_ext);
+    // First and last acc needs to be 0
+    // (because of wrapping, the acc[0] and acc[N] are the same)
+    is_first * acc_1 = 0;
+    is_first * acc_2 = 0;
+    constrain_eq_ext(update_expr, from_base(0));
```
onurinanc, line 102: I am thinking about changing the below function in the same way, but I think it is not possible since `constr` function returns `()` but I need to return `Constr[]` to be able to use `bus_interaction` to implement bus_lookup as bus lookup needs adding two constraints like `bus_send() + bus_receive()` https://github.com/powdr-labs/powdr/blob/e668cdfc6438aec58739732e82f01baf114648aa/std/protocols/bus.asm#L39-L61

PR #1798, opened 2024-09-13
Use operators for lookups.

chriseth:
We can now use the `$`, `in`, `is` and `connect` operators in any expression. Remember that `$` needs to be used when you want to have a selector.


gzanitti (approved):
:1st_place_medal:

PR #1800, opened 2024-09-13
Cleaner handle_symbol_definition

gzanitti:
`handle_symbol_definition` in `pil-analyzer/src/statement_processor.rs` is becoming to long and it will get worst after we add structs. 
This PR separate every case inside its own function to simplify and improve readability.

### DIFF
--- a/pil-analyzer/src/statement_processor.rs
+++ b/pil-analyzer/src/statement_processor.rs
@@ -472,7 +472,10 @@ where
             .iter()
             .map(|function| {
                 (
-                    vec![name.clone(), function.name.clone()],
+                    self.driver
+                        .resolve_namespaced_decl(&[&name, &function.name])
+                        .relative_to(&Default::default())
+                        .to_string(),
                     FunctionValueDefinition::TraitFunction(
                         shared_trait_decl.clone(),
                         function.clone(),
@@ -480,7 +483,7 @@ where
                 )
             })
             .collect();
-        let trait_functions = self.generate_items(source, inner_items);
+        let trait_functions = self.process_inner_definitions(source, inner_items);
 
         iter::once(PILItem::Definition(
             symbol,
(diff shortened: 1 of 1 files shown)


pil-analyzer/src/statement_processor.rs
```diff
@@ -426,110 +426,169 @@ where
...
-            return iter::once(PILItem::Definition(
-                symbol,
-                Some(FunctionValueDefinition::TypeDeclaration(enum_decl.clone())),
-            ))
-            .chain(var_items)
+        match value {
+            Some(value) => match value {
+                FunctionDefinition::TypeDeclaration(enum_decl) => {
+                    assert_eq!(symbol_kind, SymbolKind::Other());
+                    self.handle_enum_declaration_symbol(source, name, symbol, enum_decl)
```
Schaeff, line 433:
```suggestion
                    self.handle_enum_symbol_declaration(source, name, symbol, enum_decl)
```
or
```suggestion
                    self.handle_enum_declaration(source, name, symbol, enum_decl)
```
maybe?

pil-analyzer/src/statement_processor.rs
```diff
@@ -426,110 +426,169 @@ where
...
-                );
-                PILItem::Definition(var_symbol, Some(value))
-            });
-            return iter::once(PILItem::Definition(
-                symbol,
-                Some(FunctionValueDefinition::TypeDeclaration(enum_decl.clone())),
-            ))
-            .chain(var_items)
+        match value {
+            Some(value) => match value {
```
chriseth, line 430: the two matches can be combined, can't they?

pil-analyzer/src/statement_processor.rs
```diff
@@ -426,110 +426,169 @@ where
...
+        &mut self,
+        source: SourceRef,
+        name: String,
+        symbol: Symbol,
+        enum_decl: EnumDeclaration<parsed::Expression>,
+    ) -> Vec<PILItem> {
+        let enum_decl = self.process_enum_declaration(enum_decl);
+        let shared_enum_decl = Arc::new(enum_decl.clone());
+
+        let inner_items = enum_decl
```
chriseth, line 457: Maybe it would be better to move this code into the process_enum_declaration function (which is tiny) instead of creating a new function with a similar name?

pil-analyzer/src/statement_processor.rs
```diff
@@ -576,21 +599,49 @@ where
         TypeProcessor::new(self.driver, type_vars)
     }
 
-    fn process_enum_declaration(
```
chriseth, line 579: I think we should keep this name, since it's the Statement _Processor_

pil-analyzer/src/statement_processor.rs
```diff
@@ -426,110 +426,146 @@ where
...
+        let expression = self
+            .expression_processor(&Default::default())
+            .process_array_expression(value);
+        assert!(type_scheme.is_none() || type_scheme == Some(Type::Col.into()));
+        let value = FunctionValueDefinition::Array(expression);
+
+        vec![PILItem::Definition(symbol, Some(value))]
+    }
+
+    fn generate_items(
```
chriseth, line 545: Can you find a more specific name for this function and also document it?

pil-analyzer/src/statement_processor.rs
```diff
@@ -577,20 +613,48 @@ where
...
+            .map(|variant| {
+                (
+                    vec![name.clone(), variant.name.clone()],
+                    FunctionValueDefinition::TypeConstructor(
+                        shared_enum_decl.clone(),
+                        variant.clone(),
+                    ),
+                )
+            })
+            .collect();
```
chriseth, line 648: Does this need to be a vector or can it be an iterator?

gzanitti: I don't think so, but at least I simplify everything around to try to avoid some clones and make it cleaner

chriseth (approved):

PR #1801, opened 2024-09-13
Plonky3 Keccak

qwang98:
Adapted from Plonky3 Keccak. Uses prover functions for all witness generation. Inputs and outputs are both in 16-bit limbs, so works for BabyBear and small fields.

For quick iteration, currently the operation is `preimage[0], preimage[1], ..., preimage[99] -> output[0], output[1], ..., output[99]`. Will change to only take an input pointer and an output pointer in another PR.

Two things to note:
1. Main machine needs to be degree 64 or more, so that Keccak block can run at least twice (24 rows per block), or the machine is detected as an VM and weird bugs happen.
2. For some reason, when calling the Keccak submachine twice or more in the main machine, witgen marks existing value as `Incomplete` and generates a weird bug.


std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    //         for limb in 0..U64_LIMBS {
+    //             builder
+    //                 .when(not_final_step.clone())
+    //                 .when_transition()
+    //                 .assert_eq(local.preimage[y][x][limb], next.preimage[y][x][limb]);
+    //         }
+    //     }
+    // }
+
+    array::new(100, |i| active_row * (preimage[i] - preimage[i]') = 0);
```
chriseth, line 147:
```suggestion
    array::map(preimage, |p| active_row * (p - p') = 0);
```

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    // a b c xor3
+    // 0 0 0  0
+    // 0 0 1  1
+    // 0 1 0  1
+    // 0 1 1  0
+    // 1 0 0  1
+    // 1 0 1  0
+    // 1 1 0  0
+    // 1 1 1  1
+    array::new(320, |i| {
```
chriseth, line 175: Can you use map_enumerated here?

chriseth: Or actually I think it would make sense to have a helper function that already computes x and z - it is used multiple times below.

chriseth: Oh and the same helper function could do the two-dimensional access to c and c_prime.

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+        let x = i / 64;
+        let z = i % 64;
+        xor3(c[i], c[((x + 4) % 5) * 64 + z], c[((x + 1) % 5) * 64 + ((z + 63) % 64)]) - c_prime[i] = 0
```
chriseth, line 178:
```suggestion
        xor3(
            c[i],
            c[((x + 4) % 5) * 64 + z],
            c[((x + 1) % 5) * 64 + ((z + 63) % 64)]
        ) - c_prime[i] = 0
```

std/machines/hash/keccakf16.asm
chriseth: But maybe the last one is too much...

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    // }
+
+    array::map(a_prime, |i| force_bool(i));
+    array::new(100, |i| {
+        let y = i / 20;
+        let x = (i / 4) % 5;
+        let limb = i % 4;
+        let get_bit: int -> expr = |z| xor3(a_prime[y * 320 + x * 64 + z], c[x * 64 + z], c_prime[x * 64 + z]);
+        let computed_limb: expr = utils::fold(16, |z| get_bit((limb + 1) * 16 - 1 - z), 0, |acc, e| (acc * 2 + e));
+        computed_limb - a[i] = 0
```
chriseth, line 216:
```suggestion
        a[i] = utils::fold(16, |z| get_bit((limb + 1) * 16 - 1 - z), 0, |acc, e| (acc * 2 + e))
```

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    //     }
+    // }
+
+    array::map(a_prime, |i| force_bool(i));
+    array::new(100, |i| {
+        let y = i / 20;
+        let x = (i / 4) % 5;
+        let limb = i % 4;
+        let get_bit: int -> expr = |z| xor3(a_prime[y * 320 + x * 64 + z], c[x * 64 + z], c_prime[x * 64 + z]);
+        let computed_limb: expr = utils::fold(16, |z| get_bit((limb + 1) * 16 - 1 - z), 0, |acc, e| (acc * 2 + e));
```
chriseth, line 215: What about
```suggestion
        array::reverse(array::new(16, |z| get_bit((limb + 1) * 16 + z)))
```
?
(and add `reverse` to `std::array`)

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    //         for limb in 0..U64_LIMBS {
+    //             let computed_limb = (limb * BITS_PER_LIMB..(limb + 1) * BITS_PER_LIMB)
+    //                 .rev()
+    //                 .fold(AB::Expr::zero(), |acc, z| acc.double() + get_bit(z));
+    //             builder.assert_eq(computed_limb, local.a_prime_prime[y][x][limb]);
+    //         }
+    //     }
+    // }
+
+    let andn: expr, expr -> expr = |a, b| (1 - a) * b;
```
chriseth, line 260: Move to the definition of xor?

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    array::new(100, |i| {
+        let y = i / 20;
+        let x = (i / 4) % 5;
+        let limb = i % 4;
+
+        let get_bit: int -> expr = |z| {
+            xor(b(x, y, z), andn(b((x + 1) % 5, y, z), b((x + 2) % 5, y, z)))
+        };
+        let computed_limb: expr = utils::fold(16, |z| get_bit((limb + 1) * 16 - 1 - z), 0, |acc, e| (acc * 2 + e));
+        computed_limb - a_prime_prime[i] = 0
```
chriseth, line 271:
```suggestion
        a_prime_prime[i] = utils::fold(16, |z| get_bit((limb + 1) * 16 - 1 - z), 0, |acc, e| (acc * 2 + e))
```

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    //         });
+    //     let a_prime_prime_0_0_limb = local.a_prime_prime[0][0][limb];
+    //     builder.assert_eq(computed_a_prime_prime_0_0_limb, a_prime_prime_0_0_limb);
+    // }
+
+    array::map(a_prime_prime_0_0_bits, |i| force_bool(i));
+
+    array::new(4, |limb| {
+        let computed_a_prime_prime_0_0_limb = utils::fold(16, |z| a_prime_prime_0_0_bits[(limb + 1) * 16 - 1 - z], 0, |acc, e| (acc * 2 + e));
+        computed_a_prime_prime_0_0_limb - a_prime_prime[limb] = 0
```
chriseth, line 314:
```suggestion
        a_prime_prime[limb] = utils::fold(16, |z| a_prime_prime_0_0_bits[(limb + 1) * 16 - 1 - z], 0, |acc, e| (acc * 2 + e))
```

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    //             acc.double() + local.a_prime_prime_0_0_bits[z]
+    //         });
+    //     let a_prime_prime_0_0_limb = local.a_prime_prime[0][0][limb];
+    //     builder.assert_eq(computed_a_prime_prime_0_0_limb, a_prime_prime_0_0_limb);
+    // }
+
+    array::map(a_prime_prime_0_0_bits, |i| force_bool(i));
+
+    array::new(4, |limb| {
+        let computed_a_prime_prime_0_0_limb = utils::fold(16, |z| a_prime_prime_0_0_bits[(limb + 1) * 16 - 1 - z], 0, |acc, e| (acc * 2 + e));
```
chriseth, line 313: The fold with `acc * 2, e` could also be a helper function `bits_to_value` or something.

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    //             ((a_limb >> bit_in_limb) & 1) != 0
+    //         });
+    //         row.c[x][z] = F::from_bool(a.fold(false, |acc, x| acc ^ x));
+    //     }
+    // }
+
+    let query_c: int, int, int -> int = query |x, limb, bit_in_limb| utils::fold(5, |y| (int(eval(a[y * 20 + x * 4 + limb])) >> bit_in_limb) & 0x1, 0, |acc, e| acc ^ e);
+
+    query |row| {
+        let _ = array::new(320, |i| {
```
chriseth, line 441: `array::map_enumerated`?

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    //         let bit = (*acc & 1) != 0;
+    //         *acc >>= 1;
+    //         Some(bit)
+    //     })
+    //     .collect();
+    // for (i, bit) in row.a_prime_prime_0_0_bits.iter_mut().enumerate() {
+    //     *bit = F::from_bool(val_bits[i]);
+    // }
+
+    let query_a_prime_prime_0_0_bits: int, int -> int = query |limb, bit_in_limb| (int(eval(a_prime_prime[limb])) >> bit_in_limb) & 0x1;
```
chriseth, line 550: Wouldn't it be easier to inline this?

std/machines/hash/keccakf16.asm
qwang98: Done

std/machines/hash/keccakf16.asm
qwang98: Done

std/machines/hash/keccakf16.asm
qwang98: Done.

std/machines/hash/keccakf16.asm
qwang98: Done

std/machines/hash/keccakf16.asm
qwang98: Done

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    // let next: &KeccakCols<AB::Var> = (*next).borrow();
+
+    // let first_step = local.step_flags[0];
+    // let final_step = local.step_flags[NUM_ROUNDS - 1];
+    // let not_final_step = AB::Expr::one() - final_step;
+
+    let first_step: expr = step_flags[0]; // aliasing instead of defining a new fixed column
+    let final_step: expr = step_flags[NUM_ROUNDS - 1];
+    // Equals to 1 for all non-last rows in all 24-row blocks. Forced to zero for the very last row of the submachine (if partial block).
+    col fixed active_row = [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0]* + [0];
```
georgwiese, line 108: I'm not sure about this name. How about `internal_row`? Or `in_progress`?

If I understand this right, the only purpose is to activate a copy from some result in the current row to the input of the next row. Which we don't want to happen in the last row of the block *and* the last row of the trace. Think this could be explained in the comment.

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+
+    col witness operation_id;
+
+    let NUM_ROUNDS: int = 24;
+
+    // pub struct KeccakCols<T> {
```
georgwiese, line 26: Think these comments are nice, but maybe mention a link to the source code with some acknowledgement to Plonky3 then?

std/machines/hash/keccakf16.asm
qwang98: Yes. DOne

std/machines/hash/keccakf16.asm
qwang98: Yes. That's the right idea behind this column.

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,576 @@
...
+    //             local.c[(x + 1) % 5][(z + 63) % 64].into(),
+    //         );
+    //         let c_prime = local.c_prime[x][z];
+    //         builder.assert_eq(c_prime, xor);
+    //     }
+    // }
+
+    array::map(c, |i| force_bool(i));
+    let xor: expr, expr -> expr = |a, b| a + b - 2*a*b;
+    let xor3: expr, expr, expr -> expr = |a, b, c| a + b + c - 2*(a*b + b*c + c*a) + 4*a*b*c;
```
georgwiese, line 165:
```suggestion
    let xor3: expr, expr, expr -> expr = |a, b, c| xor(xor(a, b), c);
```

This would be equivalent, no? Think it's clearer.

std/machines/hash/keccakf16.asm
qwang98: Done

std/machines/hash/keccakf16.asm
qwang98: Yes.

qwang98: Done by defining `xor3 = xor(xor(a, b), c)`

std/machines/hash/keccakf16.asm
qwang98: Acknowledgement added.

std/machines/hash/keccakf16.asm
qwang98: Done and integrated

std/machines/hash/keccakf16.asm
qwang98: Created `bits_to_limb_be`

qwang98:
Ready for a final review :)

Then will decide on whether to remove Plonky3 code comments + rename variables.

std/machines/hash/keccakf16.asm
onurinanc: Maybe, you could use `unchanged_until(preimage, 1 - first_step);` in here

qwang98: You right!

qwang98: Done

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,608 @@
...
+    //         for limb in 0..U64_LIMBS {
+    //             builder
+    //                 .when(not_final_step.clone())
+    //                 .when_transition()
+    //                 .assert_eq(local.preimage[y][x][limb], next.preimage[y][x][limb]);
+    //         }
+    //     }
+    // }
+
+    array::map(preimage, |p| unchanged_until(p, internal_row));
```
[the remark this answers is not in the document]
georgwiese: Another thing that should work is:
```rs
col fixed is_last = [0]* + [1];
// I think this can be inlined, as the new name basically describes how it is defined.
// let latch_row_or_last = final_step + is_last;
```

Because the block size is not a power of 2, both conditions will never be true at the same time (even if they were, it would work, because `unchanged_until` enforces the value to be unchanged as long as the condition is nonzero).

Plonky3 has a "native" `is_last` column, so with this approach we *could* save one fixed column (if we would detect that; currently, we'd create an `is_last` fixed column anyway).

georgwiese (commented):
Nice, impressive work!

I went through the constraints and it's pretty clear that this is a direct port of Plonky3's implementation. The only thing that is ore complicated in our code is that they use multi-dimensional arrays, but you used only flat arrays. I think we can change that? I think it would help with readability, do you agree?

I also checked that the recent bug fixes of https://github.com/Plonky3/Plonky3/pull/409 are included in this port.

I haven't yet checked the prover functions and would be curious to see if we really need all of them or if at least some of the steps can be figured out by the solver. But it seems like the test is not running through ATM, so I'll wait for that to be fixed to have a closer look.

qwang98:
> Nice, impressive work!
> 
> I went through the constraints and it's pretty clear that this is a direct port of Plonky3's implementation. The only thing that is ore complicated in our code is that they use multi-dimensional arrays, but you used only flat arrays. I think we can change that? I think it would help with readability, do you agree?
> 
> I also checked that the recent bug fixes of [Plonky3/Plonky3#409](https://github.com/Plonky3/Plonky3/pull/409) are included in this port.
> 
> I haven't yet checked the prover functions and would be curious to see if we really need all of them or if at least some of the steps can be figured out by the solver. But it seems like the test is not running through ATM, so I'll wait for that to be fixed to have a closer look.

Discussed with Chris on the possibility of using nested arrays to generate column. Currently experiencing some bugs when nested array interact with prover functions in witgen, but we'll see.

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,608 @@
...
+    // }
+
+    pol commit preimage[5 * 5 * 4];
+    pol commit a[5 * 5 * 4];
+    pol commit c[5 * 64];
+    pol commit c_prime[5 * 64];
+    pol commit a_prime[5 * 5 * 64];
+    pol commit a_prime_prime[5 * 5 * 4];
+    pol commit a_prime_prime_0_0_bits[64];
+    pol commit a_prime_prime_prime_0_0_limbs[4];
```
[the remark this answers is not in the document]
qwang98: Agree.

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,608 @@
...
+    // builder.assert_bool(local.export);
+
+    // force_bool(export);
+
+    // // If this is not the final step, the export flag must be off.
+    // builder
+    //     .when(not_final_step.clone())
+    //     .assert_zero(local.export);
+
+    // not_final_step * export = 0;
```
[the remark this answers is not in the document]
qwang98: The export flag is really just the latch column `final_step`. Added a comment.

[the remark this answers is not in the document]
qwang98: FYI here's the definition of the `export` flag in Plonky3:

```
    /// The `i`th value is set to 1 if we are in the `i`th round, otherwise 0.
    pub step_flags: [T; NUM_ROUNDS],

    /// A register which indicates if a row should be exported, i.e. included in a multiset equality
    /// argument. Should be 1 only for certain rows which are final steps, i.e. with
    /// `step_flags[23] = 1`.
    pub export: T,
```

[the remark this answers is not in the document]
qwang98: To be honest I'm still a bit confused by the use of the selector column(s) in permutations. Are they submachine inputs from the main machine? Are they also needed in lookups?

std/machines/hash/keccakf16.asm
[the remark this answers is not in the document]
qwang98: > I think `unchanged_until` means the opposite: If the flag is 0, it needs to be unchanged. So it should be `unchanged_until(p, 1 - internal_row)`, right?
> 
> But maybe then it makes sense to materialize the opposite:
> 
> ```rust
> col fixed latch_row_or_last = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]* + [1];
> ```

Yes you just catched a bug and the test passes after fixing it. :)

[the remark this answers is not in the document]
qwang98: > Another thing that should work is:
> 
> ```rust
> col fixed is_last = [0]* + [1];
> // I think this can be inlined, as the new name basically describes how it is defined.
> // let latch_row_or_last = final_step + is_last;
> ```
> 
> Because the block size is not a power of 2, both conditions will never be true at the same time (even if they were, it would work, because `unchanged_until` enforces the value to be unchanged as long as the condition is nonzero).
> 
> Plonky3 has a "native" `is_last` column, so with this approach we _could_ save one fixed column (if we would detect that; currently, we'd create an `is_last` fixed column anyway).



> Another thing that should work is:
> 
> ```rust
> col fixed is_last = [0]* + [1];
> // I think this can be inlined, as the new name basically describes how it is defined.
> // let latch_row_or_last = final_step + is_last;
> ```
> 
> Because the block size is not a power of 2, both conditions will never be true at the same time (even if they were, it would work, because `unchanged_until` enforces the value to be unchanged as long as the condition is nonzero).
> 
> Plonky3 has a "native" `is_last` column, so with this approach we _could_ save one fixed column (if we would detect that; currently, we'd create an `is_last` fixed column anyway).

This does looks cleaner to me and I'm doing this with the combined flag inlined. To clarify, we are not saving a column because we delete `internal_row` and add `is_last`.

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,608 @@
...
+        let x = (i / 4) % 5;
+        let limb = i % 4;
+        let get_bit: int -> expr = |z| xor3(a_prime[y * 320 + x * 64 + z], c[x * 64 + z], c_prime[x * 64 + z]);
+        let limb_bits_be: expr[] = array::reverse(array::new(16, |z| get_bit(limb * 16 + z)));
+        a[i] = bits_to_value_be(limb_bits_be)
+    });
+
+    // // xor_{i=0}^4 A'[x, i, z] = C'[x, z], so for each x, z,
+    // // diff * (diff - 2) * (diff - 4) = 0, where
+    // // diff = sum_{i=0}^4 A'[x, i, z] - C'[x, z]
```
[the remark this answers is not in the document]
qwang98: This is basically saying that if five bits xor to one bit, then the difference between the sum of the five bits and the one bit is either 0, 2, or 4.

For example, 0 = 1 xor 1 xor 0 xor 0 xor 0, so diff = (1 + 1 + 0 + 0 + 0) - 0 = 2.

It can be proven mathematically that these two statements are equivalent.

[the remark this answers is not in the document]
georgwiese: Yes, I understood!

I think it's pretty smart, because now 5 XORs are computed with one degree-3 constraint, whereas constraining it "naively" (repeatedly applying `xor = |a, b| a + b - 2*a*b`) would yield a degree-6 constraint, which would be above the degree bound and we'd have to add witness columns to store intermediate results.

std/machines/hash/keccakf16.asm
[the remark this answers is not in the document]
qwang98: As discussed will move these to a new PR as currently prover functions for nested columns aren't stable and causes witgen bugs.

std/machines/hash/keccakf16.asm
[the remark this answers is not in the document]
georgwiese: They are not needed in lookups.

If a machine is linked via a lookup argument, it would look like this:
```
main.instr_foo $ [ <operation_id>, arg1, arg2, ... ]
in
submachine.LATCH $ [ submachine.operation_id, submachine.arg1, submachine.arg2 ]
```
(In your case, the "latch" is the `final_step` column.)

The lookup is a subset argument, all tuples on the LHS (those rows where `main.instr_foo == 1`) must be in the set defined by the RHS (those rows where `submachine.LATCH == 1`. One way to satisfy this is to create a block for each call of the main machine and then fill the remaining rows with copies of the first block.

The permutation argument is stricter: You can't have anything on the RHS that does not appear in the LHS. To still satisfy the constraint, we need to give the prover the flexibility to indicate which "latch rows" belong to actual blocks, and which are just padded blocks.

For this, we introduce a new selector witness column, constrained to be 0 or 1. The constraint is then:
```
main.instr_foo $ [ <operation_id>, arg1, arg2, ... ]
is
(submachine.selector * submachine.LATCH) $ [ submachine.operation_id, submachine.arg1, submachine.arg2 ]
```

The term `(submachine.selector * submachine.LATCH)` can only evaluate to 1 if `submachine.LATCH == 1`, but the prover additionally has the ability to remove rows from the RHS by setting the selector to zero.

You'd need a new selector column for each permutation, which is why there is an array of selector columns.

std/machines/hash/keccakf16.asm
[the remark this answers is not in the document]
qwang98: I moved the range constraints next to the column definitions though, so it's easier to see.

qwang98:
All should be fixed. I'm opening up two issues, one for the nested column syntax and another for the witgen bug from instantiating the submachine one than once.

std/machines/hash/keccakf16.asm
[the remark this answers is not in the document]
georgwiese: Yes, it's the same number of fixed columns! But, as mentioned, Plonky3 has a built-in `is_last` (because it is often needed to prevent wrapping), so we could remove this column again when compiling to the Plonky3 backend. But that's future work :)

georgwiese (commented):
Awesome! Looks good to me!

Two things until this can be merged from my side:
- Move the test to `test_data/std`, because it is testing a machine from the standard library.
- Add a test to `pipeline/tests/powdr_std.rs`, so it's actually run in the CI! I think this should add a `#[ignore = "Too slow"]`, but it should still be picked up by one of the `test_slow` jobs.

qwang98:
> Awesome! Looks good to me!
> 
> Two things until this can be merged from my side:
> 
> * Move the test to `test_data/std`, because it is testing a machine from the standard library.
> * Add a test to `pipeline/tests/powdr_std.rs`, so it's actually run in the CI! I think this should add a `#[ignore = "Too slow"]`, but it should still be picked up by one of the `test_slow` jobs.

Both done :)

qwang98:
Identified additional CI bug when running the test with BabyBear field. It error out with field modulus overflow error in witgen stage, so need additional debug efforts.

std/machines/hash/keccakf16.asm
```diff
@@ -0,0 +1,609 @@
...
+
+    let query_a_prime_prime_prime_0_0_limbs: int -> int = query |limb| 
+        int(eval(a_prime_prime[limb])) ^ 
+        ((int(eval(utils::sum(NUM_ROUNDS, |r| expr(RC[r]) * step_flags[r]))) >> (limb * 16)) & 0xffff);
+
+    query |row| {
+        let _ = array::new(4, |limb| {
+            provide_value(a_prime_prime_prime_0_0_limbs[limb], row, fe(query_a_prime_prime_prime_0_0_limbs(limb)));
+        });
+    };
```
georgwiese, line 608:
```suggestion
    let query_a_prime_prime_prime_0_0_limbs: int, int -> int = query |round, limb| 
        int(eval(a_prime_prime[limb])) ^ ((RC[round] >> (limb * 16)) & 0xffff);

    query |row| {
        let _ = array::new(4, |limb| {
            provide_value(a_prime_prime_prime_0_0_limbs[limb], row, fe(query_a_prime_prime_prime_0_0_limbs(row % NUM_ROUNDS, limb)));
        });
    };
```

This is simpler and fixes the BabyBear case!

The problem was that `expr(RC[r])` requires `RC[r]` to fit into the field, because constants in expressions are always field elements.

georgwiese (approved):
Nice work! 🚀

PR #1805, opened 2024-09-13
Array reverse helper

qwang98:
Really small. Just thought might not be appropriate to put in Keccak.


std/array.asm
```diff
@@ -29,6 +29,9 @@ let<T1, T2> fold: T1[], T2, (T2, T1 -> T2) -> T2 = |arr, initial, folder| std::u
 // use this function to flatten an array of arrays.
 let<T: Add + FromLiteral> sum: T[] -> T = |arr| fold(arr, 0, |a, b| a + b);
 
+/// Returns the reversed array.
```
chriseth, line 32: Maybe not in between sum and product?

georgwiese (approved):
LGTM

std/array.asm
qwang98: Moved

chriseth (approved):

PR #1808, opened 2024-09-16
Avoid an array copy in the input and output of the poseidon_gl function.

lvella:
The original array is mutated in-place instead.

### DIFF
--- a/riscv-runtime/src/hash.rs
+++ b/riscv-runtime/src/hash.rs
@@ -7,7 +7,7 @@ use powdr_riscv_syscalls::Syscall;
 const GOLDILOCKS: u64 = 0xffffffff00000001;
 
 /// Calls the low level Poseidon PIL machine, where the last 4 elements are the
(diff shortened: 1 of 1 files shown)


riscv-runtime/src/hash.rs
```diff
@@ -1,27 +1,30 @@
...
 use core::mem;
 
 use powdr_riscv_syscalls::Syscall;
 
 const GOLDILOCKS: u64 = 0xffffffff00000001;
 
 /// Calls the low level Poseidon PIL machine, where the last 4 elements are the
-/// "cap" and the return value is placed in data[0:4].
+/// "cap" and the return value is placed in data[..4] and the reference to this
+/// sub-array is returned.
```
leonardoalt, line 11:
```suggestion
/// "cap", the return value is placed in data[..4] and the reference to this
/// sub-array is returned.
```

leonardoalt (approved):

PR #1810, opened 2024-09-17
Port of our Poseidon GL precompile into BabyBear field.

lvella:
The same algorithm as poseidon_gl, but adjusted for BabyBear. For simplicity, I decided try this simple port first instead of implementing p3-poseidon2, thus this is not compatible to what we use in Plonky3.

Depends on PR #1790.


std/machines/hash/poseidon_bb.asm
```diff
@@ -0,0 +1,204 @@
...
+
+    // Multiply with MDS Matrix
+    let dot_product = |v1, v2| array::sum(array::zip(v1, v2, |v1_i, v2_i| v1_i * v2_i));
+    let c = array::map(M, |M_row_i| dot_product(M_row_i, b));
+
+    // Copy c to state in the next row
+    array::zip(state, c, |state, c| (state' - c) * (1-LAST) = 0);
+
+    // In the last row, the first OUTPUT_SIZE elements of the state should equal output
+    array::zip(output, state, |output, state| LASTBLOCK * (output - state) = 0);
```
chriseth, line 200: Please apply this fix accordingly, otherwise it will break once #1806 is merged:
https://github.com/powdr-labs/powdr/pull/1806/commits/12ffb9f6121dc034ea7004e6b1d658172844c331

leonardoalt (approved):

PR #1812, opened 2024-09-18
Traits: Support for generic types

gzanitti:
This PR adds support for generic types in traits.

To do this, it generates a path between the references of those traits with generic variables and those references that call them. Then, it uses this mapping to calculate the types to be replaced.


pil-analyzer/src/traits_resolver.rs
```diff
@@ -60,26 +91,176 @@ impl<'a> TraitsResolver<'a> {
...
+                if let Expression::Reference(
+                    _,
+                    Reference::Poly(PolynomialReference {
+                        name,
+                        type_args: Some(type_args),
+                        ..
+                    }),
+                ) = e
+                {
+                    if (!type_args.is_empty()) && (name != *symbol) {
```
Schaeff, line 179: Why the second condition?

pil-analyzer/src/traits_resolver.rs
```diff
@@ -60,26 +91,176 @@ impl<'a> TraitsResolver<'a> {
...
+                        references.push((name, type_args));
+                    }
+                }
+            });
+            for (name, type_args) in references {
+                result
+                    .entry(name)
+                    .or_insert_with(Vec::new)
+                    .push((*symbol, type_args));
+            }
```
Schaeff, line 189: Do we need to collect in `references` first? Why not insert into `result` directly?

pil-analyzer/src/traits_resolver.rs
```diff
@@ -60,26 +91,176 @@ impl<'a> TraitsResolver<'a> {
...
+        }
+
+        Self::combine_type_vars_paths(&result, defined_traits)
+    }
+
+    /// Solve the initial asociation and builds a dictionary that maps the
+    /// names of traits containing generic types to their represented types.
+    fn combine_type_vars_paths(
+        input: &HashMap<&String, Vec<(&String, &Vec<Type>)>>,
```
Schaeff, line 198: The type is complex and the name is too generic here, what it `input`?

pil-analyzer/src/traits_resolver.rs
```diff
@@ -60,26 +91,176 @@ impl<'a> TraitsResolver<'a> {
...
+            match input.get(t) {
+                Some(children) => {
+                    for (father, type_arg) in children {
+                        if type_arg.iter().any(|ty| matches!(ty, Type::TypeVar(_))) {
+                            let new_type = Self::solve_type_vars(input, father, type_arg);
+                            result.insert(t.to_string(), new_type);
+                        }
+                    }
+                }
+                None => continue,
```
Schaeff, line 213: Is this reachable because the keys of `input` are just the traits which contain generic types here? Related to the comment above, having all traits in both would simplify the code here as the `Some` case seems to work fine for empty vectors

Schaeff (commented):
Good addition, however I suspect the code could be simplified, and the intent would become clearer

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -361,7 +361,7 @@ impl PILAnalyzer {
     /// Creates and returns a map for every referenced trait and every concrete type to the
     /// corresponding trait implementation function.
     fn resolve_trait_impls(&mut self) -> HashMap<String, HashMap<Vec<Type>, Arc<Expression>>> {
-        let mut trait_solver = TraitsResolver::new(&self.implementations);
+        let mut trait_solver = TraitsResolver::new(&self.implementations, &self.definitions);
 
         let mut resolve_references = |expr: &Expression| {
```
chriseth, line 366: I think all the following code in this function should be simplified to this:
```rust
        // Now start resolving at the "entry points", i.e. all proof items
        // and all columns.
        // Other symbols might be unused, so we do not need to resolve them.
        let entry_points = self
            .definitions
            .values()
            // TODO simplify?
            .filter_map(|(symbol, value)| (symbol.kind == SymbolKind::Other()).then(|| value))
            .flatten()
            .flat_map(|fv| fv.children())
            .chain(&self.proof_items);
        for expr in entry_points {
            trait_solver.resolve_trait_references(&expr)?;
        }
        Ok(trait_solver.solved_impls())
```

The idea here is the following: We supply the trait resolver with all the "entry points" into the program. It follows the call graph and resolves all references to trait functions it finds. This has the advantage that we do not resolve functions (for example in the stdlib) that are never called.

Since we need to process the call graph anyway, I think this is a big simplification.

pil-analyzer/src/traits_resolver.rs
```diff
@@ -17,16 +24,38 @@ type SolvedImpl = ((String, Vec<Type>), Arc<Expression>);
 pub struct TraitsResolver<'a> {
     trait_impls: &'a HashMap<String, Vec<TraitImplementation<Expression>>>,
```
chriseth, line 25: Please document all the attributes.

gzanitti:
Discarded in favor of  https://github.com/powdr-labs/powdr/pull/1938

PR #1815, opened 2024-09-19
fix macro example

leonardoalt:


chriseth (approved):

lvella:
This will render terribly on the generated html docs. That said, I don't know another way to fix.

PR #1816, opened 2024-09-19
Rename trait functions

gzanitti:
Since we also need name and type pairs for structs, it makes sense to rename `TraitFunction` to something more general (`NamedType` in this case).

There are also `NamedExpressions` and everything could be combined into a `NamedElement` or something similar that covers both cases. Personally I prefer this option, but the discussion is open.

### DIFF
--- a/parser/src/powdr.lalrpop
+++ b/parser/src/powdr.lalrpop
@@ -728,7 +728,7 @@ PatternList: Vec<Pattern> = {
 // ---------------------------- Trait/Impl Declaration -----------------------------
 
 TraitDeclaration: TraitDeclaration<Expression> = {
-    "trait" <name:Identifier> <type_vars:("<" <TraitVars> ">")> "{" <functions:NamedTypes> "}" => TraitDeclaration { name, type_vars, functions }
+    "trait" <name:Identifier> <type_vars:("<" <TraitVars> ">")> "{" <functions:TraitFunctions> "}" => TraitDeclaration { name, type_vars, functions }
 }
 
 TraitVars: Vec<String> = {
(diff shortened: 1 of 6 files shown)


parser/src/powdr.lalrpop
```diff
@@ -719,20 +719,20 @@ PatternList: Vec<Pattern> = {
...
-    "trait" <name:Identifier> <type_vars:("<" <TraitVars> ">")> "{" <functions:TraitFunctions> "}" => TraitDeclaration { name, type_vars, functions }
+    "trait" <name:Identifier> <type_vars:("<" <TraitVars> ">")> "{" <functions:NamedTypes> "}" => TraitDeclaration { name, type_vars, functions }
 }
 
 TraitVars: Vec<String> = {
     <mut list:( <TypeVar> "," )*> <end:TypeVar> ","?  => { list.push(end); list }
 }
 
-TraitFunctions: Vec<TraitFunction<Expression>> = {
+NamedTypes: Vec<NamedType<Expression>> = {
```
chriseth, line 729: Can you move this to a more generic position?

chriseth (commented):
Looks good apart from the organization of the grammar file.

gzanitti:
I am thinking that we may also have to change the definition of the `NamedType` in the grammar, since in `Traits` we needed functions but now it is no longer the case.

```
NamedType: NamedType<Expression> = {
    <name:Identifier> ":" <params:TypeTermList<Expression>> "->" <value:TypeTermBox<Expression>> => NamedType { name, ty: Type::Function(FunctionType{params, value}) }
}
```

chriseth:
Then let's keep the two different grammar rules but combine the rust types?

chriseth:
Oh and we certainly have to allow functions, but we can just use `AStruct { a: (int -> int), b: (AStruct -> bool) }`, right?

chriseth (approved):

PR #1820, opened 2024-09-20
Simplify parsed identities and introduce the concept of proof item.

chriseth:
At the on-site, we introduced operators for all lookup-like constraints. This means that `[x, y] in s $ [a, b];` is not parsed as a lookup identity any more, but just as a regular expression with binary operators. The whole concept of a lookup or identity as a rust type is now only present after the condenser has run. Because of that, we can remove a lot of code that was concerned with parsed identities.

Since prover functions were introduced, we can have both constraints and prover functions at statement level. Because of that I extended the concept of "identity" (which we partly renamed to "constraint" already) to "proof item". A proof item is either a constraint or a prover function. Later on, we might also include fixed columns, challenges, etc.

### DIFF
--- a/pil-analyzer/src/condenser.rs
+++ b/pil-analyzer/src/condenser.rs
@@ -25,7 +25,7 @@ use powdr_ast::{
         types::{ArrayType, Type},
         visitor::{AllChildren, ExpressionVisitable},
         ArrayLiteral, BlockExpression, FunctionCall, FunctionKind, LambdaExpression,
-        LetStatementInsideBlock, Number, Pattern, TypedExpression, UnaryOperation,
+        LetStatementInsideBlock, Number, Pattern, SourceReference, TypedExpression, UnaryOperation,
     },
 };
 use powdr_number::{BigUint, FieldElement};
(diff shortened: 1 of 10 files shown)


pil-analyzer/src/condenser.rs
```diff
@@ -36,14 +36,13 @@ use crate::{
     statement_processor::Counters,
 };
 
-type ParsedIdentity = Identity<parsed::SelectedExpressions<Expression>>;
 type AnalyzedIdentity<T> = Identity<SelectedExpressions<AlgebraicExpression<T>>>;
```
gzanitti, line 39: Given that there is only one type of identity now, does it make sense to keep this? Or is it because clippy complains about the length?

gzanitti (approved):
Only one question of little relevance, the rest looks good to me :)

pil-analyzer/src/condenser.rs
chriseth: I think it makes sense, yes. I wanted to rename it to Identity but of course then we get a collision. What we could do is change the definition if Identity to be non-generic, but I would say that's another PR.

gzanitti (approved):

PR #1825, opened 2024-09-23
Use jit compiler to generate fixed column values.

chriseth:
Tries to call the JIT compiler on each fixed column when generating the values of the fixed columns. Uses the interpreter we already have for all the columns where JIT compilation failed.


chriseth:
Before this is merged, we have to tune a parameter (on the degree) to tell when to start compiling and when just interpreting is always faster.

Of course we could also compile in the background and run the interpreter until the compilation is finished...

executor/src/constant_evaluator/mod.rs
```diff
@@ -1,194 +1,57 @@
...
             for (index, (name, id)) in poly.array_elements().enumerate() {
-                let index = poly.is_array().then_some(index as u64);
-                let range = poly.degree.unwrap();
-                let values = range
-                    .iter()
-                    .map(|degree| generate_values(analyzed, degree, &name, value, index))
-                    .collect::<Vec<_>>()
-                    .into();
-                assert!(fixed_cols.insert(name, (id, values)).is_none());
+                if !fixed_cols.contains_key(&(name.clone(), id)) {
```
lvella, line 32: It would avoid one hit in the `HashMap` to use `if let hash_map::Entry::Vacant(entry) = fixed_cols.entry(...)` here.

executor/src/constant_evaluator/jit_compiler.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use powdr_ast::analyzed::{Analyzed, PolyID};
+use powdr_number::FieldElement;
+use rayon::iter::{IntoParallelIterator, ParallelIterator};
+
+use super::VariablySizedColumn;
+
+/// Tries to just-in-time compile all fixed columns in `analyzed`
+/// and then evaluates those functions to return `VariablySizedColumn`s.
+/// Ignoreds all columns where the compilation fails.
```
lvella, line 11:
```suggestion
/// Ignores all columns where the compilation fails.
```

executor/src/constant_evaluator/jit_compiler.rs
```diff
@@ -0,0 +1,62 @@
...
+                column_values.into(),
+            )
+        })
+        .collect()
+}
+
+fn symbols_to_compile<T>(analyzed: &Analyzed<T>) -> Vec<&str> {
+    analyzed
+        .constant_polys_in_source_order()
+        .into_iter()
```
lvella, line 57: Is it too common to call `.constant_polys_in_source_order().into_iter()`? I saw it two times in this PR. If so, maybe the first function should return the iter directly instead of collecting into a `Vec<_>`.

executor/src/constant_evaluator/interpreter.rs
lvella, line 1: Is this file copy-pasted from `constant_evaluator/mod.rs`? I skipped it.

lvella (changes requested):

executor/src/constant_evaluator/interpreter.rs
chriseth: yes, mostly

lvella (approved):

lvella:
I guess your other PR broke this one and now you have to merge with main.

lvella (approved):

PR #1826, opened 2024-09-23
Simplify permutation using prover functions similiar to lookups

onurinanc:
Similar to #1802, we can also simplify permutation in PIL using prover functions.


test_data/std/permutation_via_challenges_ext.asm
```diff
@@ -7,14 +7,8 @@ use std::prover::challenge;
...
 
-    let alpha1: expr = challenge(0, 1);
-    let alpha2: expr = challenge(0, 2);
-    let beta1: expr = challenge(0, 3);
-    let beta2: expr = challenge(0, 4);
-    let alpha = Fp2::Fp2(alpha1, alpha2);
-    let beta = Fp2::Fp2(beta1, beta2);
-
     col fixed first_four = [1, 1, 1, 1, 0, 0, 0, 0];
+    col fixed one_minus_first_four = [0, 0, 0, 0, 1, 1, 1, 1];
```
chriseth, line 11: Let's fix the condenser first before we merge things like that into main.

pil-analyzer/src/condenser.rs
```diff
@@ -1033,6 +1035,41 @@ fn try_value_to_expression<T: FieldElement>(value: &Value<'_, T>) -> Result<Expr
...
             ),
+
+            AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation { left, op, right }) => {
+                Expression::BinaryOperation(
+                    SourceRef::unknown(),
+                    BinaryOperation {
+                        left: Box::new(try_value_to_expression(&Value::<T>::Expression(
+                            (**left).clone(),
+                        ))?),
+                        op: match op {
```
chriseth, line 1046: You should be able to just use `op.into()` here.

pil-analyzer/src/condenser.rs
```diff
@@ -1033,6 +1035,41 @@ fn try_value_to_expression<T: FieldElement>(value: &Value<'_, T>) -> Result<Expr
...
+                        ))?),
+                    },
+                )
+            }
+
+            AlgebraicExpression::UnaryOperation(AlgebraicUnaryOperation { op, expr }) => {
+                Expression::UnaryOperation(
+                    SourceRef::unknown(),
+                    UnaryOperation {
+                        op: match op {
```
chriseth, line 1063: Same here, safer to just use`.into()`

pil-analyzer/src/condenser.rs
```diff
@@ -1033,6 +1035,41 @@ fn try_value_to_expression<T: FieldElement>(value: &Value<'_, T>) -> Result<Expr
                     type_args: None,
                 }),
             ),
+
+            AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation { left, op, right }) => {
+                Expression::BinaryOperation(
+                    SourceRef::unknown(),
+                    BinaryOperation {
+                        left: Box::new(try_value_to_expression(&Value::<T>::Expression(
```
chriseth, line 1043: If you extract this function into a specialized conversion function just for AlgebraicExpression, this code should get a lot simpler.

pil-analyzer/src/condenser.rs
onurinanc: How can I use into() here?

onurinanc: Is it just `(*op).into()`?

std/protocols/permutation.asm
```diff
@@ -53,12 +54,11 @@ let compute_next_z: Fp2<expr>, Fp2<expr>, Fp2<expr>, Constr -> fe[] = query |acc
...
+/// WARNING: This function can currently not be used multiple times since
+/// the used challenges would overlap
+/// TODO: Implement this for an array of constraints
 ///
 /// # Arguments:
-/// - acc: A phase-2 witness column to be used as the accumulator. If 2 are provided, computations
-///        are done on the F_{p^2} extension field.
-/// - alpha: A challenge used to compress the LHS and RHS values
-/// - beta: A challenge used to update the accumulator
 /// - permutation_constraint: The permutation constraint
```
chriseth, line 62: The `Arguments` part of the documentation provides no information, so I think we can just delete it.

std/protocols/permutation.asm
```diff
@@ -53,12 +54,11 @@ let compute_next_z: Fp2<expr>, Fp2<expr>, Fp2<expr>, Constr -> fe[] = query |acc
...
 ///
 /// # Returns:
```
chriseth, line 64: It now returns nothing. The constraints are just added.

chriseth (approved):

PR #1827, opened 2024-09-23
Struct parsing

gzanitti:
Struct parsing and some minor changes to maintain consistency.


analysis/src/machine_check.rs
```diff
@@ -355,10 +355,24 @@ impl TypeChecker {
                         asm::SymbolValue::Expression(e) => {
                             res.insert(ctx.clone().with_part(&name), Item::Expression(e));
                         }
-                        asm::SymbolValue::TypeDeclaration(enum_decl) => {
+                        asm::SymbolValue::TypeDeclaration(asm::TypeDeclaration::Enum(
```
chriseth, line 358: Is this code here converting anything?

ast/src/parsed/asm.rs
```diff
@@ -69,8 +69,8 @@ pub enum SymbolValue {
     Module(Module),
     /// A generic symbol / function.
     Expression(TypedExpression),
-    /// A type declaration (currently only enums)
-    TypeDeclaration(EnumDeclaration<Expression>),
+    /// A type declaration (currently only enums or structs)
```
chriseth, line 72:
```suggestion
    /// A type declaration (enums or struct)
```

ast/src/parsed/asm.rs
```diff
@@ -98,14 +98,56 @@ pub enum SymbolValueRef<'a> {
...
+impl From<StructDeclaration<Expression>> for SymbolValue {
+    fn from(value: StructDeclaration<Expression>) -> Self {
+        SymbolValue::TypeDeclaration(TypeDeclaration::Struct(value))
+    }
+}
+
+impl Children<Expression> for TypeDeclaration {
+    fn children(&self) -> Box<dyn Iterator<Item = &Expression> + '_> {
+        match self {
+            TypeDeclaration::Enum(e) => Box::new(e.variants.iter().flat_map(|v| v.children())),
```
chriseth, line 130: Please implement Children also on EnumDeclaration and StructDeclaraton.

ast/src/parsed/mod.rs
```diff
@@ -151,6 +152,13 @@ impl PilStatement {
                         .map(move |v| (name, Some(&v.name), SymbolCategory::TypeConstructor)),
                 ),
             ),
+            PilStatement::StructDeclaration(_, StructDeclaration { name, fields, .. }) => Box::new(
+                once((name, None, SymbolCategory::Type)).chain(
+                    fields
```
chriseth, line 157: I still don't think the fields should be externally accessible like this. If you return them here, then something like `StructName::field_name` is valid, but I don't think it should.

ast/src/parsed/mod.rs
```diff
@@ -239,6 +249,57 @@ impl Children<Expression> for PilStatement {
...
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct StructDeclaration<E = u64> {
+    pub name: String,
+    pub type_vars: TypeBounds,
+    pub fields: Vec<NamedType<E>>,
+}
+
+impl<E> StructDeclaration<E> {
+    pub fn type_of_field(&self, name: &str) -> Option<&Type<E>> {
```
chriseth, line 260: Not sure how this is used, but it could actually return a TypeScheme that contains the type vars of the struct

ast/src/parsed/mod.rs
```diff
@@ -239,6 +249,57 @@ impl Children<Expression> for PilStatement {
...
+impl<R> Children<Expression<R>> for StructDeclaration<u64> {
+    fn children(&self) -> Box<dyn Iterator<Item = &Expression<R>> + '_> {
+        Box::new(empty())
+    }
+    fn children_mut(&mut self) -> Box<dyn Iterator<Item = &mut Expression<R>> + '_> {
+        Box::new(empty())
+    }
+}
+
+impl<R> Children<Expression<R>> for NamedExpression<Box<Expression<R>>> {
```
chriseth, line 285: Please put this below NamedExpression

ast/src/parsed/display.rs
```diff
@@ -663,6 +664,45 @@ impl<Expr: Display> Display for SelectedExpressions<Expr> {
...
+}
+
+impl<E: Display> Display for StructExpression<E> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        write!(
+            f,
+            "{}{}",
+            self.name,
+            if self.fields.is_empty() {
+                "".to_string()
```
chriseth, line 692: I think we should still use `{}` in this case, i.e. `StructName{}`,  otherwise it gets too complicated.

ast/src/parsed/display.rs
```diff
@@ -663,6 +664,45 @@ impl<Expr: Display> Display for SelectedExpressions<Expr> {
...
+            } else {
+                format!(
+                    "{{ {} }}",
+                    self.fields
+                        .iter()
+                        .map(|named_expr| format!("{named_expr}"))
+                        .format(", ")
```
chriseth, line 699:
```suggestion
                        .join(", ")
```
(from itertools)

ast/src/parsed/display.rs
```diff
@@ -1012,14 +1053,22 @@ pub fn format_type_scheme_around_name<E: Display, N: Display>(
     if let Some(type_scheme) = type_scheme {
         format!(
             "{} {name}: {}",
-            type_scheme.type_vars_to_string(),
```
chriseth, line 1015: I think this function still makes sense (it could call the one below)

ast/src/parsed/mod.rs
```diff
@@ -1150,6 +1212,28 @@ impl<E> Children<E> for IfExpression<E> {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct StructExpression<E = Expression<NamespacedPolynomialReference>> {
+    pub name: String,
```
chriseth, line 1217: Just thinking aloud here: Shouldn't this be a SymbolPath? I know this will might create problems later on where we only work with direct Strings...

analysis/src/machine_check.rs
chriseth: This just looks like a lot of code for something very simple...

ast/src/parsed/mod.rs
```diff
@@ -151,6 +152,12 @@ impl PilStatement {
                         .map(move |v| (name, Some(&v.name), SymbolCategory::TypeConstructor)),
                 ),
             ),
+            PilStatement::StructDeclaration(
+                _,
+                StructDeclaration {
+                    name, fields: _, ..
```
chriseth, line 158:
```suggestion
                    name, ..
```

ast/src/parsed/mod.rs
```diff
@@ -1150,6 +1215,28 @@ impl<E> Children<E> for IfExpression<E> {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct StructExpression<E = Expression<NamespacedPolynomialReference>> {
+    pub name: SymbolPath,
```
chriseth, line 1220: Oh wait, I think this needs to be NamespacedPolynomialReference or something, because we can have stuff like `super::mod::StructName::<int, int[]>`.

ast/src/parsed/mod.rs
```diff
@@ -465,6 +525,7 @@ pub enum Expression<Ref = NamespacedPolynomialReference> {
     MatchExpression(SourceRef, MatchExpression<Self>),
     IfExpression(SourceRef, IfExpression<Self>),
     BlockExpression(SourceRef, BlockExpression<Self>),
+    StructExpression(SourceRef, StructExpression<Self>),
```
chriseth, line 528: concerning https://github.com/powdr-labs/powdr/pull/1827/files#r1775775696 - it might work to use `StructExpression(SourceRef, StructExpression<Ref>)` here (and then use `fields: Vec<NamedExpression<Box<Expression<Ref>>>>` or something below).

importer/src/path_canonicalizer.rs
```diff
@@ -900,6 +913,11 @@ fn check_expression(
                 None => Ok(()),
             }
         }
+        Expression::StructExpression(_, StructExpression { name: _, fields }) => fields
+            .iter()
+            .try_for_each(|NamedExpression { name: _, body }| {
```
chriseth, line 918: Are you resolving the name here?

ast/src/parsed/mod.rs
```diff
@@ -1216,18 +1213,18 @@ impl<E> Children<E> for IfExpression<E> {
 }
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
-pub struct StructExpression<E = Expression<NamespacedPolynomialReference>> {
-    pub name: SymbolPath,
-    pub fields: Vec<NamedExpression<Box<E>>>,
+pub struct StructExpression<E = NamespacedPolynomialReference> {
```
chriseth, line 1216: Please call this one `Ref`

importer/src/path_canonicalizer.rs
```diff
@@ -312,6 +312,13 @@ fn canonicalize_inside_expression(
                     canonicalize_inside_pattern(pattern, path, paths);
                 })
             }
+            Expression::StructExpression(_, StructExpression { name, .. }) => {
+                if let Some(n) = paths.get(&path.clone().join(name.path.clone())) {
+                    name.path = n.relative_to(&Default::default());
+                } else {
+                    assert!(name.path.try_to_identifier().is_some());
```
chriseth, line 319: You copied the code from the Expression::Reference case, right?

This is almost correct, the difference is that in this case, it cannot be a local variable - that would be something like `let x = 8; let y = x{a: 9}`. So you can just use `unwrap()` in lien 316.

chriseth: or `assert!(false)` here...

importer/src/path_canonicalizer.rs
```diff
@@ -913,11 +919,17 @@ fn check_expression(
                 None => Ok(()),
             }
         }
-        Expression::StructExpression(_, StructExpression { name: _, fields }) => fields
-            .iter()
-            .try_for_each(|NamedExpression { name: _, body }| {
-                check_expression(location, body, state, type_vars, local_variables)
-            }),
+        Expression::StructExpression(source_ref, StructExpression { name, fields }) => {
+            let path_to_check = location.clone().join(name.path.clone());
```
chriseth, line 930: Why not `check_path_try_prelude` here?

chriseth: (like for the Expression::Reference case)

importer/test_data/struct_expression.expected.asm
```diff
@@ -0,0 +1,10 @@
+mod other {
+    enum E1 {
+        A,
+    }
+    struct S1 {
+        a: int,
+        b: other::E1,
+    }
+}
+let s1 = other::S1{ a: 1, b: other::E1::A };
```
chriseth, line 10: Please also add a line that uses `Option` without explicitly importing it from the prelude.

chriseth:
Nice, almost there!

importer/test_data/struct_expression.expected.asm
gzanitti: It seems that `prelude` is not included in these test cases. So I had to define a small fake std on top of the test.

chriseth: No, please don't do this, it does not test what we wanted to test here. If it doesn't work as an importer test, then instead please use a pipeline test.

chriseth: Or you can call `add_std` in the test function like in `load_dependencies_and_resolve'.

The point of testing the prelude is that it should work even without explicit import and without adding `std::prelude`.

importer/test_data/struct_expression.asm
```diff
@@ -7,9 +17,16 @@ mod other {
...
+    }
 }
 
 use other::E1;
 use other::S1;
+use other::S2;
 
-let s1 = S1{ a: 1, b: E1::A };
\ No newline at end of file
+let s1 = S1{ a: 1, b: E1::A };
```
chriseth, line 31: Please also add a struct to the prelude. If you don't then the change above to `check_path_try_prelude` is not tested.

importer/src/path_canonicalizer.rs
```diff
@@ -1130,6 +1132,35 @@ mod tests {
         assert_eq!(res.map_err(|e| e.message().to_string()), expected);
     }
 
+    fn expect_with_std(path: &str, expected: Result<(), &str>) {
+        let input_path = PathBuf::from("./test_data/")
```
chriseth, line 1136: Please don't just copy the code here.

chriseth (approved):

PR #1828, opened 2024-09-23
Query input/output rework

pacheco:
This PR does a few things:
- unify `Input/DataIdentifier` into a single `Input` query that takes (channel,idx) and returns a field element
- Output query takes (channel, fe)
- change the related prover functions to reflect this
This interface is used by, but not the same, as the input/output for the riscv machine.
How to use these to input/output bytes or serialized data is a job of the runtime implementation.


pil-analyzer/src/evaluator.rs
```diff
@@ -1400,18 +1386,21 @@ fn evaluate_builtin_function<'a, T: FieldElement>(
...
             };
-            symbols.get_input_from_channel(
+            symbols.input_from_channel(
                 u32::try_from(channel).unwrap(),
                 usize::try_from(index).unwrap(),
             )?
         }
-        BuiltinFunction::OutputByte => {
+        BuiltinFunction::OutputToChannel => {
             let byte = arguments.pop().unwrap();
```
chriseth, line 1395: If we abstract it, we should do it properly:
byte -> value
fd -> channel

pipeline/src/lib.rs
```diff
@@ -97,8 +97,8 @@ pub fn serde_data_to_query_callback<T: FieldElement>(
     move |query: &str| -> Result<Option<T>, String> {
         let (id, data) = parse_query(query)?;
         match id {
-            "DataIdentifier" => {
-                let [index, cb_channel] = data[..] else {
```
chriseth, line 101: Was this a bug or did you reverse it in other places as well?

riscv/src/runtime.rs
```diff
@@ -235,20 +235,10 @@ impl Runtime {
         // Base syscalls
         r.add_syscall(
             Syscall::Input,
-            [
-                // TODO this is a quite inefficient way of getting prover inputs.
```
chriseth, line 239: Is this comment still valid?

pipeline/src/lib.rs
pacheco: this was like this here, but the other way around in the prover functions. I changed everything to (channel, idx)

riscv/src/runtime.rs
pacheco: hmmm good catch, i think it is: we use 3 asm instructions to read an input value, maybe it can be improved later. I'll put the comment back

pil-analyzer/src/evaluator.rs
pacheco: oh right, ill see if i missed this in other places

chriseth (dismissed):

chriseth (approved):

PR #1835, opened 2024-09-24
Simplify bus interactions

onurinanc:
I would like to simplify bus interactions similar to lookup and permutations

To complete it, I need to do one of the following:
- If the soundness is not broken in the bus via permutation/lookup, I'm going to delete the final sum to 0 check constraint
- If the soundness is going to be a problem, I need to find a way for the final sum to 0 check, inside the bus protocol

### DIFF
--- a/test_data/std/bus_permutation_via_challenges_ext.asm
+++ b/test_data/std/bus_permutation_via_challenges_ext.asm
@@ -1,19 +1,10 @@
 use std::prelude::Query;
 use std::convert::fe;
 use std::protocols::permutation_via_bus::permutation;
-use std::protocols::permutation_via_bus::compute_next_z_send_permutation;
-use std::protocols::permutation_via_bus::compute_next_z_receive_permutation;
 use std::math::fp2::Fp2;
 use std::prover::challenge;
 
 machine Main with degree: 8 {
-
-    let alpha1: expr = challenge(0, 1);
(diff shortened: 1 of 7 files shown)


test_data/std/bus_permutation_via_challenges_ext.asm
```diff
@@ -29,42 +20,10 @@ machine Main with degree: 8 {
...
-    u1' = u1_next;
-    u2' = u2_next;
+    permutation(1, permutation_constraint);
 
-    is_first' * (z1 + u1) = 0;
-    is_first' * (z2 + u2) = 0;
+    //let is_first: col = std::well_known::is_first;
+//
+    //is_first' * (z1 + u1) = 0;
+    //is_first' * (z2 + u2) = 0;
```
onurinanc, line 28: @chriseth Any idea how I can simplify this part to use the accumulators, which are inside the bus_interaction, also for the permutation_via_bus so that I can use these constraints `is_first' * (z1 + u1) = 0;` and
    `is_first' * (z2 + u2) = 0;`

onurinanc: Because z and u are the accumulators, which I’m going to move them into bus_interaction function. But the permutation function consists of just bus_send(); and bus_receive(); (so accumulators are used inside de bus_interaction, not in the permutation function)

PR #1836, opened 2024-09-24
Target test for JIT compiler

chriseth:


test_data/asm/expand_fixed.asm
```diff
@@ -0,0 +1,141 @@
...
+enum ArrayTerm {
+    Repeat(int[]),
+    Once(int[])
+}
+
+let MORE_THAN_ONE_REPEATED_ERROR = "unsolvable because more than one term is repeated";
+let NON_REPEATED_SIZE_EXCEEDS_DEGREE_ERROR = "non repeated array terms do not fit in degree";
+
+// returns the total size of the repeated array in this array expression
+let compute_length_of_repeated_part: ArrayTerm[], (-> int) -> Result<int, string> = |terms, degree| {
```
pacheco, line 18: could this take the degree directly instead of the function that returns the degree?

test_data/asm/expand_fixed.asm
```diff
@@ -0,0 +1,141 @@
...
+            Result::Err(e) => Result::Err(e),
+        })
+    });
+
+    match (constant_value, found_repeated_item) {
+        // we found more than one repeated term, error out
+        (_, Result::Err(s)) => Result::Err(s),
+        // This is not strict enough and would allow expressions whose size does not match the degree
+        // We should fix this by ensuring that:
+        // - the minimum size of the expanded expression is smaller than the minimum degree of the machine
```
pacheco, line 82: what happens when the expansion cannot get to the degree?
e.g.: `[1,2]* + [3]`
this would always lead to an "odd" sized expansion no?

test_data/asm/expand_fixed.asm
```diff
@@ -0,0 +1,141 @@
...
+    }
+};
+
+enum OffsetSearch {
+    Offset(int),
+    Value(int),
+}
+
+// returns a function representing the array expression
+let expand: ArrayTerm[], (-> int) -> Result<(int -> int), string> = |terms, degree| {
```
pacheco, line 52: I found this function quite hard to read... so much nesting, specially in the match.
Maybe some intermediate variables are possible?
Also, we don't support importing enum variants right? 
I think `Ok(None)` would remove some of the noise compared to `Result::Ok(Option::None)`

test_data/asm/expand_fixed.asm
```diff
@@ -0,0 +1,141 @@
...
+    // return early if all terms are constant with the same value
+    let (constant_value, found_repeated_item) = std::array::fold(terms, (Result::Ok(Option::None), Result::Ok(false)), |(r, found), term| {
```
pacheco, line 54: are there better names for `r`, `c`, `e`? had to keep double checking what they were

chriseth:
@pacheco this is not the actual code for the standard library, so I think this does not need a thorough review. This PR is just meant to check that we can JIT-compile this kind of complex code.

pacheco:
@chriseth okey! does it make sense to add a small comment to the rust side (or asm file) about the test's purpose? Doesn't look that obvious just from the code or test name.

chriseth:
I mean it's in the test directory.... But yeah, I will add a comment.

pacheco:
seems silly but the comment actually helps here hehe, lgtm

pacheco (approved):

PR #1837, opened 2024-09-24
Functions as expressions

chriseth:
Adds support for symbols of function type being assigned from an expression that is not a lambda expression.

This case is a bit more complicated because those symbols are stored as lazy statics and not as direct `fn <name>` functions. Because of that, any reference to it needs to be dereferenced. At the same time, references to functions cannot be dereferenced.

We need both flavours since only `fn`-functions can be used for recursive calls in rust.


jit-compiler/src/codegen.rs
```diff
@@ -28,32 +28,35 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
 
-    /// Request a symbol to be compiled. The code can later be retrieved
-    /// via `compiled_symbols`.
+    /// Requests code for the symbol to be generated and returns the
+    /// code that returns the value of the symbol, which is either
+    /// the escaped name of the symbol or a dereferenced name of a lazy static.
+    /// The code can later be retrieved via `compiled_symbols`.
     /// In the error case, `self` can still be used to compile other symbols.
-    pub fn request_symbol(&mut self, name: &str) -> Result<(), String> {
+    pub fn request_symbol(&mut self, name: &str) -> Result<String, String> {
```
pacheco, line 36: Found the comment for this method hard to process.
How about:
```
Try to generate code for the requested symbol.
On success, returns code for referencing the symbol.
On failure, `self` can still be used to compile other symbols.
```

pacheco: or `code for referencing the symbol's value`, if that is more accurate

pacheco: Also, i wonder if `compiled_symbols` would be better named as `generated_code`?

jit-compiler/src/codegen.rs
```diff
@@ -85,35 +88,55 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
+                assert!(type_scheme.vars.is_empty());
+                let ty = if type_scheme.ty == Type::Col {
+                    Type::Function(FunctionType {
+                        params: vec![Type::Int],
+                        value: Box::new(Type::Fe),
+                    })
+                } else {
+                    type_scheme.ty.clone()
+                };
+                format!(
```
pacheco, line 131: i'd add a small comment here about the reason for the lazy static

jit-compiler/src/codegen.rs
```diff
@@ -125,7 +148,7 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
         expr: &Expression,
```
pacheco, line 148: can't this function receive the `LambdaExpression` directly to avoid the `match` and `panic`?

jit-compiler/src/codegen.rs
```diff
@@ -316,6 +341,28 @@ fn map_type(ty: &Type) -> String {
...
+        "std::check::panic" => "(s: &str) -> ! { panic!(\"{s}\"); }".to_string(),
+        "std::field::modulus" => {
+            let modulus = T::modulus();
+            format!("() -> ibig::IBig {{ ibig::IBig::from(\"{modulus}\") }}")
+        }
+        "std::convert::fe" => "(n: ibig::IBig) -> FieldElement {\n    <FieldElement as PrimeField>::BigInt::try_from(n.to_biguint().unwrap()).unwrap().into()\n}"
+            .to_string(),
+        _ => return None,
+    };
+    Some(format!("fn {}{code}", escape_symbol(symbol)))
```
pacheco, line 363: `std::array::len` is missing from the `is_builtin` method.
What about we add `assert!(is_builtin)` here before returning to catch the bug?

chriseth: -> https://github.com/powdr-labs/powdr/pull/1847

jit-compiler/src/codegen.rs
```diff
@@ -288,6 +296,19 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
             "Compiling statements inside blocks is not yet implemented: {s}"
         ))
     }
+
+    /// Returns true if a reference to the symbol needs a deref operation or not.
+    fn symbol_needs_deref(&self, symbol: &str) -> bool {
```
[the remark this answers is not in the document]
chriseth: Hm, not really. The return logic is complicated here, and it is only called from one function, so adding another function `symbol_reference` would not make much sense.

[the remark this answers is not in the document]
pacheco: i didn't mean creating a new function, just rename this one and make it already return the needed code directly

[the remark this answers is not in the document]
chriseth: ok, let me try again

pacheco (approved):

PR #1838, opened 2024-09-24
JIT support for match expressions.

chriseth:
This PR supports match expression, while most of the work regards patterns.

Unfortunately, we cannot use rust patterns directly: The types are too incompatible. For example a `match ("abc", 7)  { ("abc", 7) => ...}` cannot be directly translated, since the rust types here are `String` and `ibig::IBig` and they cannot be used with these literals like that, at least not directly and in all circumstances.

The way it is implemented here is that each pattern is compiled to code that is supposed to evaluate to an Option. If the pattern matches, the Option is a Some-value that contains the values that are assigned to the variables in the pattern. The function also returns a string containing the variable names. The function calls itself recursively on recursive data structures.


chriseth:
Unfortunately, it looks like match will not directly map to rust's match because the types are just too incompatible (for example you cannot match a literal number and a BigInt, nor a string literal and a String).
Not yet sure how to solve this, we probably need something like
```
{
  let scrutinee = <scrutinee>;
  if let Some((var1, var2)) = try_pattern_1(scrutinee) {
    ...
  } else if let Some((var3, var4)) = try_pattern2(scrutinee) {
   ...
  } else {
  ..
  }
}
```

jit-compiler/src/codegen.rs
```diff
@@ -283,6 +284,26 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
                         .unwrap_or_default()
                 )
             }
+            Expression::MatchExpression(_, MatchExpression { scrutinee, arms }) => {
+                // TODO try to find a solution where we do not introduce a variable
```
chriseth, line 288: Not to be done in this PR.

jit-compiler/src/codegen.rs
```diff
@@ -315,6 +336,88 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
 
+/// Used for patterns in match and let statements:
+/// `value_name` is an expression string that is to be matched against `pattern`.
+/// Returns a rust pattern string (tuple of new variables, might be nested) and a code string
+/// that, when executed, returns an Option with the values for the new variables if the pattern
+/// matched `value_name` and `None` otherwise.
+///
+/// So if `let (vars, code) = check_pattern("x", pattern)?;`, then the return value
+/// can be used like this: `if let Some({vars}) = ({code}) {{ .. }}`
+fn check_pattern(value_name: &str, pattern: &Pattern) -> Result<(String, String), String> {
```
pacheco, line 347: idk about the name, maybe `match_arm_pattern`?

jit-compiler/src/codegen.rs
```diff
@@ -315,6 +336,88 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
+                        return None;
+                    }
+                    Some(if ellipsis_seen {
+                        let i_rev = items.len() - i;
+                        (format!("{value_name}[{value_name}.len() - {i_rev}]"), item)
+                    } else {
+                        (format!("{value_name}[{i}]"), item)
+                    })
+                })
+                .map(|(access_name, item)| {
```
pacheco, line 396: i wonder if this `map` being merged into the preceding `filter_map` makes it easier to read?

jit-compiler/src/codegen.rs
```diff
@@ -315,6 +336,88 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
+                .join(", ");
+            (
+                format!("({})", vars.join(", ")),
+                format!("(|| Some(({inner_code})))()"),
+            )
+        }
+        Pattern::Array(_, items) => {
+            let mut vars = vec![];
+            let mut ellipsis_seen = false;
+            let inner_code = items
```
pacheco, line 381: small comment saying what `inner_code` will be?

jit-compiler/src/codegen.rs
```diff
@@ -329,6 +432,14 @@ fn format_unsigned_integer(n: &BigUint) -> String {
     }
 }
 
+fn format_signed_integer(n: &BigInt) -> String {
+    if let Ok(n) = BigUint::try_from(n) {
+        format_unsigned_integer(&n)
+    } else {
+        format_signed_integer(&-(n.clone()))
```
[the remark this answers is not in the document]
chriseth: Because it's a bug! :)

jit-compiler/src/codegen.rs
chriseth: It will also be used in `let` statements.

jit-compiler/src/codegen.rs
chriseth: The issue is that the first is a filter, i.e. returns Option and the second one returns Result. I think it's too complicate to combine both.

jit-compiler/src/codegen.rs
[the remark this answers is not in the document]
lvella: I don't think this function should be recursive.

pacheco (dismissed):

georgwiese (dismissed):
LGTM

jit-compiler/src/codegen.rs
```diff
@@ -436,7 +441,7 @@ fn format_signed_integer(n: &BigInt) -> String {
     if let Ok(n) = BigUint::try_from(n) {
         format_unsigned_integer(&n)
     } else {
-        format_signed_integer(&-(n.clone()))
+        format!("-{}", format_signed_integer(&-(n.clone())))
```
lvella, line 444: Is clone really necessary?

Anyway, maybe it is better to avoid recursion:
```suggestion
        format!("-{}", format_unsigned_integer(&BigUint::try_from(-n).unwrap())
```

pacheco (approved):

PR #1839, opened 2024-09-24
JIT: Fix issue with Windows

gzanitti:
It makes use of `libloading` to solve Windows compilation problems.


gzanitti:
Already implemented in the original branch

PR #1841, opened 2024-09-25
JIT: Support block statements

chriseth:


jit-compiler/src/codegen.rs
```diff
@@ -312,9 +312,27 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
+                };
+                let value = self.format_expr(value)?;
+                let var_name = "scrutinee__";
+                let ty = ty
+                    .as_ref()
+                    .map(|ty| format!(": {}", map_type(ty)))
+                    .unwrap_or_default();
+
+                let (vars, code) = check_pattern(var_name, pattern)?;
+                // TODO if we want to explicitly specify the type, we need to exchange the non-captured
```
chriseth, line 330: This is just a note in case we need it in the future if rust cannot determine the type.

lvella: I don't understand this comment. It is talking about the type of `{vars}` , I assume. But what are the non-captured parts that must be exchanged by `()`?

lvella (approved):
Approved, unless you agree the TODO is a little cryptic.

chriseth:
I'll create a follow-up PR to extend the todo.

PR #1842, opened 2024-09-25
Condenser support on complex algebraic expressions to expressions

onurinanc:
Add support for Binary & Unary operations on AlgebraicExpressions -> Expressions


pil-analyzer/src/condenser.rs
```diff
@@ -1033,6 +1065,15 @@ fn try_value_to_expression<T: FieldElement>(value: &Value<'_, T>) -> Result<Expr
                     type_args: None,
                 }),
             ),
+
+            AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation { left, op, right }) => {
```
chriseth, line 1069: I would suggest to put the full RHS of `Value::Expression(e) =>` into a function, because I could imagine the code then getting simpler and shorter.

And please return a `Result` instead of `unwrap()`ing it.

pil-analyzer/tests/condenser.rs
```diff
@@ -872,18 +871,15 @@ fn capture_not_supported() {
...
     let eval = -1;
+    col witness x;
     col witness y;
     {
-        let x = std::prelude::challenge(0, 4);
-        let y = std::prover::y;
-        let t = 2;
+        let t = std::prover::x + std::prover::y;
         query |i| {
-            std::prover::provide_value(y, i, std::prover::eval(x) + t);
```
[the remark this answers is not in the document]
onurinanc: What should I do instead? It doesn't work with that tests, expected gives the following result: `let _: fe = std::prover::eval(t);`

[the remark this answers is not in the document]
chriseth: I don't follow. Are you saying that
1) the original test did not test your changes
2) the original test is still failing even with your changes
?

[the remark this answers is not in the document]
onurinanc: First one, the original test did not test your changes since I think the test should test whether let = x + y; transforms into let t = std::prover::x + std::prover::y;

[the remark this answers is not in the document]
chriseth: Ah sorry! I thought this was the input, not the expectation :)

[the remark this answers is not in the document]
onurinanc: The expected part of the original test is the same as the test for capture_challenges_and_numbers, which seems it is quite unrelated with the capture_not_supported test? https://github.com/powdr-labs/powdr/blob/718c333ff960d5c7e4f254397cc3702a0636f736/pil-analyzer/tests/condenser.rs#L836-L854

pil-analyzer/src/condenser.rs
```diff
@@ -810,6 +812,85 @@ fn try_to_function_value_definition<T: FieldElement>(
     }))
 }
 
+fn try_algebraic_expression_to_expression<T: FieldElement>(
```
chriseth, line 815: I think this fits better closer to where it is used.

pil-analyzer/src/condenser.rs
```diff
@@ -810,6 +812,85 @@ fn try_to_function_value_definition<T: FieldElement>(
...
+                type_args: None,
+            }),
+        ),
+
+        AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation { left, op, right }) => {
+            Expression::BinaryOperation(
+                SourceRef::unknown(),
+                BinaryOperation {
+                    left: Box::new(
+                        try_value_to_expression(&Value::<T>::Expression((**left).clone())).unwrap(),
```
chriseth, line 836:
```suggestion
                        try_algebraic_expression_to_expression(left)?
```

chriseth: When you need to clone something of unbounded size, there should be a better solution. Not always, but most of the time.

chriseth: Also you could try `BinaryOperation{ ... }.into()` instead of constructing `Expression::BinaryOperation` explicitly.

pil-analyzer/src/condenser.rs
```diff
@@ -810,6 +812,85 @@ fn try_to_function_value_definition<T: FieldElement>(
...
+                },
+            )
+        }
+
+        AlgebraicExpression::Number(n) => Number {
+            value: n.to_arbitrary_integer(),
+            type_: Some(Type::Expr),
+        }
+        .into(),
+        _ => {
```
chriseth, line 886: What is left here now?

onurinanc: PublicReferences?

pil-analyzer/src/condenser.rs
```diff
@@ -939,6 +941,78 @@ fn shift_local_var_refs(e: &mut Expression, shift: u64) {
...
```
Schaeff, line 1008: References with `next` set to `true` like `x'` are also not covered by the previous cases, so the error message would be incorrect in that case.

Schaeff: Maybe we shouldn't use `_` here so that we have a clear view of what is not implemented yet.

pil-analyzer/src/condenser.rs
```diff
@@ -939,6 +941,78 @@ fn shift_local_var_refs(e: &mut Expression, shift: u64) {
...
 }
 
+fn try_algebraic_expression_to_expression<T: FieldElement>(
+    e: &AlgebraicExpression<T>,
+) -> Result<Expression, EvalError> {
+    Ok(match e {
+        AlgebraicExpression::Reference(AlgebraicReference {
+            name,
+            poly_id: _,
+            next: false,
```
chriseth, line 951: Wouldn't be too difficult to implement next:

```
let e = Expression::Reference{...}; // what we currently have
if next {
  UnaryOperation{ op: UnaryOperator::Next, expr: Box::new(e)}.into()
} else {
  e
}
```

onurinanc: Not sure where to put this, also, isn't the next already assigned as false?

chriseth: If you remove the `: false`, it matches both fals and true values for next: https://doc.rust-lang.org/reference/patterns.html#destructuring

pil-analyzer/src/condenser.rs
chriseth: Yes, and adding `todos` for the ones that the compiler complains about.

onurinanc: Instead of matching value? Do you want me to return `Result<Expression>` instead of `Result<Expression, EvalError>`? and adding todo! macros?

chriseth: I just want to see what is missing. If it's just public reference, then let's implement it. It's just Expression::PublicReference.

pil-analyzer/src/condenser.rs
```diff
@@ -939,6 +941,91 @@ fn shift_local_var_refs(e: &mut Expression, shift: u64) {
...
+                }),
+            );
+            if *next {
+                Expression::UnaryOperation(
+                    SourceRef::unknown(),
+                    UnaryOperation {
+                        op: parsed::UnaryOperator::Next,
+                        expr: Box::new(e),
+                    },
+                )
```
chriseth, line 967:
```suggestion
                    UnaryOperation {
                        op: parsed::UnaryOperator::Next,
                        expr: Box::new(e),
                    }.into()
```

chriseth (approved):

PR #1845, opened 2024-09-26
Import p3 implementation

Schaeff:
This PR is only a copy paste from https://github.com/powdr-labs/plonky3 with the following exceptions:
- only the files that differ from uni-stark are copied over. Any references to other modules are targeting upstream uni-stark
- the `preprocessed_width` function is moved to the `MultiStageAir` trait from `BaseAir` to avoid changes to `p3-air`
- upstream uni-stark is nostd. Our wrapper uses things like hashmaps and mutexes, so for now I commented out `nostd`
- `lib.rs` is a merge of the existing lib.rs and that of our uni-stark fork
- the `prove` function is removed, as it is only there for compatibility with uni-stark tests
- `FieldMerkleTreeMmcs` was [renamed](https://github.com/Plonky3/Plonky3/commit/84a13454ea9a9ea82489551f081e24072e1f3e98) to `MerkleTreeMmcs` in plonky3 since we forked it. Our configs use this type and the name is updated accordingly

TODO:
- [x] extract rust update to separate PR due to the new ways clippy complains

PR #1846, opened 2024-09-26
Upgrade Rust to 1.81

Schaeff:
This is a prerequisite for #1845 

TODO:
- [x] narrow down or remove the dead_code exception

### DIFF
--- a/backend/src/field_filter.rs
+++ b/backend/src/field_filter.rs
@@ -26,7 +26,9 @@ macro_rules! generalize_factory {
                 backend_options: crate::BackendOptions,
             ) -> std::result::Result<Box<dyn crate::Backend<F>>, crate::Error> {
                 use crate::Backend;
-                use std::{any::TypeId, boxed::Box, result::Result::Ok};
+                use powdr_ast::analyzed::Analyzed;
+                use powdr_executor::constant_evaluator::VariablySizedColumn;
+                use std::{any::TypeId, boxed::Box, result::Result::Ok, sync::Arc, string::String};
 
                 let result = match TypeId::of::<F>() {
                     $(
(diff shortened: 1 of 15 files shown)


backend/src/field_filter.rs
```diff
@@ -34,10 +34,10 @@ macro_rules! generalize_factory {
                             unsafe {
                                 let result = <$restricted_factory as BackendFactory<$supported_type>>::create(
                                     &$restricted_factory,
-                                    std::mem::transmute(pil), std::mem::transmute(fixed), output_dir, setup,
+                                    std::mem::transmute::<std::sync::Arc<powdr_ast::analyzed::Analyzed<F>>, std::sync::Arc<powdr_ast::analyzed::Analyzed<$supported_type>>>(pil), std::mem::transmute::<std::sync::Arc<std::vec::Vec<(std::string::String, powdr_executor::constant_evaluator::VariablySizedColumn<F>)>>, std::sync::Arc<std::vec::Vec<(std::string::String, powdr_executor::constant_evaluator::VariablySizedColumn<$supported_type>)>>>(fixed), output_dir, setup,
```
leonardoalt, line 37: 1) what

chriseth: Wait this evil macro was merged after all?

riscv/src/asm/mod.rs
```diff
@@ -1,3 +1,8 @@
+// this is required to please clippy which finds dome dead code in the
```
chriseth, line 1: But the generated parser already has `allow(all)` (in parser.rs) - where exactly is the dead code?

Schaeff: The parser has `allow(clippy::all)` but `dead_code` is a rustc lint not a clippy lint. I moved it there!

backend/src/field_filter.rs
lvella: > Wait this evil macro was merged after all?

The first `unsafe` solution got nuked by rust. The evil macro was the only PR in town.
https://github.com/rust-lang/rust/pull/120248#issuecomment-2340261183

chriseth (approved):

PR #1847, opened 2024-09-27
Reorg builtins

chriseth:


jit-compiler/src/compiler.rs
```diff
@@ -60,6 +59,23 @@ pub fn generate_glue_code<T: FieldElement>(
     Ok(format!("{PREAMBLE}\n{glue}\n",))
 }
 
+const PREAMBLE: &str = r#"
```
chriseth, line 62: will remove, this is from another pr

onurinanc (approved):
I haven't used ibig::Ibig before, just wondering why you prefer this over others, does this have a better performance?

chriseth:
We use it in `powdr_number` and it's quite small and still fast IMO

jit-compiler/src/codegen.rs
```diff
@@ -354,25 +354,43 @@ fn map_type(ty: &Type) -> String {
...
-    matches!(
-        symbol,
-        "std::check::panic" | "std::field::modulus" | "std::convert::fe"
-    )
+fn get_builtins<T: FieldElement>() -> &'static HashMap<String, String> {
+    static BUILTINS: OnceLock<HashMap<String, String>> = OnceLock::new();
+    BUILTINS.get_or_init(|| {
+        [
+            (
+                "std::array::len",
```
pacheco, line 362: is `std::convert::fe` missing on purpose?

chriseth: It was not working as it was written any way. Will be re-added in https://github.com/powdr-labs/powdr/pull/1834

PR #1848, opened 2024-09-27
Format numbers properly.

chriseth:


jit-compiler/src/codegen.rs
```diff
@@ -290,9 +300,18 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
     }
 }
 
-pub fn escape_symbol(s: &str) -> String {
-    // TODO better escaping
-    s.replace('.', "_").replace("::", "_")
+fn format_number(n: &BigUint) -> String {
```
Schaeff, line 303:
```suggestion
fn format_integer(n: &BigUint) -> String {
```
?

jit-compiler/src/codegen.rs
```diff
@@ -211,6 +218,9 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
                     BinaryOperator::ShiftLeft => {
                         format!("(({left}).clone() << u32::try_from(({right}).clone()).unwrap())")
                     }
+                    BinaryOperator::ShiftRight => {
+                        format!("(({left}).clone() >> usize::try_from(({right}).clone()).unwrap())")
```
Schaeff, line 222: `ShiftLeft` above uses `u32::try_from`, would be good to have consistency

jit-compiler/src/codegen.rs
```diff
@@ -175,12 +179,15 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
-                    _ => unreachable!(),
+                if *type_ == Type::Int {
+                    format_number(value)
+                } else {
+                    let value = u64::try_from(value).unwrap_or_else(|_| unimplemented!());
+                    match type_ {
+                        Type::Fe => format!("FieldElement::from({value}_u64)"),
+                        Type::Expr => format!("Expr::from({value}_u64)"),
+                        _ => unreachable!(),
+                    }
```
Schaeff, line 190: I find having a single match for all types clearer

jit-compiler/src/codegen.rs
chriseth: true, right somehow didn't work with u32, will check.

jit-compiler/src/codegen.rs
chriseth: so you would prefer doubling line 185?

Schaeff: Yes

chriseth: Done.

lvella (commented):
LGTM

jit-compiler/src/codegen.rs
chriseth: Changed both to usize, since this is what ibig implements.

Schaeff (approved):

PR #1849, opened 2024-09-27
make `Elem` private to the RISCV executor

pacheco:
Bootloader/continuations now handle field elements directly instead of `Elem`. 
The bootloader can't use u32 directly because it uses register memory from the previous chunk, which may be holding values larger than 32-bit depending where the execution stops (i.e., negative Elem::Binaries or Elem::Field).

### DIFF
--- a/riscv-executor/src/lib.rs
+++ b/riscv-executor/src/lib.rs
@@ -43,7 +43,7 @@ use crate::profiler::Profiler;
 const PC_INITIAL_VAL: usize = 2;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum Elem<F: FieldElement> {
+enum Elem<F: FieldElement> {
     /// Only the ranges of i32 and u32 are actually valid for a Binary value.
     /// I.e., [-2**31, 2**32).
     Binary(i64),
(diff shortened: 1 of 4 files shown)


pacheco:
the idea here is to hide executor implementation details from the outside, in preparation for the BB executor

riscv-executor/src/lib.rs
```diff
@@ -200,7 +183,7 @@ pub struct RegWrite<F: FieldElement> {
     row: usize,
     /// Index of the register in the register bank.
     reg_idx: u16,
-    val: Elem<F>,
+    val: F,
```
leonardoalt, line 186: I think this might not be ideal for performance. Most of the register operations are actually over u32/i64, and here we're forcing it to be always F.

riscv-executor/src/lib.rs
```diff
@@ -234,7 +217,7 @@ impl<F: FieldElement> ExecutionTrace<F> {
 
 pub struct TraceReplay<'a, F: FieldElement> {
     trace: &'a ExecutionTrace<F>,
-    regs: Vec<Elem<F>>,
+    regs: Vec<F>,
```
leonardoalt, line 220: Potentially same here.

riscv-executor/src/lib.rs
leonardoalt: Or maybe I'm tripping, but wanted to bring it up anyway.

pacheco: General register operations during execution are still over Elem. This just converts `i64 -> F` when storing the reg write operation in `Trace` mode. 
Note that these are only for the actual "asm write registers" (`pc` and `query_arg_*`). I think this won't make much difference but I can double check.

leonardoalt: Ah right, makes sense then.

leonardoalt (approved):

PR #1851, opened 2024-09-30
Run Keccakf16 only on CI.

chriseth:
On my machine, the newly added keccakf16 reparse test needs too much machine stack space, when run without the pr-tests profile. Since the pr-tests profile needs way longer to compile, I usually run my tests with the default profile.

This PR excludes the keccakf16 test from running outside of a CI environment.

### DIFF
--- a/pipeline/build.rs
+++ b/pipeline/build.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::env;
 use std::fs::File;
 use std::io::{BufWriter, Write};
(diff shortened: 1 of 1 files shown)


pipeline/build.rs
```diff
@@ -8,21 +9,28 @@ use walkdir::WalkDir;
 fn main() {
     build_book_tests("asm");
     build_book_tests("pil");
-    build_reparse_test("asm", "asm");
-    build_reparse_test("pil", "pil");
-    build_reparse_test("asm", "std");
+    build_reparse_test("asm", "asm", Default::default());
+    build_reparse_test("pil", "pil", Default::default());
+    let exclude = if std::env::var("CI").is_ok() {
```
lvella, line 14: I saw another test that does it, but I really don't like this solutions. This introduces more undocumented switched when trying to run the tests.

Can you confirm if increasing the stack size will solve the problem?

chriseth: running in pr-tests solves the problem. How do you propose to increase the stack size?

chriseth: I'm also fine with flipping it around, using some env variable like `POWDR_REPARSE_EXCLUDE_LIST`

lvella: An alternative solution could be a wrapper to run tests with big stack requirements. Something like:
```
fn run_in_large_stack<F: FnOnce() -> R + Send, R: Send>(f: F) -> R {
    thread::scope(|s| {
    let handle = thread::Builder::new()
        .stack_size(16 * 1024 * 1024) // Set the stack size to 16MB
        .spawn_scoped(s, f)
        .unwrap();
    handle.join().unwrap()
}
```

chriseth: nah, I want to fix it eventually.

lvella:
If an env var you have to set is good enough for you, you can try instead setting `RUST_MIN_STACK=16777216` or some other size.

chriseth:
but it will be more complicated to implement properly, won't it?

chriseth:
cool, it works, thanks!

PR #1852, opened 2024-09-30
Remove RISC-V assembler

lvella:
Removal of a quite important feature.

Better get many approvals.


isa-utils/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+
+pub fn quote(s: &str) -> String {
+    // TODO more things to quote
+    format!("\"{}\"", s.replace('\\', "\\\\").replace('\"', "\\\""))
+}
+
+pub fn escape_label(l: &str) -> String {
+    // TODO make this proper
+    l.replace('.', "_dot_").replace('/', "_slash_")
+}
```
lvella, line 21: These is all that survived from the old `powdr-asm-utils` crate.

leonardoalt:
@chriseth had a point that if we ever add new frontends via assembly (like WASM), we'd need some of this code back. I guess we can always revive stuff in the future.

lvella:
I believe there is a binary version of WASM that is actually deployed and is easier to parse than assembly.

leonardoalt:
yea, makes sense

riscv/src/lib.rs
```diff
@@ -179,187 +119,117 @@ macro_rules! as_ref [
...
+        let bins = packages
             .iter()
-            .any(|dependency| {
-                dependency["name"] == "powdr-riscv-runtime"
-                    && dependency["features"]
-                        .as_array()
-                        .unwrap()
-                        .contains(&"std".into())
+            .flat_map(|package| {
+                package["targets"]
```
leonardoalt, line 196: where was this before? I don't see it in the removed parts

leonardoalt (commented):
Nice! Some small comments

.github/workflows/nightly-tests.yml
```diff
@@ -69,8 +69,10 @@ jobs:
       run: rustup toolchain install 1.81-x86_64-unknown-linux-gnu
     - name: Install nightly
       run: rustup toolchain install nightly-2024-08-01-x86_64-unknown-linux-gnu
-    - name: Install stdlib
+    - name: Install std source
       run: rustup component add rust-src --toolchain nightly-2024-08-01-x86_64-unknown-linux-gnu
+    - name: Install riscv target
+      run: rustup target add riscv32imac-unknown-none-elf --toolchain nightly-2024-08-01-x86_64-unknown-linux-gnu
```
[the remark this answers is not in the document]
lvella: We built the core libraries from source, since we needed the .s files. Now we can use the pre-built.

Then you ask: "Then why do we still need the line above, that installs the rust source?"

Because the risc0 target we use for std support is a lower tier and doesn't have pre-builds, so we still need to build `std` from source when using it.

.github/workflows/pr-tests.yml
```diff
@@ -198,8 +202,10 @@ jobs:
       run: sudo apt-get install -y binutils-riscv64-unknown-elf lld
     - name: Install nightly
       run: rustup toolchain install nightly-2024-08-01-x86_64-unknown-linux-gnu
-    - name: Install stdlib
+    - name: Install std source
       run: rustup component add rust-src --toolchain nightly-2024-08-01-x86_64-unknown-linux-gnu
+    - name: Install riscv target
+      run: rustup target add riscv32imac-unknown-none-elf --toolchain nightly-2024-08-01-x86_64-unknown-linux-gnu
```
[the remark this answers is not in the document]
lvella: same answer

riscv/src/lib.rs
lvella: This is new code.

The previous version used the JSON from `cargo -Zunstable-options --build-plan` to get details from every compilation unit and find the assembly files. But it is an unstable feature with a heavy run cost (it needs a new target directory to run, and redo a lot of work).

Now that we just need to find the executables, so we use the stable and and lightweight JSON from `cargo metadata` (that was already been called to figure out if this is a `std` or `no-std` build).

riscv/src/lib.rs
```diff
@@ -440,16 +292,6 @@ fn build_cargo_command(
...
-    // TODO: if asm path is removed, there are better ways to find the
-    // executable name than relying on the unstable build plan.
-    if produce_build_plan {
-        args.extend(as_ref![
-            OsStr;
-            "-Zunstable-options",
-            "--build-plan"
-        ]);
-    }
-
```
lvella, line 452: @leonardoalt Here is the cargo command we don't need anymore.

leonardoalt (approved):

PR #1857, opened 2024-09-30
Integrated plonky3 prover

Schaeff:
This is a major change to the plonky3 prover to support proving many machines.

# Sharing costs across tables
- at setup phase, fixed columns for each machine are committed to for each possible size. This happens in separate commitments, so that the prover and verifier can pick the relevant ones for a given execution
- for each phase of the proving, the corresponding traces across all machines are committed to jointly
- the quotient chunks are committed to jointly across all tables

# Multi-stage publics
The implementation supports public values for each stage of each table. This is tested internally in the plonky3 crate but not end-to-end in pipeline tests.


plonky3/src/proof.rs
```diff
@@ -23,9 +24,8 @@ pub type PcsProverData<SC> = <<SC as StarkGenericConfig>::Pcs as Pcs<
 #[serde(bound = "")]
 pub struct Proof<SC: StarkGenericConfig> {
     pub(crate) commitments: Commitments<Com<SC>>,
-    pub(crate) opened_values: OpenedValues<SC::Challenge>,
+    pub(crate) opened_values: BTreeMap<String, TableOpenedValues<SC::Challenge>>,
```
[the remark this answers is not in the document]
Schaeff: good catch, using the alias

plonky3/src/folder.rs
```diff
@@ -25,7 +25,7 @@ type ViewPair<'a, T> = VerticalPair<RowMajorMatrixView<'a, T>, RowMajorMatrixVie
 
 #[derive(Debug)]
 pub struct VerifierConstraintFolder<'a, SC: StarkGenericConfig> {
-    pub challenges: Vec<Vec<Val<SC>>>,
+    pub challenges: &'a Vec<Vec<Val<SC>>>,
```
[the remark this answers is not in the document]
Schaeff: I wished clippy had flagged this one :)

plonky3/src/verifier.rs
```diff
@@ -173,31 +195,12 @@ where
     let quotient_domains: Vec<Vec<_>> = tables
         .values()
         .zip_eq(proof.opened_values.values())
-        .map(|(input, opened_values)| {
-            let degree = 1 << opened_values.log_degree;
-            let trace_domain = pcs.natural_domain_for_degree(degree);
+        .zip_eq(trace_domains.iter())
```
lvella, line 198: Maybe use `itertools::izip!` instead of 3 chained zips? But you would lose the check that they are equal.

plonky3/src/verifier.rs
```diff
@@ -56,27 +73,33 @@ where
...
-        .map(|(name, values)| (name, values.iter().collect_vec()))
+        .map(|(name, values)| (name, values.as_slice()))
         .collect();
 
+    // sanity check that both maps have the same keys
+    assert!(program
+        .split
+        .keys()
+        .zip_eq(public_inputs.keys())
+        .all(|(a, b)| a == *b));
```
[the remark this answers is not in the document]
Schaeff: love that thanks

plonky3/src/verifier.rs
Schaeff: Yeah I do want to check the length here so will keep this as is

plonky3/src/verifier.rs
```diff
@@ -10,250 +12,423 @@ use p3_matrix::dense::RowMajorMatrixView;
...
-#[instrument(skip_all)]
-pub fn verify<SC, A>(
-    config: &SC,
-    air: &A,
-    challenger: &mut SC::Challenger,
-    proof: &Proof<SC>,
-    public_values: &Vec<Val<SC>>,
-) -> Result<(), VerificationError<PcsError<SC>>>
+/// A sub-table to be proven, in the form of an air and values for the public inputs
+struct Table<'a, T: FieldElementMap>
```
leonardoalt, line 25: why are this Table and prover::Table different?

plonky3/src/verifier.rs
```diff
@@ -10,250 +12,423 @@ use p3_matrix::dense::RowMajorMatrixView;
...
+            &self
+                .public_values_by_stage
+                .iter()
+                .map(|values| values.len())
+                .collect::<Vec<_>>(),
+        )
+    }
 }
 
 #[instrument(skip_all)]
```
leonardoalt, line 51: what's being skipped here?

plonky3/src/verifier.rs
```diff
@@ -10,250 +12,423 @@ use p3_matrix::dense::RowMajorMatrixView;
...
-        .enumerate()
-        .map(|(ch_i, ch)| {
-            ch.iter()
-                .enumerate()
-                .map(|(e_i, &c)| zps[ch_i] * SC::Challenge::monomial(e_i) * c)
-                .sum::<SC::Challenge>()
-        })
-        .sum::<SC::Challenge>();
+    pcs.verify(verify_input, opening_proof, challenger)
+        .map_err(VerificationError::InvalidOpeningArgument)?;
```
leonardoalt, line 284: do we have tests that fail each of these stages? Maybe not fully needed for this PR, but eventually...

plonky3/src/prover.rs
```diff
@@ -1,71 +1,470 @@
...
+            vec![None; state.program.table_count()]
+        };
+
+        // get values for the traces
+        let traces_by_table_by_stage: Vec<Vec<_>> = state.processed_stages.iter().fold(
+            vec![vec![]; state.program.table_count()],
+            |mut traces_by_table, _| {
+                let mut values = opened_values.next().unwrap();
+                for (values, v) in traces_by_table.iter_mut().zip_eq(values.iter_mut()) {
+                    assert_eq!(v.len(), 2);
```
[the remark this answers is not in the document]
Schaeff: Good question! The output of `pcs.open` returns an `OpenedValues<F>` defined as:
```
pub type OpenedValues<F> = Vec<OpenedValuesForRound<F>>;
pub type OpenedValuesForRound<F> = Vec<OpenedValuesForMatrix<F>>;
pub type OpenedValuesForMatrix<F> = Vec<OpenedValuesForPoint<F>>;
pub type OpenedValuesForPoint<F> = Vec<F>;
```

Here `values` is an `OpenValuesForRound<F>` obtained as a element of `opened_values`.
`v` is an element of that, so it's a `OpenedValuesForMatrix<F>`.
Now, it has size two, because for the rounds corresponding to the traces (or matrices!), we open the commitment at two points: `zeta` and `zeta_next`.
All of this is defined in the inputs to `pcs.open`, in this case [here](https://github.com/powdr-labs/powdr/blob/40e385ce8678b7825bf606995aad371a50358ebd/plonky3/src/prover.rs#L156-L160)

plonky3/src/verifier.rs
Schaeff: Good question, this is a leftover from uni-stark

plonky3/src/verifier.rs
Schaeff: After this comment I realised they are too similar. I ended up refactoring by packing in `Table` everything which relates to this particular table within the verifier: the preprocessed commitments from the vk, the air, the opened values, the publics.

plonky3/src/verifier.rs
Schaeff: Right, we do not, and I think uni-stark doesn't either.

executor/src/witgen/mod.rs
```diff
@@ -297,7 +302,7 @@ impl<'a, 'b, T: FieldElement> WitnessGenerator<'a, 'b, T> {
 pub fn extract_publics<T: FieldElement>(
     witness: &[(String, Vec<T>)],
     pil: &Analyzed<T>,
-) -> Vec<(String, T)> {
```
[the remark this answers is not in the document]
Schaeff: This is called at the end of stage 0 witgen. However some public values are only known at the end of stage 1 witgen, if they depend on stage 1 witnesses.

plonky3/src/stark.rs
```diff
@@ -279,33 +313,28 @@ mod tests {
 
         let pil = pipeline.compute_optimized_pil().unwrap();
         let witness_callback = pipeline.witgen_callback().unwrap();
-        let witness = pipeline.compute_witness().unwrap();
+        let witness = &mut pipeline.compute_witness().unwrap();
```
lvella, line 316: Again, very surprised this compiles. I don't know what is happening here.

You are taking an `&mut` from inside an `Arc`, but that should not be possible, it doesn't implement `DerefMut`.

plonky3/src/stark.rs
```diff
@@ -169,46 +196,44 @@ where
         witness: &[(String, Vec<T>)],
         witgen_callback: WitgenCallback<T>,
     ) -> Result<Vec<u8>, String> {
-        let stage_0_trace =
-            generate_matrix(witness.iter().map(|(name, value)| (name, value.as_ref())));
+        // here we need to clone the witness because the callback will modify it
+        let witness = &mut witness.to_vec();
```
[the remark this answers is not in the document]
Schaeff: chatgpt:

> Why does this work?
> The key lies in Rust's temporary lifetime extension. When a temporary is created and immediately borrowed, Rust extends the lifetime of that temporary until the end of the enclosing scope. So, in this case:

> Rust sees that you are creating a mutable reference (&mut) to a temporary value (witness.to_vec()).
> It extends the lifetime of the temporary Vec so that it lives as long as the reference (i.e., the temporary Vec will not be dropped until witness is no longer used).
> Effectively, what happens is that the Vec will remain valid for as long as the mutable reference &mut witness is in scope.

So in this particular case, Rust introduces an intermediate variable

plonky3/src/stark.rs
Schaeff: Seems like the same explanation? Would it be possible if I had introduced an intermediate variable, and taken an &mut on that? Then Rust does that automatically in this particular case, apparently.

plonky3/src/verifier.rs
```diff
@@ -10,250 +12,423 @@ use p3_matrix::dense::RowMajorMatrixView;
...
+                            *trace_domain,
+                            vec![(zeta, local.clone()), (zeta_next, next.clone())],
                         )],
                     )
                 })
+        })
+        .collect();
+
+    // for trace commitments, we have one commitment per stage, opened on each trace domain at `zeta` and `zeta_next`
+    let trace_domains_points_and_opens_by_stage: Vec<(_, Vec<(_, _)>)> = izip!(
```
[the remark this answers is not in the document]
Schaeff: Good catch! I think the reason was that I was more worried about the wrong tables being zipped together than the wrong stages. But I changed to zip_eq now

leonardoalt (dismissed):

leonardoalt (approved):

PR #1858, opened 2024-10-01
Jit enums

chriseth:


ast/src/parsed/asm.rs
```diff
@@ -167,6 +167,17 @@ impl SymbolPath {
         })
     }
 
+    /// Returns the path up to the last part and the last part individually.
+    /// Returns None if the last part is `super` or the path is empty.
+    pub fn try_split_off_last_part(&self) -> Option<(Self, String)> {
```
pacheco, line 172: i think this could be just called `pop` or `pop_last` and take `&mut self`.
Similar to `PathBuf` and `Vec`.

jit-compiler/src/codegen.rs
```diff
@@ -80,44 +81,71 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
 
-        Ok(match (&value.e, type_scheme) {
-            (Expression::LambdaExpression(_, expr), TypeScheme { vars, ty }) => {
-                assert!(vars.is_empty());
-                self.try_format_function(symbol, expr, ty)?
+        match definition {
+            FunctionValueDefinition::TypeDeclaration(EnumDeclaration {
+                type_vars,
+                variants,
+                ..
```
lvella, line 95: didn't knew this was possible

jit-compiler/src/codegen.rs
```diff
@@ -416,8 +450,37 @@ fn check_pattern(value_name: &str, pattern: &Pattern) -> Result<(String, String)
...
+                    vars.push(v);
+                    Ok(format!("({code})?"))
+                })
+                .collect::<Result<Vec<_>, String>>()?
+                .join(", ");
+
+            (
+                vars.join(", "),
+                format!(
+                    "(|| if let {}({}) = ({value_name}).clone() {{ Some({inner_code}) }} else {{ None }})()",
```
lvella, line 479: Why this must be a closure call?

chriseth: So that the inner code can use the `?` operator.

pacheco (approved):
lgtm, but there are some unrelated failing tests?

Schaeff (commented):
Just one suggestion but it's not blocking. LGMT except CI is failing now.

jit-compiler/src/codegen.rs
```diff
@@ -80,47 +81,74 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
-                    }}\n",
-                    escape_symbol(symbol),
-                    map_type(&ty),
-                    self.format_expr(&value.e)?
-                )
+                    .join(",\n")
+            )),
+            FunctionValueDefinition::TypeConstructor(decl, _) => {
+                self.request_symbol(&decl.name)?;
+                Ok(String::new())
```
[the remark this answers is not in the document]
chriseth: I don't think we should distinguish between those two cases. It returns the generated code. If no code needs to be generated, we return the empty string.

pacheco (approved):

PR #1860, opened 2024-10-01
Using Warpbuild instead of BuildJet

lvella:
BuildJet cache is broken and they won't answer support ticket.

### DIFF
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -9,7 +9,7 @@ env:
 
 jobs:
   build:
-    runs-on: buildjet-4vcpu-ubuntu-2204
+    runs-on: warp-ubuntu-2404-x64-4x
 
     steps:
     - uses: actions/checkout@v4
@@ -30,7 +30,7 @@ jobs:
     - name: Check without Halo2
       run: cargo check --all --no-default-features --profile pr-tests
     - name: ⚡ Save rust cache
-      uses: buildjet/cache/save@v4
+      uses: WarpBuilds/cache@v1
       with:
         path: |
           ~/.cargo/registry/index/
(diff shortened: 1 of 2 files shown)


leonardoalt (commented):
lgtm, but I haven't done much CI stuff here so I'll let @chriseth or @pacheco approve

.github/workflows/build-cache.yml
```diff
@@ -29,13 +29,24 @@ jobs:
...
+      uses: WarpBuilds/cache@v1
+      with:
+        path: |
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+          Cargo.lock
+        key: ${{ runner.os }}-cargo-pr-tests
+        delete-cache: true
```
pacheco, line 42: what does this delete-cache do?

pacheco: idea is to keep a single cache entry?

lvella: Yes. This cache is different from GitHub/BuildJet in two ways: costs money to keep, and you can explicitly delete it. Since we don't need more than one, we can do like this instead of using the hash trick.

chriseth: What if the build fails? Does it keep the old cache?

chriseth (approved):

.github/workflows/build-cache.yml
pacheco: maybe we can do sth like add the "date" to the cache name and then delete after saving?
also, do we need the `path` entry here for deleting the cache? just the key should be fine no?

lvella: The problem is that for deleting, we need an exact match, so it is less riskier of leaking stale cache, this way, using just one name.

About the using the path, it is in their example. I just followed.

chriseth: I mean if this only runs on main, the chances of the build failing are rather low.

lvella: Yeah, but even then, deletion only happens if the previous steps complete successfully. There is the minimal chance delete completes but the instance crashes before it can complete the upload, but then you can just run the job manually.

PR #1862, opened 2024-10-02
Return iterator instead.

chriseth:
Instead of returning a vector and turning it into an iterator, return an iterator right away.


lvella (approved):
GR6m5OhXUAApgAI

PR #1865, opened 2024-10-02
JIT: Support generic functions

chriseth:
Supports compilation of generic pil functions. This is just basic support, it might turn out that it is not possible to translate generic PIL functions directly to generic rust functions.


jit-compiler/src/codegen.rs
```diff
@@ -142,8 +142,24 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
             _ => return Err(format!("Expected function type, got {ty}")),
         };
 
+        let generics = if vars.is_empty() {
+            String::new()
+        } else {
+            // TODO support actual bounds
```
pacheco, line 148: planned for one of the following PRs? if not, maybe a more descriptive TODO?

chriseth: It's not really properly implemented on the PIL side yet. We will notice when it breaks, but I can extend the comment.

pacheco (dismissed):

jit-compiler/src/compiler.rs
```diff
@@ -117,6 +155,15 @@ pub fn call_cargo(code: &str) -> Result<PathInTempDir, String> {
     fs::write(dir.join("Cargo.toml"), CARGO_TOML).unwrap();
     fs::create_dir(dir.join("src")).unwrap();
     fs::write(dir.join("src").join("lib.rs"), code).unwrap();
+    Command::new("cargo")
+        .arg("fmt")
+        .current_dir(dir.clone())
+        .output()
+        .unwrap();
+    println!(
```
chriseth, line 163: Needs to be removed before merge.

jit-compiler/src/codegen.rs
```diff
@@ -244,6 +271,9 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
                     BinaryOperator::ShiftRight => {
                         format!("(({left}).clone() >> usize::try_from(({right}).clone()).unwrap())")
                     }
+                    BinaryOperator::Add => {
+                        format!("Add::add(({left}).clone(), ({right}).clone())")
+                    }
                     _ => format!("(({left}).clone() {op} ({right}).clone())"),
```
chriseth, line 277: Only the `Add` trait is implemented for now.

jit-compiler/src/compiler.rs
```diff
@@ -101,8 +97,46 @@ impl<Args, Ret> Callable<Args, Ret> {
...
+
 "#;
 
+fn field_specific_preamble<T: FieldElement>() -> String {
+    let modulus = u64::try_from(T::modulus().to_arbitrary_integer()).unwrap();
+    format!(
+        r#"
+        impl From<u64> for FieldElement {{
+            fn from(x: u64) -> Self {{
+                // TODO this is inefficient.
```
chriseth, line 126: Won't fix this here.

jit-compiler/src/codegen.rs
```diff
@@ -211,7 +229,16 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
                     format!("Expr::from({val}_u64)")
                 }
-                _ => return Err(format!("Unexpected type for literal number: {type_}")),
+                Type::TypeVar(tv) => {
+                    if let Ok(v) = u64::try_from(value) {
+                        format!("{tv}::from_u64({v}_u64)")
+                    } else {
+                        let v_bytes = value.to_le_bytes();
+                        let v = v_bytes.iter().map(|b| format!("{b}_u8")).format(", ");
+                        format!("{tv}::from_le_bytes({v})")
```
pacheco, line 238: should there be `[ ]` around `{v}`?

jit-compiler/src/codegen.rs
```diff
@@ -702,7 +732,7 @@ mod test {
         );
         assert_eq!(
             result,
-            "fn c((i): (ibig::IBig)) -> ibig::IBig { ((i).clone() + (ibig::IBig::from(20_u64)).clone()) }\n\
+            "fn c((i): (ibig::IBig)) -> ibig::IBig { Add::add((i).clone(), (ibig::IBig::from(20_u64)).clone()) }\n\
```
pacheco, line 735: are the `clone()`s needed here?

lvella: I don't think so, because this is a full function and both operands can be moved in.

lvella (changes requested):
What Leandro asked.

jit-compiler/src/codegen.rs
chriseth: This is juts the result. The clone is added locally, it doesn't know that `i` is only used once, for example. If you do `i + i`, you need the clone.

pacheco (approved):

jit-compiler/src/codegen.rs
lvella: The line is a complete function, you do know it is used only once.

PR #1866, opened 2024-10-02
JIT: proper function types.

chriseth:
Rust's function types are more complicated than powdr's: It has `fn`, `Fn` and closures (we ignore generic functions and trait functions for now). This PR adds an [enum](https://github.com/powdr-labs/powdr/pull/1866/files#diff-0b9128f385080144659241b59a28702f0a85a684895f59bf06f5a0b4440070fdR90) that can hold either an `fn` or an `Fn` (which is used for closures). This enum will be used as the rust type for a powdr function.

Rust closures that capture variables have to be handled specially: We need to make sure that we can move the captured values into the closure (the other option would be to use references, but we cannot say anything about the lifetime of the captured value, so this is not viable). We do this by first cloning the captured value into a new local variable (of the same name), which we can then move into the closure. In order to do this, we need to check which local variables references reach out of the lambda and which ones stay inside the lambda. This is done by keeping track of the current "variable height" during expression conversion.


jit-compiler/src/compiler.rs
```diff
@@ -117,6 +133,15 @@ pub fn call_cargo(code: &str) -> Result<PathInTempDir, String> {
     fs::write(dir.join("Cargo.toml"), CARGO_TOML).unwrap();
     fs::create_dir(dir.join("src")).unwrap();
     fs::write(dir.join("src").join("lib.rs"), code).unwrap();
+    Command::new("cargo")
```
chriseth, line 136: To remove later.

jit-compiler/src/codegen.rs
```diff
@@ -332,30 +361,38 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
...
-                !matches!(typed_expr.e, Expression::LambdaExpression(..))
+            format!("::<{}>", type_args.iter().map(map_type).join(", "))
+        };
+        if is_builtin::<T>(symbol) {
+            return format!("Callable::Fn({}{type_args})", escape_symbol(symbol));
+        }
+        let (_, def) = self.analyzed.definitions.get(symbol).as_ref().unwrap();
+        if let Some(FunctionValueDefinition::Expression(typed_expr)) = def {
+            if matches!(typed_expr.e, Expression::LambdaExpression(..)) {
+                format!("Callable::Fn({}{type_args})", escape_symbol(symbol))
```
pacheco, line 390: are there tests using type_args?

pil-analyzer/src/type_inference.rs
```diff
@@ -414,6 +414,12 @@ impl TypeChecker {
                     }
                 }
             }
+            Expression::LambdaExpression(_, LambdaExpression { param_types, .. }) => {
+                for ty in param_types {
+                    // Here, the types do not have to be concrete.
```
pacheco, line 419: why not?

pil-analyzer/src/type_inference.rs
```diff
@@ -569,7 +575,15 @@ impl TypeChecker {
...
                     .collect::<Result<_, _>>()?,
             }),
-            Expression::LambdaExpression(_, LambdaExpression { kind, params, body }) => {
+            Expression::LambdaExpression(
+                _,
+                LambdaExpression {
+                    kind,
+                    params,
+                    body,
+                    param_types,
```
pacheco, line 584: can this `param_types` already have some value here? it seems to be overridden without being looked at

jit-compiler/src/codegen.rs
```diff
@@ -142,31 +145,23 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
             _ => return Err(format!("Expected function type, got {ty}")),
         };
 
+        let var_height = params.iter().map(|p| p.variables().count()).sum::<usize>();
+
         Ok(format!(
-            "fn {}({}) -> {} {{ {} }}\n",
+            "fn {}(({}): ({})) -> {} {{ {} }}\n",
```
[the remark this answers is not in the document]
chriseth: Yes: Fn has one generic type for each parameter, so if we want to support multi-parameter functions, the only way is to pack all of them in a tuple. There is a rust marker trait called Args that could be used to combine multiple generic args into a single one, but it is only valid inside std.

jit-compiler/src/codegen.rs
chriseth: Yes, the last test calls `std::convert::fe`, which is a generic function.

pil-analyzer/src/type_inference.rs
chriseth: Will add an assertion.

pil-analyzer/src/type_inference.rs
chriseth: Added a comment.

pacheco (approved):

pacheco (approved):

PR #1867, opened 2024-10-02
new optimizer rules

leonardoalt:
This PR does 3 things:

1. Fixes a bug in the "constant lookup extractor"  where a lookup of form 0 $ [A, B] in [zero, one] generates a new constraint -1 = 0.
2. Adds a new rule that removes lookups where the left selector = 0. This is sound because it's a tautology.
3. Adds a new rule that transforms a permutation with left selector = 0 into a constraint R = 0, where R is the right selector. This is sound because the LHS of the permutation being 0 means both sides should be the empty set, which is guaranteed by the constraint for the RHS.

### DIFF
--- a/pilopt/src/lib.rs
+++ b/pilopt/src/lib.rs
@@ -8,7 +8,7 @@ use std::iter::once;
 use powdr_ast::analyzed::{
     AlgebraicBinaryOperation, AlgebraicBinaryOperator, AlgebraicExpression, AlgebraicReference,
     AlgebraicUnaryOperation, AlgebraicUnaryOperator, Analyzed, Expression, FunctionValueDefinition,
-    IdentityKind, PolyID, PolynomialReference, PolynomialType, Reference, SymbolKind,
+    Identity, IdentityKind, PolyID, PolynomialReference, PolynomialType, Reference, SymbolKind,
     TypedExpression,
 };
 use powdr_ast::parsed::types::Type;
@@ -22,6 +22,7 @@ pub fn optimize<T: FieldElement>(mut pil_file: Analyzed<T>) -> Analyzed<T> {
     remove_constant_fixed_columns(&mut pil_file);
     simplify_identities(&mut pil_file);
     extract_constant_lookups(&mut pil_file);
+    extract_empty_permutations(&mut pil_file);
     remove_constant_witness_columns(&mut pil_file);
     simplify_identities(&mut pil_file);
     remove_trivial_selectors(&mut pil_file);
@@ -382,6 +383,7 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
         .identities
         .iter_mut()
         .filter(|id| id.kind == IdentityKind::Plookup)
+        .filter(|id| id.right.selector.is_some())
     {
         let mut extracted = HashSet::new();
         for (i, (l, r)) in identity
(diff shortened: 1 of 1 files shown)


pilopt/src/lib.rs
```diff
@@ -382,6 +383,7 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
         .identities
         .iter_mut()
         .filter(|id| id.kind == IdentityKind::Plookup)
+        .filter(|id| id.right.selector.is_some())
```
leonardoalt, line 386: This is case 1.

pilopt/src/lib.rs
```diff
@@ -432,6 +434,31 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
     }
 }
 
+/// Extracts constraints from lookups where the LHS' selector = 0.
+fn extract_empty_permutations<T: FieldElement>(pil_file: &mut Analyzed<T>) {
```
leonardoalt, line 438: This is point 3.

pilopt/src/lib.rs
```diff
@@ -22,6 +22,7 @@ pub fn optimize<T: FieldElement>(mut pil_file: Analyzed<T>) -> Analyzed<T> {
     remove_constant_fixed_columns(&mut pil_file);
     simplify_identities(&mut pil_file);
     extract_constant_lookups(&mut pil_file);
+    extract_empty_permutations(&mut pil_file);
```
chriseth, line 25:
```suggestion
    extract_inactive_permutations(&mut pil_file);
```
(empty for me is `[] is []`).

pilopt/src/lib.rs
```diff
@@ -432,6 +434,31 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+        .filter(|id| id.kind == IdentityKind::Permutation)
+        .filter(|id| {
+            if let Some(AlgebraicExpression::Number(n)) = id.left.selector {
+                n.is_zero()
+            } else {
+                false
+            }
+        })
+    {
+        let new_expr = identity
```
chriseth, line 451:
```suggestion
        // If the LHS selector is 0, then the RHS selector needs to be zero as well.
        let new_expr = identity
```

pilopt/src/lib.rs
chriseth, line 438: In which way does it extract? It simplifies, doesn't it?

pilopt/src/lib.rs
chriseth: Wait, if there is no RHS selector, it should still work, shouldn't it?

chriseth: I think instead of this you should check that RHS selector is constant (either None or Some(constant)).

pilopt/src/lib.rs
leonardoalt: To be accurate, it transforms. I guess we can call it a simplification

pilopt/src/lib.rs
chriseth: You should change `pol_id` to something like
```
let pol_id = l_sel * (l.clone() - AlgebraicExpression::from(*r));
```
Assuming `r_sel` is 1 at least on a single row. If r_sel is None, this is true. If it's `Some(r_sel_i)`, then it depends on that value.

leonardoalt:
The new optimization rule causes witgen issues. I'll push the bugfix in a different PR.

Schaeff:
Should we merge some of this, or was it all fixed in #1889 ?

leonardoalt:
Yea theres one bugfix that needs to go in

pilopt/src/lib.rs
chriseth: I would expect a function called `extract_...` to return something - the stuff it extracted.

leonardoalt:
Closing in favor of https://github.com/powdr-labs/powdr/pull/2004

chriseth:
#2004 only addressed one of the items.

leonardoalt:
> #2004 only addressed one of the items.

The other one causes witgen issues, not sure it's worth pursuing it.

PR #1872, opened 2024-10-04
Poseidon BB

leonardoalt:
Depends on https://github.com/powdr-labs/powdr/pull/1871
Extracted from https://github.com/powdr-labs/powdr/pull/1790

Originally written by @lvella


lvella:
I see you removed the constants generator file. I don't think it is a good idea to have all these values opaque.

leonardoalt:
I didnt remove it, just didnt extract everything. I'll add it

lvella:
Well, yesterday it was not in the riscv-bb branch.

leonardoalt:
Then it wasnt in the same original PR, was it? Cus i didnt remove it

leonardoalt:
Ah wait, yes I did remove it. I thought it was leftover from the P3 stuff and I messed up the merge. Anyway, I'll bring it back

leonardoalt:
This PR is now the same as https://github.com/powdr-labs/powdr/pull/1810, but targeting main.

std/machines/hash/poseidon_bb.asm
```diff
@@ -0,0 +1,206 @@
...
+// This version of the Poseidon machine receives memory pointers and interacts
+// with memory directly to fetch its inputs and write its outputs.
+machine PoseidonBB(mem: Memory16, split_bb: SplitBB) with
+    latch: CLK_0,
+    operation_id: operation_id,
+    // Allow this machine to be connected via a permutation
+    call_selectors: sel,
+{
+
+    // Hashes 16 "rate" elements and 8 "capacity" elements to 8 field elements
```
pacheco, line 31: the comments in this file need to be revised... i think there are leftovers from the GL version

leonardoalt:
@lvella the memory machine now takes 2 limbs of 16 bits as the address, can you modify the poseidon_bb machine in this PR to take that into account?

lvella:
Can I just take one field element per address and use SplitBB on the address? (this would restricting the operation to ranges < P)

Or must I now take two limbs of input address and two limbs of output?

In this case, can I restrict the memory ranges to not cross the 64KB boundary of the lower limb and use SplitBB to check for that, or should I use the arithmetic machine to properly increment the address across the two limbs?

Any better idea?

lvella:
The version I just pushed works, but it takes the address and does the arithmetic as a single field element, so is restricted to address ranges that fits in one.

leonardoalt:
> Can I just take one field element per address and use SplitBB on the address? (this would restricting the operation to ranges < P)

No, @georgwiese fixed the memory machine to accept the full 32-bit address space.

> Or must I now take two limbs of input address and two limbs of output?

Yes please.

leonardoalt:
> In this case, can I restrict the memory ranges to not cross the 64KB boundary of the lower limb and use SplitBB to check for that,

That would be quite limiting for the user I think.

> or should I use the arithmetic machine to properly increment the address across the two limbs?

Maybe we can inline the increment in the machine itself?

leonardoalt:
If we want to implement it properly I dont think we have a choice, memory is already implemented with 2 limb address. Otherwise we're restricting the machine in a hacky way

std/machines/hash/poseidon_bb.asm
lvella: I am not sure what is the best way to fix this, because that is the terminology in the original Poseidon paper, that assumed the permutation would be used in sponge construction. This new way of using the permutation directly on a Merkle tree is from Poseidon2 paper.

lvella: Ah, the numbers don't even add up.

I fixed by removing confusing text.

lvella:
Please continue the discussion in PR #1926

PR #1879, opened 2024-10-08
compiler options and codegen split

leonardoalt:
Extracted from https://github.com/powdr-labs/powdr/pull/1790.

This PR:
- creates the new `CompilerOptions` that encapsulate precompiles and common options, needed by both cli-rs and tests
- splits codegen, runtime and bootloader into an outer facing dispatcher based on the compiler options, and internal versions that currently are just copies of the previous implementation that is specific to Goldilocks and larger modulus.

After this PR, the riscv-bb PR will simply add analogous 17-bit codegen, runtime and bootloader files.


riscv/src/code_gen64.rs
```diff
@@ -0,0 +1,1507 @@
...
+use crate::runtime::Runtime;
+use crate::runtime64::Runtime64;
+
+use crate::code_gen::{
+    InstructionArgs, MemEntry, Register, RiscVProgram, SourceFileInfo, Statement,
+};
+use crate::CompilerOptions;
+
+/// Translates a RISC-V program to POWDR ASM
+/// with constraints that work for a field >= the Goldilocks modulus.
```
leonardoalt, line 18: This file contains everything that was "removed" according to GH from the file above `code_gen.rs`, and I changed the text in this line. Everything else is untouched.

riscv/src/continuations/bootloader64.rs
```diff
@@ -0,0 +1,649 @@
+use crate::code_gen::Register;
+
+use crate::continuations::bootloader::{
+    BOOTLOADER_INPUTS_PER_PAGE, BYTES_PER_WORD, DEFAULT_PC, MEMORY_HASH_START_INDEX,
+    MERKLE_TREE_DEPTH, NUM_PAGES_INDEX, N_LEAVES_LOG, PAGE_INPUTS_OFFSET, PAGE_NUMBER_MASK,
+    PAGE_SIZE_BYTES, PC_INDEX, REGISTER_NAMES, SHUTDOWN_START, WORDS_PER_HASH, WORDS_PER_PAGE,
+};
+
```
leonardoalt, line 8: This "new" file contains everything "removed" from file above `bootloader.rs` that is specific to the 64 bits version of the constraints, unmodified.

riscv/src/runtime64.rs
```diff
@@ -0,0 +1,662 @@
+use std::{collections::BTreeMap, convert::TryFrom};
+
+use powdr_riscv_syscalls::Syscall;
+
+use itertools::Itertools;
+
+use crate::code_gen::Register;
+
```
leonardoalt, line 8: This file contains everything that was "removed" from file above `runtime.rs`, with all the 64-bit specific stuff, unmodified.

riscv/src/runtime64.rs
```diff
@@ -0,0 +1,662 @@
...
+            runtime = runtime.with_poseidon(continuations);
+        }
+        if libs.keccak {
+            runtime = runtime.with_keccak();
+        }
+        if libs.arith {
+            runtime = runtime.with_arith();
+        }
+        runtime
+    }
```
leonardoalt, line 36: This type, its constructor and its usage are the only changes.

riscv/src/runtime.rs
```diff
@@ -73,635 +68,20 @@ impl SubMachine {
...
-            tmp1.addr(),
-            tmp1.addr(),
-            tmp2.addr()
-        ),
-        format!("mstore {addr_reg_id}, 0, {offset}, {};", tmp1.addr()),
-    ]
+#[derive(Clone)]
+pub struct SyscallImpl(pub Vec<FunctionStatement>);
+
+pub trait Runtime {
```
leonardoalt, line 74: This trait could even go away potentially, since we don't use it as dyn or Box anymore.

riscv/tests/common/mod.rs
```diff
@@ -1,45 +1,52 @@
...
-use powdr_riscv::Runtime;
+use powdr_riscv::CompilerOptions;
 use std::{
     path::{Path, PathBuf},
     process::Command,
 };
 
 /// Like compiler::test_util::run_pilcom_asm_string, but also runs RISCV executor.
-pub fn verify_riscv_asm_string<S: serde::Serialize + Send + Sync + 'static>(
+pub fn verify_riscv_asm_string<T: FieldElement, S: serde::Serialize + Send + Sync + 'static>(
```
leonardoalt, line 14: Unfortunately we do need T around here because of Pipeline and inputs.

cli-rs/src/main.rs
```diff
@@ -260,63 +272,40 @@ fn run_command(command: Commands) {
...
+    let libs = coprocessors_to_options(coprocessors)?;
+    let options = CompilerOptions::new(field, libs, continuations);
+    powdr_riscv::compile_riscv_elf(input_file, Path::new(input_file), options, output_dir, true)
+        .ok_or_else(|| vec!["could not translate RISC-V executable".to_string()])?;
 
     Ok(())
 }
 
 #[allow(clippy::too_many_arguments)]
 fn execute<F: FieldElement>(
```
pacheco, line 306: execute already takes `F`, it doesn't need the `field: KnownField` parameter

pipeline/src/test_util.rs
```diff
@@ -57,21 +57,21 @@ pub fn make_prepared_pipeline<T: FieldElement>(
...
 
     let inputs_bn = inputs.iter().map(|x| Bn254Field::from(*x)).collect();
     let pipeline_bn = make_prepared_pipeline(file_name, inputs_bn, vec![]);
     test_halo2(pipeline_bn);
 
     let inputs_bb = inputs.iter().map(|x| BabyBearField::from(*x)).collect();
-    let mut pipeline_bb = make_prepared_pipeline(file_name, inputs_bb, vec![]);
-    pipeline_bb.compute_witness().unwrap();
+    let pipeline_bb = make_prepared_pipeline(file_name, inputs_bb, vec![]);
+    test_plonky3_pipeline(pipeline_bb);
```
pacheco, line 74: should this go into this PR?

pipeline/tests/asm.rs
```diff
@@ -24,7 +24,7 @@ fn slice_to_vec<T: FieldElement>(arr: &[i32]) -> Vec<T> {
 fn sqrt_asm() {
     let f = "asm/sqrt.asm";
     let i = [3];
-    regular_test(f, &i);
+    regular_test_without_babybear(f, &i);
```
pacheco, line 27: why `without_babybear` again?

cli-rs/src/main.rs
leonardoalt: yea I kept both because they are used in different ways, so didn't want to mix them up. Fine with removing, but want to double check after explaining why. Should I still remove it?

pipeline/src/test_util.rs
leonardoalt: I can remove, it just came with the same patch when I extracted

pipeline/tests/asm.rs
leonardoalt: since I added the change in your other comment above to also make proofs with p3, some asm tests fail because they have lookups/permutations inside the machine which we don't support

riscv/src/continuations.rs
```diff
@@ -216,6 +216,7 @@ pub struct DryRunResult<F: FieldElement> {
 /// - The number of rows after which the prover should jump to the shutdown routine.
 pub fn rust_continuations_dry_run<F: FieldElement>(
     pipeline: &mut Pipeline<F>,
+    field: KnownField,
```
pacheco, line 219: can get `KnownField` from `F`

riscv/src/lib.rs
```diff
@@ -8,34 +8,140 @@ use std::{
...
+    pub fn new() -> Self {
+        Self {
+            arith: false,
+            keccak: false,
+            poseidon: false,
+        }
+    }
+
+    pub fn with_arith(self) -> Self {
+        Self {
```
pacheco, line 41: you can do the following:
```
Self {
    arith: true,
    ..self
}
```
and it will move all the other fields to the new object

riscv/src/lib.rs
```diff
@@ -8,34 +8,140 @@ use std::{
...
+    pub fn new_32() -> Self {
+        Self {
+            field: KnownField::GoldilocksField,
+            libs: RuntimeLibs::new(),
+            continuations: false,
+        }
+    }
+
+    pub fn with_continuations(self) -> Self {
+        Self {
```
pacheco, line 89: same for these

riscv/tests/common/mod.rs
pacheco: either that or passing a `KnownField` argument?

riscv/tests/instructions.rs
```diff
@@ -4,12 +4,17 @@ mod instruction_tests {
     use std::path::Path;
 
     use crate::common::verify_riscv_asm_file;
-    use powdr_riscv::Runtime;
+    use powdr_riscv::CompilerOptions;
     use test_log::test;
 
     fn run_instruction_test(path: &Path) {
+        let options_32 = CompilerOptions::new_32();
+        run_instruction_test_with_options(path, options_32);
```
pacheco, line 12: is the extra function needed?

riscv/src/continuations.rs
leonardoalt: yea but I wanted to keep them decoupled on purpose, in case we manage to remove F here. But I'll remove it

riscv/tests/instructions.rs
leonardoalt: in the riscv-bb PR we also run for babybear here, then it made sense. I removed it after the extraction to keep this PR GL only

riscv/tests/riscv.rs
```diff
@@ -13,13 +12,12 @@ use test_log::test;
 
 use powdr_riscv::{
     continuations::{rust_continuations, rust_continuations_dry_run},
-    Runtime,
+    CompilerOptions, RuntimeLibs,
 };
 
 /// Compiles and runs a rust program with continuations, runs the full
 /// witness generation & verifies it using Pilcom.
 pub fn test_continuations(case: &str) {
```
pacheco, line 20: should these tests be `_64`? or should we also have BB inside here?

pacheco: or maybe  `_gl`?

cli-rs/src/main.rs
pacheco: yes, i'd call `F::known_field` inside

pacheco: there's a similar situation in another part of the code, but there I think `F` can be removed, because there you `match` on the known field to then pick a concrete field

riscv/tests/riscv.rs
leonardoalt: eventually we should have bb as well, but for now only 64. Do you mean the test name?

pacheco: yes, was just thinking if we should have different `test_continuations_foo` tests or just test all variants in a single test

riscv/src/runtime.rs
```diff
@@ -32,25 +26,26 @@ pub fn parse_function_statement(input: &str) -> FunctionStatement {
         .expect("invalid function statement")
 }
 
-struct SubMachine {
+#[derive(Clone)]
+pub struct SubMachine {
     /// Full path to machine (e.g, `path::to::Machine`)
-    path: SymbolPath,
+    pub path: SymbolPath,
```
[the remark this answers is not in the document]
leonardoalt: actually not

riscv/src/runtime.rs
leonardoalt: removed now to not overengineer the abstraction

cli-rs/src/main.rs
leonardoalt: done here. Can you comment on the other instance?

riscv/src/continuations.rs
leonardoalt: ah this one

cli-rs/src/main.rs
leonardoalt: actually this one ended up being removed completely when I changed the other one

riscv/tests/riscv.rs
leonardoalt: will probably add a different one at that point, but I'd wait until we get there to change it

leonardoalt:
This is ready for review again

pacheco (approved):
ltgm

Issue #1853, opened 2024-09-30
Parser: Allow trailing `,` in all lists.


chriseth:
Arrays, tuples, function arguments, etc.

Schaeff:
Fixed in #1882

PR #1882, opened 2024-10-10
Parser: Allow trailing `,` in all lists.

gzanitti:
Solves #1853

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -550,4 +550,84 @@ namespace N(8);
         let printed = format!("{}", parse(Some("input"), input).unwrap_err_to_stderr());
         assert_eq!(expected.trim(), printed.trim());
     }
+
+    #[test]
+    fn parse_trailing_commas() {
+        let input = r#"
+    let<T1, T2,> left: T1, T2, -> T1 = |a, b,| a;
+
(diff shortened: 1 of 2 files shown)


chriseth (approved):
Looks good! If this is all lists where commas are useful, then please change the main comment from "relates to" to "solves" or "closes". Then the issue will be automatically closed when the PR is merged.

chriseth:
Why does it not run the tests? Maybe because you are not using a branch on the powdr-labs repository but in your own fork?

gzanitti:
> Why does it not run the tests? Maybe because you are not using a branch on the powdr-labs repository but in your own fork?

No idea, never happened before. But I had to solve a conflict and they are running now.

parser/src/lib.rs
```diff
@@ -551,6 +551,86 @@ namespace N(8);
...
+"#;
+        let expected = r#"
+machine Main(a: Byte, b: Byte) {
+    reg pc[@pc];
+    reg X[<=];
+    function get a, step -> b {
+    A <== mload(a, step);
+    return A;
+    }
+    instr assert_eq X, Y{ X = Y }
```
chriseth, line 627: This doesn't look right, was it like that before?

chriseth (approved):

PR #1884, opened 2024-10-10
Rotate (Circular shift) machine 

onurinanc:
SHA256 uses ROTR, so I added a rotate machine that have 2 operations: rotl and rotr


leonardoalt:
can you also add this to the std tests in pipeline tests powdr_std so that it's run in CI

std/machines/rotate.asm
```diff
@@ -0,0 +1,66 @@
...
+    degree: 65536
+{
+    operation run<0> P_operation, P_A, P_B, P_ROW -> P_C;
+
+    col fixed latch = [1]*;
+    col fixed operation_id = [0]*;
+
+    let bit_counts = [256, 32, 4, 2];
+    let min_degree = std::array::product(bit_counts);
+    std::check::assert(std::prover::min_degree() >= std::array::product(bit_counts), || "The shift machine needs at least 65536 rows to work.");
```
Schaeff, line 19:
```suggestion
    std::check::assert(std::prover::min_degree() >= std::array::product(bit_counts), || "The ByteRotate machine needs at least 65536 rows to work.");
```

leonardoalt:
@onurinanc actually, is this for goldilocks or babybear?

onurinanc:
> @onurinanc actually, is this for goldilocks or babybear?

Now, it's for both

std/machines/large_field/rotate.asm
```diff
@@ -0,0 +1,66 @@
+use std::utils::unchanged_until;
+use std::utils::cross_product;
+use std::convert::int;
+
+// Rotate for single bytes using an exhaustive table
+// TODO this way, we cannot prove anything that rotates by more than 31 bits.
```
chriseth, line 6:
```suggestion
/// Rotate for single bytes using an exhaustive table
/// TODO this way, we cannot prove anything that rotates by more than 31 bits.
```

std/machines/large_field/rotate.asm
```diff
@@ -0,0 +1,66 @@
...
+machine ByteRotate with
+    latch: latch,
+    operation_id: operation_id,
+    degree: 65536
+{
+    operation run<0> P_operation, P_A, P_B, P_ROW -> P_C;
+
+    col fixed latch = [1]*;
+    col fixed operation_id = [0]*;
+
```
chriseth, line 16: Can you also add a minimum field size assertion?

std/machines/large_field/rotate.asm
```diff
@@ -0,0 +1,66 @@
...
+    let bit_counts = [256, 32, 4, 2];
+    let min_degree = std::array::product(bit_counts);
+    std::check::assert(std::prover::min_degree() >= std::array::product(bit_counts), || "The rotate machine needs at least 65536 rows to work.");
+    let inputs = cross_product(bit_counts);
+    let a: int -> int = inputs[0];
+    let b: int -> int = inputs[1];
+    let row: int -> int = inputs[2];
```
chriseth, line 23: I think we should find a better name for this. Maybe 'out_byte_offset'?

std/machines/small_field/rotate.asm
```diff
@@ -0,0 +1,71 @@
+use std::utils::unchanged_until;
+use std::utils::cross_product;
+use std::convert::int;
+
+// Rotate for single bytes using an exhaustive table, returning two 16-bit values as result.
```
chriseth, line 5: Similar to above.

std/machines/small_field/rotate.asm
```diff
@@ -0,0 +1,71 @@
...
+    // P_C0 and P_C1 are both 16 bit limbs of P_C, where P_C0 is the less significant limb.
+    operation run<0> P_operation, P_A, P_B, P_ROW -> P_C0, P_C1;
+
+    col fixed latch = [1]*;
+    col fixed operation_id = [0]*;
+
+    let bit_counts = [256, 32, 4, 2];
+    let min_degree = std::array::product(bit_counts);
+    std::check::assert(std::prover::min_degree() >= std::array::product(bit_counts), || "The rotate machine needs at least 65536 rows to work.");
+    std::check::assert(std::field::modulus() >= 65535, || "The field modulo should be at least 2^16 - 1 to work in the rotate machine.");
```
chriseth, line 21:
```suggestion
    std::check::assert(std::field::modulus() >= 2**16, || "The field modulo should be at least 2^16 - 1 to work in the rotate machine.");
```

std/machines/small_field/rotate.asm
```diff
@@ -0,0 +1,71 @@
...
+    let inputs = cross_product(bit_counts);
+    let a: int -> int = inputs[0];
+    let b: int -> int = inputs[1];
+    let row: int -> int = inputs[2];
+    let op: int -> int = inputs[3];
+    let P_A: col = a;
+    let P_B: col = b;
+    let P_ROW: col = row;
+    let P_operation: col = op;
+    let c: int -> int = |i| match op(i) {
```
chriseth, line 31: Can we extract his to a "common" file?

std/machines/large_field/rotate.asm
onurinanc: Yes, that would have a better explanation

std/machines/large_field/rotate.asm
onurinanc: Inside the Rotate right? Not, inside the ByteRotate

std/machines/small_field/rotate.asm
onurinanc: Or, this `require_field_bits(16, || "The field modulo should be at least 2^16 - 1 to work in the shift16 machine.");`?

onurinanc: Or, this? `require_field_bits(16, || "The field modulo should be at least 2^16 - 1 to work in the rotate machine.");` Which one? It's better to have similar concept with the shift machine, which uses https://github.com/powdr-labs/powdr/blob/main/std/machines/small_field/shift.asm#L24

chriseth: I think using the helper function is more descriptive.

std/machines/small_field/rotate.asm
onurinanc: not sure what's best idea to do it

std/machines/large_field/rotate.asm
onurinanc: yes, it would be better

std/machines/small_field/rotate.asm
onurinanc: Isn't it better to refactor this later? Cause this one also includes in the shift machine. I think this can be in a different PR

onurinanc: The problem is we have both the same generic for the 1. different fields and the 2. different machines. For poseidon, we need to share the same code across the "different fields" and for shift and rotate, we can share all the  code both "different fields" and "different machines". So, if we put the generic.asm inside the machine folder for now, it would be better?

onurinanc: https://github.com/powdr-labs/powdr/issues/1903 Opened a new issue for it

onurinanc: Cause, I need to test the copy-pasted code for the bigger circuit without getting rid of productivity, but I need to be conscious that I'm adding a new code which needs to be refactor with another circuit

onurinanc: @chriseth Should we do it now or in a different PR?

onurinanc: @chriseth What to do next?

chriseth: If you actually do it, then it's fine.

std/machines/large_field/rotate.asm
```diff
@@ -0,0 +1,73 @@
...
+    latch: latch,
+    operation_id: operation_id,
+    degree: 65536
+{
+    operation run<0> P_operation, P_A, P_B, P_ROW -> P_C;
+
+    col fixed latch = [1]*;
+    col fixed operation_id = [0]*;
+
+    require_field_bits(16, || "The field modulo should be at least 2^16 - 1 to work in the rotate machine.");
```
chriseth, line 18: Don't you need at least 32 bits if you return a 32 bit value?

std/machines/large_field/rotate.asm
```diff
@@ -0,0 +1,73 @@
+use std::utils::unchanged_until;
+use std::utils::cross_product;
+use std::convert::int;
+use std::check::require_field_bits;
+
+/// Rotate on single byte input using an exhaustive table, returning 32-bit value as result.
+/// TODO this way, we cannot prove anything that rotates by more than 31 bits.
```
chriseth, line 7: I would just turn this into a comment that says that we can rotate by at most 31.

std/machines/large_field/rotate.asm
onurinanc: Sorry, mixed up with the small_field

std/machines/small_field/rotate.asm
```diff
@@ -0,0 +1,76 @@
...
+    col fixed P_C1(i) { (P_C(i) >> 16) & 0xffff };
+}
+
+machine Rotate(byte_rotate: ByteRotate) with
+    latch: latch,
+    operation_id: operation_id,
+    // Allow this machine to be connected via a permutation
+    call_selectors: sel,
+{
+    require_field_bits(17, || "Rotate requires a field that fits any 17-Bit value.");
```
chriseth, line 48: Why does this need 17 and the sub-machine only 16?

std/machines/small_field/rotate.asm
```diff
@@ -0,0 +1,76 @@
...
+    latch: latch,
+    operation_id: operation_id,
+    degree: 65536
+{
+    // P_C0 and P_C1 are both 16 bit limbs of P_C, where P_C0 is the less significant limb.
+    operation run<0> P_operation, P_A, P_B, P_ROW -> P_C0, P_C1;
+    col fixed latch = [1]*;
+    col fixed operation_id = [0]*;
+
+    require_field_bits(16, || "The field modulo should be at least 2^16 - 1 to work in the rotate machine.");
```
chriseth, line 18:
```suggestion
    require_field_bits(16, || "The field modulus should be at least 2^16 - 1 to work in the rotate machine.");
```

std/machines/small_field/rotate.asm
onurinanc: Hmm, good question, I'm not sure actually. I supposed it's a requirement as it is in the shift machine

chriseth: @leonardoalt I think you added the 17 for the shift machine - are you sure about it?

onurinanc: https://github.com/powdr-labs/powdr/blob/main/std/machines/small_field/shift.asm#L48

onurinanc: I thought it is related to f17 f48 thing which I don't know about it

onurinanc: Arith also has 48 bits requirement

onurinanc: https://github.com/powdr-labs/powdr/blob/main/std/machines/large_field/arith.asm#L25

std/machines/large_field/rotate.asm
```diff
@@ -14,9 +14,6 @@ machine ByteRotate with
 
     col fixed latch = [1]*;
     col fixed operation_id = [0]*;
-
-    require_field_bits(16, || "The field modulo should be at least 2^16 - 1 to work in the rotate machine.");
```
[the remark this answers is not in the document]
onurinanc: And, also large_field/shift.asm doesn't have this requirement?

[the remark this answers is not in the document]
onurinanc: @leonardoalt

chriseth (approved):

Issue #1854, opened 2024-09-30
JIT: Avoid clone or use ref counting smart pointer


chriseth:
Currently, we clone everything and everywhere. Sometimes this is not even consistent. Especially for arrays, we should introduce a reference counting smart pointer. For other types, it might not be that important.

Since in pil, all data types are read-only, this should be quite efficient and doable.

PR #1885, opened 2024-10-11
JIT: Efficient vector

chriseth:
Closes https://github.com/powdr-labs/powdr/issues/1854

### DIFF
--- a/jit-compiler/src/compiler.rs
+++ b/jit-compiler/src/compiler.rs
@@ -103,12 +103,47 @@ impl<Args, Ret> Callable<Args, Ret> {
     }
 }
 
+#[derive(Clone)]
+struct PilVec<T>(std::sync::Arc<Vec<T>>);
+
+impl<T> PilVec<T> {
+    fn len(&self) -> usize {
+        self.0.len()
+    }
+}
+impl<T> From<Vec<T>> for PilVec<T> {
+    fn from(v: Vec<T>) -> Self {
+        PilVec(std::sync::Arc::new(v))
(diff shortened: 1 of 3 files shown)


chriseth:
Compare the benchmark with https://github.com/powdr-labs/powdr/pull/1855 which does not use Arc.

chriseth:
Some statistics:

After this PR:
```
evaluator-benchmark/sort_2700 time:   [124.46 ms 124.63 ms 124.81 ms]
jit-benchmark/sort_2700 time:   [18.175 ms 18.189 ms 18.205 ms]
```

lvella (approved):

jit-compiler/src/compiler.rs
```diff
@@ -111,6 +134,18 @@ impl Add for ibig::IBig {
     fn add(a: Self, b: Self) -> Self { a + b }
 }
 
+impl<T: Clone> Add for PilVec<T> {
+    fn add(a: Self, b: Self) -> Self {
+        // TODO for a regular "push" or array::map this is very slow.
+        // We could optimize this by sharing a larger backing vector
+        // across prefix instances, allowing to extend the backing vector if
+        // our view is the full vector.
```
lvella, line 142: Maybe it should implement "push" and "extend", but then it would have to be Arc<RwLock<Vec<T>>>.

chriseth: Yeah, we can do that later, after we checked what the performance bottlenecks actually are.

PR #1886, opened 2024-10-11
JIT: Fix pattern matching for enums.

chriseth:
The pattern matching code for enums used comma-separated variables instead of proper tuples.
Also type constructors can only be `Callable::Fn` when they take a single tuple of field values instead of multiple parameters.

### DIFF
--- a/jit-compiler/src/codegen.rs
+++ b/jit-compiler/src/codegen.rs
@@ -433,10 +433,12 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
                     escape_symbol(&decl.name),
                     escape_symbol(&variant.name)
                 );
-                if variant.fields.is_none() {
-                    formatted_variant
+                if let Some(fields) = &variant.fields {
+                    // Callable::Fn takes only a single argument, so we pack the fields into a tuple.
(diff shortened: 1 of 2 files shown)


jit-compiler/src/codegen.rs
```diff
@@ -433,10 +433,12 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
                     escape_symbol(&decl.name),
                     escape_symbol(&variant.name)
                 );
-                if variant.fields.is_none() {
-                    formatted_variant
+                if let Some(fields) = &variant.fields {
+                    // Callable::Fn takes only a single argument, so we pack the fields into a tuple.
+                    let field_vars = (0..fields.len()).map(|i| format!("v_{i}")).join(", ");
```
pacheco, line 438: very minor, but i just for "uniformity", somewhere else there's double underscore for generated names (`item__{i}`), maybe here too?

chriseth: The double underscore was actually in "reference" to `scrutinee__`. These variables could potentially clash with existing variables or even with similar variables in other patterns. Added a comment to `item__{i}` regarding that.

For the variables here, it is different, they cannot clash - at least I think... maybe if the enum is called `v_1`, but maybe not even then.

pacheco (approved):

PR #1887, opened 2024-10-11
Change jit error level to debug.

chriseth:
The JIT compiler errors were spamming the log. This reduces the verbosity in the default log level, but also prints all symbols that failed to compile.


leonardoalt (approved):
lgtm but lint failing

leonardoalt (approved):

PR #1889, opened 2024-10-11
RISCV BabyBear machine

leonardoalt:
Replaces https://github.com/powdr-labs/powdr/pull/1790

This PR:
- [X] implements the basic RISCV BB machine without precompiles
- [X] runs the instruction tests with BB as well
- [x] runs the Rust tests except for continuations and tests that require keccak, poseidon, ec ops
- [x] Missing: inputs and outputs with 2 limbs. There is already a failing test that covers that.


leonardoalt:
@georgwiese I fixed most of your comments from https://github.com/powdr-labs/powdr/pull/1790 in here

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1793 @@
...
+        link ~> (tmp1_h, tmp1_l) = regs.mload(0, XL, STEP)
+        link ~> (tmp2_h, tmp2_l) = regs.mload(0, YL, STEP + 1)
+        link ~> (tmp3_h, tmp3_l) = add_sub.sub(tmp1_h, tmp1_l, tmp2_h, tmp2_l)
+    {
+        // TODO this could be optimized by removing the sub call above.
+        // What we want to do is to compare tmp1_h == tmp2_h && tmp1_l == tmp2_l.
+        XXIsZero = 1 - XX * XX_inv,
+        XX = (tmp3_h + tmp3_l),
+        pc' = (1 - XXIsZero) * l + XXIsZero * (pc + 1)
+    }
```
Schaeff, line 395: Maybe related: https://github.com/valida-xyz/valida/blob/a39c986f328457239aadcf8917e943dfaf4df875/cpu/src/columns.rs#L24

Schaeff:
```suggestion
    instr branch_if_not_equal XL, YL, l: label
        link ~> (tmp1_h, tmp1_l) = regs.mload(0, XL, STEP)
        link ~> (tmp2_h, tmp2_l) = regs.mload(0, YL, STEP + 1)
    {
        XXIsZero = 1 - XX * XX_inv,
        XX = (tmp1_h - tmp1_l)**2 + (tmp2_h - tmp2_l)**2,
        pc' = (1 - XXIsZero) * l + XXIsZero * (pc + 1)
    }
```

Would it reduce to this?

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1793 @@
...
+    // ================= logical instructions =================
+
+    // Stores 1 in register W if the value in register X is zero,
+    // otherwise stores 0.
+    instr is_equal_zero XL, WL
+        link ~> (tmp1_h, tmp1_l) = regs.mload(0, XL, STEP)
+        link ~> regs.mstore(0, WL, STEP + 2, 0, XXIsZero)
+    {
+        XXIsZero = 1 - XX * XX_inv,
+        XX = (tmp1_h + tmp1_l)
```
Schaeff, line 533: Is this sound? I would have thought `tmp1_h * tmp1_l`

leonardoalt: It's different semantics:
- `tmp1_h * tmp1_l` is 0 if either is 0
- but here we want to test whether both limbs are zero at the same time (so the whole number is 0), and that is only true if the sum is 0

Schaeff, line 533: This is sound because `tmp1_h` and `tmp1_l` cannot be opposites of each other?

leonardoalt: not sure I follow. They are unrelated to each other, but they do compose the 32-bit number `tmp1_h * 2**16 + tmp1_l`. We want to check whether that 32-bit number is 0, which is only true if both limbs are 0

Schaeff: My previous comment was wrong, my bad!
Right, and checking the sum works because the limbs cannot be negative?

leonardoalt: yes each limb is always unsigned 2 bytes. The whole 32-bit number may be negative in 2s complement

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1793 @@
...
+    // Jump to the address in register X and store the return program counter in register W.
+    instr jump_dyn XL, WL
+        link ~> (tmp1_h, tmp1_l) = regs.mload(0, XL, STEP)
+        link ~> regs.mstore(0, WL, STEP, tmp2_h, tmp2_l)
+        // pc is capped at 24 bits, so for this instruction 
+        // we restrict the higher limbs to 1 byte
+        link => byte.check(tmp1_h)
+        link => byte.check(tmp2_h)
+    {
+        // We need a temporary col here because the compiler doesn't like
```
chriseth, line 377: We could also fix it. The only reason we have this is because we are excessively trying to reduce the multplicative degree.

chriseth: Here is the fix: https://github.com/powdr-labs/powdr/pull/1893

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1793 @@
...
+    link => byte2.check(tmp3_l);
+    link => byte2.check(tmp4_h);
+    link => byte2.check(tmp4_l);
+    link => byte2.check(tmp5_h);
+    link => byte2.check(tmp5_l);
+    link => byte2.check(tmp6_h);
+    link => byte2.check(tmp6_l);
+
+    // We need to add these inline instead of using std::utils::is_zero
+    // because when XX is not constrained, witgen will try to set XX,
```
chriseth, line 337: Why was this not a problem in the other machine?

chriseth: You could also just add a hint.

riscv/src/small_field/code_gen.rs
leonardoalt: Thanks!

riscv/src/small_field/code_gen.rs
leonardoalt: Yea that's a nice trick! I think that implementation indeed has soundness issues due to overflows. I wrote another version that further decomposes the 16-bit limbs into bytes, I *think* it should be sound, please check.

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1825 @@
...
+        // 32bits tmp2 = tmp5_h * 2**24 + tmp5_l * 2**16 + tmp6_h * 2**8 + tmp6_l
+
+        // The constraint below leads to a plonky3 translation error:
+        // thread '<unnamed>' panicked at /home/leo/devel/powdr2/plonky3/src/circuit_builder.rs:259:25:
+        // internal error: entered unreachable code: exponentiations should have been evaluated
+        // So we inline the squares.
+
+        // This should fit in 19 bits.
+        //XX = (tmp3_h - tmp5_h)**2 + (tmp3_l - tmp5_l)**2 + (tmp4_h - tmp6_h)**2 + (tmp4_l - tmp6_l)**2,
+        XX = (tmp3_h - tmp5_h)*(tmp3_h - tmp5_h) + (tmp3_l - tmp5_l)*(tmp3_l - tmp5_l) + (tmp4_h - tmp6_h)*(tmp4_h - tmp6_h) + (tmp4_l - tmp6_l)*(tmp4_l - tmp6_l),
```
leonardoalt, line 424: @Schaeff @georgwiese this here

riscv/src/small_field/code_gen.rs
leonardoalt: > You could also just add a hint.

I think we tried this while prototyping the machine and for some reason it didn't work. Can you give me an example in the asm machine?

leonardoalt:
There are some optimizations we can still make, but I think this is mergeable and we can open follow up PRs with optimizations.

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+        link => byte.check(tmp4_h)
+        link => byte.check(tmp4_l)
+        link => byte.check(tmp5_h)
+        link => byte.check(tmp5_l)
+        link => byte.check(tmp6_h)
+        link => byte.check(tmp6_l)
+    {
+        XXIsZero = 1 - XX * XX_inv,
+
+        // What we want to do is to compare tmp1_h == tmp2_h && tmp1_l == tmp2_l.
```
Schaeff, line 396:
```suggestion
        // We want to check `tmp1_h == tmp2_h && tmp1_l == tmp2_l`.
```

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+        // This could be done with the constraint below, but overflows causes soundness issues:
+        // XX = (tmp1_h - tmp2_h)**2 + (tmp1_l - tmp2_l)**2,
+        // One solution is to further decompose into single bytes.
+
+        // Decompose tmp1_h into 2 bytes
+        tmp1_h = tmp3_h * 2**8 + tmp3_l,
```
Schaeff, line 402: This is a good example of why we should allow links inside bodies, so that the range checks are next to these decompositions. Not for this PR, but is there anything blocking this?

leonardoalt: No this is working, just wanted more people to check soundness. The byte range checks are in the links above also.

riscv/src/small_field/code_gen.rs
Schaeff: Defining the euclidean norm (squared) in PIL could make this neater.

In pseudo-PIL:

```rust
let squared_euclidean_distance = |left: fe[], right: fe[]| {
   assert_eq(len(left), len(right));
   array::fold(array::map(array::zip(left, right), |(left, right)| (left - right) * (left - right)), |acc, e| acc + e)
};

XX = squared_euclidean_distance([tmp3_h, tmp3_l, tmp5_h, tmp5_l], [tmp4_h, tmp4_l, tmp6_h, tmp6_l]),
```

Maybe this is not needed if we only use this once though.

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+        // now we have that
+        // 32bits tmp1 = tmp3_h * 2**24 + tmp3_l * 2**16 + tmp4_h * 2**8 + tmp4_l
+        // 32bits tmp2 = tmp5_h * 2**24 + tmp5_l * 2**16 + tmp6_h * 2**8 + tmp6_l
+
+        // The constraint below leads to a plonky3 translation error:
+        // thread '<unnamed>' panicked at /home/leo/devel/powdr2/plonky3/src/circuit_builder.rs:259:25:
+        // internal error: entered unreachable code: exponentiations should have been evaluated
+        // So we inline the squares.
+
+        // This should fit in 19 bits.
```
Schaeff, line 419: What is the meaning of "should" here

riscv/src/small_field/code_gen.rs
Schaeff: Oh I don't mean that it doesn't work, just that I had to jump back up to the links the check the range checks and it would be great they were here instead. Unrelated to this PR.

riscv/src/small_field/code_gen.rs
leonardoalt: "it fits, as long as I didn't miss anything" will rephrase

riscv/src/small_field/code_gen.rs
leonardoalt: ah ok

backend/src/composite/split.rs
```diff
@@ -173,13 +173,16 @@ fn referenced_namespaces<F: FieldElement>(
 fn split_by_namespace<F: FieldElement>(
     pil: &Analyzed<F>,
 ) -> BTreeMap<String, Vec<StatementIdentifier>> {
+    let mut current_namespace = "".to_string();
```
pacheco, line 176: are the changes to this file BB related?

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+
+use crate::code_gen::{
+    InstructionArgs, MemEntry, Register, RiscVProgram, SourceFileInfo, Statement,
+};
+use crate::CompilerOptions;
+
+use crate::small_field::runtime::Runtime;
+
+/// Translates a RISC-V program to POWDR ASM
+/// with constraints that work for a field >16 bits.
```
pacheco, line 19: is the `>16` part correct? i guess it is for the asm machine right? only the submachines are more field specific?

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+///
+/// Will call each of the methods in the `RiscVProgram` just once.
+pub fn translate_program(program: impl RiscVProgram, options: CompilerOptions) -> String {
+    let runtime = Runtime::new(options.libs, options.continuations);
+
+    // Do this in a separate function to avoid most of the code being generic on F.
+    // TODO I think this isn't needed anymore
```
pacheco, line 26: does this todo still make sense?

riscv/src/small_field/code_gen.rs
leonardoalt: The asm machines are also specific to field size, though some require 16, 17, 18, or more bits. Poseidon is field specific. The RISCV asm machine here requires 16 bits for the limbs, but in some places it does require more. I'll update this comment.

riscv/src/small_field/code_gen.rs
leonardoalt: I wrote this TODO actually, because now the functions aren't generic on F anymore so we could do what the comment says. But I think it's actually fine as is so I'd just remove both comments.

riscv/src/small_field/runtime.rs
```diff
@@ -0,0 +1,354 @@
...
+        // Base syscalls
+        r.add_syscall(
+            // TODO this is a quite inefficient way of getting prover inputs.
+            // We need to be able to access the register memory within PIL functions.
+            Syscall::Input,
+            [
+                "query_arg_1_h, query_arg_1_l <== get_reg(10);",
+                "query_arg_2_h, query_arg_2_l <== get_reg(11);",
+                // TODO fix for BabyBear
+                "set_reg 10, 0, ${ std::prelude::Query::Input(std::convert::int(std::prover::eval(query_arg_1_l)), std::convert::int(std::prover::eval(query_arg_2_l))) };",
```
leonardoalt, line 199: @pacheco I think these are fine for this PR to keep going but we should make a plan soon.

riscv/src/small_field/code_gen.rs
pacheco: do we need to restrict the limbs to 16-bit here to avoid `(1+(modulus-1))=0`? i guess not because the values come from memory?

pacheco: nevermind, just saw the closed discussion

riscv/tests/riscv.rs
```diff
@@ -96,46 +96,43 @@ fn bn254_sanity_check() {
...
-    verify_riscv_crate_gl(case, Default::default())
+    verify_riscv_crate_bb_gl(case, Default::default())
 }
 
 #[test]
 #[ignore = "Too slow"]
 fn zero_with_values() {
     let case = "zero_with_values";
-    verify_riscv_crate_gl(case, Default::default())
+    verify_riscv_crate_bb_gl(case, Default::default())
```
pacheco, line 113: why not just name this `verify_riscv_crate`?

riscv/src/small_field/runtime.rs
pacheco: yeah, left a couple of comments but the PR lgtm so we can move on and iterate

Schaeff (commented):
Made it halfway through codegen, partial review

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+        XXIsZero * 0xffff + (1 - XXIsZero) * tmp4_h = tmp6_h,
+        XXIsZero * 0xffff + (1 - XXIsZero) * tmp4_l = tmp6_l
+    }
+"# + mul_instruction
+}
+
+fn mul_instruction() -> &'static str {
+    r#"
+    // Computes V = val(X) * val(Y) and
+    // stores the lower 32 bits in register Z and the upper 32 bits in register W.
```
Schaeff, line 663:
```suggestion
    // stores the lower 32 bits in register ZL and the upper 32 bits in register WL.
```

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+            }
+            .push(comment);
+        }
+        match value {
+            SingleDataValue::Value(v) => {
+                if continuations {
+                    // Instead of generating the data loading code, we store it
+                    // in the variable that will be used as the initial memory
+                    // snapshot, committed by the bootloader.
+                    // TODO for BabyBear
```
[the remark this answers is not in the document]
pacheco: this PR doesn't handle continuations for BB, TODO is a reminder of that afaiu

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+    memory_machine.to_string()
+        + r#"
+
+    // Increased by 4 in each step, because we do up to 4 register memory accesses per step
+    col fixed STEP(i) { 4 * i };
+
+    // ============== memory instructions ==============
+
+    /// Loads one word from an address V = val(X) + Y and rounds it down to the next multiple of 4.
+    /// Writes the loaded word and the remainder of the division by 4 to registers Z and W,
```
Schaeff, line 691:
```suggestion
    /// Writes the loaded word and the remainder of the division by 4 to registers ZL and WL,
```

I don't know if these are left here on purpose

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+            .map(|s| format!("\t\t{s}\n"))
+            .join("")
+        + &bootloader_preamble_if_included
+        + &memory(with_bootloader)
+        + r#"
+    // =============== Register memory =======================
+"# + "std::machines::small_field::memory::Memory regs(bit12, byte2);"
+        + r#"
+    // Get the value in register Y.
+    instr get_reg YL -> XH, XL link ~> (XH, XL) = regs.mload(0, YL, STEP);
```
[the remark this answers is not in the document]
pacheco: yeah, we currently use a single field for addresses in BB, and for register memory in particular bits from the L part are enough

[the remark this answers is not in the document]
pacheco: actually, I'm wrong, this PR was updated to use 2 field elements... I was going by an older version of this. But in any case, for registers, the low bits of the addresses are enough

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+    col witness XX, XX_inv, XXIsZero;
+    std::utils::force_bool(XXIsZero);
+    XXIsZero * XX = 0;
+
+    // ============== control-flow instructions ==============
+
+    // Load the value of label `l` into register X.
+    // We restrict the address to 24 bits to avoid overflows.
+    instr load_label XL, l: label
+        link ~> regs.mstore(0, XL, STEP, tmp1_h, tmp1_l)
```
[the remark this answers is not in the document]
pacheco: it reuses the memory machine, but only needs the low bits for referencing register addresses

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+}
+
+/// Push register into the stack
+
+pub fn push_register(name: &str) -> Vec<String> {
+    assert!(name.starts_with('x'), "Only x registers are supported");
+    let reg = Register::from(name);
+    vec![
+        // x2 + x0 - 4 => x2
+        format!("add_wrap 2, 0, 0xffff, 0xfffc, 2;",),
```
Schaeff, line 743: Would it be useful to have a utility function to turn -4 into (0xffff, 0xfffc)?

Schaeff: Based on `i32_high(-4), i32_low(-4)` ?

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+                    rd.addr(),
+                ),
+            )
+        }
+        "mulh" => {
+            let (rd, r1, r2) = args.rrr()?;
+            only_if_no_write_to_zero_vec(
+                rd,
+                vec![
+                    //format!("to_signed {}, {};", r1.addr(), tmp1.addr()),
```
Schaeff, line 910: leftover? and below

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+                    ),
+                ],
+            )
+        }
+        "mulhsu" => {
+            let (rd, r1, r2) = args.rrr()?;
+            only_if_no_write_to_zero_vec(
+                rd,
+                vec![
+                    //format!("to_signed {}, {};", r1.addr(), tmp1.addr()),
```
Schaeff, line 977: Leftover?

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+
+        tmp4_l = (1 - wrap_bit) * (1 - wrap_bit_2),
+        tmp4_h = wrap_bit * wrap_bit_2,
+        tmp5_l = (1 - wrap_bit) * wrap_bit_2,
+        //... = wrap_bit * (1 - wrap_bit_2),
+
+        X_b3 = tmp4_l * (1 - tmp5_h) +
+               tmp4_h * (1 - tmp5_h) +
+               tmp5_l, /* * 1 */
+               /* ...* 0, */
```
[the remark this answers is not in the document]
pacheco: This would be much nicer if we could have "aliases" for columns inside instructions.
The idea is as follows:
We need to compare two signed numbers: `wrap_bit` and `wrap_bit_2` are the signs.
`tmp4_l`: are both numbers positive?
`tmp4_h`: are both numbers negative?
`tmp5_l`: is the first number positive and the second number negative?

`X_b3` does the math for whether we should jump or not (i.e., X >= Y), considering the different combinations of signs and the result of the comparison operation in the submachine (add_sub.gt)

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+        // bitwise
+        "xor" => {
+            let (rd, r1, r2) = args.rrr()?;
+            only_if_no_write_to_zero(
+                rd,
+                format!("xor {}, {}, 0, 0, {};", r1.addr(), r2.addr(), rd.addr()),
+            )
+        }
+        "xori" => {
+            let (rd, r1, imm) = args.rri()?;
```
Schaeff, line 1053: Idea: if `rri` returned `(Register, Register, U32)` where U32 in a wrapper around U32 whose Display is implemented as `"{}, {}", u32_high(self.0), u32_low(self.0)` then there would be no diff here vs large fields. Not required for this PR though.

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+            // arithmetic shift right
+            // TODO see if we can implement this directly with a machine.
+            // Now we are using the equivalence
+            // a >>> b = (a >= 0 ? a >> b : ~(~a >> b))
+            let (rd, rs, amount) = args.rri()?;
+            assert!(amount <= 31);
+            only_if_no_write_to_zero_vec(
+                rd,
+                vec![
+                    //format!("to_signed {}, {};", rs.addr(), tmp1.addr()),
```
Schaeff, line 1168: Leftover?

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+        "snez" => {
+            let (rd, rs) = args.rr()?;
+            only_if_no_write_to_zero(rd, format!("is_not_equal {}, 0, {};", rs.addr(), rd.addr()))
+        }
+        "slti" => {
+            let (rd, rs, imm) = args.rri()?;
+            only_if_no_write_to_zero_vec(
+                rd,
+                vec![
+                    //format!("to_signed {}, {};", rs.addr(), tmp1.addr()),
```
Schaeff, line 1210: Leftover?

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+                ],
+            )
+        }
+        "slt" => {
+            let (rd, r1, r2) = args.rrr()?;
+            only_if_no_write_to_zero_vec(
+                rd,
+                vec![
+                    //format!("to_signed {}, {};", r1.addr(), tmp1.addr()),
+                    //format!("to_signed {}, {};", r2.addr(), tmp2.addr()),
```
Schaeff, line 1239: Leftover?

Schaeff (commented):
Finished code_gen

riscv/src/small_field/runtime.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub fn ecall_handler(&self) -> Vec<String> {
+        let ecall = [
+            "// ecall handler".to_string(),
+            "__ecall_handler:".to_string(),
+        ]
+        .into_iter();
+
+        let jump_table = self.syscalls.keys().map(|s| {
+            let s32_h = ((*s as u32) >> 16) as u16;
+            let s32_l = ((*s as u32) & 0xffff) as u16;
```
Schaeff, line 335: there are utils for this `u32_low` `u32_high`

riscv/src/small_field/runtime.rs
```diff
@@ -0,0 +1,354 @@
...
+        let mut runtime = Runtime::base();
+        if libs.poseidon {
+            runtime = runtime.with_poseidon(continuations);
+        }
+        if libs.keccak {
+            runtime = runtime.with_keccak();
+        }
+        if libs.arith {
+            runtime = runtime.with_arith();
+        }
```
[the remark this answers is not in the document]
Schaeff: Ok seems like calling these functions panics with todo, fine

riscv/src/small_field/code_gen.rs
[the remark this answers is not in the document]
Schaeff: Ah I see, regs and memory are the same machine type! Closing

riscv/src/small_field/code_gen.rs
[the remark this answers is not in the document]
Schaeff: What about the commented out parts? What's the reasoning there?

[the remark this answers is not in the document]
leonardoalt: They were left as documentation to show the full truth table, but not needed cus mul by 0

riscv/src/small_field/runtime.rs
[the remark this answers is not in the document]
leonardoalt: yep, will be implemented in follow up PRs. These tests are also GL only for now.

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+    program: Vec<String>,
+) -> String {
+    format!(
+        r#"
+{}
+use std::machines::small_field::add_sub::AddSub;
+use std::machines::small_field::arith::Arith;
+machine Main with min_degree: {}, max_degree: {} {{
+AddSub add_sub(byte2);
+Arith arith_mul(byte, byte2);
```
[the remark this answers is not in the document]
leonardoalt: Not sure, I assume because they're more fundamental and used by a bunch of basic instructions of the machine. I could split it as well, but would prefer to do that in a different PR.

riscv/src/small_field/code_gen.rs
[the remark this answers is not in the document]
leonardoalt: yes registers are small so we always use one limb only

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+        link => byte.check(tmp1_h)
+    {
+        tmp1_h * 2**16 + tmp1_l = l
+    }
+
+    // Jump to `l` and store the return program counter in register W.
+    // We restrict the address to 24 bits to avoid overflows.
+    instr jump l: label, WL
+        link ~> regs.mstore(0, WL, STEP, tmp1_h, tmp1_l)
+        link => byte.check(tmp1_h)
```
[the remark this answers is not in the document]
leonardoalt: yep that's actually the reason why we're constraining this here

riscv/src/small_field/code_gen.rs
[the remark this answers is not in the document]
leonardoalt: added some text comments to clarify

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+               /* ...* 0, */
+
+        pc' = X_b3 * l + (1 - X_b3) * (pc + 1)
+    }
+
+    // Stores 1 in register W if val(X) >= val(Y) <=> not(val(X) < val(Y)).
+    instr is_greater_or_equal XL, YL, WL
+        link ~> (tmp1_h, tmp1_l) = regs.mload(0, XL, STEP)
+        link ~> (tmp2_h, tmp2_l) = regs.mload(0, YL, STEP + 1)
+        link ~> wrap_bit = add_sub.gt(tmp1_h, tmp1_l, tmp2_h, tmp2_l)
```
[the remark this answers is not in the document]
leonardoalt: yep, for now had to keep it like this for witgen. Eventually it should be constrained inside each instruction that needs it

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+               tmp4_h * (1 - tmp6_l) +
+               tmp5_l /* * 1 */
+               /* tmp5_h * 0, */
+    }
+
+    // Stores val(X) * Z + W in register Y.
+    instr affine XL, YL, ZH, ZL, WH, WL
+        link ~> (tmp1_h, tmp1_l) = regs.mload(0, XL, STEP)
+        link ~> (tmp3_h, tmp3_l, tmp2_h, tmp2_l) = arith_mul.mul(tmp1_h, tmp1_l, 0, 0, ZH, ZL)
+        // we ignore tmp3 because that's the high 32 bits of the 64 bits multiplication result
```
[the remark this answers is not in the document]
leonardoalt: oh we're just not interested in those values

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+        .submachines_instructions()
+        .into_iter()
+        .map(|s| format!("    {s}"))
+        .join("\n")
+        + r#"
+    col witness X_b1;
+    col witness X_b2;
+    col witness X_b3;
+
+    // TODO Remove these once witgen supports Boolean
```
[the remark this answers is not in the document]
leonardoalt: Boolean comes from George Boole so some people prefer using the capitalized version (me included). If you have a strong opinion I'll change.

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+        tmp3_l = tmp1_l
+    }
+
+    // ======================= assertions =========================
+
+    instr fail 
+        link ~> (tmp1_h, tmp1_l) = regs.mload(0, 0, STEP)
+    {
+      tmp1_h = 1
+    }
```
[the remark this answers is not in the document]
leonardoalt: it does but the witgen error message is impossible to understand that this is what happened. This makes it a bit more clear

riscv/src/small_field/code_gen.rs
leonardoalt: fixed all of these to use the same register name as the argument

riscv/src/small_field/code_gen.rs
leonardoalt: removed

riscv/src/small_field/code_gen.rs
leonardoalt: done

riscv/src/small_field/runtime.rs
leonardoalt: done

leonardoalt:
@Schaeff @pacheco I think I've fixed all comments

riscv/src/small_field/code_gen.rs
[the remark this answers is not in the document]
Schaeff: Fair point, no strong opinion, closing

riscv/src/small_field/code_gen.rs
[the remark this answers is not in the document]
Schaeff: the Altan syntax. fine

riscv/src/small_field/code_gen.rs
```diff
@@ -0,0 +1,1822 @@
...
+
+        tmp4_l = (1 - wrap_bit) * (1 - wrap_bit_2),
+        tmp4_h = wrap_bit * wrap_bit_2,
+        tmp5_l = (1 - wrap_bit) * wrap_bit_2,
+        //tmp5_h = wrap_bit * (1 - wrap_bit_2),
+ 
+        X_b3 = tmp4_l * (1 - tmp6_l) +
+               tmp4_h * (1 - tmp6_l) +
+               tmp5_l /* * 1 */
+               /* tmp5_h * 0, */
```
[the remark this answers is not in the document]
Schaeff:
```
tmp5_l /* * 1 */
/* + tmp5_h * 0, */
```

riscv/src/small_field/code_gen.rs
Schaeff: Comment still there and squares not there yet

leonardoalt: Oh weird, will check

Schaeff (approved):
LGTM based on comparison with the large implementation. It's a large change though so I could have missed something.

PR #1897, opened 2024-10-14
Fix typo.

chriseth:


leonardoalt (approved):

std/machines/small_field/shift.asm
```diff
@@ -21,7 +21,7 @@ machine ByteShift with
     let bit_counts = [256, 32, 4, 2];
     let min_degree = std::array::product(bit_counts);
     std::check::assert(std::prover::min_degree() >= std::array::product(bit_counts), || "The shift16 machine needs at least 65536 rows to work.");
-    require_field_bits(16, || "The field modulo should be at least 2^16 - 1 to work in the shift16 machine.");
+    require_field_bits(16, || "The field modulus should be at least 2^16 - 1 to work in the shift16 machine.");
```
onurinanc, line 24:
```suggestion
    require_field_bits(16, || "The field modulus should be at least 2^16 - 1 to work in the shift machine.");
```

std/machines/small_field/shift.asm
```diff
@@ -21,7 +21,7 @@ machine ByteShift with
     let bit_counts = [256, 32, 4, 2];
     let min_degree = std::array::product(bit_counts);
     std::check::assert(std::prover::min_degree() >= std::array::product(bit_counts), || "The shift16 machine needs at least 65536 rows to work.");
```
onurinanc, line 23:
```suggestion
    std::check::assert(std::prover::min_degree() >= std::array::product(bit_counts), || "The shift machine needs at least 65536 rows to work.");
```

onurinanc (changes requested):

std/machines/small_field/shift.asm
chriseth: I think this is fine to distinguish the small field and the large field shift machine.

PR #1899, opened 2024-10-14
JIT: Implement the degree builtin.

chriseth:


jit-compiler/src/compiler.rs
```diff
@@ -59,7 +58,14 @@ pub fn generate_glue_code<T: FieldElement>(
 }
 
 const PREAMBLE: &str = r#"
-#![allow(unused_parens)]
+#![allow(unused_parens, unused_variables)]
+
+static DEGREE: std::sync::RwLock<Option<ibig::IBig>> = std::sync::RwLock::new(None);
+
+#[no_mangle]
+pub extern "C" fn __set_degree(degree: u64) {
```
chriseth, line 66: This is my proposal for how to do built-in symbols. Since degree is so simple, it can just be a stored number.

For more complicated builtin functions, we probably need a proper callback to the host code.

lvella: Why `RwLock`? Is the jit executor multithreaded? `RefCell` isn't enough?

lvella (commented):
I what situation the same JIT compiled code will have to run with different degrees?

Or from another perspective, why does it has to be dynamically settable?

jit-compiler/src/compiler.rs
chriseth: Yes it is. Fixed columns are computed in parallel.

chriseth: I mean we have the case that we never change the degree while having multiple threads, but that's just the promise of RwLock, isn't it?

chriseth:
The degree can be different for different fixed columns.

jit-compiler/src/compiler.rs
lvella: Yes, `RefCell` not `!Sync`, so it can't be used from different threads, it has to be `RwLock`.

lvella (approved):

lvella (dismissed):

lvella (approved):

PR #1900, opened 2024-10-14
Need to clone local variables as well.

chriseth:


jit-compiler/src/codegen.rs
```diff
@@ -189,7 +189,7 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
 
     fn format_expr(&mut self, e: &Expression, var_height: usize) -> Result<String, String> {
         Ok(match e {
-            Expression::Reference(_, Reference::LocalVar(_id, name)) => name.clone(),
+            Expression::Reference(_, Reference::LocalVar(_id, name)) => format!("({name}).clone()"),
```
chriseth, line 192: ah wait, I might have a different solution.

PR #1901, opened 2024-10-14
Some more logging.

chriseth:


executor/src/constant_evaluator/mod.rs
```diff
@@ -42,6 +44,9 @@ pub fn generate<T: FieldElement>(analyzed: &Analyzed<T>) -> Vec<(String, Variabl
             }
         }
     }
+    if !used_interpreter && !fixed_cols.is_empty() {
```
pacheco, line 47: i think if `used_interpreter==true` then `generate_values` was called and `fixed_cols` can't be empty, no?

chriseth: We could have no columns at all, then the statement would still be true, but...

pacheco (approved):

PR #1906, opened 2024-10-14
Structs: Definitions & Stmt processor

gzanitti:


ast/src/parsed/mod.rs
```diff
@@ -1476,6 +1476,12 @@ impl<Ref> Children<Expression<Ref>> for ArrayExpression<Ref> {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub enum TypeDeclaration {
```
chriseth, line 1480: In which way is this TypeDeclaration different from the one above? Can we make it generic over the type of expression?

pil-analyzer/src/statement_processor.rs
```diff
@@ -420,6 +436,36 @@ where
         }
     }
 
+    fn process_struct_declaration(
```
chriseth, line 439: Why not put this next to `process_enum_declaration`?

ast/src/parsed/mod.rs
chriseth: Also why not put this next to StructDeclaration and EnumDeclaration? How is it related to Pattern / ArrayExpression?

pil-analyzer/src/statement_processor.rs
```diff
@@ -420,6 +436,36 @@ where
...
+        &mut self,
+        symbol: Symbol,
+        struct_decl: StructDeclaration<parsed::Expression>,
+    ) -> Vec<PILItem> {
+        let type_vars = struct_decl.type_vars.vars().collect();
+        let fields = struct_decl
+            .fields
+            .into_iter()
+            .map(|v| NamedType {
+                name: v.name.to_string(),
```
chriseth, line 449: Can't you move here? Please use `.clone()` instead of `to_string()` if you are just cloning.

pilopt/src/lib.rs
```diff
@@ -123,6 +128,12 @@ impl ReferencedSymbols for EnumDeclaration {
     }
 }
 
+impl ReferencedSymbols for StructDeclaration {
```
chriseth, line 131: Can you also implement this for TypeDeclaration and then just call it there above?

pil-analyzer/src/statement_processor.rs
```diff
@@ -647,6 +614,86 @@ where
...
+        &mut self,
+        symbol: Symbol,
+        struct_decl: StructDeclaration<parsed::Expression>,
+    ) -> Vec<PILItem> {
+        let type_vars = struct_decl.type_vars.vars().collect();
+        let fields = struct_decl
+            .fields
+            .into_iter()
+            .map(|v| NamedType {
+                name: v.name.clone(),
```
chriseth, line 627: Why do you need to clone here? Maybe you need a destructuring assignment at the start of the function
```
let StructDeclaration{name, type_vars, fields} = struct_decl;
```

pil-analyzer/src/statement_processor.rs
```diff
@@ -647,6 +614,91 @@ where
...
+        let struct_decl = StructDeclaration {
+            name: self.driver.resolve_decl(&name),
+            type_vars,
+            fields,
+        };
+
+        iter::once(PILItem::Definition(
+            symbol,
+            Some(FunctionValueDefinition::TypeDeclaration(
+                TypeDeclaration::Struct(struct_decl.clone()),
```
chriseth, line 646:
```suggestion
                TypeDeclaration::Struct(struct_decl),
```

chriseth (approved):

PR #1907, opened 2024-10-15
Structs: Fields check

gzanitti:
Note: Like #1910, this PR also includes the same modifications to expression processing, in order to run tests.


pil-analyzer/tests/types.rs
```diff
@@ -734,3 +734,39 @@ fn prover_functions() {
...
+    let input = "
+    struct Dot { x: int, y: int }
+    let f: int -> Dot = |i| Dot{a: 0, y: i};
+    let x = f(0);
+    ";
+    type_check(input, &[]);
+}
+
+#[test]
+#[should_panic = "Field 'z' is declared but never used in struct 'Point' initialization"]
```
chriseth, line 761: I would say something like "forgot field z" - please check what rustc's errors look like.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +258,69 @@ impl PILAnalyzer {
         }
     }
 
+    pub fn structs_check(&self) -> Result<(), Vec<Error>> {
```
chriseth, line 261: Every public function needs documentation.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -56,6 +56,7 @@ fn analyze<T: FieldElement>(files: Vec<PILFile>) -> Result<Analyzed<T>, Vec<Erro
     let mut analyzer = PILAnalyzer::new();
     analyzer.process(files)?;
     analyzer.side_effect_check()?;
+    analyzer.structs_check()?;
```
chriseth, line 59: Maybe `struct_fields_check` to be more precise?

pil-analyzer/src/pil_analyzer.rs
chriseth: Ah wait, it doesn't need to be public. But please document it nevertheless :)

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +258,69 @@ impl PILAnalyzer {
...
+                ))) = value
+                {
+                    Some((name.clone(), struct_decl.clone()))
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        for (_, def) in self.definitions.values() {
```
chriseth, line 278: This is not enough, we also need to traverse proof items and impls. Maybe the easiest is to implement `Children<Expression>` on `PilAnalyzer` and then call `self.all_children()`.

pil-analyzer/tests/types.rs
```diff
@@ -734,3 +734,39 @@ fn prover_functions() {
     ";
     type_check(input, &[("a", "", "int"), ("b", "", "()[]")]);
 }
+
```
chriseth, line 737: Please add a test like `enum A { X } let a = A{x: 8};`, i.e. where you use an enum for a struct expression.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +258,69 @@ impl PILAnalyzer {
...
+                    let Some(struct_decl) = struct_declarations.get(&struct_expr.name.to_string())
+                    else {
+                        errors.push(sr.with_error(format!(
+                            "Struct '{}' has not been declared",
+                            struct_expr.name
+                        )));
+                        return;
+                    };
+
+                    let mut used_fields = HashSet::new();
```
chriseth, line 292: What about something like
```
let declared_fields: HashSet<_> = struct_decl.fields.iter().map(|f| f.name).collect();
let used_fields: HashSet<_> = sturct_expr.fields...;
for f in declared_fields.difference(used_field) { ... }
for f in used_fields.difference(declared_fields) { ... }
```

chriseth: The last two lines could even be
```
errors.extend(declared_fields.difference(used_field).map(|f| ... ))
```

chriseth:
Oh and please also add a test where you mention a field twice.

pil-analyzer/tests/types.rs
```diff
@@ -768,32 +768,68 @@ fn wrong_struct() {
...
+        y: int,
+    }
+    let x = A{ y: 0, x: 2, y: 1 };
+";
+
+    type_check(input, &[]);
+}
+
+#[test]
+#[should_panic = "Field 'x' is already declared"]
```
chriseth, line 814: Why is this a different error message than above?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +258,64 @@ impl PILAnalyzer {
         }
     }
 
+    /// Verifies that all struct instantiations match their corresponding declarations (existence of field names, completeness)
+    fn struct_fields_check(&self) -> Result<(), Vec<Error>> {
+        let mut errors = vec![];
+
+        self.all_children().for_each(|expr| {
+            if let Expression::StructExpression(sr, struct_expr) = expr {
```
chriseth, line 266: You could decrease the indentation here by either using `filter_map` or `for expr in ... { let Expression::StructExpression{ ... } else { continue };` or something else. This is good because when someone reads until here, it is good to know that there will be no `else` clause.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +258,64 @@ impl PILAnalyzer {
...
+                )) = self.definitions.get(&struct_expr.name.to_string())
+                else {
+                    panic!("Struct '{}' has not been declared", struct_expr.name);
+                };
+
+                let mut declared_fields = HashSet::new();
+                for field in &struct_decl.fields {
+                    if !declared_fields.insert(&field.name) {
+                        errors.push(
+                            sr.with_error(format!("Field '{}' is already declared", field.name)),
```
chriseth, line 281: Oh I see where this error message comes from. If you do it this way, then 1) we won't get an error for a struct declaration that is never used and 2) we will get the error at the location of the use and not the declaration and 3) we will get multiple errors for each use.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -542,6 +601,24 @@ impl PILAnalyzer {
     }
 }
 
+impl AllChildren<Expression> for PILAnalyzer {
```
chriseth, line 604: If you just implement Children, will you get AllChildren for free?

chriseth: If you implement `Children`, please take care to return the expression itself instead of calling children on them, because in the latter case, it will skip the expression itself.

pil-analyzer/tests/types.rs
gzanitti: Both errors are for the case of repeated fields, but "specified more than once" is for the instantiation and "is already declared" is for the declaration. It's the way rust handles them.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +257,113 @@ impl PILAnalyzer {
...
+        let structs_exprs = self.all_children().filter_map(|expr| {
+            if let Expression::StructExpression(
+                sr,
+                StructExpression {
+                    name: Reference::Poly(poly),
+                    fields,
+                },
+            ) = expr
+            {
+                Some((sr, poly.name.clone(), fields))
```
chriseth, line 274: Please don't clone if not necessary.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +257,113 @@ impl PILAnalyzer {
...
+                let duplicate_fields: Vec<_> =
+                    fields.iter().map(|f| &f.name).duplicates().collect();
+
+                errors.extend(
+                    duplicate_fields
+                        .into_iter()
+                        .map(|f| sr.with_error(format!("Field '{f}' specified more than once"))),
+                );
+            } else {
+                errors.push(sr.with_error(format!("Struct '{name}' has not been declared")));
```
chriseth, line 317: Can this happen? Shouldn't it be a panic? Or should it rather be an error that we exected a struct and got something else?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +257,113 @@ impl PILAnalyzer {
...
+                ))
+            )
+        });
+
+        for (name, (symbol, def)) in structs_decls {
+            if !visited_structs.contains(name) {
+                errors.push(
+                    symbol
+                        .source
+                        .with_error(format!("warning: struct '{name}' is never constructed")), // TODO: Warningx
```
chriseth, line 337: I think it should be totally fine for a struct to never be constructed.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +257,113 @@ impl PILAnalyzer {
...
+                    .fields
+                    .iter()
+                    .map(|f| &f.name)
+                    .duplicates()
+                    .collect();
+
+                errors.extend(duplicate_declaration_fields.into_iter().map(|f| {
+                    symbol
+                        .source
+                        .with_error(format!("Field '{f}' is already declared",))
```
chriseth, line 353: Declared twice? Or do we have the source reference directly on the second field declaration?

gzanitti: No, the source is for the complete struct. Maybe `is declared more than once` then?

ast/src/parsed/mod.rs
```diff
@@ -42,6 +42,8 @@ pub enum SymbolCategory {
     TypeConstructor,
     /// A trait declaration, which can be used as a type.
     TraitDeclaration,
+    /// A struct
```
chriseth, line 45: If you introduce this and it cannot be unified with `Type` then please also rename `Type`.

ast/src/parsed/mod.rs
```diff
@@ -54,6 +56,9 @@ impl SymbolCategory {
                 request == SymbolCategory::TypeConstructor || request == SymbolCategory::Value
             }
             SymbolCategory::TraitDeclaration => request == SymbolCategory::TraitDeclaration,
+            SymbolCategory::Struct => {
+                request == SymbolCategory::Struct || request == SymbolCategory::Type
```
chriseth, line 60: Please add a comment like the one above.

chriseth: And maybe sort them so that the more complicated ones go first.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -542,6 +560,40 @@ impl PILAnalyzer {
...
+                .filter_map(|(_, def)| def.as_ref())
+                .flat_map(|def| def.children())
+                .chain(
+                    self.trait_impls
+                        .values()
+                        .flat_map(|impls| impls.iter())
+                        .flat_map(|impl_| impl_.children()),
+                )
+                .chain(self.proof_items.iter()),
+        )
```
chriseth, line 577: Maybe also the publics?

chriseth: Ah no it's fine.

pil-analyzer/src/structural_checks.rs
```diff
@@ -0,0 +1,93 @@
...
+                duplicate_fields
+                    .into_iter()
+                    .map(|f| sr.with_error(format!("Field '{f}' specified more than once"))),
+            );
+        }
+
+        visited_structs.insert(name);
+    }
+
+    let structs_decls = definitions.iter().filter(|(_, (_, def))| {
```
chriseth, line 61: Could be a new function starting here?
If you do that, please also create a wrapper function in this file so that we have only one call into this file from pil_analyzer.

pil-analyzer/src/structural_checks.rs
```diff
@@ -0,0 +1,93 @@
...
+    parsed::{StructExpression, TypeDeclaration},
+};
+use powdr_parser_util::{Error, SourceRef};
+
+/// Verifies that all struct instantiations match their corresponding declarations
+/// (existence of field names, completeness) and ensures that both are correct.
+pub fn check_structs_fields<'a>(
+    structs_exprs: impl Iterator<Item = (&'a SourceRef, &'a StructExpression<Reference>)>,
+    definitions: &HashMap<String, (Symbol, Option<FunctionValueDefinition>)>,
+) -> Vec<Error> {
```
chriseth, line 15:
```suggestion
) -> Result<(), Vec<Error>> {
```

pil-analyzer/src/structural_checks.rs
```diff
@@ -0,0 +1,93 @@
...
+                .duplicates()
+                .collect();
+
+            errors.extend(duplicate_declaration_fields.into_iter().map(|f| {
+                symbol
+                    .source
+                    .with_error(format!("Field '{f}' is declared more than once"))
+            }));
+        } else {
+            unreachable!()
```
chriseth, line 88: In that case I would use a `let ... else {unreachable!()};`

pil-analyzer/src/structural_checks.rs
```diff
@@ -0,0 +1,93 @@
...
+use powdr_parser_util::{Error, SourceRef};
+
+/// Verifies that all struct instantiations match their corresponding declarations
+/// (existence of field names, completeness) and ensures that both are correct.
+pub fn check_structs_fields<'a>(
+    structs_exprs: impl Iterator<Item = (&'a SourceRef, &'a StructExpression<Reference>)>,
+    definitions: &HashMap<String, (Symbol, Option<FunctionValueDefinition>)>,
+) -> Vec<Error> {
+    let mut errors = Vec::new();
+    let mut visited_structs = HashSet::new();
```
chriseth, line 17: Does not seem to be needed.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -257,6 +258,23 @@ impl PILAnalyzer {
         }
     }
 
+    pub fn struct_fields_check(&self) -> Result<(), Vec<Error>> {
+        let structs_exprs = self.all_children().filter_map(|expr| {
```
chriseth, line 262: I'm wondering if it would be better to do that inside `check_structs_fields`?

pil-analyzer/src/structural_checks.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    for (sr, st) in structs_exprs {
```
chriseth, line 19:
```suggestion
    for (source, st) in structs_exprs {
```

chriseth: I didn't remember what it was

pil-analyzer/src/pil_analyzer.rs
gzanitti: Not sure, but the idea of having to pass `PILAnalyzer` (to make it possible to do `self.all_children()`), doesn't really convince me.

ast/src/parsed/mod.rs
```diff
@@ -42,23 +42,24 @@ pub enum SymbolCategory {
     TypeConstructor,
     /// A trait declaration, which can be used as a type.
```
chriseth, line 43: Is this implemented like that? I actually don't think we currently do that, and for simplicity, I would prefer to keep it like `let<T: MyTrait<T>> f: T -> int` for now. We can introduce a shorthand like `let f: MyTrait -> int` later like rust does it.

pil-analyzer/src/structural_checks.rs
```diff
@@ -25,39 +24,61 @@ pub fn check_structs_fields<'a>(
...
+}
+
+fn check_struct_expression(
+    definitions: &HashMap<String, (Symbol, Option<FunctionValueDefinition>)>,
+    name: &String,
+    fields: &[powdr_ast::parsed::NamedExpression<Box<powdr_ast::parsed::Expression<Reference>>>],
+    source: &SourceRef,
+) -> Vec<Error> {
+    let mut errors = Vec::new();
+    if let (
```
chriseth, line 46: Shouldn't this panic if it's not a `FunctionValueDefinition::TypeDeclaration`?

gzanitti: Since names comes from StructExpressions, None shouldn't be a possible case, hence the direct unwrap.

pil-analyzer/src/structural_checks.rs
```diff
@@ -25,39 +24,61 @@ pub fn check_structs_fields<'a>(
             unreachable!()
```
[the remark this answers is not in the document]
gzanitti: I think it is mainly due to the `Reference::Poly(...)`. Which I understand is the only possible case, that's why the  `unreachable!()` else

pil-analyzer/src/structural_checks.rs
```diff
@@ -0,0 +1,117 @@
...
+    for expr in structs_exprs {
+        let Expression::StructExpression(
+            source,
+            StructExpression {
+                name: Reference::Poly(PolynomialReference { name, .. }),
+                fields,
+            },
+        ) = expr
+        else {
+            unreachable!()
```
chriseth, line 27: To be honest, instead of this I would prefer just filtering the expressions here instead of in the analyzer, but you can do this in a separate pr.

pil-analyzer/src/structural_checks.rs
chriseth: What I meant is that you have an `if let { ... }` here, but there is no else case. Shouldn't it panic in the else case?

pil-analyzer/src/structural_checks.rs
```diff
@@ -0,0 +1,117 @@
...
+    errors
+}
+
+fn check_struct_declarations(
+    definitions: &HashMap<String, (Symbol, Option<FunctionValueDefinition>)>,
+) -> Vec<Error> {
+    let structs_decls = definitions.iter().filter(|(_, (_, def))| {
+        matches!(
+            def,
+            Some(FunctionValueDefinition::TypeDeclaration(
```
chriseth, line 88: Exactly the same pattern is matched 10 lines below. Why not iterate over all definitions, use `if let Some(FunctionValueDefinition::TypeDeclaration...) = def else { contine; }` ?

chriseth (approved):

PR #1908, opened 2024-10-15
change toolchain to nightly

ShuangWu121:
related to [this PR](https://github.com/powdr-labs/powdr/pull/1898)
we need to change to nightly toolchain to integrate stwo

I kept the toolchain related to riscv to be "nightly-2024-08-01" as it is handled separately in workflow, so I made the least change to make stwo integrate-able for now.

fix some clippy issues about the comment format on some files.


Schaeff:
@ShuangWu121 if you search for "1.81" in the whole codebase, you should get a bunch of finds inside the CI folder. These need to be updated as well

ShuangWu121:
> @ShuangWu121 if you search for "1.81" in the whole codebase, you should get a bunch of finds inside the CI folder. These need to be updated as well. Or maybe you did and didn't commit the files?

aha, on my way

executor/src/constant_evaluator/mod.rs
```diff
@@ -8,6 +8,7 @@ mod interpreter;
 mod jit_compiler;
 
 /// Generates the fixed column values for all fixed columns that are defined
+///
```
chriseth, line 11: Please undo, this adds a paragraph inside a sentence.

chriseth:
Can you explain the reason for doing this and the consequences? Does this mean any project using our crates will need nightly, too?
Does it mean that we don't get any warning about using unstable features? Can we at least restrict feature use by us (instead of the crates we use) to stable-only?

lvella:
Maybe it is too much work, but in every toplevel `lib.rs` and `main.rs` we probably should have a fat warning saying the unstable features should not be enabled, and we are nightly just for the dependencies.

chriseth:
Is there another way? Hallucinating now, but maybe we can link to external crates as binaries or something? We could have a adapter crate that takes its input as binary or something?

lvella:
> Can you explain the reason for doing this and the consequences? Does this mean any project using our crates will need nightly, too? Does it mean that we don't get any warning about using unstable features? Can we at least restrict feature use by us (instead of the crates we use) to stable-only?

If we gate the backends that require nightly behind features, we can still support stable builds. We will need a `cargo check` job with the features disabled using the stable toolchain.

executor/src/constant_evaluator/mod.rs
ShuangWu121: I made the sentence to be one line.
The empty line is to solve this version of Clippy’s lint warning "too_long_first_doc_paragraph", which is warning you that the first paragraph of the documentation comment is too long.

ShuangWu121:
> > Can you explain the reason for doing this and the consequences? Does this mean any project using our crates will need nightly, too? Does it mean that we don't get any warning about using unstable features? Can we at least restrict feature use by us (instead of the crates we use) to stable-only?
> 
> If we gate the backends that require nightly behind features, we can still support stable builds. We will need a `cargo check` CI job with the features disabled using the stable toolchain.

Yes, there is a option to only use nightly to build stwo, the rest of the features still use stable, I try to make this with CI, and it make things very complicated ([check how the pre-test.yml file changed here](https://github.com/powdr-labs/powdr/pull/1898/files)), but I don't know how to use 'cargo check' to do it, maybe it is a better solution? I need to check it out

lvella:
> but I don't know how to use 'cargo check' to do it, maybe it is a better solution? I need to check it out

We don't need to do a full matrix test with different toolchains. We just need to, in one of the jobs (maybe a new one), call `cargo +stable check <plus options to set the correct features>`

.github/workflows/pr-tests.yml
```diff
@@ -39,11 +39,15 @@ jobs:
         key: ${{ runner.os }}-cargo-pr-tests
     - name: Date of the restored cache
       run: cat target/cache-build-date.txt
-      continue-on-error: true
-    - name: Install Rust toolchain 1.81 (with clippy and rustfmt)
-      run: rustup toolchain install 1.81-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain 1.81-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain 1.81-x86_64-unknown-linux-gnu
+      continue-on-error: true
```
chriseth, line 42:
```suggestion
      continue-on-error: true
```

Schaeff (dismissed):

leonardoalt (approved):

PR #1909, opened 2024-10-15
Replacing the old cache without deleting it first.

lvella:
Working on the assumption that there is a race condition between the delete command and the push new cache command, that is why some days we have no cache.


pacheco:
so the cache is not deleted anymore?

lvella:
> so the cache is not deleted anymore?

From my tests, it seems it is being replaced when you save a new cache with the same key. But I made the build job print the date of the cache, so we can monitor this.

chriseth (approved):

PR #1910, opened 2024-10-15
Structs: Type inference & Expression processor

gzanitti:


pil-analyzer/src/pil_analyzer.rs
```diff
@@ -343,7 +358,7 @@ impl PILAnalyzer {
             expressions.push((expr, constr_function_statement_type()));
         }
 
-        let inferred_types = infer_types(definitions, &mut expressions)?;
+        let inferred_types = infer_types(definitions, &mut expressions, struct_declarations)?;
```
chriseth, line 361: You pass the definitions, why do you need to pass the structs as well?

pil-analyzer/src/type_inference.rs
```diff
@@ -729,8 +737,30 @@ impl TypeChecker {
                 self.local_var_types.truncate(original_var_count);
                 result?
             }
-            Expression::StructExpression(_sr, _struct_expr) => {
-                unimplemented!("Struct expressions are not yet supported")
+            Expression::StructExpression(sr, StructExpression { name, fields }) => {
+                let Reference::Poly(PolynomialReference { name, .. }) = name else {
```
chriseth, line 741: What about the type args?

pil-analyzer/src/type_inference.rs
```diff
@@ -729,8 +737,30 @@ impl TypeChecker {
...
+                    unreachable!()
+                };
+                let struct_decl = self.struct_declarations.get(name).unwrap();
+
+                let field_types: HashMap<_, _> = fields
```
chriseth, line 746: Is it helpful to create this hash map?

gzanitti: Helps to avoid borrowing problems with the `&mut self` of `expect_type`

pil-analyzer/src/pil_analyzer.rs
gzanitti: Because definitions no longer have what I need. I have to somehow keep the info inside`FunctionValueDefinition::StructDeclaration`, because they are the ones that contain the type of the Struct fields. 
At this point, definitions has only Expressions.

pil-analyzer/src/type_inference.rs
chriseth: Hm, I don't get it. `type_of_field` returns no references. Where do you borrow `self`?

chriseth: Ah, `struct_decl`?

chriseth: Ok!

pil-analyzer/src/type_inference.rs
```diff
@@ -729,8 +737,30 @@ impl TypeChecker {
...
+                    .iter()
+                    .map(|named_expr| {
+                        let scheme = struct_decl.type_of_field(&named_expr.name).unwrap();
+                        (named_expr.name.clone(), scheme.ty)
+                    })
+                    .collect();
+
+                for named_expr in fields.iter_mut() {
+                    if let Some(field_type) = field_types.get(&named_expr.name) {
```
chriseth, line 755: You can unwrap here, we already checked that.

pil-analyzer/src/type_inference.rs
```diff
@@ -738,11 +734,33 @@ impl TypeChecker {
                 result?
             }
             Expression::StructExpression(sr, StructExpression { name, fields }) => {
-                let Reference::Poly(PolynomialReference { name, .. }) = name else {
+                let Reference::Poly(PolynomialReference { name, type_args }) = name else {
```
chriseth, line 737: Please add a test like this:
```
let x: int = 1;
let y = x{a: 2};
```

pil-analyzer/src/type_inference.rs
```diff
@@ -738,11 +734,33 @@ impl TypeChecker {
...
+                let (ty, args) = self.unifier.instantiate_scheme(scheme);
+                if let Some(requested_type_args) = type_args {
+                    if requested_type_args.len() != args.len() {
+                        return Err(sr.with_error(format!(
+                            "Expected {} type arguments for struct {name}, but got {}: {}",
+                            args.len(),
+                            requested_type_args.len(),
+                            requested_type_args.iter().join(", ")
+                        )));
+                    }
```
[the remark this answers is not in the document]
chriseth: like `instantiate_type_scheme_with_args` or something.

pil-analyzer/tests/types.rs
```diff
@@ -744,3 +744,13 @@ fn simple_struct() {
...
 }
+
+#[test]
+fn simple_struct_type_arg() {
+    let input = "
+    struct Dot<T> { x: int, y: T }
+    let f: int -> Dot = |i| Dot{x: 0, y: i};
+    let x = f(0);
+    ";
+    type_check(input, &[("x", "", "Dot")]);
```
chriseth, line 755: Please also add a test that shows that the generic type for the struct is properly inferred.

pil-analyzer/src/type_inference.rs
```diff
@@ -739,46 +723,86 @@ impl TypeChecker {
...
-                            requested_type_args.len(),
-                            requested_type_args.iter().join(", ")
-                        )));
-                    }
-                    for (requested, inferred) in requested_type_args.iter_mut().zip(&args) {
-                        requested.substitute_type_vars(&self.declared_type_vars);
+                    for (fresh_var, requested) in
+                        fresh_type_vars.iter().zip(requested_type_args.iter())
+                    {
                         self.unifier
```
chriseth, line 736: Why didn't you call the new helper function?

pil-analyzer/tests/types.rs
```diff
@@ -749,8 +749,38 @@ fn simple_struct() {
...
+    let x = f(n);
+    
+    let m: bool = false;
+    let y = Dot{x: 0, y: m};
+    ";
+    type_check(input, &[("x", "", "Dot<int>"), ("y", "", "Dot<bool>")]);
+}
+
+#[test]
+#[should_panic = "Cannot unify types Dot<int> and Dot"]
```
chriseth, line 764: I don't think this is the right error. `Dot` is generic with no constraints on its type parameter. Why should it not unify? Or does the type system see `Dot` as a different type than `Dot<T>`? That should also not happen. It looks like a bug where we do not detect that `Dot` needs a type argument. Is that the same with enums?

gzanitti: Yes, it is a bug, the test is simply to catch it while I was doing some explorations. Sorry, I should have put this PR back as a `Draft`.

The main problem is the following:
If I want the type of a struct `Dot<T>`, to be for example, `Dot<int>`, I need to use a `NamedType`. In that case, tests like

``` rust
fn simple_struct_type_arg() {
  let input = "
  struct Dot<T> { x: int, y: T }
  
  let f: int -> Dot<int> = |i| Dot{x: 0, y: i};
  let n = 1;
  let x = f(n);
  
  let m: bool = false;
  let y = Dot{x: 0, y: m};
  ”;
  type_check(input, &[((“x”, “”, “Dot<int>”), (“y”, “”, “Dot<bool>”)]);
}
```

work correctly. But then, this will fail:

``` rust
fn simple_struct() {
    let input = "
    struct Dot { x: int, y: int }
    let f: int -> Dot = |i| Dot{x: 0, y: i};
    let x = f(0);
    ";
    type_check(input, &[("x", "", "Dot")]);
}
```

becasuse the `Dot` in `int -> Dot` seems to be a `TypeVar` so the unifier will complain: `Cannot unify types Dot<> and Dot`. 

In the case of enums, both examples work correctly, but it is true that 

``` rust
#[test]
fn simple_enum_test() {
    let input = ”
    enum Option<T> {
        Some(T),
        None
    }
    let f: int -> Option = |i| Option::Some(i);
    let x = f(0);
    ”;
    type_check(input, &[((“x”, “”, “Option<int>”)]));
}
```
will fail in the same way (`Cannot unify types Option<>...`) if the function is `int -> Option` instead of `int -> Option<int>`.

So, we could emulate the behavior that enums have by returning a `TypeVar` or `NamedType` during type inference as needed (depending on whether there are type_args in the struct or not), but that might be more of a workaround than a solution. Also, if we want something like 
``` rust 
let f: int -> Option = |i| Option::Some(i);
```
to be valid (and typed as `Option<int>`), I think we should change how the signature is created.

what do you think?

gzanitti: I have been doing some research on this issue. I honestly don't see it being problematic to force the user to do this (and for my taste it's less error prone) and I also adapted type inference so that the Struct cases accommodate these variants (type_args -> NamedType, type_args.empty() -> TypeVar). 

But if you consider it necessary, we can discuss it and I can go deeper into the task in a new clean PR (since enums already have the same problem). 

Otherwise, the PR can be reviewed.

pil-analyzer/src/type_inference.rs
gzanitti: The function made it harder to instantiate free type_vars every time, so I reverted the code to its original place and removed the helper function.

pil-analyzer/tests/types.rs
chriseth: Regardless of what the cause is, I think we should either require full type arguments for function signatures or deduce the type arguments. But I don't think we want the current behaviour. Or what did you want to test in this test? If you agree, then please remove the test.

pil-analyzer/tests/types.rs
```diff
@@ -762,6 +762,66 @@ fn prover_functions() {
...
+    
+    let f: int -> Dot = |i| Dot{x: 0, y: i};
+    let n = 1;
+    let x = f(n);
+    ";
+    type_check(input, &[]);
+}
+
+#[test]
+#[should_panic = "Expected symbol of kind Struct but got Value: x"]
```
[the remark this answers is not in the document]
chriseth: Oh and another one: `enum A { X } let y = A::X{a: 2}`.

chriseth, line 816: And please also add such tests (symbol lookup) starting from asm.

pil-analyzer/src/type_inference.rs
chriseth: The code here is way too long with nothing "interesting" going on, please try to combine it with other code.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -297,6 +297,21 @@ impl PILAnalyzer {
...
 
+        let struct_declarations: HashMap<String, StructDeclaration> = self
+            .definitions
+            .iter()
+            .filter_map(|(name, (_, value))| {
+                if let Some(FunctionValueDefinition::TypeDeclaration(TypeDeclaration::Struct(
+                    struct_decl,
+                ))) = value
+                {
+                    Some((name.clone(), struct_decl.clone()))
```
chriseth, line 308: Is it possible to avoid this clone? We are iterating oven the definitions below, is it possible to collect references to the struct declarations there?

chriseth: Wait, I think this is not the right way at all.

pil-analyzer/tests/types.rs
chriseth: Can you create an issue where this appears for the enum, provide some examples etc? I think I already have an idea where the bug is in the code.

pil-analyzer/src/type_inference.rs
```diff
@@ -220,15 +260,41 @@ impl TypeChecker {
...
+                        DeclaredType {
+                            source,
+                            vars: builtin.vars.clone(),
+                            ty: TypeDeclaredType::Type(builtin.ty.clone()),
+                        }
                     }
                     // Store an (uninstantiated) type scheme for symbols with a declared polymorphic type.
-                    (None, Some(type_scheme)) => type_scheme.clone(),
+                    (None, Some(type_scheme)) => {
+                        let ty = match self.struct_declarations.get(name) {
```
gzanitti, line 271: I'm not 100% sure about this, and my plan was to get rid of structs_declarations. But, since I don't have the structs expressions, I need some way to match them.

pil-analyzer/src/type_inference.rs
```diff
@@ -60,28 +62,66 @@ impl From<Type> for ExpectedType {
...
+    fn type_mut(&mut self) -> &mut Type {
+        match &mut self.ty {
+            TypeDeclaredType::Struct(ty, _) => ty,
+            TypeDeclaredType::Type(ty) => ty,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+enum TypeDeclaredType {
```
chriseth, line 95: Hm, not the nicest name, but also don't know a better one :)

pil-analyzer/src/type_inference.rs
```diff
@@ -60,28 +62,66 @@ impl From<Type> for ExpectedType {
...
+}
+
+impl DeclaredType {
+    fn scheme(&self) -> TypeScheme {
+        match &self.ty {
+            TypeDeclaredType::Struct(ty, _) => TypeScheme {
+                vars: self.vars.clone(),
+                ty: ty.clone(),
+            },
+            TypeDeclaredType::Type(ty) => TypeScheme {
```
chriseth, line 79:
```suggestion
            TypeDeclaredType::Struct(ty, _) | TypeDeclaredType::Type(ty) => TypeScheme {
```

pil-analyzer/src/type_inference.rs
chriseth: Shouldn't the struct declaration be inside `value`?

pil-analyzer/src/type_inference.rs
gzanitti: mmm, what about `DeclaredTypeKind`?

pil-analyzer/src/type_inference.rs
chriseth: Ah no it's not. But then we just have to pull the DeclaredType up into pil_analyzer and construct it there.

leonardoalt:
Closing for now.

PR #1919, opened 2024-10-17
Split building test binary from other build checks.

chriseth:


.github/workflows/pr-tests.yml
```diff
@@ -55,19 +91,8 @@ jobs:
...
-    - name: Create tests archive
-      run: cargo nextest archive --archive-file tests.tar.zst --cargo-profile pr-tests --workspace --all-features
-    - name: Upload build artifacts
-      uses: actions/upload-artifact@v4
-      with:
-        name: tests_archive
-        path: |
-          tests.tar.zst
-          pil-stark-prover-deps.tar.zst
+      run: cargo check --all-targets --all --all-features --profile pr-tests
```
lvella, line 94: This looks redundant with:
```
cargo clippy --all --all-targets --all-features --profile pr-tests -- -D warnings
```

.github/workflows/pr-tests.yml
```diff
@@ -22,6 +22,42 @@ jobs:
...
+      run: tar --zstd -cf pil-stark-prover-deps.tar.zst target/pr-tests/build/pil-stark-prover-*/out
+    - name: Upload build artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: tests_archive
+        path: |
+          tests.tar.zst
+          pil-stark-prover-deps.tar.zst
+
+  build_check:
```
lvella, line 58: Maybe this should just be called "check".

lvella (changes requested):
What I dislike about this is that there is a lot of redundant work done between the two jobs, as `cargo clippy` run is almost entirely included in `cargo build`.

Maybe the `--all-features` clippy run should go on the build job...

lvella:
Looking at the execution times, even if there is redundant work, the split seems to pay off.

chriseth:
I think the execution time does not include the waiting time, so maybe this does not pay off after all if we just have one build runner.

PR #1920, opened 2024-10-17
Fix type arg formatting for trait impl.

chriseth:
If we re-format `impl X<(int -> int)> { ... }`, the parentheses are removed without this change and it does not parse.

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -446,7 +446,7 @@ namespace N(2);
     }"#;
 
         let expected = r#"
-    impl<T> Iterator<ArrayIterator<T>, T> {
+    impl<T> Iterator<(ArrayIterator<T>), T> {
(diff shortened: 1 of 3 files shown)


parser/src/lib.rs
```diff
@@ -462,7 +462,7 @@ namespace N(2);
     }"#;
 
         let expected = r#"
-    impl<A, B> Iterator<ArrayIterator<A>, B> {
+    impl<A, B> Iterator<(ArrayIterator<A>), B> {
```
gzanitti, line 465: Are these brackets necessary? The documentation of 
`format_type_args` says that they should only go in the last case if they are necessary (although that's not what the code actually does).

chriseth: Hm, interesting! I got a parse error, but only for a single parameter, let me check.

PR #1921, opened 2024-10-17
riscv bb executor + continuations

pacheco:
Just for running tests. PR will be broken down


riscv/src/continuations/bootloader.rs
```diff
@@ -31,6 +32,199 @@ pub const MEMORY_HASH_START_INDEX: usize = 2 * REGISTER_NAMES.len();
 pub const NUM_PAGES_INDEX: usize = MEMORY_HASH_START_INDEX + WORDS_PER_HASH * 2;
 pub const PAGE_INPUTS_OFFSET: usize = NUM_PAGES_INDEX + 1;
 
+/// This trait provides all the field specific types and implementations that the bootloader needs.
+///
+/// For now, this trait is implemented directly by each `FieldElement` type, because the hash functions (i.e., poseidon) are field-specific.
+pub trait BootloaderImpl {
+    type Fe: FieldElement;
+    const FE_PER_WORD: usize;
```
leonardoalt, line 40:
```suggestion
    const FE_PER_RISCV32_WORD: usize;
```
just to make it more obvious?

riscv/src/continuations/bootloader.rs
```diff
@@ -31,6 +32,199 @@ pub const MEMORY_HASH_START_INDEX: usize = 2 * REGISTER_NAMES.len();
...
+        self.0[root_hash_start..root_hash_start + WORDS_PER_HASH * B::FE_PER_WORD]
+            .copy_from_slice(hash);
+    }
+
+    pub fn pc(&mut self) -> B::Fe {
+        // the PC is a powdr asm register, which is a single fe, but for the
+        // bootloader inputs, we store it as a "word", the same representation
+        // as the other registers in memory, which may be more than one fe
+        // (e.g., 2 for small fields such as BabyBear).
+        // Here we return the composed fe value.
```
leonardoalt, line 182: I think we could keep the pc as a single FE because in asm we do that right now.

riscv/src/small_field/bootloader.rs
```diff
@@ -1 +1,792 @@
...
+
+impl BootloaderImpl for BabyBearField {
+    type Fe = BabyBearField;
+    const FE_PER_WORD: usize = 2;
+    type Page = [Self::Fe; 2 * WORDS_PER_PAGE];
+    type Hash = [Self::Fe; 8];
+
+    fn update_page(page: &mut Self::Page, idx: usize, word: u32) {
+        let (hi, lo) = bb_split_word(word);
+        // TODO: check proper endianess here!
```
leonardoalt, line 31: The actual memory is little endian

leonardoalt: I think

leonardoalt (commented):
Some small comments here and there, was mainly checking the new bootloader abstraction as requested by @pacheco

riscv/src/continuations/bootloader.rs
pacheco: I think its better to keep everything as 2 fe in the bootloader itself, because it all becomes "bootloader inputs", which are two columns in the small field machine.

riscv/src/continuations/bootloader.rs
pacheco: i think just WORD is fine, we already use "WORD" with this same meaning in the existing bootloader constants (e.g., WORDS_PER_PAGE). I'll add a small comment

leonardoalt:
Closing for now.

PR #1925, opened 2024-10-18
Make plonky3 no-std

Schaeff:
uni-stark was no-std, but we removed this in #1857.
This change moves the parts which require std (threads, io) to the `backend` crate and makes the `plonky3` crate no-std.
Also, it simplified the interface of the verifier to only requiring the necessary data (the circuit for each table).

PR #1926, opened 2024-10-18
Poseidon BB machine.

lvella:
Working with small memory machine.


lvella:
This time targeting main.

test_data/std/poseidon_gl_memory_test.asm
```diff
@@ -137,7 +137,7 @@ machine Main with degree: 65536 {
         mstore_le 180, 0, 0;
         mstore_le 188, 0, 0;
  
-        // This will read bytes [100, 191] and write the result to bytes [104, 131]
+        // This will read bytes [100, 195] and write the result to bytes [104, 131]
```
lvella, line 140: I think this line is yours, @leonardoalt . Is it correct? Should be left here?

leonardoalt: it came from your original poseidon BB PR I think?

lvella: I remember now. I fixed the value when porting the test to poseidon-bb. I re-ran the numbers and it is correct.

leonardoalt:
Linter is failing. I guess that's why @Schaeff added https://github.com/powdr-labs/powdr/pull/1925/files#diff-e9cc0c11ed0e05388d68419f06b32c0bb9844f29dcde2fa503fd79ace4e32c49R2 here.

lvella:
how this got into main?

leonardoalt:
> how this got into main?

also no idea

std/machines/hash/poseidon_bb.asm
```diff
@@ -0,0 +1,254 @@
...
+
+    // Increment the low limb if address is not being set:
+    (1 - is_addr_set') * (
+        // If low limb is about to overflow, next value must be 0:
+        low_overflow * low_addr' +
+        // Otherwise, next value is current plus 4:
+        (1 - low_overflow) * (low_addr + 4 - low_addr')
+    ) = 0;
+    // Set high limb, incremented if low overflowed:
+    (1 - is_addr_set') * (high_addr' - high_addr - low_overflow) = 0;
```
leonardoalt, line 120: Does this block assume addresses to be 4-aligned?

leonardoalt:
Lint still failing. There are a few more that the other PR also fixes.

std/machines/hash/poseidon_bb.asm
lvella: Yes. Starting from this line:
```
    let low_diff = low_addr - (0x10000 - 4);
```

 My hope is that address alignment is enforced somewhere else (in memory machine?). I know it is enforced for poseidon_gl when the call comes from RISC-V.

leonardoalt: yea I think it makes sense

leonardoalt (approved):

PR #1928, opened 2024-10-18
Proof of plonky3 verification

Schaeff:
This is an attempt at running our plonky3 verifier inside powdr

So far:
- move `WitgenCallback` and `VariableDegreeColumn` out of the executor which uses console which doesn't play well with our target
- force indexmap to use std: is this required?


leonardoalt:
why do we need witgen changes for this?

PR #1929, opened 2024-10-21
Add KoalaBear.

chriseth:


leonardoalt (approved):
Could also add M31 (either here or next PR) since we also have that in the CLI

PR #1935, opened 2024-10-22
RISCV bootloader types

pacheco:
Refactor to support multiple implementations of the bootloader and executor.
It introduces a `BootloaderImpl`, with types/functions that must be given by the concrete bootloader implementation (large/small).
It also creates a `BootloaderInputs` type to encapsulate all the indexing logic for manipulating bootloader inputs.


pacheco:
I'm open for suggestions on the design here.
My idea was to extract the "generic" parts into the trait, so as to avoid code duplication, since the code in these parts is already quite complex/fragile.

riscv-executor/src/large_field/mod.rs
```diff
@@ -0,0 +1,1363 @@
...
+//! help with making partition decisions.
+//!
+//! WARNING: the general witness generation/execution code over the polynomial
+//! constraints try to ensure the determinism of the instructions. If we bypass
+//! much of witness generation using the present module, we lose the
+//! non-determinism verification.
+//!
+//! TODO: perform determinism verification for each instruction independently
+//! from execution.
+
```
pacheco, line 11: Most of the code here didn't change, was just moved to the `large_field` module.
There are some small changes I'll point out.

riscv-executor/src/large_field/mod.rs
```diff
@@ -0,0 +1,1363 @@
...
+}
+
+#[derive(Default)]
+pub struct RegisterMemory<F: FieldElement> {
+    last: HashMap<u32, Elem<F>>,
+    second_last: HashMap<u32, Elem<F>>,
+}
+
+impl<F: FieldElement> RegisterMemory<F> {
+    pub fn for_bootloader(&self) -> HashMap<u32, Vec<F>> {
```
pacheco, line 165: The return type here was `HashMap<u32, F>`. It became `Vec<F>` because the `small_field` machine needs two `F` for a register value.
Ideally, this would be `u32` as is the `MemoryState`, but the goldilocks machine actually uses the full `F` during its execution (e.g., for `to_signed`).

riscv-executor/src/lib.rs
```diff
@@ -42,125 +30,14 @@ use crate::profiler::Profiler;
...
-    }
+pub enum ExecMode {
+    Fast,
+    Trace,
 }
 
 pub type MemoryState = HashMap<u32, u32>;
-pub type RegisterMemoryState<F> = HashMap<u32, F>;
+/// Value of registers is Vec<F> to unify the output for different field sizes
+pub type RegisterMemoryState<F> = HashMap<u32, Vec<F>>;
```
pacheco, line 40: see comment on `for_bootloader` function in the `large_field` executor code

riscv/src/continuations/bootloader.rs
```diff
@@ -32,6 +37,225 @@ pub const MEMORY_HASH_START_INDEX: usize = 2 * REGISTER_NAMES.len();
 pub const NUM_PAGES_INDEX: usize = MEMORY_HASH_START_INDEX + WORDS_PER_HASH * 2;
 pub const PAGE_INPUTS_OFFSET: usize = NUM_PAGES_INDEX + 1;
 
+/// This trait provides all the field specific types and implementations that the bootloader needs.
+///
+/// For now, this trait is implemented directly by each `FieldElement` type, because the hash functions (i.e., poseidon) are field-specific.
+pub(crate) trait BootloaderImpl<F> {
```
pacheco, line 43: main reason for this trait here are the `Page` and `Hash` types, which are just arrays of `F` but with different sizes for the small/large machines.
These are used in the merkle tree and bootloader code, and this is the way I found to make it generic.

riscv/src/large_field/code_gen.rs
```diff
@@ -197,7 +197,7 @@ machine Main with min_degree: {}, max_degree: {} {{
 
 {}
 
-let initial_memory: (fe, fe)[] = [
```
pacheco, line 200: this needs to be `(int, int)` because we cant represent `u32` values using `fe` in small field

riscv/src/large_field/code_gen.rs
```diff
@@ -197,7 +197,7 @@ machine Main with min_degree: {}, max_degree: {} {{
...
+let initial_memory: (int, int)[] = [
```
pacheco, line 200: this needs to be `(int, int)` because we cant represent `u32` values using `fe` in small field

pacheco:
@lvella

riscv/src/continuations/bootloader.rs
```diff
@@ -32,6 +37,225 @@ pub const MEMORY_HASH_START_INDEX: usize = 2 * REGISTER_NAMES.len();
...
+        BootloaderInputs {
+            inputs,
+            phantom: PhantomData,
+        }
+    }
+
+    /// The bootloader input that is equivalent to not using a bootloader, i.e.:
+    /// - No pages are initialized
+    /// - All registers are set to 0 (including the PC, which causes the bootloader to do nothing)
+    /// - The state at the end of the execution is the same as the beginning
```
leonardoalt, line 103: why is the state the same?

riscv/src/continuations/bootloader.rs
```diff
@@ -32,6 +37,225 @@ pub const MEMORY_HASH_START_INDEX: usize = 2 * REGISTER_NAMES.len();
...
+    pub fn default_for(accessed_pages: &[u64]) -> BootloaderInputs<F, B> {
+        // Set all registers and the number of pages to zero
+        let register_values = default_register_values::<_, B>();
+        let merkle_tree = MerkleTree::<_, B>::new();
+
+        // TODO: We don't have a way to know the memory state *after* the execution.
+        // For now, we'll just claim that the memory doesn't change.
+        // This is fine for now, because the bootloader does not yet enforce that the memory
+        // state is actually as claimed. In the future, the `accessed_pages` argument won't be
+        // supported anymore (it's anyway only used by the benchmark).
```
leonardoalt, line 113: I don't really understand this comment block.

riscv/src/continuations/bootloader.rs
```diff
@@ -32,6 +37,225 @@ pub const MEMORY_HASH_START_INDEX: usize = 2 * REGISTER_NAMES.len();
...
+        self.inputs[root_hash_start..root_hash_start + WORDS_PER_HASH * B::FE_PER_WORD]
+            .copy_from_slice(hash);
+    }
+
+    pub fn pc(&mut self) -> F {
+        // the PC is a powdr asm register, which is a single fe, but for the
+        // bootloader inputs, we store it as a "word", the same representation
+        // as the other registers in memory, which may be more than one fe
+        // (e.g., 2 for small fields such as BabyBear).
+        // Here we return the composed fe value.
```
leonardoalt, line 190: In both machines we use a single FE for the pc. For BB 24 bits, so it doesn't need 2.

riscv/src/continuations.rs
```diff
@@ -45,6 +47,11 @@ fn transposed_trace<F: FieldElement>(trace: &ExecutionTrace<F>) -> HashMap<Strin
 }
 
 fn render_memory_hash<F: FieldElement>(hash: &[F]) -> String {
+    // TODO: support for small field, which uses two F per word
```
leonardoalt, line 50: here or later?

riscv/src/small_field/bootloader.rs
```diff
@@ -1 +1,100 @@
...
+
+use crate::continuations::bootloader::BootloaderImpl;
+
+use crate::continuations::bootloader::WORDS_PER_PAGE;
+
+pub fn split_word<F: FieldElement>(v: u32) -> (F, F) {
+    ((v & 0xffff).into(), (v >> 16).into())
+}
+
+pub fn split_fe<F: FieldElement>(v: &F) -> [F; 2] {
```
leonardoalt, line 14: So this takes 1 FE and splits into 2 16-bit limbs still in FE?

riscv/src/small_field/bootloader.rs
```diff
@@ -1 +1,100 @@
...
+pub struct SmallFieldBootloader<F: FieldElement>(PhantomData<F>);
+
+impl<F: FieldElement> BootloaderImpl<F> for SmallFieldBootloader<F> {
+    const FE_PER_WORD: usize = 2;
+    type Page = [F; 2 * WORDS_PER_PAGE];
+    type Hash = [F; 8];
+
+    fn update_page(page: &mut Self::Page, idx: usize, word: u32) {
+        let (hi, lo) = split_word(word);
+        // TODO: check proper endianess here!
```
leonardoalt, line 28: I think RISCV memory is [low, high], so this should fail

riscv/src/small_field/bootloader.rs
pacheco: yes: this is quite confusing, but is like that because of the poseidon machine which takes F instead of half-words.
So a `Hash` for the small field a `[F;8]`, the full `F`. But that eventually needs to become [F;16] (i.e., 16 half words), that our machine uses.
This similar for Goldilocks: poseidon takes `[F;4]`, full `F`, but we then need to split into `[F;8]` (i.e., 8 words).

riscv/src/small_field/bootloader.rs
pacheco: yes, riscv is little endian, but i was not sure here (we use hi,lo in our interface to the memory machine), so I left the todo

riscv/src/continuations/bootloader.rs
pacheco: this comment was there already, I just kept the functionality

riscv/src/continuations/bootloader.rs
pacheco: i don't know, did not dig into this yet, just kept the bootloader behaviour as it was, but adapting for the small_field memory layout

leonardoalt: so these comments were there before already?

riscv/src/continuations/bootloader.rs
leonardoalt: ah ok

riscv/src/continuations/bootloader.rs
leonardoalt: ok

riscv/src/continuations.rs
leonardoalt: is this already written in the other branch? or should we make an issue so we don't forget?

riscv/src/small_field/bootloader.rs
leonardoalt: yea I see

riscv/src/small_field/bootloader.rs
leonardoalt: actually yea, it's tricky. We use big-endian in the machine, but Rust uses little endian when representing u64 as 2 words in RISCV32, for example. So this is probably correct actually

leonardoalt:
lgtm, but I'll let @lvella take a look too

riscv/src/continuations.rs
pacheco: the assertion is enough: you'll get a panic if you try to run with small field

riscv-executor/src/lib.rs
```diff
@@ -1497,8 +164,40 @@ pub fn execute<F: FieldElement>(
...
+    program: &AnalysisASMFile,
+    initial_memory: MemoryState,
+    inputs: &Callback<F>,
+    bootloader_inputs: &[F],
+    max_steps_to_execute: usize,
+    mode: ExecMode,
+    profiling: Option<ProfilerOptions>,
+) -> (ExecutionTrace<F>, MemoryState, RegisterMemoryState<F>) {
+    match F::known_field() {
+        Some(KnownField::BabyBearField | KnownField::Mersenne31Field) => small_field::execute_ast(
```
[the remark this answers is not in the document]
pacheco: it's already there as `known_field().unwrap().field_size()`, this PR predates that so I forgot to update here, will change

riscv-executor/src/large_field/mod.rs
pacheco: I couldn't use the `BootloaderImpl` trait because it makes the crate dependencies circular, but maybe a `const` parameter might be possible... ill try

riscv-executor/src/lib.rs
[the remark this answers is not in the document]
lvella: But the problem is the `known_field()`. Field size should be a property of the `FieldElement` trait itself. But maybe this is out of scope for this PR.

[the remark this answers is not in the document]
leonardoalt: yea there are other places that could benefit from that, but we could do it after

leonardoalt:
Closing for now.

PR #1937, opened 2024-10-22
Add4 Machine

onurinanc:
Add4 machine which can be used for sha256bb machine


Schaeff:
It doesn't seem like we need this right now. We should close and revive if we need it in the future, if we end up delegating this to a submachine instead of inlining it inside the caller.

PR #1940, opened 2024-10-23
AnalysisDriver: Result<..>

gzanitti:
Change the AnalysisDriver to return Result<...>. Also, add some source references.


pil-analyzer/src/expression_processor.rs
```diff
@@ -41,78 +41,71 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
...
                 ArrayLiteral {
-                    items: self.process_expressions(items),
+                    items: self.process_expressions(items)?,
                 },
-            ),
+            )),
             PExpression::LambdaExpression(src, lambda_expression) => {
-                Expression::LambdaExpression(src, self.process_lambda_expression(lambda_expression))
+                Ok(Expression::LambdaExpression(
+                    src.clone(),
```
chriseth, line 106: Do you need to clone here?

pil-analyzer/src/expression_processor.rs
```diff
@@ -121,54 +114,60 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
...
             PExpression::MatchExpression(src, MatchExpression { scrutinee, arms }) => {
-                Expression::MatchExpression(
+                let scrutinee = self.process_expression(*scrutinee)?;
+                let arms = arms
+                    .into_iter()
+                    .map(|MatchArm { pattern, value }| {
+                        let vars = self.save_local_variables();
+                        let processed_pattern = self.process_pattern(pattern)?;
+                        let processed_value = self.process_expression(value)?;
+                        self.reset_local_variables(vars);
```
chriseth, line 156: I always try to keep the state correct even in case of an error. If we have an error here, the local variables still contain the state they have inside the match arm. What you could do is somethign like
```
let processed_pattern = self.process_pattern(pattern);
let processed_value = ...;
self.reset_local_variables(vars);
let processed_pattern = processed_pattern?;
```
(or use `?` when you construct the MatchArm below.

chriseth:
I added some comments, but this PR is again way to large.

gzanitti:
> I added some comments, but this PR is again way to large.

Sorry about that. I will split it.

gzanitti:
I have simplified the PR a little bit. Basically I removed several places where we trapped the new result and resolved the error (several of them in the expression_processor). Now most of them are unwraps, so the behavior is the same as before. I will update the changes in successive PRs (also I want to start adding SourceRef where we need them).

ast/src/parsed/mod.rs
```diff
@@ -188,6 +188,43 @@ impl PilStatement {
     }
 }
 
+impl SourceReference for PilStatement {
```
chriseth, line 191: Can you use the macro `impl_source_reference` in this module?

pil-analyzer/src/expression_processor.rs
```diff
@@ -196,10 +196,13 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
                     .map(|args| args.into_iter().map(|t| self.process_type(t)).collect());
 
                 Expression::StructExpression(
-                    src,
+                    src.clone(),
```
chriseth, line 199: Why?

pil-analyzer/src/expression_processor.rs
```diff
@@ -196,10 +196,13 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
...
                     StructExpression {
                         name: Reference::Poly(PolynomialReference {
-                            name: self.driver.resolve_ref(&name.path, SymbolCategory::Struct),
+                            name: self
+                                .driver
+                                .resolve_ref(&name.path, SymbolCategory::Struct)
+                                .unwrap(),
```
chriseth, line 205: I'm still quite uneasy about this because it is difficult to search for. Can you maybe use `expect` with something like `"TODO pass this up"`?

pil-analyzer/src/lib.rs
```diff
@@ -29,27 +29,32 @@ pub use pil_analyzer::{analyze_ast, analyze_file, analyze_string};
 
 pub trait AnalysisDriver: Clone + Copy {
     /// Turns a declaration into an absolute name.
-    fn resolve_decl(&self, name: &String) -> String {
+    fn resolve_decl(&self, name: &String) -> Result<String, String> {
         self.resolve_namespaced_decl(&[name])
```
chriseth, line 33: Why not use the `?` operator?

pil-analyzer/src/lib.rs
```diff
@@ -29,27 +29,32 @@ pub use pil_analyzer::{analyze_ast, analyze_file, analyze_string};
...
+    fn resolve_ref(
+        &self,
+        path: &SymbolPath,
+        symbol_category: SymbolCategory,
+    ) -> Result<String, String> {
         let (path, cat) = self
             .try_resolve_ref(path)
-            .unwrap_or_else(|| panic!("{symbol_category} symbol not found: {path}"));
+            .ok_or_else(|| format!("{symbol_category} symbol not found: {path}"))?;
         if !cat.compatible_with_request(symbol_category) {
```
chriseth, line 52: Sorry if this is pedantic, but I would apply two suggestions for readability here: (1) Avoid negation in conditions if possible (turn around the cases instead) and (2) only use explicit `return` if an early return makes sense. So this would become something like
```rust
if cat.compatible... {
  Ok(path)
} else {
  Err(format!(..))
}
```

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -435,16 +435,22 @@ impl PILAnalyzer {
...
                 let names = statement
                     .symbol_definition_names_and_contained()
                     .map(|(name, sub_name, symbol_category)| {
                         (
                             match sub_name {
-                                None => self.driver().resolve_decl(name),
+                                None => self
+                                    .driver()
+                                    .resolve_decl(name)
+                                    .map_err(|e| stmt.source_reference().with_error(e))
```
chriseth, line 447: Is it possible for this to create an error? We are iterating over all defined names in this statement and only call `resolve_decl` to make it absolute - do I understand that correctly? So we can use `unwrap` here.

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -435,16 +435,22 @@ impl PILAnalyzer {
...
+                                    .unwrap(),
                                 Some(sub_name) => self
                                     .driver()
                                     .resolve_namespaced_decl(&[name, sub_name])
+                                    .map_err(|e| stmt.source_reference().with_error(e))
```
chriseth, line 452: Same here.

pil-analyzer/src/lib.rs
```diff
@@ -29,27 +29,32 @@ pub use pil_analyzer::{analyze_ast, analyze_file, analyze_string};
...
+    fn resolve_decl(&self, name: &String) -> Result<String, String> {
         self.resolve_namespaced_decl(&[name])
-            .relative_to(&Default::default())
-            .to_string()
+            .map(|path| path.relative_to(&Default::default()).to_string())
     }
     /// Turns a nested declaration into an absolute name.
-    fn resolve_namespaced_decl(&self, path: &[&String]) -> AbsoluteSymbolPath;
-    fn resolve_value_ref(&self, path: &SymbolPath) -> String {
+    fn resolve_namespaced_decl(&self, path: &[&String]) -> Result<AbsoluteSymbolPath, String>;
```
chriseth, line 37: Actually I think this is not supposed to fail because it does not do a lookup, it just makes the ref absolute.

pil-analyzer/src/pil_analyzer.rs
chriseth: Actually see my new comment on `resolve_decl` - this is not supposed to fail anyway.

ast/src/parsed/mod.rs
```diff
@@ -650,25 +613,42 @@ pub trait SourceReference {
...
                 }
             }
         }
-    }
+    };
 }
 
 impl_source_reference!(
-    Expression,
+    Expression<E>,
```
chriseth, line 651: Why is expression also mentioned below?

ast/src/parsed/mod.rs
```diff
@@ -650,25 +613,42 @@ pub trait SourceReference {
 }
 
 macro_rules! impl_source_reference {
+    // Version for types with generic parameter
+    ($enum:ident<$generic:ident>, $($variant:ident),*) => {
+        impl<$generic> SourceReference for $enum<$generic> {
+            fn source_reference(&self) -> &SourceRef {
```
chriseth, line 619: Maybe extract the lines starting here into a macro `impl_source_reference_inner`?

pil-analyzer/src/expression_processor.rs
```diff
@@ -262,7 +265,8 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
                 }
             }
             Pattern::Enum(source_ref, name, fields) => {
-                self.process_enum_pattern(source_ref, self.driver.resolve_value_ref(&name), fields)
+                let name = self.driver.resolve_value_ref(&name).unwrap();
```
chriseth, line 268: Should also be expect (and some others as well)

ast/src/parsed/mod.rs
gzanitti: The expression below is the Expression case inside the PilStatement enum.

pil-analyzer/src/lib.rs
```diff
@@ -29,27 +29,36 @@ pub use pil_analyzer::{analyze_ast, analyze_file, analyze_string};
 
 pub trait AnalysisDriver: Clone + Copy {
     /// Turns a declaration into an absolute name.
-    fn resolve_decl(&self, name: &String) -> String {
-        self.resolve_namespaced_decl(&[name])
+    fn resolve_decl(&self, name: &String) -> Result<String, String> {
```
chriseth, line 32: I don't think we should allow this to fail.

ast/src/parsed/mod.rs
```diff
@@ -613,25 +613,38 @@ pub trait SourceReference {
...
 
 macro_rules! impl_source_reference {
+    // Version for types with generic parameter
+    ($enum:ident<$generic:ident>, $($variant:ident),*) => {
+        impl<$generic> SourceReference for $enum<$generic> {
+            impl_source_reference_inner!($enum, $($variant),*);
+        }
+    };
+    // Version for types without generic parameter
     ($enum:ident, $($variant:ident),*) => {
```
chriseth, line 623: Maybe not in this PR, but I was not aware that the first item in the list is special. Maybe we should change the syntax to
`impl_source_reference!(PilStatement { Include, Namespace, ... });` (i.e. but the variantes inside `{}`)

chriseth: Another idea could be to think about a derive macro...

chriseth: Yeah, I think this is quite a lot of work.

chriseth (approved):

PR #1941, opened 2024-10-23
Rust constr enum

Schaeff:
`Identity` is currently a struct which acts as an union: all kinds of identities are represented using the same struct fields, with runtime checks that the right fields are used.

In the context of #1934 where we add new kinds of identities, it seems advantageous to move to an enum on the Rust side as well.

Todo:
- [x] reimplement ids
- [x] minimize code duplication between lookups and permutations, if possible
- [x] remove set_id
- [x] find better way to compare identities
- [x] fix connect


ast/src/analyzed/mod.rs
```diff
@@ -715,96 +711,296 @@ impl<Expr> Default for SelectedExpressions<Expr> {
...
+    }
+}
+
+impl<T> From<ConnectIdentity<T>> for Identity<T> {
+    fn from(value: ConnectIdentity<T>) -> Self {
+        Self::Connect(value)
+    }
+}
+
+// TODO This is the only version of Identity left.
```
Schaeff, line 926: @chriseth can I remove this comment?

executor/src/witgen/machines/machine_extractor.rs
```diff
@@ -309,6 +336,9 @@ fn all_row_connected_witnesses<T>(
                         witnesses.extend(in_rhs);
                     }
                 }
+                Identity::Permutation(..) | Identity::Connect(..) => {
+                    todo!("same as plookup, avoid repetition?")
+                }
```
Schaeff, line 341: The comment of this function does not mention connect, is this something we use?

executor/src/witgen/mod.rs
```diff
@@ -227,7 +228,13 @@ impl<'a, 'b, T: FieldElement> WitnessGenerator<'a, 'b, T> {
             // as they are assumed to be handled in stage 0.
             let polynomial_identities = identities
```
Schaeff, line 229: I'm curious what happens here if `identities` still contains anything else than polynomial identities? Due to the filtering they would get ignored? Or do we not have any non-polynomial identities, in which case this can be a hard cast rather than a filter?

ast/src/analyzed/mod.rs
chriseth: The idea behind this comment was to refactor Identity so that it is only generic over the field, and not about what a selected expressions or an expression / algebraic expression is. If you did this change, then you can remove it.

Schaeff: Oh I didnt get that but yes that's exactly what I did :)

ast/src/analyzed/mod.rs
```diff
@@ -715,96 +710,240 @@ impl<Expr> Default for SelectedExpressions<Expr> {
...
+    pub fn id(&self) -> u64 {
+        match self {
+            Identity::Polynomial(i) => i.id,
+            Identity::Lookup(i) => i.id,
+            Identity::Permutation(i) => i.id,
+            Identity::Connect(i) => i.id,
+        }
+    }
+
+    pub fn source(&self) -> &SourceRef {
```
chriseth, line 901: There is a trait called SourceReference, I think we should implement that instead.

executor/src/witgen/machines/mod.rs
```diff
@@ -151,6 +150,66 @@ impl<'a, T: FieldElement> Machine<'a, T> for KnownMachine<'a, T> {
...
+    }
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+enum ConnectionType {
+    Lookup,
+    Permutation,
+}
+
+impl<'a, T> ConnectingIdentityRef<'a, T> {
```
chriseth, line 175: Is there any way we could re-write the code that uses these Refs so that they are not needed?

pilopt/src/lib.rs
```diff
@@ -552,29 +561,151 @@ fn remove_trivial_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
     pil_file.remove_identities(&to_remove);
 }
 
 fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
+    // TODO: there must be a less verbose way to do this...
+    /// Wrapper around `Identity` that implements `PartialEq` and `Ord` for canonical comparison, ignoring source information and id.
+    struct CanonicalIdentity<'a, T>(&'a Identity<T>);
+
+    impl<T: PartialEq> PartialEq for CanonicalIdentity<'_, T> {
+        fn eq(&self, other: &Self) -> bool {
```
chriseth, line 581: At least this one can be implemented from PartialCmp, right?

ast/src/analyzed/mod.rs
Schaeff: Done

executor/src/witgen/machines/mod.rs
Schaeff: Hmm.. So assuming we only rely on the overlap of the two (left and right selected expressions) we could just pass that, but I don't really find that cleaner and we lose the ability to do things specifically for each case. 

Also this assumption doesn't seem to hold since we have code like this for `WriteOnceMemory`

```rust
if !parts.connecting_identities.values().all(|i| i.is_lookup()) {
    return None;
}
```

Or this for `BlockMachine`
```
    // Connecting identities should either all be permutations or all lookups.
    let connection_type = connecting_identities
        .values()
        .map(|id| id.kind())
        .unique()
        .exactly_one()
        .ok()?;
```

If your point is that we could refactor these, maybe? I think the current way turns what was a runtime invariant into a compile-time one, with minimal changes to the rest of the code.

pilopt/src/lib.rs
Schaeff: Excellent point

Schaeff: Done using Ord

Schaeff:
Reverting to draft as I realised I broke `connect`

pilopt/src/lib.rs
```diff
@@ -552,29 +561,92 @@ fn remove_trivial_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+    }
+
     // Set of (left, right) tuples.
     let mut identity_expressions = BTreeSet::new();
     let to_remove = pil_file
         .identities
         .iter()
         .enumerate()
         .filter_map(|(index, identity)| {
-            match identity_expressions.insert((&identity.left, &identity.right)) {
```
Schaeff, line 577: By the way, this seems like a bug, as these would be considered duplicates:
```
[a] in [b]
[a] connect [b]
```

executor/src/witgen/machines/mod.rs
chriseth: I think we are not talking about the same thing. My question was if we can avoid adding a `..Ref` type for every type.

chriseth: Ah sorry, I thought this was the code used in path canonicalizer which is also sprinkled with ...Refs.

executor/src/witgen/mod.rs
chriseth: I think the filtering here could be improved in general. I don't see polynomial constraints between lower-stage witness column being filtered out, for example. I also don't see a reason why this should only contain polynomial constraints at this stage, e.g. we shouldn't filter out the lookups. Of course, usually when we use a multi-stage approach the idea is to remove the lookups, but conceptually, we could still have a lookup at a higher stage (although the backend probably would not support it).

pilopt/src/lib.rs
```diff
@@ -552,29 +561,92 @@ fn remove_trivial_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+            discriminant(self)
+                .cmp(&discriminant(other))
+                .then_with(|| match (self.0, other.0) {
+                    (
+                        Identity::Polynomial(PolynomialIdentity { expression: a, .. }),
+                        Identity::Polynomial(PolynomialIdentity { expression: b, .. }),
+                    ) => a.cmp(b),
+                    (
+                        Identity::Lookup(LookupIdentity {
+                            left: a, right: b, ..
```
chriseth, line 599: You could use a generic `left` and `right` function for any non-polynomial identity.

executor/src/witgen/machines/mod.rs
```diff
@@ -151,6 +150,66 @@ impl<'a, T: FieldElement> Machine<'a, T> for KnownMachine<'a, T> {
...
 }
 
+#[derive(Copy, Clone)]
+/// A reference to a connecting identity, which is either a permutation or a lookup.
+pub enum ConnectingIdentityRef<'a, T> {
+    Lookup(&'a LookupIdentity<T>),
+    Permutation(&'a PermutationIdentity<T>),
+}
+
+impl<'a, T: Display> Display for ConnectingIdentityRef<'a, T> {
```
chriseth, line 160: Could use `derive_more` here.

executor/src/witgen/machines/mod.rs
chriseth: Hm, I still don't particularly like the introduction of this type. Could this also be done using a trait? Then `left` and `right` could be function that are generally useful...

executor/src/witgen/machines/machine_extractor.rs
```diff
@@ -78,11 +87,18 @@ pub fn split_out_machines<'a, T: FieldElement>(
             base_identities.iter().cloned().partition(|i| {
                 // The identity's left side has at least one machine witness, but
                 // all referenced witnesses are machine witnesses.
-                // Note that expressions of "simple" polynomial identities are stored
-                // in i.left.selector, so those are covered as well.
                 // For lookups, any lookup calling from the current machine belongs
                 // to the machine; lookups to the machine do not.
-                let all_refs = &refs_in_selected_expressions(&i.left) & (&all_witnesses);
+                let all_refs = match i {
+                    Identity::Polynomial(identity) => {
```
chriseth, line 93: Maybe extract that into a function?

executor/src/witgen/machines/machine_extractor.rs
```diff
@@ -92,15 +108,25 @@ pub fn split_out_machines<'a, T: FieldElement>(
...
             .cloned()
-            .filter(|i| {
-                refs_in_selected_expressions(&i.right)
-                    .intersection(&machine_witnesses)
-                    .next()
-                    .is_some()
+            .filter_map(|i| {
+                let id = i.id();
+                match i {
+                    Identity::Polynomial(_) => None,
```
chriseth, line 114: Maybe extract that into a function?

executor/src/witgen/machines/machine_extractor.rs
Schaeff: done, with all_children

executor/src/witgen/machines/mod.rs
Schaeff: done

executor/src/witgen/machines/mod.rs
Schaeff: I don't really like it either to be honest.

What about

```rust
struct ConnectingIdentity<'a, T> {
   left: &'a SelectedExpressions<T>,
   right: &'a SelectedExpressions<T>,
   kind: ConnectingIdentityKind
}
```

Where `kind` is `Lookup` or `Permutation` and will not need to be extended for `PhantomLookup` and `PhantomPermutation`?

Then we can remove some duplicated code and still rely on `kind` for the cases I pointed out above?
I don't really see how traits help here, since at the end of the day we need to be able to store either of these two possibilities anywhere, so we'd have to use `dyn ConnectingIdentity` which seems unnatural here. But maybe I'm not zooming out enough here.

executor/src/witgen/machines/machine_extractor.rs
Schaeff: simplified due to refactor of `ConnectingIdentity`

executor/src/witgen/machines/mod.rs
Schaeff: I tried the above and I like it more

pilopt/src/lib.rs
Schaeff: I thought about it but given that these are in different variants, I don't see how we could use it:
now:
```rust
Identity::Lookup(LookupIdentity { left, right, .. }) | Identity::Permutation(PermutationIdentity { left, right, ..}) => {
   left
}
```

With a trait, we can't treat both variants at the same time because we can only bind to variables of the same type in patterns:
```rust
// does NOT work
Identity::Lookup(i) | Identity::Permutation(i) => {
   i.left()
```

For this to be possible, we need to use the same struct inside the variant:

```
Identity::Lookup(ConnectingIdentity<Lookup>),
Identity::Permutation(ConnectingIdentity<Permutation>),
```

... but this doesn't work either because the generic types also need to match!

So we go one step further and use the same variant

```
Identity::Connection(ConnectingIdentity)
```

And store the kind inside the `ConnectingIdentity`.

This is nice but my issue is that at this point we drifted away completely from the PIL enum.

My conclusion is that relying on patterns to work across connecting identities, while verbose, is fine as it preserves the structure of the PIL enum. If we refactor the PIL enum to be nested, then the nested approach would work in Rust as well. I don't think this refactor is necessarily beneficial (because of its nested nature), nor a priority.

executor/src/witgen/machines/machine_extractor.rs
Schaeff: That was wrong so done, just extracting as suggested

Schaeff:
This is ready for another review @chriseth

executor/src/witgen/machines/machine_extractor.rs
chriseth: Connect is very sporadically implemented.

executor/src/witgen/machines/mod.rs
chriseth: Cool! Don't we need the ID? Oh, it's in a hash map from id anyway, right?

ast/src/analyzed/mod.rs
```diff
@@ -715,96 +710,231 @@ impl<Expr> Default for SelectedExpressions<Expr> {
...
+    fn children_mut(&mut self) -> Box<dyn Iterator<Item = &mut AlgebraicExpression<T>> + '_> {
+        Box::new(iter::once(&mut self.expression))
+    }
+    fn children(&self) -> Box<dyn Iterator<Item = &AlgebraicExpression<T>> + '_> {
+        Box::new(iter::once(&self.expression))
+    }
+}
+
+impl<T> PolynomialIdentity<T> {
+    pub fn new(id: u64, source: SourceRef, expression: AlgebraicExpression<T>) -> Self {
```
chriseth, line 742: Do we acutally need this when the fields are all `pub`?

ast/src/analyzed/mod.rs
chriseth: Can you remove the TODO?

executor/src/witgen/machines/mod.rs
```diff
@@ -151,6 +150,67 @@ impl<'a, T: FieldElement> Machine<'a, T> for KnownMachine<'a, T> {
     }
 }
 
+#[derive(Clone, Copy)]
+pub struct ConnectingIdentity<'a, T> {
```
chriseth, line 154: Maybe one line of documentation?

ast/src/analyzed/mod.rs
Schaeff: No, personal preference but happy to remove these!

executor/src/witgen/machines/mod.rs
Schaeff: Yeah I put the id for now but it did seem redundant, let's see.

Schaeff: It's gone!

ast/src/analyzed/mod.rs
Schaeff: done

executor/src/witgen/machines/mod.rs
Schaeff: done

ast/src/analyzed/mod.rs
Schaeff: removed

executor/src/witgen/machines/mod.rs
chriseth: Hm, not sure I would agree to this. A polynomial identity can also cross machine boundaries (depending on the definition, it would be the same machine, but still...). Shouldn't we be more explicit about the kind? Or the fact that it has two parts that do not need to match on the same row or something like that if you want to be abstract?

chriseth: I would just be explicit and say, lookup or permutation.

chriseth (approved):
I still think the doc is confusing (people will wonder if it is related to Identity or not), but ok.

PR #1942, opened 2024-10-24
Stwo support "addition and equal test"

ShuangWu121:
## Initial Support for `stwo` Backend

This PR introduces the initial support for the `stwo` backend in the Powdr pipeline.

### Summary
- **Constraint Proving**: Added functionality to prove and verify constraints between columns.
- **Test Added**: A test case has been added to validate the above functionality.

### Remaining Tasks
- Implement the `setup` function.
- Use `logup` for the next reference (now the next reference creates a new column).
- Check the constant columns and public values in stwo and implement to Powdr.

### How to Test
To test the `stwo` backend, use the following command:

```bash
cargo test --features stwo --package powdr-pipeline --test pil -- stwo_add_and_equal --exact --show-output

### CODE REVIEW

File: backend/src/stwo/circuit_builder.rs [hunk 1]
```diff
@@ -0,0 +1,269 @@
...
+                eval.add_constraint(expr);
+            });
+        }
+        eval
+    }
+}
+
+fn to_stwo_expression<T: FieldElement, E: EvalAtRow>(
+    expr: &AlgebraicExpression<T>,
+    eval: &mut E,
```
Schaeff, line 196: I think this cannot take a mutable reference: if we call this function twice on witness x, we want to receive the same stwo witness. Now it would create a new witness each time. My intuition is that we should first allocate all witnesses into a vector by calling next_interaction as many times as needed. Then in this function when we see a reference, we get the associated stwo reference. The plonky3 code can be helpful here: the vector of plonky3 witnesses is returned by the builder. In stwo, we have to build it ourselves using E.

backend/src/stwo/prover.rs
```diff
@@ -18,15 +40,102 @@ impl<F: FieldElement> StwoProver<F> {
...
+        //Trace
+        let circuit = PowdrCircuit::new(self.analyzed.clone())
+            .with_witgen_callback(witgen_callback.clone())
+            .with_witness(witness)
+            .generate_stwo_circuit_trace();
+       
+
+        let circuitEval = PowdrCircuit::new(self.analyzed.clone())
+            .with_witgen_callback(witgen_callback.clone())
+            .with_witness(witness);
```
Schaeff, line 69: Ideally we'd create the circuit a single time, and then do everything based on that one instantiation. These two are the same except the first one stores the precomputed stwo trace. There is another instantiation further down which gets consumed into the trace. All of these should rely on the same instance, with methods returning stuff rather than mutating the circuit.

So:

```
let circuit = PowdrCircuit::new(self.analyzed).with_witgen_callback(...).....;
let trace = circuit.generate_stwo_trace(); // this takes a reference to the circuit and returns the trace
```

backend/src/stwo/prover.rs
```diff
@@ -18,15 +40,102 @@ impl<F: FieldElement> StwoProver<F> {
...
+
+        //     println!("proving time for fibo length of {:?} is {:?}",fibonacci_y_length, duration);
+        //     println!("proof size is {:?} bytes",proof.size_estimate());
+
+        //     let verifystart = Instant::now();
+        stwo_prover::core::prover::verify(&[&component], verifier_channel, commitment_scheme, proof).unwrap();
+        
+        //     println!("verify time is {:?} ",verifyduration);
+
+        println!("prove_stwo in prover.rs is not complete yet");
```
Schaeff, line 139: return the proof here

pipeline/src/test_util.rs
```diff
@@ -516,3 +518,19 @@ pub fn run_reparse_test_with_blacklist(file: &str, blacklist: &[&str]) {
         .compute_optimized_pil()
         .unwrap();
 }
+
+// TODO: Add #[cfg(feature = "stwo")] to conditionally compile this function. now when I add it, this code is disabled. how to eable by default?
```
Schaeff, line 522: Yes we should have this. Check the other cases above for example for plonky3, it has a version with and without the feature, where the one without the feature is a no-op.
What do you mean by the code being disabled? If it shows as disabled in your IDE it just means that the feature flag is not active, which would make sense if stwo is off by default.

pipeline/src/test_util.rs
```diff
@@ -516,3 +518,19 @@ pub fn run_reparse_test_with_blacklist(file: &str, blacklist: &[&str]) {
...
+    let backend = powdr_backend::BackendType::Stwo;
+    let mut pipeline = Pipeline::default()
+        .with_tmp_output()
+        .from_file(resolve_test_file(file_name))
+        .with_prover_inputs(inputs.clone())
+        .with_backend(backend, None)
+        .set_witness(vec![("add::a".to_string(),
+        vec![Mersenne31Field::from(1),Mersenne31Field::from(2),Mersenne31Field::from(1),Mersenne31Field::from(2)]),("add::b".to_string(), 
+        vec![Mersenne31Field::from(1),Mersenne31Field::from(2),Mersenne31Field::from(1),Mersenne31Field::from(2)]),("add::c".to_string(), 
+        vec![Mersenne31Field::from(2),Mersenne31Field::from(4),Mersenne31Field::from(2),Mersenne31Field::from(4)])]);
```
Schaeff, line 533: Do we have to build the witness ourselves here? `Pipeline` has a `compute_witness` method which does it automatically for any program, as opposed to the custom approach here.

Schaeff (commented):
Review of the code architecture mostly.
Suggestion for next steps:
- clean up using `cargo fmt` and `cargo clippy --fix`
- go through function signatures and fix them. It's fine to use `unimplemented` internally. For example, the prover should return a proof.

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+        
+        let witness_columns:BTreeMap<PolyID, usize> = analyzed
+            .definitions_in_source_order(PolynomialType::Committed)
+            .flat_map(|(symbol, _)| symbol.array_elements())
+            .enumerate()
+            .map(|(index, (_, id))| (id, index))
+            .collect();
+
+        Self {
+            log_n_rows: analyzed.degree().ilog2(),
```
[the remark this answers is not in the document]
ShuangWu121: no log_n_rows anymore :)

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+                .iter()
+                .map(|(name, values)| {
+                    let values = values
+                        .iter()
+                        .map(|v| {
+                            let ptr = v as *const T as *const u32;
+
+                            let value = unsafe {
+                                *ptr // Dereference the pointer to get the u32 value
+                            };
```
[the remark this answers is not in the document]
ShuangWu121: use v.try_into_i32() now, nice tip!

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+                            value.into()
+                        })
+                        .collect();
+                    (name.clone(), values)
+                })
+                .collect(),
+        );
+        Self {
+            elements: element,
```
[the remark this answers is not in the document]
ShuangWu121: no element in struct anymore :)

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+            ..self
+        }
+    }
+
+    pub fn gen_trace(
+        self,
+    ) -> ColumnVec<CircleEvaluation<SimdBackend, BaseField, BitReversedOrder>> {
```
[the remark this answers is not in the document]
ShuangWu121: now only a function "into_stwo_circuit_trace()" is used.

backend/src/stwo/prover.rs
```diff
@@ -18,15 +40,102 @@ impl<F: FieldElement> StwoProver<F> {
...
     ) -> Result<Self, io::Error> {
         Ok(Self {
-            _analyzed,
+            analyzed,
             _fixed,
             _proving_key: None,
             _verifying_key: None,
         })
     }
+    pub fn prove(&self, witness: &[(String, Vec<F>)], witgen_callback: WitgenCallback<F>) {
```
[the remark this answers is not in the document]
ShuangWu121: function signature is fixed, a verify function is added as well

backend/src/stwo/prover.rs
ShuangWu121: circuit is generated only once now.

backend/src/stwo/prover.rs
```diff
@@ -18,15 +35,102 @@ impl<F: FieldElement> StwoProver<F> {
...
+        let twiddles = SimdBackend::precompute_twiddles(
+            CanonicCoset::new(
+                (self.analyzed.degree() as u32) + 1 + config.fri_config.log_blowup_factor,
+            )
+            .circle_domain()
+            .half_coset,
+        );
+
+        // Setup protocol.
+        let prover_channel = &mut Poseidon252Channel::default();
```
Schaeff, line 69: Let's move Poseidon252Channel to `config.rs` as well, see plonky3. Using type declarations so we can swap it easily in the future.

backend/src/stwo/prover.rs
```diff
@@ -18,15 +35,102 @@ impl<F: FieldElement> StwoProver<F> {
...
+        let commitment_scheme =
```
Schaeff, line 70: Same for `CommitmentSchemeProver`, to `config.rs`

backend/src/stwo/prover.rs
```diff
@@ -18,15 +35,102 @@ impl<F: FieldElement> StwoProver<F> {
...
+        tree_builder.extend_evals(trace);
+        tree_builder.commit(prover_channel);
+
+        let component = PowdrComponent::new(
+            &mut TraceLocationAllocator::default(),
+            PowdrEval::new(
+                self.analyzed.clone(),
+                self.analyzed.commitment_count()
+                    + self.analyzed.constant_count()
+                    + self.analyzed.publics_count(),
```
Schaeff, line 90: We do not support publics not fixed columns, so this should be only witness columns and we should fail before this if we have any fixed columns or publics

backend/src/stwo/prover.rs
```diff
@@ -18,15 +35,102 @@ impl<F: FieldElement> StwoProver<F> {
...
+            ),
+        );
+
+        //let start = Instant::now();
+        let proof = stwo_prover::core::prover::prove::<SimdBackend, Poseidon252MerkleChannel>(
```
Schaeff, line 95: Same, to `config.rs`, we can define our own `Proof` type there so that we don't need to pass generics here.

backend/src/stwo/prover.rs
```diff
@@ -18,15 +35,102 @@ impl<F: FieldElement> StwoProver<F> {
...
+    }
+
+    pub fn verify(&self, proof: &[u8], _instances: &[F]) -> Result<(), String> {
+        let proof: StarkProof<<Poseidon252MerkleChannel as MerkleChannel>::H> =
+            bincode::deserialize(proof).map_err(|e| format!("Failed to deserialize proof: {e}"))?;
+
+        let config = PcsConfig {
+            pow_bits: 16,                          // Any value you want to set for pow_bits
+            fri_config: FriConfig::new(0, 1, 100), // Using different numbers for FriConfig
+        };
```
Schaeff, line 112: `get_config`, cf above

backend/src/stwo/prover.rs
```diff
@@ -18,15 +35,102 @@ impl<F: FieldElement> StwoProver<F> {
...
+            &mut CommitmentSchemeVerifier::<Poseidon252MerkleChannel>::new(config);
+
+        //Constraints that are to be proved
+        let component = PowdrComponent::new(
+            &mut TraceLocationAllocator::default(),
+            PowdrEval::new(
+                self.analyzed.clone(),
+                self.analyzed.commitment_count()
+                    + self.analyzed.constant_count()
+                    + self.analyzed.publics_count(),
```
Schaeff, line 125: cf above

pipeline/src/test_util.rs
```diff
@@ -532,3 +532,55 @@ pub fn run_reparse_test_with_blacklist(file: &str, blacklist: &[&str]) {
...
+                vec![
+                    Mersenne31Field::from(2),
+                    Mersenne31Field::from(4),
+                    Mersenne31Field::from(2),
+                    Mersenne31Field::from(4),
+                ],
+            ),
+        ]);
+    // Generate a proof
+    let proof = pipeline.compute_proof().cloned().unwrap();
```
Schaeff, line 577:
```suggestion
    let proof = Pipeline::default()
        .with_tmp_output()
        .from_file(resolve_test_file(file_name))
        .with_prover_inputs(inputs.clone())
        .with_backend(backend, None)
        .compute_proof().cloned().unwrap();
```

this might work, the witness gets computed if needed

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+                AlgebraicBinaryOperator::Sub => {
+                    lhe - rhe
+                }
+                AlgebraicBinaryOperator::Mul => {
+                    lhe * rhe
+                }
+                AlgebraicBinaryOperator::Pow => {
+                    let AlgebraicExpression::Number(e) = powdr_rhe.as_ref() else {
+                        panic!("Expected number in exponent.")
+                    };
```
Schaeff, line 178: I think plonky3 supports more complex expressions here, we should check there.

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+                    let e: u32 = e
+                        .to_arbitrary_integer()
+                        .try_into()
+                        .unwrap_or_else(|_| panic!("Exponent has to fit 32 bits."));
+                    if e == 0 {
+                        //Expression::Constant(F::from(1))
+                        println!("This is the power");
+                        unimplemented!()
+                    } else {
+                        (0..e).fold(lhe.clone(), |acc, _| acc * lhe.clone())
```
Schaeff: Same here, this is implemented in p3 iirc

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,236 @@
...
+                    }
+                }
+            }
+        }
+        AlgebraicExpression::Challenge(challenge) => {
+            unimplemented!()
```
Schaeff, line 193: Let's have a more descriptive error message here

backend/src/stwo/mod.rs
```diff
@@ -55,6 +57,11 @@ impl<T: FieldElement> Backend<T> for StwoProver<T> {
         if prev_proof.is_some() {
             return Err(Error::NoAggregationAvailable);
         }
+        let circuit = PowdrCircuit::new(self.analyzed.clone())
+        .with_witgen_callback(witgen_callback.clone())
+        .with_witness(witness);
+
+        self.prove(witness, witgen_callback);
         unimplemented!()
```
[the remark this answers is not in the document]
Schaeff: unimplemented can be removed now that prove returns a proof

backend/src/stwo/prover.rs
```diff
@@ -18,15 +35,102 @@ impl<F: FieldElement> StwoProver<F> {
...
+        let prover_channel = &mut Poseidon252Channel::default();
+        let commitment_scheme =
+            &mut CommitmentSchemeProver::<SimdBackend, Poseidon252MerkleChannel>::new(
+                config, &twiddles,
+            );
+
+        let trace = PowdrCircuit::new(self.analyzed.clone())
+            .with_witgen_callback(witgen_callback)
+            .with_witness(witness)
+            .into_stwo_circuit_trace();
```
Schaeff, line 78: Is this the only place where we instantiate PowdrCircuit? It gets consumed by the end of this statement, so if it's the case maybe we do not need a struct and rather this can all be one function. We can discuss this.

pipeline/src/test_util.rs
ShuangWu121: it works, but it always set witness to be all 0, would this be enough for test case?

backend/src/stwo/circuit_builder.rs
```diff
@@ -122,36 +86,56 @@ impl<'a, T: FieldElement> PowdrCircuit<'a, T> {
...
 
-impl<'a, T: FieldElement> FrameworkEval for PowdrCircuit<'a, T> {
+pub struct PowdrEval<T> {
+    analyzed: Arc<Analyzed<T>>,
+    col_count: usize,
+    witness_columns: BTreeMap<PolyID, usize>,
+}
+
+impl<T: FieldElement> PowdrEval<T> {
+    pub fn new(analyzed: Arc<Analyzed<T>>, col_count: usize) -> Self {
```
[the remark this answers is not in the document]
ShuangWu121: col_count is removed.
we don't have col_count function, I need to get the total number of columns, now I compute col_count as 
```
  analyzed.commitment_count()
                    + self.analyzed.constant_count()
                    + self.analyzed.publics_count()
```
is this a correct count for total columns?

backend/src/stwo/circuit_builder.rs
```diff
@@ -122,36 +86,56 @@ impl<'a, T: FieldElement> PowdrCircuit<'a, T> {
...
-        println!("This is the witness eval {:?}", self.witness);
 
         // Add polynomial identities
-        let identities = self
+        let polynomial_identities: Vec<PolynomialIdentity<_>> = self
             .analyzed
             .identities_with_inlined_intermediate_polynomials()
             .into_iter()
-            .filter(|id| id.kind == IdentityKind::Polynomial)
+            .filter_map(|id| id.try_into().ok())
```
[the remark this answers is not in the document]
ShuangWu121: placeholder added for other type of identities, with unimplemented!()

pipeline/src/test_util.rs
Schaeff: Right, so this is due to the strategy witness generation uses. It tries to find a unique witness on each row by solving the constraints. However, in the case where it cannot find a unique solution, it makes a last attempt setting everything to zero. If the constraints are satisfied, it outputs this witness.

This program does not have a unique witness, so it applies this strategy and sets everything to zero.

I think that's fine for this test, we'll have less trivial examples once we support fixed columns. Also we already have an `add` test in `pipeline/tests/pil` so we should add stwo there and we can remove this new test.

backend/src/stwo/prover.rs
ShuangWu121: prove function of StwoProver is generic for "backend, Merkelchannel, and channel" now.  

in the create() function of StwoProverFactory in mod.rs, StwoProver is initialised as 
```
stwo: Box<StwoProver<F, SimdBackend, Blake2sMerkleChannel, Blake2sChannel>> =
            Box::new(StwoProver::new(pil, fixed)?); 
```
generic type can be specified here.  
Fri parameters are set in prover.rs, store in StwoProver struct so it can be accessed by prove and verify functions. 

I will chek if make a separate struct StwoProverConfig for choosing hash and choosing FRI parameters would be better

pipeline/src/test_util.rs
ShuangWu121: I see. make sense. witness generation is automatic now.

backend/src/stwo/circuit_builder.rs
ShuangWu121: now the to_stwo_expression() function follows the  implementation in plonky3

backend/src/stwo/circuit_builder.rs
```diff
@@ -79,6 +88,11 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
         self.analyzed.degree().ilog2() + 1
     }
     fn evaluate<E: EvalAtRow>(&self, mut eval: E) -> E {
+        assert!(
+            self.analyzed.constant_count() == 0 && self.analyzed.publics_count() == 0,
+            "Error: Expected constant_count and publics_count to be 0, as there are not supported yet.",
```
Schaeff, line 93:
```suggestion
            "Error: Expected no fixed columns nor public inputs, as they are not supported yet.",
```

backend/src/stwo/circuit_builder.rs
```diff
@@ -95,21 +109,34 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
...
+        });
 
-        for id in self.analyzed.identities_with_inlined_intermediate_polynomials() {
+        for id in self
+            .analyzed
+            .identities_with_inlined_intermediate_polynomials()
+        {
             match id {
                 // Already handled above
                 Identity::Polynomial(..) => {}
```
Schaeff, line 124: It would be clearer to treat the polynomial identities here, instead of extracting them before and ignoring them here.

backend/src/stwo/circuit_builder.rs
```diff
@@ -122,53 +149,67 @@ fn to_stwo_expression<T: FieldElement, E: EvalAtRow>(
...
+                        let index = witness_columns[&poly_id];
+                        witness_eval[index][0]
+                    }
+                    true => {
+                        let index = witness_columns[&poly_id];
+                        witness_eval[index][1]
+                    }
+                },
+                PolynomialType::Constant => {
+                    unimplemented!("Constant polynomials are not supported in this stwo yet")
```
Schaeff, line 169:
```suggestion
                    unimplemented!("Constant polynomials are not supported in stwo yet")
```

backend/src/stwo/circuit_builder.rs
```diff
@@ -122,53 +149,67 @@ fn to_stwo_expression<T: FieldElement, E: EvalAtRow>(
...
                 }
-                true => {
-                    let index = witness_columns[&poly_id];
-                    witness_eval[index][1].into()
+                PolynomialType::Intermediate => {
+                    unimplemented!("Intermediate polynomials are not supported in this stwo yet")
```
Schaeff, line 172:
```suggestion
                    unimplemented!("Intermediate polynomials are not supported in stwo yet")
```

backend/src/stwo/circuit_builder.rs
```diff
@@ -122,53 +149,67 @@ fn to_stwo_expression<T: FieldElement, E: EvalAtRow>(
...
                 }
             }
         }
+        AlgebraicExpression::PublicReference(..) => {
+            unimplemented!("Public references are not supported in this stwo yet")
```
Schaeff, line 177:
```suggestion
            unimplemented!("Public references are not supported in stwo yet")
```

backend/src/stwo/circuit_builder.rs
```diff
@@ -122,53 +149,67 @@ fn to_stwo_expression<T: FieldElement, E: EvalAtRow>(
     witness_eval: &Vec<[<E as EvalAtRow>::F; 2]>,
     _eval: &E,
```
Schaeff, line 150: If this is unused it should be removed

backend/src/stwo/circuit_builder.rs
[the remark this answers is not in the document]
Schaeff: now we only support witness columns so only the commitment_count

backend/src/stwo/prover.rs
Schaeff: I see, you implemented the fully generic way already, that's fine I guess!

backend/src/stwo/prover.rs
```diff
@@ -1,32 +1,144 @@
...
+    _channel_marker: PhantomData<C>,
+    _channel_marker_backend: PhantomData<B>,
+    _channel_marker_merkelchannel: PhantomData<MC>,
 }
 
-impl<F: FieldElement> StwoProver<F> {
+impl<'a, F: FieldElement, B: Backend + Send, MC: MerkleChannel + Send, C: Channel + Send>
+    StwoProver<F, B, MC, C>
+where
+    B: Backend + BackendForChannel<MC>, // Ensure B implements BackendForChannel<MC>
```
Schaeff, line 50: Here we have bounds both in the generic type declarations and in the where clause. Let's have them all in the where clause.

backend/src/stwo/prover.rs
```diff
@@ -1,32 +1,144 @@
...
+            PowdrEval::new(self.analyzed.clone()),
+        );
+
+        let proof = stwo_prover::core::prover::prove::<B, MC>(
+            &[&component],
+            &mut prover_channel,
+            commitment_scheme,
+        )
+        .unwrap();
+
```
Schaeff, line 115: can we run the verifier here as well? Then we know that any created proof is valid

backend/src/stwo/prover.rs
```diff
@@ -1,32 +1,144 @@
...
+        //  setup: Option<&mut dyn io::Read>,
     ) -> Result<Self, io::Error> {
+        let config = PcsConfig {
+            pow_bits: FRI_PROOF_OF_WORK_BITS as u32,
+            fri_config: FriConfig::new(
+                LOG_LAST_LAYER_DEGREE_BOUND as u32,
+                FRI_LOG_BLOWUP as u32,
+                FRI_NUM_QUERIES,
+            ),
+        };
```
Schaeff, line 70: I'm not convinced this should be stored in `StwoProver`: it is constant and cannot be set in the constructor. Instead we could have a function which returns it.

Schaeff:
```
fn get_config() -> PcsConfig {
    ...
}
```

backend/src/stwo/prover.rs
```diff
@@ -1,32 +1,144 @@
...
             _verifying_key: None,
+            _channel_marker_backend: PhantomData,
+            _channel_marker_merkelchannel: PhantomData,
+            pcs_config: config,
+            _channel_marker: PhantomData,
         })
     }
+    pub fn prove(&self, witness: &[(String, Vec<F>)]) -> Result<Vec<u8>, String> {
+        // twiddles are used for FFT, it is computed in a bigger group than the eval domain.
+        // eval domain half coset G_{2n} + <G_{n/2}>
```
Schaeff, line 85:
```suggestion
        // the eval domain is the half coset G_{2n} + <G_{n/2}>
```
?

backend/src/stwo/prover.rs
```diff
@@ -1,32 +1,144 @@
...
             _proving_key: None,
             _verifying_key: None,
+            _channel_marker_backend: PhantomData,
+            _channel_marker_merkelchannel: PhantomData,
+            pcs_config: config,
+            _channel_marker: PhantomData,
         })
     }
+    pub fn prove(&self, witness: &[(String, Vec<F>)]) -> Result<Vec<u8>, String> {
+        // twiddles are used for FFT, it is computed in a bigger group than the eval domain.
```
Schaeff, line 84:
```suggestion
        // twiddles are used for FFT, they are computed in a bigger group than the eval domain.
```

backend/src/stwo/prover.rs
Schaeff: Also what does `<...>` mean here?

ShuangWu121: short answer: when creating twiddles, all the examples given in stwo, has this +1 check[ here](https://github.com/starkware-libs/stwo/blob/64270dc51420a6887e7c0695c71584949711dcb2/crates/prover/src/examples/wide_fibonacci/mod.rs#L176), some related comments in[ stwo ](https://github.com/starkware-libs/stwo/blob/64270dc51420a6887e7c0695c71584949711dcb2/crates/prover/src/core/poly/twiddles.rs#L4) and [here](https://github.com/starkware-libs/stwo/blob/64270dc51420a6887e7c0695c71584949711dcb2/crates/prover/src/core/circle.rs#L332)

long answer:
<...> means the generator of the subgroup of the circle group, <G_{n}> generates a group of size 2^n.  in coset (eval domain is a coset) context, it is referred as step size, G_{2n} is the offset of the coset. 
say a trace domain of size 4 in normal STARK, the cyclic group can be written as <G_2>, eval domain is a coset, then it can be written as G_3 + <G_2>, or G_4 + <G_2>, as long as the offset is not in <G_2>
In circle STARK, trace domain itself is a coset, for coset of size 4,  it can be written as G_4+<G_2> as well, but it can also be written as G_4+<G_1> U G_4^{-1}+<G_1>, this is the twin coset/ standard position coset, mentioned often in circle STARK tutorials. The half coset is  G_4+<G_1>, its size is 2. 
eval coset will be G_8+<G_2> U G_8^{-1}+<G_2> of size 8
twiddles are used for FFT, and they will be computed on a half coset G_5+<G_3> of size 8, it comes from the coset G_5+<G_3> U G_5^{-1}+<G_3> of size 16.   one log size bigger than eval domain  
I am writing some blogs to give some explanations for this, including FFT in circle domain. TBD...

backend/src/stwo/prover.rs
ShuangWu121: function added

backend/src/stwo/prover.rs
ShuangWu121: StarkProof doesn't have Clone method, when move the proof to verify function, I cannot reuse it, serialised it and return. I tried some method recommended by ChatGPT, using Arc, Box, no luck, then I make a request to stwo repo to ask for Clone method. 
Do you have other alternatives maybe I can try?

backend/src/stwo/prover.rs

backend/src/stwo/circuit_builder.rs
ShuangWu121: Done. Now I know I can simply use "function_name::<T, E>(...)" as explicit type parameter specification, didn't mange to find this solution before, Rust always tell me to give type annotations if I don't pass this unused _eval ^^'

backend/src/stwo/circuit_builder.rs
ShuangWu121: done

backend/src/stwo/circuit_builder.rs
```diff
@@ -1,30 +1,39 @@
...
+pub(crate) fn gen_stwo_circuit_trace<T, B, F>(
+    witness: Option<&[(String, Vec<T>)]>,
     analyzed: Arc<Analyzed<T>>,
-) -> ColumnVec<CircleEvaluation<SimdBackend, BaseField, BitReversedOrder>> {
-    let element: Option<Vec<(String, BaseColumn)>> = Some(
+) -> ColumnVec<CircleEvaluation<B, BaseField, BitReversedOrder>>
+where
+    T: FieldElement,
+    B: FieldOps<M31> + ColumnOps<F>, // Ensure B implements FieldOps for M31
+    F: ExtensionOf<BaseField>,
```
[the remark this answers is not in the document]
ShuangWu121: I added comment and added generalize_factory!(Factory <- RestrictedFactory, [Mersenne31Field]); in mod.rs, would this be enough for the runtime check?

backend/src/stwo/circuit_builder.rs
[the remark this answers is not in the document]
ShuangWu121: constant_count() and publics_count are removed. I added
`assert!(
            self.analyzed.constant_count() == 0 && self.analyzed.publics_count() == 0,
            "Error: Expected no fixed columns nor public inputs, as they are not supported yet.",
        );`
before count cols

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,204 @@
...
+use stwo_prover::core::fields::m31::M31;
+use stwo_prover::core::fields::{ExtensionOf, FieldOps};
+use stwo_prover::core::poly::circle::{CanonicCoset, CircleEvaluation};
+use stwo_prover::core::poly::BitReversedOrder;
+use stwo_prover::core::ColumnVec;
+
+pub type PowdrComponent<'a, F> = FrameworkComponent<PowdrEval<F>>;
+
+pub(crate) fn gen_stwo_circuit_trace<T, B, F>(
+    witness: Option<&[(String, Vec<T>)]>,
```
Schaeff, line 28: This does not seem like it needs to be an option

backend/src/stwo/prover.rs
```diff
@@ -1,32 +1,153 @@
...
+            &[&component],
+            &mut prover_channel,
+            commitment_scheme,
+        )
+        .unwrap();
+
+        Ok(bincode::serialize(&proof).unwrap())
+    }
+
+    pub fn verify(&self, proof: &[u8], _instances: &[F]) -> Result<(), String> {
```
Schaeff, line 116: Let's assert that the instance slice is empty.

backend/src/stwo/prover.rs
Schaeff: Let's not run the verifier for now here then

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,204 @@
...
+                }
+                Identity::Permutation(..) => {
+                    unimplemented!("Permutation is not implemented in stwo yet")
+                }
+                Identity::PhantomPermutation(..) => {
+                    unimplemented!("PhantomPermutation is not implemented in stwo yet")
+                }
+                Identity::PhantomLookup(..) => {
+                    unimplemented!("PhantomLookup is not implemented in stwo yet")
+                }
```
Schaeff, line 129: These two are actually unreachable

Schaeff, line 129: These two should not panic, but rather do nothing, since these identities are only used in witness generation

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,204 @@
...
+pub(crate) fn gen_stwo_circuit_trace<T, B, F>(
+    witness: Option<&[(String, Vec<T>)]>,
+    analyzed: Arc<Analyzed<T>>,
+) -> ColumnVec<CircleEvaluation<B, BaseField, BitReversedOrder>>
+where
+    T: FieldElement, //only Merenne31Field is supported, checked in runtime
+    B: FieldOps<M31> + ColumnOps<F>, // Ensure B implements FieldOps for M31
+    F: ExtensionOf<BaseField>,
+{
+    let element: Option<Vec<(String, Col<B, M31>)>> = Some(
```
Schaeff, line 36: Why do we need to wrap the value in Some here? It seems like we unwrap a few lines below

Schaeff: Also do we need to collect this into a vector of pairs, only to iterate through it and ignore the first element of the pair below?

Schaeff: This compiles:

```rust
pub(crate) fn gen_stwo_circuit_trace<T, B, F>(
    witness: &[(String, Vec<T>)],
    analyzed: Arc<Analyzed<T>>,
) -> ColumnVec<CircleEvaluation<B, BaseField, BitReversedOrder>>
where
    T: FieldElement, //only Merenne31Field is supported, checked in runtime
    B: FieldOps<M31> + ColumnOps<F>, // Ensure B implements FieldOps for M31
    F: ExtensionOf<BaseField>,
{
    let domain = CanonicCoset::new(analyzed.degree().ilog2()).circle_domain();
    witness
        .iter()
        .map(|(name, values)| {
            let values = values
                .iter()
                .map(|v| v.try_into_i32().unwrap().into())
                .collect();
            CircleEvaluation::new(domain, values)
        })
        .collect()
}
```

backend/src/stwo/circuit_builder.rs
Schaeff: Having to use explicit type parameters is often a hint that the generics can be simplified. In this case, the function is generic over `E` but the signature only uses `E::F`. This requires explicit type parameters, because you pass a `F` but there could be many `E` which have this `F` as their `E::F`, and the compiler needs to know which `E` you're talking about. I suspect it could be possible to change function to be generic over `F`, in which case we would not need explicit type parameters anymore.

backend/src/stwo/circuit_builder.rs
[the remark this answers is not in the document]
Schaeff: Yes I think this is fine!

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,204 @@
...
+        assert!(
+            self.analyzed.constant_count() == 0 && self.analyzed.publics_count() == 0,
+            "Error: Expected no fixed columns nor public inputs, as they are not supported yet.",
+        );
+
+        let col_count = self.analyzed.commitment_count();
+        let mut witness_eval = Vec::with_capacity(col_count);
+        for _ in 0..col_count {
+            witness_eval.push(eval.next_interaction_mask(0, [0, 1]));
+        }
```
Schaeff, line 100: This can be simplified by building a map of the values, instead of a vector:

```rust
        let witness_eval = self
            .witness_columns
            .iter()
            .map(|(poly_id, index)| (*poly_id, eval.next_interaction_mask(0, [0, 1])))
            .collect();
```

Then we only need to pass this `(poly_id -> value)` map to `to_stwo_expression`, instead of currently passing the `(poly_id -> index)` and `(index -> value)` maps

ShuangWu121: Nice simplification! Done

backend/src/stwo/circuit_builder.rs
ShuangWu121: remove explicit type for to_stwo_expression as well

backend/src/stwo/prover.rs
ShuangWu121: Done

backend/src/stwo/circuit_builder.rs

backend/src/stwo/prover.rs
ShuangWu121: Done as above : constant_count() and publics_count are removed. I added
assert!( self.analyzed.constant_count() == 0 && self.analyzed.publics_count() == 0, "Error: Expected no fixed columns nor public inputs, as they are not supported yet.", );
before count cols

backend/src/stwo/circuit_builder.rs
ShuangWu121: Done. Nice tip to help me discover potential simplification, thanks!

backend/src/lib.rs
```diff
@@ -107,15 +107,7 @@ impl BackendType {
...
             #[cfg(feature = "stwo")]
-            BackendType::Stwo => Box::new(stwo::StwoProverFactory),
-            #[cfg(not(any(
-                feature = "halo2",
-                feature = "estark-polygon",
-                feature = "estark-starky",
-                feature = "plonky3",
-                feature = "stwo"
-            )))]
-            _ => panic!("Empty backend."),
```
Schaeff, line 118: This should probably stay

backend/src/stwo/prover.rs
```diff
@@ -1,32 +1,159 @@
...
+
+        //Constraints that are to be proved
+        let component = PowdrComponent::new(
+            &mut TraceLocationAllocator::default(),
+            PowdrEval::new(self.analyzed.clone()),
+        );
+
+        // Retrieve the expected column sizes in each commitment interaction, from the AIR.
+        let sizes = component.trace_log_degree_bounds();
+        commitment_scheme.commit(proof.commitments[0], &sizes[0], &mut verifier_channel);
```
Schaeff, line 138: Clarify these 0s

pipeline/src/test_util.rs
```diff
@@ -312,7 +312,7 @@ pub fn test_plonky3_with_backend_variant<T: FieldElement>(
     let mut pipeline = Pipeline::default()
         .with_tmp_output()
         .from_file(resolve_test_file(file_name))
-        .with_prover_inputs(inputs)
+        .with_prover_inputs(inputs.clone())
```
Schaeff, line 315: revert?

pipeline/src/test_util.rs
```diff
@@ -539,3 +539,27 @@ pub fn run_reparse_test_with_blacklist(file: &str, blacklist: &[&str]) {
         .compute_optimized_pil()
         .unwrap();
 }
+
+// TODO: Add #[cfg(feature = "stwo")] to conditionally compile this function. now when I add it, this code is disabled. how to eable by default?
```
Schaeff, line 543: Add noop version when the feature is not active and remove comment

pipeline/tests/pil.rs
```diff
@@ -264,6 +266,12 @@ fn add() {
     );
 }
 
+#[cfg(feature = "stwo")]
```
Schaeff, line 269: remove feature

test_data/pil/add_and_equal.pil
```diff
@@ -0,0 +1,6 @@
+namespace Add(4);
+    let a;
```
Schaeff, line 2: use query functions to set a non-trivial witness

pipeline/tests/pil.rs
ShuangWu121: Done

pipeline/src/test_util.rs
ShuangWu121: Done

backend/src/stwo/prover.rs
ShuangWu121: add assert to make sure the sizes has length 1 for now. 
Later when constant columns get implemented, witness columns and constant columns will be committed separately and sizes will have length 2
On my way to make constant columns in another PR

backend/src/stwo/circuit_builder.rs
```diff
@@ -0,0 +1,192 @@
...
+use stwo_prover::core::fields::{ExtensionOf, FieldOps};
+use stwo_prover::core::poly::circle::{CanonicCoset, CircleEvaluation};
+use stwo_prover::core::poly::BitReversedOrder;
+use stwo_prover::core::ColumnVec;
+
+pub type PowdrComponent<'a, F> = FrameworkComponent<PowdrEval<F>>;
+
+pub(crate) fn gen_stwo_circuit_trace<T, B, F>(
+    witness: &[(String, Vec<T>)],
+    analyzed: Arc<Analyzed<T>>,
```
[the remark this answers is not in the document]
ShuangWu121: analyzed is removed.

backend/src/stwo/prover.rs
```diff
@@ -1,32 +1,159 @@
...
+        //  setup: Option<&mut dyn io::Read>,
     ) -> Result<Self, io::Error> {
+        let config = PcsConfig {
+            pow_bits: FRI_PROOF_OF_WORK_BITS as u32,
+            fri_config: FriConfig::new(
+                LOG_LAST_LAYER_DEGREE_BOUND as u32,
+                FRI_LOG_BLOWUP as u32,
+                FRI_NUM_QUERIES,
+            ),
+        };
```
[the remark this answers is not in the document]

Schaeff (approved):
let's go!

backend/src/lib.rs
leonardoalt: why was it ok to remove this?

ShuangWu121: It seems somehow at least one backend feature will always be enabled (I don't know why, maybe it is related to default features being "default = ["halo2", "plonky3"]" in powdr/Cargo.toml, or defalut=[plonky3] in cli/Cargo.toml?). Then this code become redundant. At one point this code gave me error (I forgot what error it is). Removing it seems fine so it ends up here

leonardoalt: Yes but that only applies if you're compiling the backend crate together with these other crates, if you try to compile only the backend crate it will fail, which I think we should not allow

PR #1946, opened 2024-10-24
regenerate artifacts if needed

leonardoalt:


powdr/src/lib.rs
```diff
@@ -64,25 +65,53 @@ impl Session {
     }
 
     pub fn prove(&mut self) {
-        match self.pipeline.read_constants_mut(Path::new(&self.out_path)) {
-            Ok(_) => (),
-            Err(_) => {
-                self.pipeline.compute_fixed_cols().unwrap();
+        let asm_name = self.pipeline.asm_string().unwrap().0.clone().unwrap();
+        let pil_file = create_pil_file_path(&asm_name);
```
pacheco, line 69: nitpick: i think just `pil_file_path` is clearer

powdr/src/lib.rs
```diff
@@ -64,25 +65,53 @@ impl Session {
...
+
+        let generate_artifacts = if let Some(existing_pil) = read_file_if_exists(&pil_file) {
```
pacheco, line 71: i think the `if_exists` can be removed also, you can't read the file if it doesn't exist :P

powdr/src/lib.rs
```diff
@@ -103,6 +132,23 @@ impl Session {
...
 }
 
+fn create_pil_file_path(asm_name: &Path) -> PathBuf {
+    let file_stem = asm_name.file_stem().unwrap().to_str().unwrap();
+    let opt_file_stem = format!("{file_stem}_opt");
+    asm_name.with_file_name(opt_file_stem).with_extension("pil")
+}
+
+fn read_file_if_exists(path: &Path) -> Option<String> {
+    if path.exists() {
```
pacheco, line 142: the `File::open` will fail exactly if the file doesn't exist, so i'd check its result instead...

powdr/src/lib.rs
```diff
@@ -64,25 +65,53 @@ impl Session {
...
-        if pkey.exists() && vkey.exists() {
-            self.pipeline.set_pkey_file(pkey.clone());
-            self.pipeline.set_vkey_file(vkey.clone());
-            self.pipeline.setup_backend().unwrap();
-        } else {
+        if generate_artifacts {
+            self.pipeline.compute_fixed_cols().unwrap();
             self.pipeline.setup_backend().unwrap();
             self.export_setup();
             self.pipeline.set_pkey_file(pkey.clone());
```
pacheco, line 92: this `set_pkey_file` and `set_vkey_file` are done in all if/else paths, can they be outside?

leonardoalt: The annoying part here is whether they're before or after `self.pipeline.setup_backend()`. If they exist, we want them to be before setup_backend so they're loaded during setup. If they don't, we want to do the setup before (which writes them to disk) and then set them to pipeline in case it gets cloned etc. I'll check if the latter can be done inside pipeline.

powdr/src/lib.rs
pacheco: actually, there's a function in rust `std::fs::read_to_string` which could be used in place of `read_file_if_exists`

powdr/src/lib.rs
leonardoalt: Yea I think we have to set it after for now, otherwise it's gonna try to get the keys from non-existing or empty files

powdr/src/lib.rs
leonardoalt: replaced this function by `read_to_string`

pacheco (approved):

PR #1948, opened 2024-10-24
Use lower max degree by default and let user change it

leonardoalt:
Depends on https://github.com/powdr-labs/powdr/pull/1946

### DIFF
--- a/powdr/src/lib.rs
+++ b/powdr/src/lib.rs
@@ -29,10 +29,10 @@ const DEFAULT_PKEY: &str = "pkey.bin";
 const DEFAULT_VKEY: &str = "vkey.bin";
 
 // Minimum and maximum log of number of rows for the RISCV machine.
-const DEFAULT_LOG_MIN_DEGREE: u32 = 5;
-const DEFAULT_LOG_MAX_DEGREE: u32 = 20;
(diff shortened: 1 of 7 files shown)


powdr/src/lib.rs
```diff
@@ -29,10 +29,35 @@ pub struct Session {
 const DEFAULT_PKEY: &str = "pkey.bin";
 const DEFAULT_VKEY: &str = "vkey.bin";
 
+// Minimum and maximum log of number of rows for the RISCV machine.
+const DEFAULT_LOG_MIN_DEGREE: u32 = 5;
```
pacheco, line 33: `MIN/MAX_DEGREE_LOG` to be consistent with the cmd line option?

pacheco: same for the related names of args and variables

pacheco:
what's the usage for the user setting the MIN_DEGREE?

leonardoalt:
> what's the usage for the user setting the MIN_DEGREE?

not much, that's why I didn't expose it to the end user

leonardoalt:
updated, waiting for the other PR to be merged

leonardoalt:
this is ready now

pacheco (approved):

PR #1949, opened 2024-10-24
Fix small continuations issue

leonardoalt:
I ran into this unwrapping a None when using continuations with a very low max degree. It's kinda hard to trigger without triggering a bunch of other issues, but I think this fix is enough for now: if a register is never used it is not contained in the snapshot, so it should be 0.

### DIFF
--- a/riscv/src/continuations.rs
+++ b/riscv/src/continuations.rs
@@ -424,7 +424,9 @@ pub fn rust_continuations_dry_run<F: FieldElement>(
             .map(|reg| {
                 let reg = reg.strip_prefix("main.").unwrap();
                 let id = Register::from(reg).addr();
-                *register_memory_snapshot.get(&(id as u32)).unwrap()
+                *register_memory_snapshot
+                    .get(&(id as u32))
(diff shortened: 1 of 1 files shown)


riscv/src/continuations.rs
```diff
@@ -424,7 +424,9 @@ pub fn rust_continuations_dry_run<F: FieldElement>(
             .map(|reg| {
                 let reg = reg.strip_prefix("main.").unwrap();
                 let id = Register::from(reg).addr();
-                *register_memory_snapshot.get(&(id as u32)).unwrap()
+                *register_memory_snapshot
+                    .get(&(id as u32))
+                    .unwrap_or(&0.into())
```
pacheco, line 429: this is fine i think, but i wonder is the executor itself should take care of this (zero out unused registers)

leonardoalt: for that it would have to know which registers we're using

pacheco: mmm right

pacheco (approved):

PR #1950, opened 2024-10-24
Make selector non-optional in analyzed

Schaeff:
We represent lookup and permutation selectors as `Option<AlgebraicExpression<T>>`.
The issue with this is that there are two ways to represent 1: `Some(1)` and `None`.
This leads to issues in witgen where we check `is_none` but not against `Some(1)`.
This already led to an awkward optimizer step which reduces one to the other so that witgen only sees one of the two options.

This PR makes the selector non-optional. Therefore, 1 is only 1. Display is adjusted to not print the selector if it prints to "1". This str-level comparison is there to avoid introducing invasive bounds on T which would contaminate everything.

### DIFF
--- a/ast/src/analyzed/display.rs
+++ b/ast/src/analyzed/display.rs
@@ -307,10 +307,13 @@ impl<T: Display> Display for SelectedExpressions<T> {
         write!(
             f,
             "{}[{}]",
-            self.selector
-                .as_ref()
-                .map(|s| format!("{s} $ "))
-                .unwrap_or_default(),
+            {
+                let s = self.selector.to_string();
+                match s.as_str() {
+                    "1" => "".to_string(),
+                    _ => format!("{s} $ "),
+                }
+            },
             self.expressions.iter().format(", ")
         )
     }
--- a/ast/src/analyzed/display.rs
+++ b/ast/src/analyzed/display.rs
@@ -311,7 +311,7 @@ impl<T: Display> Display for SelectedExpressions<T> {
                 let s = self.selector.to_string();
                 match s.as_str() {
                     "1" => "".to_string(),
-                    _ => format!("{s} $ "),
+                    s => format!("{s} $ "),
                 }
             },
             self.expressions.iter().format(", ")
--- a/ast/src/analyzed/display.rs
+++ b/ast/src/analyzed/display.rs
@@ -308,6 +308,7 @@ impl<T: Display> Display for SelectedExpressions<T> {
             f,
             "{}[{}]",
             {
+                // we only print the selector if it is not 1. The comparison is string-based to avoid introducing invasive type bounds on T.
                 let s = self.selector.to_string();
                 match s.as_str() {
                     "1" => "".to_string(),
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -10,6 +10,7 @@ use std::ops::{self, ControlFlow};
 use std::sync::Arc;
 
 use itertools::Itertools;
+use num_traits::One;
 use powdr_number::{DegreeType, FieldElement};
 use powdr_parser_util::SourceRef;
 use schemars::JsonSchema;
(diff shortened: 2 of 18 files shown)


pilopt/src/lib.rs
```diff
@@ -371,27 +370,6 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
...
-                    left.selector = None;
-                }
-                if right.selector.as_ref() == Some(&one) {
-                    right.selector = None;
-                }
-            }
-            _ => {}
-        }
-    }
-}
```
Schaeff, line 393: This is the awkward optimisation step in question, which this PR gets rid of.

pil-analyzer/src/condenser.rs
```diff
@@ -759,17 +759,17 @@ fn to_selected_exprs<'a, T: Clone + Debug>(
...
 
-fn to_option_expr<T: Clone + Debug>(value: &Value<'_, T>) -> Option<AlgebraicExpression<T>> {
+fn to_option_expr<T: FieldElement>(value: &Value<'_, T>) -> AlgebraicExpression<T> {
     let Value::Enum(enum_value) = value else {
         panic!("Expected option but got {value:?}")
     };
     assert_eq!(enum_value.enum_decl.name, "std::prelude::Option");
     match enum_value.variant {
-        "None" => None,
+        "None" => T::one().into(),
```
Schaeff, line 768: Note to self: this looks wrong

Schaeff: Fixed

ast/src/analyzed/display.rs
```diff
@@ -307,10 +307,13 @@ impl<T: Display> Display for SelectedExpressions<T> {
         write!(
             f,
             "{}[{}]",
-            self.selector
-                .as_ref()
-                .map(|s| format!("{s} $ "))
-                .unwrap_or_default(),
+            {
+                let s = self.selector.to_string();
```
chriseth, line 311: Can you add a short comment why the comparison is string-based?

ast/src/analyzed/mod.rs
```diff
@@ -757,14 +757,14 @@ impl PublicDeclaration {
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct SelectedExpressions<T> {
-    pub selector: Option<AlgebraicExpression<T>>,
+    pub selector: AlgebraicExpression<T>,
     pub expressions: Vec<AlgebraicExpression<T>>,
 }
 
-impl<T> Default for SelectedExpressions<T> {
+impl<T: FieldElement> Default for SelectedExpressions<T> {
```
chriseth, line 764: Is `T: One` sufficient?

ast/src/analyzed/mod.rs
```diff
@@ -969,8 +969,7 @@ impl<T> SelectedExpressions<T> {
     /// @returns true if the expression contains a reference to a next value of a
     /// (witness or fixed) column
     pub fn contains_next_ref(&self) -> bool {
-        self.selector
```
chriseth, line 972: What about `self.children().any(|e| e.contains_next_ref())`?

chriseth: I mean (not as part of this PR), but we could also get rid of the `contains_next_ref` functions above Expression and just work with `all_children()` called on an AST node.

chriseth (approved):
All comments are optional.

ast/src/analyzed/display.rs
Schaeff: done

ast/src/analyzed/mod.rs
Schaeff: It is!

ast/src/analyzed/mod.rs
Schaeff: Changed to use children

PR #1952, opened 2024-10-25
Better prints to the end user

leonardoalt:
Since RUST_LOG=info is quite verbose, I think we should have a few messages that are always shown to the user.
The main point is to give some results about what they asked (run/proof), but also let them know when a heavier process is going on. I ran this locally and felt the messages and flow were ok.

Depends on https://github.com/powdr-labs/powdr/pull/1948

### DIFF
--- a/powdr/src/lib.rs
+++ b/powdr/src/lib.rs
@@ -29,10 +29,10 @@ const DEFAULT_PKEY: &str = "pkey.bin";
 const DEFAULT_VKEY: &str = "vkey.bin";
 
 // Minimum and maximum log of number of rows for the RISCV machine.
-const DEFAULT_LOG_MIN_DEGREE: u32 = 5;
-const DEFAULT_LOG_MAX_DEGREE: u32 = 20;
+const DEFAULT_MIN_DEGREE_LOG: u32 = 5;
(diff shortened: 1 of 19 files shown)


powdr/src/lib.rs
```diff
@@ -112,12 +112,14 @@ impl Session {
         let vkey = Path::new(&self.out_path).join(DEFAULT_VKEY);
 
         if generate_artifacts {
+            println!("Creating program ZK setup. This has to be done only once per program.");
```
lvella, line 115: > "This has to be done only once per program."
This is the kind of stuff that should printed when you run `powdr --help".

But it is good to have it here, too.

powdr/src/lib.rs
```diff
@@ -219,8 +221,8 @@ pub fn run(pipeline: &mut Pipeline<GoldilocksField>) {
     );
 
     let duration = start.elapsed();
-    log::info!("Fast executor took: {:?}", duration);
-    log::info!("Trace length: {}", trace.len);
+    println!("Fast executor took: {duration:?}");
```
lvella, line 224: Debug print `{:?}` isn't really meant for end user printing. But in this case, I don't think there is a better option aside from formatting manually.

lvella (approved):

PR #1956, opened 2024-10-25
Properly re-print trait impls.

chriseth:
Before this PR, the list of trait implementations was not present in Analyzed, but we need them for re-printing the PIL (and other things).

This PR also contains a simplification in that it only keeps a list of all trait impls, not sorted into which trait they implement. This information is only relevant during trait impl solving, where we re-build that data structure. Keeping the pre-sorted layout would make re-printing in source order much more difficult.

And now it also formats paths to the current namespace properly.

### DIFF
--- a/pil-analyzer/src/traits_resolver.rs
+++ b/pil-analyzer/src/traits_resolver.rs
@@ -74,7 +74,7 @@ impl<'a> TraitsResolver<'a> {
         }
         let Some(trait_impls) = self.trait_impls.get(trait_decl_name) else {
             return Err(format!(
-                "Could not find an implementation for the trait function {reference}"
+                "Could not find an implementation for the trait function {reference} (trait is not implemented at all)"
             ));
         };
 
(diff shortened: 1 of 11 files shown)


pil-analyzer/src/traits_resolver.rs
```diff
@@ -52,8 +68,14 @@ impl<'a> TraitsResolver<'a> {
...
             return Ok(());
         };
-        let Some(trait_impls) = self.trait_impls.get(trait_decl_name) else {
+        if !self.traits.contains(trait_decl_name) {
+            // Not a trait function.
             return Ok(());
+        }
+        let Some(trait_impls) = self.trait_impls.get(trait_decl_name) else {
+            return Err(format!(
+                "Could not find an implementation for the trait function {reference}"
```
gzanitti, line 77: maybe

```suggestion
                "Could not find an implementation for the trait declaration {reference}"
```

gzanitti (approved):
LGTM

chriseth:
I'll create a follow-up PR for the error message.

chriseth:
We have a merge conflict, so I'll add the follow-up PR here.

gzanitti (approved):

PR #1957, opened 2024-10-25
change jit logs to debug for now

leonardoalt:


executor/src/constant_evaluator/jit_compiler.rs
```diff
@@ -14,7 +14,8 @@ pub fn generate_values<T: FieldElement>(
 ) -> HashMap<(String, PolyID), VariablySizedColumn<T>> {
     let compiled_pil = match powdr_jit_compiler::compile(analyzed, &symbols_to_compile(analyzed)) {
         Err(err) => {
-            log::error!("Failed to compile some constant columns: {}", err);
+            // TODO this should be changed back to Error after the introduction of the ToCol trait.
+            log::debug!("Failed to compile some constant columns: {}", err);
```
leonardoalt, line 18: this error is leaking to stderr during any run

jit-compiler/src/lib.rs
```diff
@@ -70,7 +71,8 @@ pub fn compile<T: FieldElement>(
 
     if successful_symbols.len() < requested_symbols.len() {
         let successful_hash = successful_symbol_names.iter().collect::<HashSet<_>>();
-        log::info!(
+        // TODO this should be changed back to Info after the introduction of the ToCol trait.
+        log::debug!(
```
leonardoalt, line 75: this one generates quite a large message because of all symbols that didn't work

jit-compiler/src/lib.rs
```diff
@@ -53,7 +53,8 @@ pub fn compile<T: FieldElement>(
     analyzed: &Analyzed<T>,
     requested_symbols: &[&str],
 ) -> Result<CompiledPIL, String> {
-    log::info!("JIT-compiling {} symbols...", requested_symbols.len());
+    // TODO this should be changed back to Info after the introduction of the ToCol trait.
+    log::debug!("JIT-compiling {} symbols...", requested_symbols.len());
```
leonardoalt, line 57: changed this one as well for context

chriseth (approved):

jit-compiler/src/lib.rs
chriseth: This is not related to the ToCol trait.

leonardoalt: Yea but it gets implied from the other ones

PR #1959, opened 2024-10-25
Attempt to make cache more useful for PR builds.

lvella:
The same build steps is now performed in both cache build and PR build.


lvella:
We will only know if it worked after merging.

pacheco (approved):
fingers crossed

PR #1960, opened 2024-10-26
Convert Keccak non memory circuit endianness

qwang98:
Was little endian before, now converted to big endian to match with all other machines.

Simple PR.


leonardoalt:
since we're changing this, can you also:
- move the std machine to the `small_field` directory and change the `mod.asm` files accordingly
- rename the test file
- probably change the filename in pipeline `powdr_std` tests

leonardoalt:
This still needs changes

qwang98:
This can be merged now. @leonardoalt

leonardoalt (approved):

PR #1961, opened 2024-10-27
split simd features in backend

leonardoalt:
Unfortunately, cargo weirdness attacks again. Even with `estark-starky-simd = ["starky?/avx512"]` enabling the feature only if `starky` is being used, cargo still selects `starky` for compilation.

Then, because `starky` itself doesn't check the architecture being used at compile time to enable AVX and only checks the features, compilation fails for https://github.com/powdr-labs/powdr-template with `simd` if AVX is not available.

Therefore, we can't group both p3 and starky's simd features in the same backend `simd` feature, and need one simd feature per prover. This is what this PR does. The result has been tested with the template repo and starky is correctly *not* selected for compilation both with and without the `simd` feature of the template crate.

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -27,8 +27,7 @@ stwo = ["dep:stwo-prover"]
 
 # Enable AVX or Neon accordingly in backends that support them.
 plonky3-simd = ["powdr-plonky3?/nightly-features"]
-estark-starky-simd = ["starky/avx512"]
-simd = ["plonky3-simd", "estark-starky-simd"]
+estark-starky-simd = ["starky?/avx512"]
 
 [dependencies]
 powdr-ast.workspace = true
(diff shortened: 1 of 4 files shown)


Schaeff:
LGTM!

Related non blocking questions:
- could `powdr_cli` be built on top of `powdr` instead of them being "siblings" now?
- would removing the simd flags and always turning simd on an option? What's the use case for non-simd?

Schaeff (approved):

Schaeff (changes requested):
I noticed the book mentions the `simd` feature.

leonardoalt:
> could powdr_cli be built on top of powdr instead of them being "siblings" now?

yea makes sense!

> would removing the simd flags and always turning simd on an option? What's the use case for non-simd?

Hmm we can check what that entails. Plonky3 checks the architecture anyway so there it'd be safe, but starky wouldn't compile.

leonardoalt:
> I noticed the book mentions the `simd` feature.

Will fix tomorrow/

Schaeff (approved):

PR #1967, opened 2024-10-28
Fix p3 challenge collection

Schaeff:
When building the `ConstraintSystem` struct which caches `Analyzed`, we incorrectly count usage of the same challenge many times. Use a set to deduplicate.

Enable the relevant tests to trigger phase 2 witgen.


georgwiese (approved):
LGTM!

@Schaeff Please review my changes before you merge :)

PR #1969, opened 2024-10-28
cargo powdr

leonardoalt:
cargo CLI tool to generate powdr host/guest templates, install tooling, etc. Currently only does the former.

The template was copied from the latest version of https://github.com/powdr-labs/powdr-template . Should we also copy README?

It can be installed via the cmd below, and it's quite quick:
```console
cargo install --git https://github.com/powdr-labs/powdr --branch cargo-powdr cargo-powdr
```
(without the branch after it's merged)

It can be used via any of these options:
```console
$ cargo-powdr new # creates project 'my-powdr-host' with guest dir 'guest' and guest name 'my-powdr-guest'
$ cargo-powdr new host # creates project 'host' with guest dir 'guest' and guest name 'my-powdr-guest'
$ cargo-powdr new host111 --guest-name guest222 # creates project 'host111' with guest dir 'guest' and guest name 'guest222'
```


pacheco:
I guess if we merge this we can drop the `template` project no?

cargo-powdr/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[command(subcommand)]
+    command: Option<Commands>,
+}
+
+#[derive(Subcommand)]
+enum Commands {
+    /// Create a new powdr project.
+    New {
+        /// The name of the project
+        name: Option<String>,
```
pacheco, line 23: i think the name should be mandatory, as in `cargo powdr new <PROJ>`

cargo-powdr/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        /// The name of the guest
+        #[arg(short, long)]
+        #[arg(default_value_t = String::from("my-powdr-guest"))]
```
pacheco, line 27: maybe just `powdr-guest` as a default? most likely this name wont be changed unless people are doing something more complicated like having more than one guest in the same project?

cargo-powdr/template/guest/src/main.rs
```diff
@@ -0,0 +1,12 @@
+use powdr_riscv_runtime;
+use powdr_riscv_runtime::io::read;
+
+fn main() {
+    // Read some data from channel 1.
```
pacheco, line 5: this could be more specific maybe? could be nice to make it clear that this is a `serde` object

pacheco:
I am getting a couple of errors when doing `cargo install`:
```
error: invalid character `{` in package name: `{{PROJECT_NAME}}`, the first character must be a Unicode XID start character (most letters or `_`)
 --> ../../.cargo/git/checkouts/powdr-5ad669c296f4e377/47f65a8/cargo-powdr/template/Cargo.toml:2:8
  |
2 | name = "{{PROJECT_NAME}}"
  |        ^^^^^^^^^^^^^^^^^^
  |
error: invalid character `{` in package name: `{{GUEST_NAME}}`, the first character must be a Unicode XID start character (most letters or `_`)
 --> ../../.cargo/git/checkouts/powdr-5ad669c296f4e377/47f65a8/cargo-powdr/template/guest/Cargo.toml:2:8
  |
2 | name = "{{GUEST_NAME}}"
  |        ^^^^^^^^^^^^^^^^
  |
```

leonardoalt:
weird, what exactly did you run?

cargo-powdr/template/guest/src/main.rs
leonardoalt: added a comment above this one

leonardoalt:
Fixed the comments and added README.me and rust_toolchain.toml files too to the template.

leonardoalt:
So weird, locally `cargo install --git https://github.com/powdr-labs/powdr cargo-powdr --branch cargo-powdr`  works for me, but on the server it doesn't.

leonardoalt:
Ok, fixed by renaming the template `Cargo.toml` files to `Cargo.toml.template`, which also works on the server. The issue is that even though those crates are not supposed to be compiled, cargo still tries to parse any `Cargo.toml` files found.

pacheco (approved):

PR #1970, opened 2024-10-28
Abstract solved trait impls

chriseth:


ast/src/analyzed/mod.rs
```diff
@@ -565,6 +565,65 @@ pub fn type_from_definition(
...
+        &mut self,
+        trait_function_name: String,
+        type_args: Vec<Type>,
+        index: usize,
+        function: Arc<Expression>,
+    ) {
+        self.impls
+            .entry(trait_function_name)
+            .or_default()
+            .insert(type_args, ImplData { index, function });
```
gzanitti, line 623: We never merge the code to check for overlaps. Couldn't we have a problem if this `.insert(type_args, ...)` overlaps?

gzanitti (approved):

ast/src/analyzed/mod.rs
chriseth: Yes, but that's why we still need to do the overlap check :)

PR #1973, opened 2024-10-28
fix `--export-csv` and skip witgen when full witness is provided

pacheco:
This PR does two things:
- make `--export-csv` not export fixed columns, so it can be directly used as external witness
- skip witness generation when all witness columns are provided externally

### DIFF
--- a/number/src/serialize.rs
+++ b/number/src/serialize.rs
@@ -74,6 +74,10 @@ pub fn read_polys_csv_file<T: FieldElement>(file: impl Read) -> Vec<(String, Vec
     for result in reader.records() {
         let record = result.unwrap();
         for (idx, value) in record.iter().enumerate() {
+            // shorter polys/columns end in empty cells
+            if value.trim().is_empty() {
+                continue;
+            }
             let value = if let Some(value) = value.strip_prefix("0x") {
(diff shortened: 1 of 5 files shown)


number/src/serialize.rs
```diff
@@ -74,6 +74,10 @@ pub fn read_polys_csv_file<T: FieldElement>(file: impl Read) -> Vec<(String, Vec
     for result in reader.records() {
         let record = result.unwrap();
         for (idx, value) in record.iter().enumerate() {
+            // shorter polys/columns end in empty cells
```
leonardoalt, line 77: wdym shorter here? sounds like it's just empty in the end?

lvella:
Won't this make the debugging harder? Since it don't display the constants...

leonardoalt:
there's an option if you want to see the constants too

number/src/serialize.rs
pacheco: with VADCOP, the cells for rows larger than a machine's degree are empty

leonardoalt: But then the entire col isnt written?

leonardoalt (approved):

leonardoalt (approved):

PR #1981, opened 2024-10-29
Use more parallel jobs.

chriseth:
Two very long running tests currently end up in the same job, this tries to fix this. if the slow tests are faster than 32 minutes, this is successful.


chriseth:
Weird. `evm` and `features` still end up in the same job.

PR #1983, opened 2024-10-29
Pass degree to submachines mvp

Schaeff:
When passing arguments to a machine which has n submachines, the first n arguments are interpreted as the submachines, then two more optional arguments are allowed, which are interpreted as the min and max degree for that submachine.


airgen/src/lib.rs
```diff
@@ -68,12 +68,15 @@ pub fn compile(input: AnalysisASMFile) -> PILGraph {
         let machine = &input.get_machine(&ty).unwrap();
 
         queue.extend(machine.submachines.iter().map(|def| {
+            let called_machine = &input.get_machine(&def.ty).unwrap();
```
leonardoalt, line 71: @Schaeff  shadowing sucks :upside_down_face:

leonardoalt (approved):

PR #1985, opened 2024-10-30
Improve fingerprint expression performance.

chriseth:


georgwiese (commented):
Awesome! This is a massive improvement over the main branch, and folding ~70 elements (as we do for the PC lookup in the RISC-V machine) is now feasible.

Repeating [this benchmark](https://gist.github.com/georgwiese/cb5f1436bf5951ce3dddc0987d90497e), I get this curve:
Screenshot 2024-10-30 at 13 00 54

So, it's still exponential, but blows up later.

I wonder if implementing repeated squaring would get us to polynomial scaling behavior.

In either case, I think before merging we should add a test of `fingerprint()` first, so we can be sure that the implementations are equivalent. I can have a look in a separate PR.

georgwiese (commented):
Nice, using the optimized square gave another performance boost, making it feasible to fold ~100 expressions:
Screenshot 2024-10-31 at 12 10 35

My final request is, please add a test for `pow_ext` :)

georgwiese (approved):
LGTM if you wanna add the test after #1999 ;)

PR #1994, opened 2024-10-30
Fix utility function filter.

chriseth:


airgen/src/lib.rs
```diff
@@ -178,31 +178,7 @@ fn utility_functions(asm_file: AnalysisASMFile) -> BTreeMap<AbsoluteSymbolPath,
...
-                        | PilStatement::StructDeclaration(..)
-                        | PilStatement::LetStatement(..) => true,
-                        PilStatement::Include(..) => false,
-                        PilStatement::Namespace(..) => false,
-                        PilStatement::PolynomialDefinition(..) => false,
-                        PilStatement::PublicDeclaration(..) => false,
-                        PilStatement::PolynomialConstantDeclaration(..) => false,
-                        PilStatement::PolynomialConstantDefinition(..) => false,
-                        PilStatement::PolynomialCommitDeclaration(..) => false,
-                        PilStatement::TraitImplementation(..) => false,
```
chriseth, line 199: Trait impls and trait declarations were incorrectly removed here.

I first changed them to `true` and replaced all other `false` cases by `unreachable!()` and did not hit a panic. So I guess this filter is just redundant.

airgen/src/lib.rs
```diff
@@ -220,31 +220,7 @@ fn utility_functions(asm_file: AnalysisASMFile) -> BTreeMap<AbsoluteSymbolPath,
     asm_file
```
gzanitti, line 220: What is a utility function? Maybe a comment or a more descriptive name (if that makes sense)?

Schaeff (approved):

PR #1998, opened 2024-10-30
SourceRef: path_canonilizer (includes `Param`)

gzanitti:
Add source references for path_canonilizer. Also, as a requirement for this, add source references to `Param`.


importer/src/path_canonicalizer.rs
```diff
@@ -682,11 +682,11 @@ fn check_pil_statement_inside_module(
...
-            check_enum_declaration(&location, enum_decl, state)
+        PilStatement::EnumDeclaration(src_ref, enum_decl) => {
+            check_enum_declaration(src_ref, &location, enum_decl, state)
         }
-        PilStatement::StructDeclaration(_, struct_decl) => {
-            check_struct_declaration(&location, struct_decl, state)
+        PilStatement::StructDeclaration(src_ref, struct_decl) => {
+            check_struct_declaration(src_ref, &location, struct_decl, state)
         }
         PilStatement::TraitImplementation(_, trait_impl) => {
```
Schaeff, line 691: is this one left out on purpose?

chriseth: It only calls `check_type_scheme` and `check_expression` who operate on data structures that have their own (more specific) source references.

PR #1999, opened 2024-10-31
Test runner.

chriseth:
Simple command line utility that can run tests in a `.asm` file.


cli/src/main.rs
```diff
@@ -363,6 +364,24 @@ enum Commands {
...
+        /// The field to use
+        #[arg(long)]
+        #[arg(default_value_t = FieldArgument::Gl)]
+        #[arg(value_parser = clap_enum_variants!(FieldArgument))]
+        field: FieldArgument,
+
+        /// Also run the tests inside the standard library.
+        #[arg(long)]
+        #[arg(default_value_t = false)]
+        include_std_tests: bool,
```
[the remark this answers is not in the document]
chriseth: It filters tests based on their namespace. Since `std` is always included, we need to filter the tests out. If you run `powdr test std` with the filter active, it will not run any tests.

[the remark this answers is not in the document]
georgwiese: Hmm, but does this need to leak to the CLI? Or could we just say we never want to run the std tests, unless we pass `std` explicitly?

pipeline/tests/powdr_std.rs
```diff
@@ -396,57 +396,19 @@ fn ff_inv_big() {
...
-
-#[test]
-fn fingerprint() {
-    let analyzed = std_analyzed::<GoldilocksField>();
-    evaluate_function(
-        &analyzed,
-        "std::protocols::fingerprint::test::test_fingerprint",
-        vec![],
+fn std_tests() {
+    let test_count = 9;
```
georgwiese, line 400: Is it important to assert that there are 9 tests? In what scenario would it not run all tests? I don't see a good reason and it will be slightly annoying that we have to change this number when adding a new test.

pipeline/tests/powdr_std.rs
```diff
@@ -396,57 +396,19 @@ fn ff_inv_big() {
...
+        test_count,
+        run_tests(&std_analyzed::<GoldilocksField>(), true).unwrap()
+    );
+    assert_eq!(
+        test_count,
+        run_tests(&std_analyzed::<Bn254Field>(), true).unwrap()
+    );
+    assert_eq!(
+        test_count,
+        run_tests(&std_analyzed::<BabyBearField>(), true).unwrap(),
```
georgwiese, line 411: We also have BabyBear and KoalaBear, actually!

georgwiese: Also, this approach assumes that all tests should be run on all fields. I think we might have tests in the future that are only supposed to pass on some fields. But we we can also fix that when we get there.

pipeline/src/test_runner.rs
```diff
@@ -0,0 +1,109 @@
...
+            " ".to_string()
+        } else {
+            " ".repeat(76 - name_len)
+        };
+        print!("{name}...");
+        let function = symbols.lookup(name, &None).unwrap();
+        match evaluator::evaluate_function_call::<F>(function, vec![], &mut symbols) {
+            Err(e) => {
+                let msg = e.to_string();
+                println!("{padding}failed\n  {msg}");
```
georgwiese, line 89: beautiful 😍

pipeline/tests/powdr_std.rs
chriseth: Sure but then the test function should return early. We might find a more convenient way, though.

pipeline/tests/powdr_std.rs
chriseth: It's just a safeguard against missing some tests because of the naming scheme or something.

georgwiese (approved):
LGTM

PR #2002, opened 2024-10-31
fix registers in continuations

leonardoalt:


riscv/src/continuations.rs
```diff
@@ -419,21 +421,24 @@ pub fn rust_continuations_dry_run<F: FieldElement>(
         }
 
         // Go over all registers except the PC
```
pacheco, line 423: // Go over all register in memory

riscv/src/continuations.rs
```diff
@@ -419,21 +421,24 @@ pub fn rust_continuations_dry_run<F: FieldElement>(
...
+                let reg = reg.strip_prefix("main::").unwrap();
                 let id = Register::from(reg).addr();
                 *register_memory_snapshot
                     .get(&(id as u32))
                     .unwrap_or(&0.into())
             })
             .collect::<Vec<_>>();
 
-        register_values.push(*chunk_trace["main::pc"].last().unwrap());
+        for reg in REGISTER_NAMES {
```
pacheco, line 435: // Go over machine registers

riscv/src/continuations/bootloader.rs
```diff
@@ -90,48 +90,49 @@ pub fn bootloader_and_shutdown_routine(
...
-    "main.x29",
-    "main.x30",
-    "main.x31",
-    "main.tmp1",
-    "main.tmp2",
-    "main.tmp3",
-    "main.tmp4",
-    "main.lr_sc_reservation",
-    "main.pc",
+pub const REGISTER_MEMORY_NAMES: [&str; 36] = [
```
pacheco, line 93: specify in the comment that these are registers in memory only?

riscv/src/continuations/bootloader.rs
```diff
@@ -90,48 +90,49 @@ pub fn bootloader_and_shutdown_routine(
...
+    "main::x30",
+    "main::x31",
+    "main::tmp1",
+    "main::tmp2",
+    "main::tmp3",
+    "main::tmp4",
+    "main::lr_sc_reservation",
 ];
 
+pub const REGISTER_NAMES: [&str; 3] = ["main::query_arg_1", "main::query_arg_2", "main::pc"];
```
pacheco, line 132: specify in a comment that these are machine registers?

leonardoalt:
fixed all

riscv/src/large_field/bootloader.rs
```diff
@@ -523,15 +544,26 @@ shutdown_start:
...
     );
 
-    // Go over all registers except the PC
-    let register_iter = REGISTER_NAMES.iter().take(REGISTER_NAMES.len() - 1);
+    // Go over all memory registers
+    for (i, reg) in REGISTER_MEMORY_NAMES.iter().enumerate() {
+        let reg = reg.strip_prefix("main::").unwrap();
+        bootloader.push_str(&format!(
+            "assert_bootloader_input 0, {}, 1, {};\n",
+            Register::from(reg).addr(),
```
pacheco, line 552: can you check and leave a TODO on this `Register::from` implementation to maybe use the `REGISTER*NAMES`? It would be less likely to break in the future should one of them change without the other

pacheco (approved):

PR #2004, opened 2024-10-31
Fix pilopt select

chriseth:
Fixes item 1 in https://github.com/powdr-labs/powdr/pull/1867


pilopt/src/lib.rs
```diff
@@ -333,6 +333,15 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
             ..
         }) = identity
         {
+            let r_sel = &right.selector;
+            // We can only do this if we know that the selector is one in at least
+            // one row, but this is too complicated to detect (especially if we
+            // have a dynamic degree), so we just do this for constants to be safe.
+            match r_sel {
+                AlgebraicExpression::Number(r_sel) if r_sel == &T::zero() => continue,
+                AlgebraicExpression::Number(_) => {}
```
chriseth, line 342: Just wondering, since this is field-independent, if we put the modulus here as a number, it could still be zero...

chriseth: Maybe just do it for explicit 1?

leonardoalt (dismissed):

leonardoalt:
has conflicts

leonardoalt (approved):

PR #2008, opened 2024-10-31
DeclaredType in TypeChecker

gzanitti:
As part of the Struct integration (mainly typing), we need to improve the way “declared types” are handled so that we can type struct fields without passing new data structures to type checking.

This PR introduces the necessary modification to then add Struct typing on top of it, preventing #1910 from being huge.

### DIFF
--- a/pil-analyzer/src/type_inference.rs
+++ b/pil-analyzer/src/type_inference.rs
@@ -263,7 +263,7 @@ impl TypeChecker {
 
         match declared_type.ty {
             DeclaredTypeKind::Struct(_, _) => {
-                unreachable!("Struct declarations should have been removed at this point")
+                unreachable!("Declared types for Structs should have been removed at this point")
             }
             DeclaredTypeKind::Type(_) => declared_type.scheme(),
         }
--- a/pil-analyzer/src/type_inference.rs
+++ b/pil-analyzer/src/type_inference.rs
@@ -185,7 +185,7 @@ impl TypeChecker {
         // After we setup declared types, every definition
         // related with a Struct Declaration is not nedded any more
         let mut definitions: HashMap<_,_> = definitions
-        .into_iter()
+        .iter_mut()
         .filter(|(_, (ty, _))| {
             !matches!(ty, Some(declared) if matches!(declared.ty, DeclaredTypeKind::Struct(_, _)))
         })
--- a/pil-analyzer/src/type_inference.rs
+++ b/pil-analyzer/src/type_inference.rs
@@ -263,7 +263,7 @@ impl TypeChecker {
 
         match declared_type.ty {
             DeclaredTypeKind::Struct(_, _) => {
-                unreachable!("Struct declarations should be removed")
+                unreachable!("Struct declarations should have been removed at this point")
             }
             DeclaredTypeKind::Type(_) => declared_type.scheme(),
         }
--- a/pil-analyzer/src/type_inference.rs
+++ b/pil-analyzer/src/type_inference.rs
@@ -29,7 +29,7 @@ use crate::{
 /// Sets the generic arguments for references and the literal types in all expressions.
 /// Returns the types for symbols without explicit type.
 pub fn infer_types(
-    definitions: HashMap<String, (Option<TypeScheme>, Option<&mut Expression>)>,
+    definitions: HashMap<String, (Option<DeclaredType>, Option<&mut Expression>)>,
     expressions: &mut [(&mut Expression, ExpectedType)],
 ) -> Result<Vec<(String, Type)>, Vec<Error>> {
     TypeChecker::new().infer_types(definitions, expressions)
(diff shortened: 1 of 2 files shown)


pil-analyzer/src/type_inference.rs
```diff
@@ -977,6 +1037,11 @@ impl TypeChecker {
     pub fn local_var_type(&self, id: u64) -> Type {
         self.local_var_types[id as usize].clone()
     }
+
+    fn instantiate_scheme_by_declared_name(&mut self, name: &str) -> (Type, Vec<Type>) {
+        self.unifier
+            .instantiate_scheme(self.declared_types[&name.to_string()].scheme().clone())
```
chriseth, line 1043:
```suggestion
            .instantiate_scheme(self.declared_types[name].scheme().clone())
```

pil-analyzer/src/type_inference.rs
```diff
@@ -977,6 +1037,11 @@ impl TypeChecker {
     pub fn local_var_type(&self, id: u64) -> Type {
         self.local_var_types[id as usize].clone()
     }
+
+    fn instantiate_scheme_by_declared_name(&mut self, name: &str) -> (Type, Vec<Type>) {
```
chriseth, line 1041:
```suggestion
    /// Returns a new type scheme for the given name.
    fn instantiate_scheme_by_declared_name(&mut self, name: &str) -> (Type, Vec<Type>) {
```

pil-analyzer/src/type_inference.rs
```diff
@@ -913,18 +972,19 @@ impl TypeChecker {
         inferred_types: HashMap<String, Type>,
     ) -> Result<HashMap<String, HashMap<String, Type>>, Error> {
         inferred_types.into_iter().map(|(name, inferred_type)| {
-            let (source_ref, declared_type) = self.declared_types[&name].clone();
+            let declared_type = self.declared_types[&name].clone();
```
chriseth, line 975: Can you create a function here that abstracts some of this functionality so that you don't need to know what `DeclaredType` is in this function? I.e. already call `simplify_type_vars()` in that function etc.

pil-analyzer/src/type_inference.rs
chriseth: Oh and you also need to explain what the vector is.

pil-analyzer/src/type_inference.rs
```diff
@@ -145,10 +206,10 @@ impl TypeChecker {
                 continue;
             };
 
-            let (_, declared_type) = self.declared_types[&name].clone();
+            let declared_type = self.declared_types[&name].clone();
```
chriseth, line 209: Please abstract this way of access. It could just return a TypeScheme (and panic for struct because we don't get here for structs).

pil-analyzer/src/type_inference.rs
```diff
@@ -168,13 +229,13 @@ impl TypeChecker {
 
         // Now we check for all symbols that are not declared as a type scheme that they
         // can resolve to a concrete type.
-        for (name, (source_ref, declared_type)) in &self.declared_types {
```
chriseth, line 171: Instead of accessing `self.declared_types` directly, return an iterator over all concrete symbols.

pil-analyzer/src/type_inference.rs
```diff
@@ -912,28 +985,51 @@ impl TypeChecker {
...
-                    .into_iter()
-                    .cloned()
-                    .zip(declared_type_vars.into_iter().map(|tv| Type::TypeVar(tv.clone())))
-                    .collect(),
-            ))
-        }).collect::<Result<_, Error>>()
+        inferred_types
+            .into_iter()
+            .map(|(name, inferred_type)| {
+                let declared_type = self.declared_types[&name].clone();
```
chriseth, line 991: Why not move this into `compare_type_schemes` as well? Maybe also name it `compare_inferred_type_to_declared`?

pil-analyzer/src/type_inference.rs
```diff
@@ -168,23 +230,25 @@ impl TypeChecker {
...
-                                name,
-                                &Some(inferred_scheme),
-                            )
-                        )));
-                }
-            }
-        }
+        self.declared_types
+            .iter()
+            .filter(|(_, declared_type)| declared_type.vars.is_empty())
```
chriseth, line 235: I think `vars.is_empty()` is a very common operation on DeclaredType - maybe this could be a function `is_concrete`?

pil-analyzer/src/type_inference.rs
```diff
@@ -237,7 +313,7 @@ impl TypeChecker {
         for (name, scheme) in builtin_schemes() {
             self.declared_types
                 .entry(name.clone())
-                .or_insert_with(|| (SourceRef::unknown(), scheme.clone()));
+                .or_insert_with(|| Into::<DeclaredType>::into(scheme.clone()));
```
chriseth, line 316: `DeclaredType::from(...`?

chriseth:
This looks very good now, thanks! Now my main remaining question is: Can we turn the two enums into a single one?

gzanitti:
> This looks very good now, thanks! Now my main remaining question is: Can we turn the two enums into a single one?

Do you mean DeclaredType and DeclaredTypeKind? Maybe adding the type and an Option<HashMap<...>> to the Struct? I find the Option more cumbersome than having another Enum, but whatever you prefer :)

leonardoalt:
Closing for now.

PR #2012, opened 2024-11-01
add new Fib example with many chunks

leonardoalt:


examples/multiple_chunks/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+        .out_path("powdr-target")
+        .chunk_size_log2(18)
+        .build()
+        .write(1, &data)
+        .write(2, &data.iter().sum::<u32>());
+
+    // Fast dry run to test execution.
+    session.run();
+
+    // Uncomment to compute the proof.
```
Schaeff, line 19: it's uncommented

riscv/src/continuations.rs
```diff
@@ -141,6 +146,11 @@ where
                     ),
                 ]);
                 pipeline_callback(pipeline)?;
+
+                if let Some(original_dir) = parent_dir {
+                    pipeline.set_output(original_dir, force_overwrite);
+                }
```
leonardoalt, line 152: Before this fix continuations would do:

```console
[2024-11-01T09:43:02Z INFO  powdr_pipeline::pipeline] Writing powdr-target/chunk_0/chunk_1/guest_proof.bin.                                                   
```

leonardoalt: Now:
```console
Writing powdr-target/chunk_1/guest_proof.bin.
```

examples/multiple_chunks/src/main.rs
leonardoalt: removed comment

leonardoalt:
After talking to @Schaeff  I'll add a different example where the guest has some explicit computation that span across multiple chunks.

Schaeff (approved):

PR #2014, opened 2024-11-01
Make some types explicit for columns.

chriseth:
Add some explicit types or explicit conversions to prepare for the `ToCol` trait.


jit-compiler/src/codegen.rs
```diff
@@ -150,9 +150,7 @@ impl<'a, T: FieldElement> CodeGenerator<'a, T> {
                     }
                 })
             }
-            _ => Err(format!(
-                "Definition of this kind not supported: {symbol} - {definition}"
```
chriseth, line 154: Removed this because this might call the wrong display function and lead to a panic.

lvella (approved):

PR #2020, opened 2024-11-01
Bring back p3 composite

Schaeff:
This makes PIL splitting idempotent since in the composite case it is called twice: in the composite backend, and inside the integrated plonky3 prover, which leads to a duplicate dummy column before this change.

Tests still use the integrated prover. Usage as expected:
```console
cargo run pil test_data/asm/empty.asm --prove-with plonky3-composite --force
```


pipeline/tests/asm.rs
```diff
@@ -38,7 +39,11 @@ fn simple_sum_asm() {
     let f = "asm/simple_sum.asm";
     let i = [16, 4, 1, 2, 8, 5];
     regular_test(f, &i);
-    test_plonky3::<GoldilocksField>(f, slice_to_vec(&i));
+    test_plonky3_with_backend_variant::<GoldilocksField>(
+        f,
+        slice_to_vec(&i),
+        BackendVariant::Monolithic,
```
leonardoalt, line 45: Why Monolithic?

pipeline/tests/pil.rs
```diff
@@ -90,7 +91,11 @@ fn permutation_with_selector() {
 fn fibonacci() {
     let f = "pil/fibonacci.pil";
     regular_test(f, Default::default());
-    test_plonky3::<GoldilocksField>(f, Default::default());
+    test_plonky3_with_backend_variant::<GoldilocksField>(
+        f,
+        Default::default(),
+        BackendVariant::Monolithic,
```
leonardoalt, line 97: same here

leonardoalt (approved):

pipeline/tests/asm.rs
Schaeff: it's the integrated prover

PR #2023, opened 2024-11-04
Split common parts of poseidon2 out of poseidon2_bb

lvella:
In preparation for #2018.

### DIFF
--- a/std/machines/small_field/pointer_arith.asm
+++ b/std/machines/small_field/pointer_arith.asm
@@ -1,3 +1,5 @@
+use std::array;
+
 /// Creates constraints that increments a 32-bits pointer by `amount` bytes.
 ///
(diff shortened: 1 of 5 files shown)


lvella:
I am still not sure if the new argument "output_capacity" is a net positive.

std/machines/small_field/pointer_arith.asm
```diff
@@ -37,3 +39,28 @@ let increment_ptr: expr, expr, expr, expr, expr -> Constr[] = constr |
         post_high = pre_high + carry
     ]
 };
+
+// Creates a sequence of 4-byte sparsed addresses.
```
chriseth, line 43: What is a `4-byte sparsed address`?

Also please use triple slashes.

PR #2027, opened 2024-11-04
Expression processor: Results & Error Handling

gzanitti:


pil-analyzer/src/untyped_evaluator.rs
```diff
@@ -17,7 +17,9 @@ pub fn evaluate_expression_to_int(
     expr: parsed::Expression,
 ) -> Result<BigInt, EvalError> {
     evaluator::evaluate_expression::<GoldilocksField>(
-        &ExpressionProcessor::new(driver, &Default::default()).process_expression(expr),
+        &ExpressionProcessor::new(driver, &Default::default())
+            .process_expression(expr)
+            .map_err(|e| EvalError::TypeError(e.message().to_string()))?, //TODO: better error type
```
gzanitti, line 22: It's necessary to add a new variant to EvalError for this case. What name, generic enough to be a “processing error”. should we give it?

pil-analyzer/src/expression_processor.rs
```diff
@@ -41,51 +41,57 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
...
             .into_iter()
             .map(|e| self.process_expression(e))
-            .collect()
+            .collect::<Result<Vec<_>, _>>()
     }
 
-    pub fn process_expression(&mut self, expr: parsed::Expression) -> Expression {
+    pub fn process_expression(&mut self, expr: parsed::Expression) -> Result<Expression, Error> {
         use parsed::Expression as PExpression;
         match expr {
```
chriseth, line 72: I often use the style of `Ok(match expr{...`, then you don't need an `Ok(` in every single branch.

pil-analyzer/src/expression_processor.rs
```diff
@@ -94,54 +100,59 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
...
                             .map(|MatchArm { pattern, value }| {
                                 let vars = self.save_local_variables();
-                                let pattern = self.process_pattern(pattern);
-                                let value = self.process_expression(value);
+                                let result = match (
+                                    self.process_pattern(pattern),
+                                    self.process_expression(value),
+                                ) {
+                                    (Ok(pattern), Ok(value)) => Ok(MatchArm { pattern, value }),
+                                    (Err(e), _) | (_, Err(e)) => Err(e),
```
chriseth, line 148: Why not use the `?` operator here?

pil-analyzer/src/expression_processor.rs
```diff
@@ -150,55 +161,60 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
...
-                            type_args,
-                        }),
+                        name: Reference::Poly(PolynomialReference { name, type_args }),
                         fields: fields
                             .into_iter()
-                            .map(|named_expr| NamedExpression {
-                                name: named_expr.name,
-                                body: Box::new(self.process_expression(*named_expr.body)),
+                            .map(|named_expr| {
+                                let processed_body = self.process_expression(*named_expr.body);
```
chriseth, line 193: `?`?

pil-analyzer/src/expression_processor.rs
```diff
@@ -227,153 +244,198 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
...
-                }
-                StatementInsideBlock::Expression(expr) => {
-                    StatementInsideBlock::Expression(self.process_expression(expr))
-                }
-            })
-            .collect::<Vec<_>>();
+        let processed_statements =
+            statements
+                .into_iter()
+                .try_fold(Vec::new(), |mut acc, statement| {
```
chriseth, line 371: `collect` does not work here?

pil-analyzer/src/pil_analyzer.rs
```diff
@@ -519,8 +519,14 @@ impl PILAnalyzer {
...
     fn handle_namespace(&mut self, name: SymbolPath, degree: Option<parsed::NamespaceDegree>) {
         let evaluate_degree_bound = |e| {
-            let e =
-                ExpressionProcessor::new(self.driver(), &Default::default()).process_expression(e);
+            let e = match ExpressionProcessor::new(self.driver(), &Default::default())
+                .process_expression(e)
+            {
+                Ok(e) => e,
+                Err(e) => {
+                    panic!("Failed to evaluate degree bound: {e}");
```
chriseth, line 527:
```suggestion
                    // TODO propagate this error up
                    panic!("Failed to evaluate degree bound: {e}");
```

pil-analyzer/src/statement_processor.rs
```diff
@@ -213,7 +213,8 @@ where
             }
             PilStatement::Expression(_, expr) => vec![PILItem::ProofItem(
                 self.expression_processor(&Default::default())
-                    .process_expression(expr),
+                    .process_expression(expr)
+                    .expect("Expression processing failed"),
```
chriseth, line 217:
```suggestion
                    // TODO propagate this error up
                    .expect("Expression processing failed"),
```

pil-analyzer/src/untyped_evaluator.rs
chriseth: I mean in the end, this function should return two different errors. I think it would be better to change this function to just take an `analyzed::Expression`, and do the expression processing at the caller side.

Three of the four call sites are inside StatementProcessor which makes regular use of Expression processor anyway.

pil-analyzer/tests/types.rs
```diff
@@ -480,7 +480,7 @@ fn enum_pattern() {
 }
 
 #[test]
-#[should_panic = "Only one \"..\"-item allowed in array pattern"]
+#[should_panic = "Only one \\\"..\\\"-item allowed in array pattern"]
```
chriseth, line 483: Is this change correct?

chriseth:
Looks good! The only comment left is the updated test expectation. Is it just a temporary change?

The untyped evaluator can be done in its own PR.

pil-analyzer/tests/types.rs
gzanitti: Is the way it's parsed now that the error is a proper error and not an string, any ideas on how to solve it?

chriseth:
Did you check if the escaping issue is a real issue (by running the commandline interface) or just a testing issue?

gzanitti:
> Did you check if the escaping issue is a real issue (by running the commandline interface) or just a testing issue?
Oh, yes, the three lines only occur when parsing the error during testing. When running the command line interface, the error is the same as the one that is created in the message

chriseth (approved):

PR #2046, opened 2024-11-05
Implements publics in Rust and RISCV

leonardoalt:
Fixes https://github.com/powdr-labs/powdr/issues/2039
Fixes https://github.com/powdr-labs/powdr/issues/2038

- This PR adds one public hash (8 RISCV32 words) to the RISCV machine.
- Rust users can commit any amount of u32s using `powdr_riscv_runtime::commit`
- The Rust runtime automatically invokes a finalization routine after the program, that takes the hash of the whole thing and exposes to the public cells
- Because the Rust runtime stuff is the same for both GL and BB machines, I had to add a fake poseidon to the BB Runtime so that the RISCV BB tests don't fail compilation.
- Poseidon is enabled by default now in RuntimeLibs because of publics.


cli-rs/src/main.rs
```diff
@@ -364,7 +364,6 @@ fn coprocessors_to_options(coprocessors: Option<String>) -> Result<RuntimeLibs,
         let names = list.split(',').collect::<Vec<_>>();
         for name in names {
             match name {
-                "poseidon_gl" => libs = libs.with_poseidon(),
```
leonardoalt, line 367: Removed from here because it's now default

riscv-executor/src/lib.rs
```diff
@@ -1138,6 +1138,12 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
 
                 vec![]
             }
+            "commit_public" => {
```
leonardoalt, line 1141: We should probably return these at some point when the clients of riscv-executor need it.

riscv-syscalls/src/lib.rs
```diff
@@ -48,11 +48,12 @@ macro_rules! syscalls {
 syscalls!(
     (1, Input, "input"),
     (2, Output, "output"),
-    (3, PoseidonGL, "poseidon_gl"),
+    (3, Poseidon, "poseidon"),
```
leonardoalt, line 51: Renamed for common use between GL and BB machines. The interface is the same.

riscv/src/large_field/code_gen.rs
```diff
@@ -296,6 +296,24 @@ fn preamble(field: KnownField, runtime: &Runtime, with_bootloader: bool) -> Stri
...
+    /*
+    public hash_0 = main_publics::value(0);
+    public hash_1 = main_publics::value(1);
+    public hash_2 = main_publics::value(2);
+    public hash_3 = main_publics::value(3);
+    public hash_4 = main_publics::value(4);
+    public hash_5 = main_publics::value(5);
+    public hash_6 = main_publics::value(6);
+    public hash_7 = main_publics::value(7);
+    */
```
leonardoalt, line 314: Commented out to get the tests to run. When uncommented we get this:

```console
thread 'main' panicked at pil-analyzer/src/expression_processor.rs:374:18:
TODO: Handle this up in the code: "Value symbol not found: main_publics::value"
```

We need to decide if we want to go ahead without these or fix first.

riscv/src/small_field/runtime.rs
```diff
@@ -272,8 +274,15 @@ impl Runtime {
...
     }
 
-    fn with_poseidon(self, _continuations: bool) -> Self {
-        todo!()
+    // TODO This implementation is just a placeholder,
+    // needed by the runtime "finalize" code.
+    fn with_poseidon(mut self, continuations: bool) -> Self {
+        assert!(!continuations);
+
+        let implementation = std::iter::once("affine 0, 0, 0, 0, 0, 0;".to_string());
```
leonardoalt, line 282: This could be a no-op, but the submachine mechanism checks that this parses as a `FunctionStatement`.

leonardoalt:
This version should work.

riscv-runtime/src/commit.rs
```diff
@@ -0,0 +1,55 @@
...
+pub fn commit(n: u32) {
+    unsafe { PUBLICS.commit(n) }
+}
+
+pub(crate) fn finalize() -> [u64; 4] {
+    unsafe { PUBLICS.finalize() }
+}
+
+pub struct CommittedPublics {
+    state: [u64; 12],
```
Schaeff, line 12: Why 12?

riscv-runtime/src/commit.rs
```diff
@@ -0,0 +1,55 @@
...
+            buffer_size: 0,
+        }
+    }
+
+    pub fn commit(&mut self, n: u32) {
+        self.state[self.buffer_size as usize + 4] = n as u64;
+
+        self.buffer_size += 1;
+
+        if self.buffer_size == 4 {
```
Schaeff, line 29: Why 4?

riscv/src/large_field/code_gen.rs
```diff
@@ -296,6 +296,13 @@ fn preamble(field: KnownField, runtime: &Runtime, with_bootloader: bool) -> Stri
     col witness tmp3_col;
     col witness tmp4_col;
 
+    // ================ Publics ==================
+    std::machines::write_once_memory_with_8_publics::WriteOnceMemoryWith8Publics publics;
```
Schaeff, line 300: indent?

std/machines/write_once_memory_with_8_publics.asm
```diff
@@ -0,0 +1,32 @@
+// Very simple write-once memory that let's you store any field element at
```
Schaeff, line 1:
```suggestion
// Very simple write-once memory that lets you store any field element at
```

std/machines/write_once_memory_with_8_publics.asm
```diff
@@ -0,0 +1,32 @@
...
+// an address from 0 to N-1. Any value can be written only once, writing two
```
Schaeff, line 2:
```suggestion
// an address from 0 to 7. Any value can be written only once, writing two
```

riscv/src/lib.rs
```diff
@@ -33,7 +33,8 @@ impl RuntimeLibs {
         Self {
             arith: false,
             keccak: false,
-            poseidon: false,
+            // Poseidon is required for publics.
+            poseidon: true,
```
Schaeff, line 37: I don't have full context here but if it's required, does it still make sense for it to be in this struct? Is the value ever going to be false?

riscv-runtime/src/commit.rs
leonardoalt: That's the size of the poseidon state which we replicate here in this pseudo-sponge construction

riscv-runtime/src/commit.rs
leonardoalt: Every time we get 4 words we compute a hash, it's the same size as the output which becomes the input of the next round

riscv/src/lib.rs
leonardoalt: hm yea you could still want non-continuations cases without publics. I'll revert this and change the tests/APIs accordingly

riscv/src/large_field/code_gen.rs
leonardoalt: looks fine to me? or what's wrong?

std/machines/write_once_memory_with_8_publics.asm
leonardoalt: updated, also in write_once_memory which was the base for this file

riscv/src/lib.rs
leonardoalt: yea actually it's not only about continuations, it's mainly about publics, so it is indeed always required by the RISCV machine

leonardoalt: Ok removed `poseidon` from RuntimeLibs field, it's now set inside Runtime itself as default

riscv-runtime/src/hash.rs
```diff
@@ -14,7 +14,7 @@ const GOLDILOCKS: u64 = 0xffffffff00000001;
 /// Goldilocks field.
 pub fn poseidon_gl_unsafe(data: &mut [u64; 12]) -> &[u64; 4] {
     unsafe {
-        asm!("ecall", in("a0") data as *mut [u64; 12], in("t0") u32::from(Syscall::PoseidonGL));
+        asm!("ecall", in("a0") data as *mut [u64; 12], in("t0") u32::from(Syscall::Poseidon));
```
lvella, line 17: Please leave it `PoseidonGL`. When we introduce an interface for `PoseidonBB`, we must not give the impression that it is the same function with the same results.

leonardoalt: This ends up in the RISCV BB machine too though, so what do we do there?

leonardoalt: Or well, the new `finalize` function calls this

lvella: Panic as unimplemented.

If the user doesn't care about what specific hash function is being used, maybe there should be a "fn hash_native()" that redirects to whatever specialized hash implementation for that field.

leonardoalt: We have tons of tests for the RISCV BB machine that would panic that way, I don't think it's reasonable to just remove them.

lvella: How do they work if we don't have poseidon on BB yet? It should only panic if called.

leonardoalt: In the current version of this PR it translates to the `poseidon` instruction which is a no-op internally.

leonardoalt:
After discussion with @lvella we decided to add another syscall called `HashNative` that is not specific to a field, and each machine has to propagate to its own poseidon internally

riscv-syscalls/src/lib.rs
```diff
@@ -55,4 +55,6 @@ syscalls!(
     (7, KeccakF, "keccakf"),
     (8, Mod256, "mod_256"),
     (9, Halt, "halt"),
+    (10, CommitPublic, "commit_public"),
+    (11, PoseidonNative, "poseidon_native"),
```
lvella, line 59: Has to be poseidon? Can't be some other permutation function?

For the lack of a better name, I suggest "NativeCryptoPermutation".

leonardoalt: currently it will be poseidon, and only poseidon for now has that property, so why not name it accurately?

lvella: So, poseidon2 is a kind of poseidon?

lvella: Also, Keccak fits the bill, and might be the fastest on some field (binary field?)

leonardoalt: I'll go with NativeHash for now, I think it's a bit nicer

riscv/tests/common/mod.rs
```diff
@@ -58,6 +59,7 @@ pub fn verify_riscv_asm_string<T: FieldElement, S: serde::Serialize + Send + Syn
         let pipeline = pipeline.add_external_witness_values(executor_trace);
         test_plonky3_pipeline::<T>(pipeline);
     }
+    */
```
leonardoalt, line 62: The commented out block tests fast witgen from riscv-executor. It doesn't implement Poseidon yet, which is always enabled from this PR. In those cases, auto witgen should be able to use the partial witness and go from there, but it fails in these tests.

Schaeff (dismissed):
I think this is good to go but maybe @lvella has an opinion?

Schaeff (dismissed):

lvella (approved):
The NativeHash interface may need some polishing to be generic across fields, but it works with Goldilocks for now.

PR #2053, opened 2024-11-06
Fix p3 publics

Schaeff:
p3 publics currently incorrectly use the column name as a key. This explains why this was working as long as all public inputs were the same column and value, even with different rows.

### DIFF
--- a/plonky3/src/circuit_builder.rs
+++ b/plonky3/src/circuit_builder.rs
@@ -43,8 +43,8 @@ pub struct ConstraintSystem<T> {
     // for each intermediate polynomial, the expression
     intermediates: BTreeMap<PolyID, AlgebraicExpression<T>>,
     identities: Vec<Identity<T>>,
-    // for each public column, the name, poly_id, index in the witness columns, and stage
-    pub(crate) publics_by_stage: Vec<Vec<(String, PolyID, usize)>>,
+    // for each stage, for each public input of that stage, the name, the column name, the poly_id, the row index
+    pub(crate) publics_by_stage: Vec<Vec<(String, String, PolyID, usize)>>,
     constant_count: usize,
     // for each stage, the number of witness columns. There is always a least one stage, possibly empty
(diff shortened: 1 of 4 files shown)


leonardoalt:
I just tested with https://github.com/powdr-labs/powdr/pull/2046 and now I get this error:

```console
thread 'main' panicked at /home/leo/devel/powdr/plonky3/src/prover.rs:459:36:
no entry found for key
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

asm: https://gist.github.com/leonardoalt/7f59c09b7bc5229ef72e0416e348fda7

Schaeff:
My bad, should have open draft, was still in progress. Thanks for testing though

Schaeff:
> I just tested with #2046 and now I get this error:
> 
> ```
> thread 'main' panicked at /home/leo/devel/powdr/plonky3/src/prover.rs:459:36:
> no entry found for key
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> asm: https://gist.github.com/leonardoalt/7f59c09b7bc5229ef72e0416e348fda7

This is unrelated, will tackle in another PR.

plonky3/src/circuit_builder.rs
```diff
@@ -418,13 +418,13 @@ where
...
         // public constraints
-        let public_vals_by_id = self
+        let public_vals_by_name = self
             .constraint_system
             .publics_by_stage
             .iter()
             .zip_eq(public_input_values_by_stage)
             .flat_map(|(publics, values)| publics.iter().zip_eq(values.iter()))
-            .map(|((id, _, _), pi)| (id, *pi))
+            .map(|((name, _, _, _), pi)| (name, *pi))
```
Schaeff, line 427: This is where the issue was, we were addressing publics by the column name, not the public name

leonardoalt (approved):

PR #2058, opened 2024-11-07
Improve fixed lookup

chriseth:
The change from storing just the row to storing the row values changes the time reported for the fixed machine on the keccak example from 1.5s to 1.2 seconds.

The pc lookup plus copying the values from the columns took 3700 ns, now the lookup takes 500 ns.

Using a hash map takes it down to 1 second.

Current numbers:

```
460 ns: Evaluating LHS
621 ns: Splitting into known and unknown values
201 ns: Actual index lookup
1182 ns: creating the EvalValue result
```

Note that the "1.5 to 1.0 seconds" overall speed improvement includes the time it takes to build the lookup table.

### DIFF
--- a/executor/src/witgen/machines/fixed_lookup_machine.rs
+++ b/executor/src/witgen/machines/fixed_lookup_machine.rs
@@ -1,3 +1,4 @@
+use bit_vec::BitVec;
 use num_traits::One;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::iter::Peekable;
@@ -29,7 +30,7 @@ struct Application {
     pub identity_id: u64,
     /// Booleans indicating if the respective column is a known input column (true)
     /// or an unknown output column (false).
-    pub inputs: Vec<bool>,
+    pub inputs: BitVec,
 }
 
 type Index<T> = HashMap<Vec<T>, IndexValue<T>>;
@@ -254,7 +255,7 @@ impl<'a, T: FieldElement> FixedLookup<'a, T> {
 
         // Split the left-hand-side into known input values and unknon output expressions.
         let mut input_values = vec![];
-        let mut known_inputs = vec![];
+        let mut known_inputs: BitVec = Default::default();
         let mut output_expressions = vec![];
 
         for l in left {
--- a/executor/src/witgen/machines/fixed_lookup_machine.rs
+++ b/executor/src/witgen/machines/fixed_lookup_machine.rs
@@ -23,6 +23,7 @@ use crate::Identity;
 
 use super::{Connection, ConnectionKind, Machine};
 
+/// An Application specifies a lookup cache.
 #[derive(Hash, Eq, PartialEq, Ord, PartialOrd, Clone)]
 struct Application {
     pub identity_id: u64,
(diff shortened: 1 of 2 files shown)


executor/src/witgen/machines/fixed_lookup_machine.rs
```diff
@@ -24,154 +23,146 @@ use crate::Identity;
...
+/// Create an index for a set of columns to be queried, if does not exist already
+/// `input_fixed_columns` is assumed to be sorted
+fn create_index<T: FieldElement>(
+    fixed_data: &FixedData<T>,
+    application: &Application,
+    connections: &BTreeMap<u64, Connection<'_, T>>,
+) -> HashMap<Vec<T>, IndexValue<T>> {
+    let right = connections[&application.identity_id].right;
+
+    // TODO use partition
```
chriseth, line 60: still need to do this.

executor/src/witgen/machines/fixed_lookup_machine.rs
```diff
@@ -272,50 +263,49 @@ impl<'a, T: FieldElement> FixedLookup<'a, T> {
             );
         }
 
+        // TODO
```
chriseth, line 266: todo

executor/src/witgen/machines/fixed_lookup_machine.rs
```diff
@@ -272,50 +252,46 @@ impl<'a, T: FieldElement> FixedLookup<'a, T> {
             );
         }
 
-        // split the fixed columns depending on whether their associated lookup variable is constant or not. Preserve the value of the constant arguments.
-        // [1, 2, x] in [A, B, C] -> [[(A, 1), (B, 2)], [C, x]]
-
-        let mut input_assignment = vec![];
-        let mut output_columns = vec![];
+        // Split the left-hand-side into known input values and unknon output expressions.
```
leonardoalt, line 255:
```suggestion
        // Split the left-hand-side into known input values and unknown output expressions.
```

executor/src/witgen/machines/fixed_lookup_machine.rs
```diff
@@ -24,154 +23,135 @@ use crate::Identity;
...
 
-type Application = (Vec<PolyID>, Vec<PolyID>);
-type Index<T> = BTreeMap<Vec<T>, IndexValue>;
+/// An Application specifies a lookup cache.
+#[derive(Hash, Eq, PartialEq, Ord, PartialOrd, Clone)]
+struct Application {
+    pub identity_id: u64,
+    /// Booleans indicating if the respective column is a known input column (true)
+    /// or an unknown output column (false).
+    pub inputs: Vec<bool>,
```
leonardoalt, line 32: not necessary to change here, but I wonder if this would be faster with a bitvector

chriseth: I can check. I think bitvector is mostly compact, but slow.

leonardoalt: yea fine with keeping

executor/src/witgen/machines/fixed_lookup_machine.rs
leonardoalt: still there

leonardoalt:
conflicts

leonardoalt (dismissed):

leonardoalt (approved):

leonardoalt (approved):

PR #2077, opened 2024-11-12
Patch plonky3 to use Powdr accelerated Poseidon2

lvella:


leonardoalt:
we should to extract some stuff from `p3-recursion-leo` into another PR to get it merged, and also write separate tests for the patched P3

plonky3/src/params/poseidon2/goldilocks/mod.rs
```diff
@@ -0,0 +1,23 @@
+#[cfg(not(all(target_os = "zkvm", target_arch = "riscv32")))]
+mod software_impl;
+#[cfg(not(all(target_os = "zkvm", target_arch = "riscv32")))]
+pub use software_impl::*;
+
+#[cfg(all(target_os = "zkvm", target_arch = "riscv32"))]
+mod powdr_accel_impl;
+#[cfg(all(target_os = "zkvm", target_arch = "riscv32"))]
```
leonardoalt, line 8: I think we should also add a feature `powdr` to the crate and these checks, in case we want to run inside powdr but with software impl. Wdyt?

plonky3/Cargo.toml
```diff
@@ -67,3 +67,12 @@ serde = { version = "1.0", default-features = false, features = [
...
 
+[target.'cfg(all(target_os = "zkvm", target_arch = "riscv32"))'.dependencies]
+powdr-riscv-runtime = { path = "../riscv-runtime", features = [
+  "std",
+  "getrandom",
+  "allow_fake_rand",
+] }
+indexmap = { version = "1.9.3", features = [
+  "std",
+] }
```
leonardoalt, line 78: why this one?

lvella: I copied from your example. Without this, this indexmap crate that is an indirect dependency fails because it thinks it is in a no-std env.

plonky3/src/params/poseidon2/goldilocks/mod.rs
lvella: I prefer with a `powdr` feature, but I did the way you told me to do.

leonardoalt: maybe we should just use the 3? I mean, the Cargo.toml powdr-riscv-runtime dep needs the target anyway, we could be restrictive here too

lvella: 3 what? 3 conditions in the `all`?

leonardoalt: yea, the current 2 in the PR plus the feature

lvella: I don't see it would be better than just using the feature.

If compiling to powdr, both with and without the feature would be valid.

But if not, the feature should not be used and the user should get an error if attempted.

lvella:
Something between p2-recursion-leo and main broke this build.

lvella:
Ah, it is the powdr-executor-utils split. Maybe it is best to not merge this into `main`, because the changes don't even compile.

plonky3/src/params/poseidon2/goldilocks/mod.rs
leonardoalt: Exactly. I'm fine with only using the feature, but then we allow the user to try to compile to native using the feature, which will fail ofc but will be ugly. If we use the target as well the user actively needs to be using powdr to go that path

lvella: Imagine the user builds for x86-64 and sets the feature. What do you want to happen?

I want a compilation error, and that is what will happen if we only the use the feature.

If we use the 3 conditions, the path is not taken and the feature is silently ignored.

leonardoalt:
lint fails

plonky3/Cargo.toml
```diff
@@ -22,6 +22,8 @@ nightly-features = [
   "p3-mersenne-31/nightly-features",
   "p3-poseidon2/nightly-features",
 ]
+# As a guest in powdr, use accelerated machines.
+powdr-accel = []
```
leonardoalt, line 26: powdr-accel sounds a bit weird, why not just powdr?

lvella: Because the thing is accelerated with powdr.

I think a feature named just `powdr` in a library of the powdr project, in a repo named powdr, is confusing.

If not `powdr-accel`, maybe `guest-in-powdr` or something like it?

plonky3/Cargo.toml
```diff
@@ -67,3 +69,12 @@ serde = { version = "1.0", default-features = false, features = [
   "alloc",
 ] }
 
+[target.'cfg(all(target_os = "zkvm", target_arch = "riscv32"))'.dependencies]
+powdr-riscv-runtime = { path = "../riscv-runtime", features = [
```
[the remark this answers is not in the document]
lvella: Then the feature would be required to build for powdr target.

This way, the feature is just about the accelerated operations, and the target itself "just works".

plonky3/Cargo.toml
leonardoalt: ok, then `powdr-accel` it is

plonky3/Cargo.toml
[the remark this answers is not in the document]
leonardoalt: ah yea true, this is still imported by the guest's main or something closer to it which will have the import itself

plonky3/src/params/poseidon2/goldilocks/mod.rs
lvella: Ok, this is not possible unless we accept to break `--all-features`, which I think we don't. The fact that the feature could not be enabled at x86-64 broke our CI, because we build powdr with `--all-features`. I changed for the feature to be silently ignored.

leonardoalt:
merged main

PR #2086, opened 2024-11-13
pilopt: equal-constrained witness columns removal

### LINKED ISSUE #2081: Optimizer should remove equal-constrained witness columns.
The optimizer seems to not remove equal-constrained witness columns.

When run on the `poseidon_gl` std test:
```bash
cargo run   pil -f test_data/std/poseidon_gl_test.asm 
```
The optimized pil contains the following code:
```
col witness X11_const;
main::X11 = main::X11_const;
```

Both `X11` and `X11_const` are witness column.

I would propose the following:
If we have a polynomial constraint of the form `w = v`, where both `w` and `v` are witness columns, then replace all occurrences of `w` by `v` (replace the longer witness columns with the shorter one) and remove `w`.

Note that we cannot do that for array elements currently.

gzanitti:
Resolves #2081


chriseth:
Nice! Did you try what the result on `test_data/std/poseidon_gl_test.asm ` is?

pilopt/src/lib.rs
```diff
@@ -620,3 +621,96 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+            if let Some((replacement_name, replacement_id)) =
+                substitutions_by_id.get(&reference.poly_id)
+            {
+                reference.poly_id = *replacement_id;
+                reference.name = replacement_name.clone();
+            }
+        }
+    });
+
+    let identities_to_remove = substitutions
```
chriseth, line 704: The removal can also be done by the `remove...` steps. I think it would be better to just do replacement here, keeping all the steps as simple as possible.

gzanitti:
> Nice! Did you try what the result on `test_data/std/poseidon_gl_test.asm ` is?

Yes, it went from `Removed 164 witness and 154 fixed columns. Total count now: 99 witness and 50 fixed columns.` to `
Removed 175 witness and 154 fixed columns. Total count now: 88 witness and 50 fixed columns.`

pilopt/src/lib.rs
```diff
@@ -495,6 +496,22 @@ fn remove_trivial_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
                     // Otherwise the constraint is not satisfiable,
                     // but better to get the error elsewhere.
                 }
+                if let AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation {
```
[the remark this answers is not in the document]
chriseth: Ah, maybe turn this into a `match expression { ... }`?

pilopt/src/lib.rs
```diff
@@ -620,3 +637,66 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
         .collect();
     pil_file.remove_identities(&to_remove);
 }
+
+fn equal_constrained<T: FieldElement>(
```
chriseth, line 641: can you document this?

pilopt/src/lib.rs
```diff
@@ -620,3 +637,66 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+    poly_data: (&u64, &AlgebraicExpression<T>),
```
chriseth, line 642: Why not two separate parameters (I think it would also make it clearer what `poly_data` is).

pilopt/src/lib.rs
```diff
@@ -480,23 +481,42 @@ fn constrained_to_constant<T: FieldElement>(
     None
 }
 
-/// Removes identities that evaluate to zero and lookups with empty columns.
+/// Removes identities that evaluate to zero, the same column and lookups with empty columns.
```
chriseth, line 484:
```suggestion
/// Removes identities that evaluate to zero and lookups with empty columns.
```
(I think one can argue that `X - X` evaluates to zero :)

chriseth: Maybe you can add `(including constraints of the form "X = X")`

pilopt/src/lib.rs
```diff
@@ -620,3 +640,61 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
         .collect();
     pil_file.remove_identities(&to_remove);
 }
+
+/// Identifies witness columns that are directly constrained to be equal to other witness columns
+/// through polynomial identities of the form "x = y".
```
chriseth, line 645: What does it return?

pilopt/src/lib.rs
```diff
@@ -620,3 +640,61 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+        .filter_map(|id| {
+            if let Identity::Polynomial(PolynomialIdentity { expression: e, .. }) = id {
+                equal_constrained(e)
+            } else {
+                None
+            }
+        })
+        .collect::<HashMap<PolyID, _>>();
+
+    pil_file.post_visit_expressions_in_identities_mut(&mut |e: &mut AlgebraicExpression<_>| {
```
chriseth, line 691: I'm wondering if this is enough. I think we also need to visit references in definitions. Can you add a test where you have `x = y` and use `provide_value` on both in a prover function?

chriseth: Ah is that why you commented out `set_hint` below? :)

gzanitti: Yes, but also because `remove_unreferenced_definitions` did not seem to take into account the variables inside `definitions`. I added a small case inside this function and set_hint now seems to works perfectly, but I'm sure that the way in which I identify these variables/columns is not 100% correct/sufficiently accurate.

gzanitti:
More on this: There is also the issue that some backends expects to have columns with specific names. If I'm not wrong, this should be the case of RISCV and the `many_chunks` tests.

Since there is an equality between `Z` & `Z_const`:
```
col witness Z_const;
main::Z = main::Z_const;
```
one of the columns will be replaced. But both cases, keeping Z or Z_const, give us the same problem: The other column cannot be find. Is there a way to identity this variables? Or should we use some kind of blackilst during optimization?

pilopt/src/lib.rs
```diff
@@ -166,6 +167,14 @@ fn collect_required_symbols<'a, T: FieldElement>(
             }
         });
     }
+    for (name, (sym, def)) in &pil_file.definitions {
```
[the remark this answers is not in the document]
gzanitti: Okay, I see. I thought that the variable on which the hint is applied should be preserved as well. It should be fixed then.

riscv-executor/src/lib.rs
```diff
@@ -1039,10 +1043,22 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
 
         self.proc.backup_reg_mem();
 
-        set_col!(X, get_col!(X_const));
-        set_col!(Y, get_col!(Y_const));
-        set_col!(Z, get_col!(Z_const));
-        set_col!(W, get_col!(W_const));
+        if self.proc.col_is_defined("X_const") {
```
pacheco, line 1046: Column name needs the namespace here `main::X_const`.
Still this is not quite right: `X_const` is being set from the "rom" (`p_X_const`), which is then being used to set `X` here.

riscv-executor/src/lib.rs
```diff
@@ -626,6 +626,10 @@ mod builder {
             }
         }
 
+        pub fn col_is_defined(&self, name: &str) -> bool {
```
pacheco, line 629: can you make this a `try_get_col` which return an Option instead? And checks that its in ExecMode::Trace? (see `get_col`)

riscv-executor/src/lib.rs
pacheco: But for the purposes of this PR just fixing the column names here is fine (witness is not being checked in tests for now)

riscv-executor/src/lib.rs
```diff
@@ -1039,10 +1051,22 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
 
         self.proc.backup_reg_mem();
 
-        set_col!(X, get_col!(X_const));
-        set_col!(Y, get_col!(Y_const));
-        set_col!(Z, get_col!(Z_const));
-        set_col!(W, get_col!(W_const));
+        if self.proc.try_get_col("main::X_const").is_some() {
```
pacheco, line 1054: you could use `if let Some()` here to avoid the call to get_col inside

chriseth:
This looks great now! One last annoying thing to test: Array elements. The issue is that we cannot remove them (because we would need to shorten the array by one element and update the indices, etc - it's very tricky).

Can you add a test of the form `let w: col[20]; w[4] = w[7];` and see what happens? Please also include some references to the (potentially) removed witness column both in other identities and in hints / prover functions.

chriseth:
I think `remove_constant_witness_columns` already avoids removing arrays, but I think it does not correctly avoid removing array elements (`.is_array()` only returns true on the first element of the array or on the array itself). The function `build_poly_id_to_definition_name_lookup` already creates a lookup table that could be extended to also supply the information if the PolyID is an array element or not.

pilopt/src/lib.rs
```diff
@@ -141,6 +141,26 @@ fn build_poly_id_to_definition_name_lookup(
     poly_id_to_definition_name
 }
 
+/// Builds a lookup-table that can be used to turn array elements
+/// (in form of their poly ids) into the names of the arrays.
+fn build_poly_id_to_array_elem_lookup(
```
chriseth, line 146: This is very similar to the function above. Could you please find a way to reduce code duplication? We could have a single lookup tables that also contains a boolean flag, for example.

ast/src/analyzed/mod.rs
```diff
@@ -759,6 +759,24 @@ impl Symbol {
         })
     }
 
+    /// Returns an iterator producing either an empty iterator (if it is not an array),
+    /// or all the elements of the array with their names in the form `array[index]`.
+    pub fn array_elements_only(&self) -> impl Iterator<Item = (String, PolyID)> + '_ {
```
chriseth, line 764: This function looks a bit too specific for a single use case to me. Instead of calling `array_elements_only`, you could use
```
symbol.is_array().then(|| symbol.array_elements()).flatten()
```

pilopt/src/lib.rs
```diff
@@ -120,31 +121,47 @@ fn remove_unreferenced_definitions<T: FieldElement>(pil_file: &mut Analyzed<T>)
     pil_file.remove_trait_impls(&impls_to_remove);
 }
 
-/// Builds a lookup-table that can be used to turn array elements
-/// (in form of their poly ids) into the names of the arrays.
+/// Builds a lookup-table that can be used to turn all symbols
+/// (including array elements) in the form of their poly ids, into the names of the symbols.
```
chriseth, line 125:
```suggestion
/// Builds a lookup-table that can be used to turn all poly ids into the names of the symbols that define them.
/// For array elements, this is the name of the array.
```

pilopt/src/lib.rs
```diff
@@ -120,31 +121,47 @@ fn remove_unreferenced_definitions<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
+/// The boolean flag indicates whether the symbol belongs to an array or not
```
chriseth, line 126: Would it makes sense to return `Option<usize>` instead and return the array index?

pilopt/src/lib.rs
```diff
@@ -120,31 +121,47 @@ fn remove_unreferenced_definitions<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
-            });
+            if symbol.is_array() {
+                symbol.array_elements().for_each(|(_, id)| {
+                    poly_id_to_info.insert(id, (name, true));
+                });
+            } else {
+                symbol.array_elements().for_each(|(_, id)| {
+                    poly_id_to_info.insert(id, (name, false));
+                });
+            }
```
chriseth, line 142: Why not
```suggestion
                symbol.array_elements().for_each(|(_, id)| {
                    poly_id_to_info.insert(id, (name, symbol.is_array()));
                });
```
?

pilopt/src/lib.rs
```diff
@@ -531,23 +547,44 @@ fn constrained_to_constant<T: FieldElement>(
...
+                    left,
+                    op: AlgebraicBinaryOperator::Sub,
+                    right,
+                }) => {
+                    if let (
+                        AlgebraicExpression::Reference(left),
+                        AlgebraicExpression::Reference(right),
+                    ) = (left.as_ref(), right.as_ref())
+                    {
+                        if left.is_witness() && right.is_witness() && left == right {
```
chriseth, line 577:
```suggestion
                        if lleft == right {
```

chriseth: oh and maybe require that both are "not next".

pilopt/src/lib.rs
```diff
@@ -671,3 +708,78 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+            left,
+            op: AlgebraicBinaryOperator::Sub,
+            right,
+        }) => match (left.as_ref(), right.as_ref()) {
+            (AlgebraicExpression::Reference(l), AlgebraicExpression::Reference(r)) => {
+                let is_valid = |x: &AlgebraicReference| {
+                    x.is_witness()
+                        && !x.next
+                        && poly_id_to_array_elem
+                            .get(&x.poly_id)
```
chriseth, line 730: Can this `get` fail?

pilopt/src/lib.rs
```diff
@@ -671,3 +708,78 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+        .iter()
+        .filter_map(|id| {
+            if let Identity::Polynomial(PolynomialIdentity { expression, .. }) = id {
+                equal_constrained(expression, &poly_id_to_array_elem)
+            } else {
+                None
+            }
+        })
+        .collect();
+
```
chriseth, line 763: I think you need to process `substitutions` a little just in case we get `b -> c` and `c -> a`, then we also need to modify `b -> c` to be `b -> a`.

gzanitti:
Replaced by #2224 , which can run bench tests.

PR #2087, opened 2024-11-13
Inlining ecalls when constant.

lvella:


riscv-syscalls/src/lib.rs
```diff
@@ -26,15 +36,15 @@ macro_rules! syscalls {
             }
         }
 
-        impl From<Syscall> for u32 {
+        impl From<Syscall> for u8 {
             fn from(syscall: Syscall) -> Self {
                 syscall as Self
             }
         }
```
lvella, line 43: This looks a bit dangerous. I think this should be removed, and if you need to convert number to enum, use the TryFrom.

riscv/src/large_field/runtime.rs
```diff
@@ -17,7 +17,7 @@ use crate::RuntimeLibs;
 #[derive(Clone)]
 pub struct Runtime {
     submachines: BTreeMap<String, SubMachine>,
-    syscalls: BTreeMap<Syscall, SyscallImpl>,
+    syscalls: BTreeMap<&'static str, SyscallImpl>,
```
lvella, line 20: This will change the order the "jump table" will be emitted, but it is still deterministic.

riscv/Cargo.toml
```diff
@@ -10,8 +10,12 @@ repository = { workspace = true }
 
 [features]
 default = []                                       # complex-tests is disabled by default
+
+# features below only affects tests
```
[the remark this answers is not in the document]
lvella: That it won't affect the built library, just the tests for the crate.

riscv/tests/riscv.rs
```diff
@@ -628,6 +690,7 @@ fn profiler_sanity_check() {
     assert!(!callgrind.unwrap().is_empty());
 }
 
+#[cfg(feature = "plonky3")]
```
chriseth, line 693: why does it need ploky3 now?

lvella: It already needed, but without this switch, build fails if plonky3 is not activated in the backend crate (and it isn't, by default).

lvella: These two changes in the features you commented is fixing unrelated issues I found while trying to debug the test I wrote inside the IDE. The building of the tests was failing when building just the crate, turns out because required features were not being enabled.

leonardoalt:
poseidon riscv test on main:
Execution trace length: 2807

on this branch:
Execution trace length: 2637

leonardoalt:
test failing

leonardoalt (approved):

leonardoalt (approved):

PR #2088, opened 2024-11-13
Use contiguous data for finalized rows.

chriseth:
Change FinalizableData so that finalized rows are stored as one contiguous array instead of an array of arrays. Furthermore, if the column IDs are a contiguous sequence, the row will only have as many field elements as there are columns.

I don't expect this PR to improve performance, but we can directly re-use this data structure for JIT-compiled executors (which need contiguous cell data, it was one of the main performance boosts). This means we can decide at runtime if we want to use a JIT-compiled executor or the interpreted one.

And this in turn allows us to delay the actual JIT-compiling until we get a request to the machine in the form of a bit field of known columns in the lookup. And once we know which columns are known, we can try to JIT-compile. If it fails, we store that this combination failed and will only use the interpreted one on that one in the future. This is especially useful for things like poseidon_gl, where we don't want to try all `2^13` combinations...

### DIFF
--- a/executor/src/witgen/data_structures/finalizable_data.rs
+++ b/executor/src/witgen/data_structures/finalizable_data.rs
@@ -205,50 +205,60 @@ impl<T: FieldElement> FinalizableData<T> {
     }
 
     pub fn truncate(&mut self, len: usize) {
-        if len <= self.pre_finalized_data.len() {
-            self.pre_finalized_data.truncate(len);
-            self.finalized_data.clear();
-            self.post_finalized_data.clear();
-        } else if len <= self.pre_finalized_data.len() + self.finalized_data.len() {
-            self.finalized_data
-                .truncate(len - self.pre_finalized_data.len());
-            self.post_finalized_data.clear();
+        if len == 0 {
+            self.clear();
         } else {
-            self.post_finalized_data
-                .truncate(len - self.pre_finalized_data.len() - self.finalized_data.len());
+            match self.location_of_row(len - 1) {
+                Location::PreFinalized(local) => {
+                    self.pre_finalized_data.truncate(local + 1);
+                    self.finalized_data.clear();
+                    self.post_finalized_data.clear();
+                }
+                Location::Finalized(local) => {
+                    self.finalized_data.truncate(local + 1);
+                    self.post_finalized_data.clear();
+                }
+                Location::PostFinalized(local) => {
+                    self.post_finalized_data.truncate(local + 1);
+                }
+            }
         }
     }
 
+    pub fn clear(&mut self) {
+        self.pre_finalized_data.clear();
+        self.finalized_data.clear();
+        self.post_finalized_data.clear();
+    }
+
     pub fn get_mut(&mut self, i: usize) -> Option<&mut Row<T>> {
-        if i < self.pre_finalized_data.len() {
-            Some(&mut self.pre_finalized_data[i])
-        } else if i < self.pre_finalized_data.len() + self.finalized_data.len() {
-            panic!("Row {i} already finalized.");
-        } else {
-            self.post_finalized_data
-                .get_mut(i - self.pre_finalized_data.len() - self.finalized_data.len())
+        match self.location_of_row(i) {
+            Location::PreFinalized(local) => self.pre_finalized_data.get_mut(local),
+            Location::Finalized(_) => panic!("Row {i} already finalized."),
+            Location::PostFinalized(local) => self.post_finalized_data.get_mut(local),
         }
     }
 
     pub fn last(&self) -> Option<&Row<T>> {
-        if !self.post_finalized_data.is_empty() {
-            self.post_finalized_data.last()
-        } else if !self.finalized_data.is_empty() {
-            panic!("Row already finalized");
-        } else {
-            self.pre_finalized_data.last()
+        match self.location_of_last_row()? {
+            Location::PreFinalized(local) => self.pre_finalized_data.get(local),
+            Location::Finalized(_) => panic!("Row already finalized"),
+            Location::PostFinalized(local) => self.post_finalized_data.get(local),
         }
     }
 
     pub fn mutable_row_pair(&mut self, i: usize) -> (&mut Row<T>, &mut Row<T>) {
-        let (before, after) = if i + 1 < self.pre_finalized_data.len() {
-            self.pre_finalized_data.split_at_mut(i + 1)
-        } else if i < self.pre_finalized_data.len() + self.finalized_data.len() {
-            panic!("Row {i} or {} (or both) already finalized.", i + 1);
-        } else {
-            self.post_finalized_data
-                .split_at_mut(i - self.pre_finalized_data.len() - self.finalized_data.len() + 1)
+        let (before, after) = match self.location_of_row(i) {
+            Location::PreFinalized(local) => {
+                if local + 1 >= self.pre_finalized_data.len() {
+                    panic!("Row {} already finalized.", i + 1);
+                }
+                self.pre_finalized_data.split_at_mut(local + 1)
+            }
+            Location::Finalized(_) => panic!("Row {i} already finalized."),
+            Location::PostFinalized(local) => self.post_finalized_data.split_at_mut(local + 1),
         };
+
         let current = before.last_mut().unwrap();
         let next = after.first_mut().unwrap();
         (current, next)
--- a/executor/src/witgen/data_structures/finalizable_data.rs
+++ b/executor/src/witgen/data_structures/finalizable_data.rs
@@ -179,12 +179,15 @@ impl<T: FieldElement> FinalizableData<T> {
     }
 
     pub fn extend(&mut self, other: Self) {
-        if other.finalized_data.is_empty() && other.post_finalized_data.is_empty() {
-            self.post_finalized_data.extend(other.pre_finalized_data);
-        } else if self.finalized_data.is_empty() && self.post_finalized_data.is_empty() {
+        if self.finalized_data.is_empty() {
+            self.pre_finalized_data
+                .extend(self.post_finalized_data.drain(..));
             self.pre_finalized_data.extend(other.pre_finalized_data);
             self.finalized_data = other.finalized_data;
             self.post_finalized_data = other.post_finalized_data;
+        } else if other.finalized_data.is_empty() {
+            self.post_finalized_data.extend(other.pre_finalized_data);
+            self.post_finalized_data.extend(other.post_finalized_data);
         } else if self.post_finalized_data.is_empty() && other.pre_finalized_data.is_empty() {
             self.finalized_data.data.extend(other.finalized_data.data);
             self.post_finalized_data = other.post_finalized_data;
--- a/executor/src/witgen/data_structures/finalizable_data.rs
+++ b/executor/src/witgen/data_structures/finalizable_data.rs
@@ -181,7 +181,7 @@ impl<T: FieldElement> FinalizableData<T> {
     pub fn extend(&mut self, other: Self) {
         if self.finalized_data.is_empty() {
             self.pre_finalized_data
-                .extend(self.post_finalized_data.drain(..));
+                .append(&mut self.post_finalized_data);
             self.pre_finalized_data.extend(other.pre_finalized_data);
             self.finalized_data = other.finalized_data;
             self.post_finalized_data = other.post_finalized_data;
--- a/executor/src/witgen/data_structures/finalizable_data.rs
+++ b/executor/src/witgen/data_structures/finalizable_data.rs
@@ -177,7 +177,8 @@ impl<T: FieldElement> FinalizableData<T> {
         }
     }
 
-    /// Removes the last row, even if it has been finalized.
+    /// Tries to remove the last row, even if it has been finalized.
+    /// Returns `false` if there are no rows.
     pub fn try_remove_last_row(&mut self) -> bool {
         if let Some(loc) = self.location_of_last_row() {
             let removed = match loc {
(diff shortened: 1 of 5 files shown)


executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -4,47 +4,118 @@ use std::{
...
+    /// Bit vector of known cells, stored in row-major order.
     known_cells: BitVec,
 }
 
-impl<T> FinalizedRow<T> {
-    pub fn new(values: Vec<T>, known_cells: BitVec) -> Self {
+impl<T: FieldElement> CompactData<T> {
+    /// Creates a new empty compact data storage. The column IDs have to be sorted.
+    fn new(column_ids: &[PolyID]) -> Self {
+        let first_column_id = column_ids.first().map_or(0, |id| id.id);
```
pacheco, line 29: maybe an assertion here that they are indeed sorted?

pacheco: the ids also need to be contiguous?

chriseth: No, they only need to be sorted.

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -58,84 +129,185 @@ impl<T: FieldElement> FinalizableData<T> {
...
+            self.finalized_data.try_remove_last_row();
+        } else {
+            self.pre_finalized_data.pop();
         }
     }
 
     pub fn extend(&mut self, other: Self) {
-        self.data.extend(other.data);
+        if self.finalized_data.is_empty() {
+            self.pre_finalized_data
```
pacheco, line 183: can you have pre and post rows with empty finalized in the middle?

chriseth: Usually not, but I did not want to rely on that assumption.

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -58,84 +129,185 @@ impl<T: FieldElement> FinalizableData<T> {
...
+            self.post_finalized_data = other.post_finalized_data;
+        } else if other.pre_finalized_data.is_empty() {
+            self.finalize_range(
+                (self.pre_finalized_data.len() + self.finalized_data.len())..self.len(),
+            );
+            self.finalized_data.data.extend(other.finalized_data.data);
+            self.post_finalized_data = other.post_finalized_data;
+        } else {
+            panic!(
+                "Cannot extend. Please try to ensure that `other` does not contain finalized rows."
```
pacheco, line 203: could this be an assert at the beginning of the function?

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -58,84 +129,185 @@ impl<T: FieldElement> FinalizableData<T> {
...
+            self.post_finalized_data
+                .remove(i - self.pre_finalized_data.len() - self.finalized_data.len())
         }
     }
 
     pub fn truncate(&mut self, len: usize) {
-        self.data.truncate(len);
+        if len <= self.pre_finalized_data.len() {
+            self.pre_finalized_data.truncate(len);
+            self.finalized_data.clear();
```
pacheco, line 225: i don't have the outer context: you can't remove finalized data, but you can truncate it out?

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -58,84 +129,185 @@ impl<T: FieldElement> FinalizableData<T> {
...
+                .for_each(|(i, row)| {
+                    if start + i < end {
+                        self.finalized_data.push(row)
+                    } else {
+                        self.post_finalized_data.push(row)
+                    }
+                });
+            end - start
+        } else {
+            // If we are asked to finalize the pre-finalized data, we just don't do it.
```
pacheco, line 295: why is that? this differs from how it behaved before no?

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -146,39 +318,41 @@ impl<T: FieldElement> FinalizableData<T> {
...
-        for i in 0..self.data.len() {
-            if self.finalize(i) {
-                counter += 1;
-            }
-        }
-        log::debug!("Needed to finalize {} / {} rows.", counter, self.data.len());
+        let counter = self.finalize_range(
+            (self.pre_finalized_data.len() + self.finalized_data.len())..self.len(),
+        );
+        log::debug!("Needed to finalize {} / {} rows.", counter, self.len());
```
pacheco, line 329: this is also different from before right? there was no concept of pre/post and everything was finalized here?

executor/src/witgen/data_structures/finalizable_data.rs
chriseth: I also don't fully understand, but the point is that `pop()` returns a `Row`, which is unfinalized data. So removing the finalized row is not a problem, but we would need to turn it into an unfinalized row again (which is also not a problem per se...)

executor/src/witgen/data_structures/finalizable_data.rs
chriseth: Yes it differs, but just not finalizing data we were asked to is not a correctness issue, it could just be a performance (memory requirement) issue.

chriseth: But note that the memory requirements of CompactData is vastly better than before: We used to store all columns in the whole system for every machine (even if they are not used in the machine), plus we stored a full vector "head" in each row.

executor/src/witgen/data_structures/finalizable_data.rs
chriseth: I even considered removing this call to `finalize`. The reason is that we copy out the data anyway, so it doesn't really matter if it has been finalized or not.

executor/src/witgen/data_structures/finalizable_data.rs
chriseth: Oh and it actually does work if other contain finalized rows. I can change the panic message if you like.

georgwiese (commented):
Hmm, I find the `FinalizedData` a bit hard to read with all the cases. It's also pretty repetitive.

I think something like this would help:
```rs
enum Location {
    PreFinalized(usize),
    Finalized(usize),
    PostFinalized(usize),
}


impl<T: FieldElement> FinalizableData<T> {
    fn get_location(&self, i: usize) -> Location {
        if i < self.pre_finalized_data.len() {
            Location::PreFinalized(i)
        } else if i < self.pre_finalized_data.len() + self.finalized_data.len() {
            Location::Finalized(i - self.pre_finalized_data.len())
        } else {
            Location::PostFinalized(i - self.pre_finalized_data.len() - self.finalized_data.len())
        }
    }

    fn get_last_location(&self) -> Location {
        self.get_location(self.len() - 1)
    }

    pub fn push(&mut self, row: Row<T>) {
        match self.get_last_location() {
            Location::PreFinalized(_) => {
                self.pre_finalized_data.push(row);
            }
            Location::Finalized(_) | Location::PostFinalized(_) => {
                self.post_finalized_data.push(row);
            }
        }
    }

    pub fn pop(&mut self) -> Option<Row<T>> {
        match self.get_last_location() {
            Location::PreFinalized(_) => self.pre_finalized_data.pop(),
            Location::Finalized(_) => {
                panic!("Row already finalized");
            }
            Location::PostFinalized(_) => self.post_finalized_data.pop(),
        }
    }
}

impl<T: FieldElement> Index<usize> for FinalizableData<T> {
    type Output = Row<T>;

    fn index(&self, index: usize) -> &Self::Output {
        match self.get_location(index) {
            Location::PreFinalized(i) => &self.pre_finalized_data[i],
            Location::Finalized(_) => panic!("Row {index} already finalized."),
            Location::PostFinalized(i) => &self.post_finalized_data[i],
        }
    }
}
```

And so on. Because `match` arms need to be exhaustive and the conditions & index arithmetic would only be defined in one place, I'd find it easier to verify the implementation.

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -4,47 +4,119 @@ use std::{
...
+        self.data.clear();
+        self.known_cells.clear();
+    }
+
+    /// Appends a non-finalized row to the data, turning it into a finalized row.
+    #[inline]
+    fn push(&mut self, row: Row<T>) {
+        self.data.reserve(self.data.len() + self.column_count);
+        self.known_cells
+            .reserve(self.known_cells.len() + self.column_count);
```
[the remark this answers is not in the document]
chriseth: Oh interesting, I didn't actually know that, I always thought it's the same as [vector.reserve](https://cplusplus.com/reference/vector/vector/reserve/) in C++

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -4,47 +4,119 @@ use std::{
...
+        self.known_cells.truncate(len * self.column_count);
+    }
+
+    fn clear(&mut self) {
+        self.data.clear();
+        self.known_cells.clear();
+    }
+
+    /// Appends a non-finalized row to the data, turning it into a finalized row.
+    #[inline]
```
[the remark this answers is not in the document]
chriseth: Yeah not sure, it's more of a gut feeling that I think that pulling the code into the caller allows more optimization. But maybe I should remove all of them again until we have some measurements.

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -4,47 +4,119 @@ use std::{
...
+        } else {
+            self.data.truncate(self.data.len() - self.column_count);
+            self.known_cells
+                .truncate(self.known_cells.len() - self.column_count);
+            true
+        }
+    }
+
+    #[inline]
+    fn get(&self, row: usize, col: u64) -> (T, bool) {
```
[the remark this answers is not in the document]
chriseth: I really would like to avoid any complication in the interface here. If we pass a PolID it suggests that we can also query for fixed columns which is just not true.

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -58,84 +130,185 @@ impl<T: FieldElement> FinalizableData<T> {
...
+        if !self.post_finalized_data.is_empty() {
+            self.post_finalized_data.pop();
+        } else if !self.finalized_data.is_empty() {
+            self.finalized_data.try_remove_last_row();
+        } else {
+            self.pre_finalized_data.pop();
         }
     }
 
     pub fn extend(&mut self, other: Self) {
```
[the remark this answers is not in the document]
chriseth: I thought I tried that, but maybe it was `self.finalized_data.is_empty()`. I'l try!

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -58,84 +130,185 @@ impl<T: FieldElement> FinalizableData<T> {
...
+            self.finalized_data
+                .truncate(len - self.pre_finalized_data.len());
+            self.post_finalized_data.clear();
+        } else {
+            self.post_finalized_data
+                .truncate(len - self.pre_finalized_data.len() - self.finalized_data.len());
+        }
     }
 
     pub fn get_mut(&mut self, i: usize) -> Option<&mut Row<T>> {
```
chriseth, line 237: Why does this return an Option? It's always `Some`, right?

executor/src/witgen/data_structures/finalizable_data.rs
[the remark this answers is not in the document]
chriseth: It worked!

executor/src/witgen/data_structures/finalizable_data.rs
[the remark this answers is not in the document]
georgwiese: Hmm, but now nothing prevents a caller from passing `some_fixed_poly_id.id`. If we received a `PolyID`, we could add an assertion that it is for a committed polynomial.

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -58,84 +124,197 @@ impl<T: FieldElement> FinalizableData<T> {
...
+            Location::Finalized(_) => panic!("Row already finalized"),
+            Location::PostFinalized(_) => self.post_finalized_data.pop(),
         }
     }
 
-    pub fn extend(&mut self, other: Self) {
-        self.data.extend(other.data);
+    /// Tries to remove the last row, even if it has been finalized.
+    /// Returns `false` if there are no rows.
+    pub fn try_remove_last_row(&mut self) -> bool {
```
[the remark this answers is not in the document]
chriseth: I made the change, let's see if I need to add it again in the next PR :)

executor/src/witgen/data_structures/finalizable_data.rs
```diff
@@ -150,120 +144,132 @@ impl<T: FieldElement> FinalizableData<T> {
...
+        match self.location_of_last_row() {
+            None | Some(Location::PreFinalized(_)) => {
+                self.pre_finalized_data.extend(other.pre_finalized_data);
+                self.finalized_data = other.finalized_data;
+                self.post_finalized_data = other.post_finalized_data;
+            }
+            Some(Location::Finalized(_)) | Some(Location::PostFinalized(_)) => {
+                assert!(other.finalized_data.is_empty() && other.post_finalized_data.is_empty());
+                self.post_finalized_data.extend(other.pre_finalized_data);
+            }
```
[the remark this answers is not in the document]
chriseth: That didn't work, the assertion fails.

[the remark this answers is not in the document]
georgwiese: Oh 😢

georgwiese (approved):
LGTM! 🚀

executor/src/witgen/generator.rs
```diff
@@ -72,7 +72,7 @@ impl<'a, T: FieldElement> Machine<'a, T> for Generator<'a, T> {
             log::trace!("End processing VM '{}' (successfully)", self.name());
             // Remove the last row of the previous block, if it exists,
             // as it is the first row of the current block.
-            self.data.try_remove_last_row();
+            self.data.pop().unwrap();
```
georgwiese, line 75: The `unwrap()` is wrong (think that's why the test fails).

georgwiese (approved):

PR #2090, opened 2024-11-14
lower default max degree in riscv

leonardoalt:


riscv-executor/src/lib.rs
```diff
@@ -724,12 +724,6 @@ mod builder {
 
             let main_degree = {
                 let range = namespace_degree_range(pil, "main");
-                assert!(
```
leonardoalt, line 727: These assertions were too restrictive

leonardoalt: and caused many_chunks to fail

chriseth, line 727: Why can we remove those?

pacheco (approved):

riscv-executor/src/lib.rs
pacheco: for the continuations "dry run" it needs to run for the full execution (which will have a higher degree)

PR #2093, opened 2024-11-14
Remove monolithic inference in the linker

Schaeff:
- Replace the rule which sets all degrees to that of main if it is static and use a CLI flag instead: `--degrees-mode`
- Adjust tests to run mostly on vadcop
- Tweak tests to reduce wasted rows


pipeline/src/test_util.rs
```diff
@@ -74,13 +73,14 @@ pub fn regular_test(file_name: &str, inputs: &[i32]) {
 pub fn regular_test_without_small_field(file_name: &str, inputs: &[i32]) {
     let inputs_gl = inputs.iter().map(|x| GoldilocksField::from(*x)).collect();
     let pipeline_gl = make_prepared_pipeline(file_name, inputs_gl, vec![]);
+
     test_mock_backend(pipeline_gl.clone());
-    test_pilcom(pipeline_gl.clone());
-    gen_estark_proof(pipeline_gl);
+    run_pilcom_with_backend_variant(pipeline_gl.clone(), BackendVariant::Composite).unwrap();
```
leonardoalt, line 78: I thought pilcom only works with Monolithic?

leonardoalt (commented):
Nice!

pipeline/src/test_util.rs
Schaeff: My bad, I meant the non-composite pilcom backend. The non-composite plonky3 backend does support vadcop for example. The composite pilcom backend is of limited value since the links get removed when splitting the pil.

PR #2102, opened 2024-11-16
optimizer: deduplicate fixed columns

Schaeff:
With bus lookups, we insert `is_first` once per lookup.
Only keep a single column per syntactic value.
Depends on #2103


pilopt/src/lib.rs
```diff
@@ -215,6 +217,39 @@ fn constant_value(function: &FunctionValueDefinition) -> Option<BigUint> {
     }
 }
 
+/// Deduplicate fixed columns which share the same value.
+/// This uses the `Display` implementation of the function value, so `|i| i` is different from `|j| j`
```
chriseth, line 221: Is it that much more expensive to check for syntactic equality? I think Display is very expensive.

Schaeff: I can have a go at it, seemed annoying because of sourcemaps.

pilopt/src/lib.rs
```diff
@@ -215,6 +217,44 @@ fn constant_value(function: &FunctionValueDefinition) -> Option<BigUint> {
     }
 }
 
+/// Deduplicate fixed columns of the same namespace which share the same value.
+/// This compares the function values, so `|i| i` is different from `|j| j`
```
chriseth, line 221: We could actually quite easily support `|i| i` being equal to `|j| j`, because all references are already the same. The only thing we need to change is the comparison and hash function for `Pattern` to ignore the variable name.

pilopt/src/lib.rs
```diff
@@ -215,6 +217,44 @@ fn constant_value(function: &FunctionValueDefinition) -> Option<BigUint> {
...
+/// This is enough for use cases where exactly the same function is inserted many times
+/// This only replaces the references inside expressions and does not clean up the now unreachable fixed column definitions
+fn deduplicate_fixed_columns<T: FieldElement>(pil_file: &mut Analyzed<T>) {
+    // build a map of `poly_id` to the `(name, poly_id)` they can be replaced by
+    let replacement_map: BTreeMap<PolyID, (String, PolyID)> = pil_file
+        .constant_polys_in_source_order()
+        // group symbols by common namespace and function value
+        .into_group_map_by(|(symbol, value)| {
+            (
+                symbol.absolute_name.split("::").next().unwrap(),
```
chriseth, line 231: Use `SymbolPath::name` or at least introduce a helper function somewhere?

chriseth: Ah sorry, this is the namespace, not the name.

pilopt/src/lib.rs
Schaeff: Oh great, I was hoping for something like this

pilopt/src/lib.rs
```diff
@@ -215,6 +217,44 @@ fn constant_value(function: &FunctionValueDefinition) -> Option<BigUint> {
...
+            group[1..].iter().flat_map(|from| {
+                from.0
+                    .array_elements()
+                    .map(|(_, from_id)| from_id)
+                    .zip_eq(group[0].0.array_elements())
+            })
+        })
+        .collect();
+
+    // substitute all occurences in expressions.
```
chriseth, line 247: I think you also have to substitute in definitions. This has some overlap with #2086.

pilopt/src/lib.rs
```diff
@@ -215,6 +217,44 @@ fn constant_value(function: &FunctionValueDefinition) -> Option<BigUint> {
...
+/// This only replaces the references inside expressions and does not clean up the now unreachable fixed column definitions
+fn deduplicate_fixed_columns<T: FieldElement>(pil_file: &mut Analyzed<T>) {
+    // build a map of `poly_id` to the `(name, poly_id)` they can be replaced by
+    let replacement_map: BTreeMap<PolyID, (String, PolyID)> = pil_file
+        .constant_polys_in_source_order()
+        // group symbols by common namespace and function value
+        .into_group_map_by(|(symbol, value)| {
+            (
+                symbol.absolute_name.split("::").next().unwrap(),
+                value.as_ref().unwrap(),
```
chriseth, line 232: I think the parser also allows fixed columns without value, but I don't know how far we advance for those.

pilopt/src/lib.rs
Schaeff: So I'm looking into this, I don't think we should make all `Pattern::Variable` equal, right?
One level up I was looking at
```rust
#[derive(
    Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, JsonSchema, Hash,
)]
pub struct LambdaExpression<E = Expression<NamespacedPolynomialReference>> {
    pub kind: FunctionKind,
    pub params: Vec<Pattern>,
    pub body: Box<E>,
    /// Type of the parameters, filled in during type inference.
    pub param_types: Vec<Type>,
}
```
Here I would expect `body` to refer to the parameters by index, in which case it would be the same no matter what the parameters are called. Then we could ignore them in comparisons.
Unless I'm missing obvious, this seems like it should be a different PR?

pilopt/src/lib.rs
Schaeff: It does not seem like this is something we expect, so it would be fine to fail if it does not fail before?

pilopt/src/lib.rs
Schaeff: added function

pilopt/src/lib.rs
chriseth: I mean the question is if we want to do that in the "default" comparison function. Is there an easy way to provide a different comparison - probably quite cumbersome to do that recursively, right?

pilopt/src/lib.rs
chriseth: Maybe we should remove that option from the parser.

Schaeff: https://github.com/powdr-labs/powdr/pull/2126

pilopt/src/lib.rs
Schaeff: Opened https://github.com/powdr-labs/powdr/issues/2128 for this

chriseth (approved):

PR #2103, opened 2024-11-17
Make all `SourceRef` equal

Schaeff:
When comparing pil expressions, the source information gets in the way. Experiment with ignoring it when comparing.
This respects the requirements [here](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html)


chriseth:
Looks good, I'm just wondering if we would assume them not be equal at some point. Can't think of a situation though. 

Could you maybe document that in `SourceRef` and maybe also in `Expression`?

chriseth (approved):

PR #2108, opened 2024-11-18
keccak with memory for goldilocks

leonardoalt:


qwang98:
Works up to witgen. Not sure what to do with the CI error happening for PIL verifier in the form of ":0:  permutation not found w=24 values: 1:0,0,32,0". @leonardoalt

leonardoalt:
@qwang98 the error seems related to the keccak test, so I would assume the error is legit and there's something wrong with the constraints

leonardoalt:
@qwang98 after https://github.com/powdr-labs/powdr/pull/2121 is merged you can also use it instead of pilcom to check why the constraints fail

PR #2111, opened 2024-11-18
lower degree of arith machine

leonardoalt:


std/machines/large_field/arith.asm
```diff
@@ -324,20 +324,25 @@ machine Arith with
...
     *******/
-    
-    col eq0_sum = sum(32, |i| eq0(i) * CLK32[i]);
-    col eq1_sum = sum(32, |i| eq1(i) * CLK32[i]);
-    col eq2_sum = sum(32, |i| eq2(i) * CLK32[i]);
-    col eq3_sum = sum(32, |i| eq3(i) * CLK32[i]);
-    col eq4_sum = sum(32, |i| eq4(i) * CLK32[i]);
+
+    col witness eq0_sum;
+    eq0_sum = sum(32, |i| eq0(i) * CLK32[i]);
```
chriseth, line 337: I'm wondering if "evaluating" `eq0` somehow fails if `selEq[0]` is not one. As long as `eq0_sum` is not a witness column, this is irrelevant (because it is unused) but as soon as it becomes a witness column, witgen needs to assign some value. And it could be that it's just stuck there or assigns the wrong one.

PR #2112, opened 2024-11-19
Add support for constant column in STWO backend

ShuangWu121:
# Add Support for Constant Columns in Stwo Backend

Stwo's recent development introduced new APIs to support constant/pre-processed columns. However, the `dev` branch of Stwo is still using an older nightly toolchain, which is incompatible with Powdr.

Currently, Stwo has two open PRs [PR1](https://github.com/starkware-libs/stwo/pull/847), [PR2](https://github.com/starkware-libs/stwo/pull/871) aimed at updating the toolchain to nightly `11-06`. Previously, our Stwo backend dependency relied on one of these PRs' branches. However, this branch is no longer updated with the latest commits from their `dev` branch. These new commits are required to support constant columns.

### Temporary Solution
To address this issue temporarily:
- I moved Stwo's dependency to my fork of Stwo, where the branch is updated with both the latest `dev` branch commits and the newer toolchain.

---

### Tasks in this PR
1. **Add APIs to support constant columns**:
2. **Update test cases**:
   - Modify and add test cases to validate constant column support.

---


backend/src/stwo/circuit_builder.rs
```diff
@@ -80,14 +92,29 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
...
+                )
+            })
+            .collect();
+        let constant_eval: BTreeMap<PolyID, <E as EvalAtRow>::F> = self
+            .constant_columns
+            .keys()
+            .map(|poly_id| {
+                (
+                    *poly_id,
+                    eval.get_preprocessed_column(PreprocessedColumn::Plonk(3)),
```
Schaeff, line 115: Why 3?

backend/src/stwo/circuit_builder.rs
```diff
@@ -145,7 +174,7 @@ where
                     true => witness_eval[&poly_id][1].clone(),
                 },
                 PolynomialType::Constant => {
-                    unimplemented!("Constant polynomials are not supported in stwo yet")
+                    constant_eval[&poly_id].clone()
```
Schaeff, line 177: Similar to committed polynomials, we want to be able to access the next row depending on `r.next`

backend/src/stwo/circuit_builder.rs
```diff
@@ -80,14 +92,29 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
...
         );
 
         let witness_eval: BTreeMap<PolyID, [<E as EvalAtRow>::F; 2]> = self
             .witness_columns
             .keys()
-            .map(|poly_id| (*poly_id, eval.next_interaction_mask(0, [0, 1])))
+            .map(|poly_id| {
+                (
+                    *poly_id,
+                    eval.next_interaction_mask(ORIGINAL_TRACE_IDX, [0, 1]),
```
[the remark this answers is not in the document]
ShuangWu121: not yet, I will keep looking at their commitments maybe there will be, get_preprocessed_column is added around 2 weeks ago

backend/src/stwo/circuit_builder.rs
ShuangWu121: It can be random number, stwo uses this number to mark how many preprocessed columns are added in [plonk](https://github.com/starkware-libs/stwo/blob/77de4d79b08dd89600994e4dc3dc06f77ddce25b/crates/prover/src/examples/plonk/mod.rs#L52) example. 
[here](https://github.com/starkware-libs/stwo/blob/77de4d79b08dd89600994e4dc3dc06f77ddce25b/crates/prover/src/constraint_framework/mod.rs#L80) shows the parameter PreprocessedColumn is not used actually 
and [here](https://github.com/starkware-libs/stwo/blob/77de4d79b08dd89600994e4dc3dc06f77ddce25b/crates/prover/src/constraint_framework/preprocessed_columns.rs#L11) is the definition of PreprocessedColumn, no declaration of what the number means.

backend/src/stwo/circuit_builder.rs
ShuangWu121: currently stwo only allows user to access pre-processed columns via get_preprocess_colums defined like this 
```
fn get_preprocessed_column(&mut self, _column: PreprocessedColumn) -> Self::F {
        let [mask_item] = self.next_interaction_mask(PREPROCESSED_TRACE_IDX, [0]);
        mask_item
    }
```
it is not able to pass offset, and if directly using next_interaction_mask(PREPROCESSED_TRACE_IDX, [0,1]) to get next reference for constant, it gives error "preprocessed columns can only be accessed by get_preprocessed_column"

Schaeff: Uff, that's unfortunate. Let's focus on same row identities everywhere for now then.

backend/src/stwo/prover.rs
```diff
@@ -46,16 +52,111 @@ where
...
+                                let bit_reversed_fixed_columns: BTreeMap<
+                                    DegreeType,
+                                    Vec<(String, Vec<F>)>,
+                                > = fixed_columns
+                                    .iter()
+                                    .map(|(key, vec)| {
+                                        let transformed_vec: Vec<(String, Vec<F>)> = vec
+                                            .iter()
+                                            .map(|(name, slice)| {
+                                                let mut values: Vec<F> = slice.to_vec(); // Clone the slice into a Vec
```
Schaeff, line 108: Cloning is not ideal here, these can be quite large. A better approach would be to introduce a type, something like

```rust
struct BitReversedColumn<'a> {
   column: &'a [F]
}
```

and implement `IntoIterator`, returning an iterator which goes through `column` in bit-reversed order.

However it seems like this is only needed for next references, which I'm not sure we should support in this PR?

backend/src/stwo/prover.rs
```diff
@@ -46,16 +52,111 @@ where
...
+                                let mut tree_builder = commitment_scheme.tree_builder();
+                                tree_builder.extend_evals(constant_trace.clone());
+                                tree_builder.commit(prover_channel);
+                                let trees = commitment_scheme.trees;
+
+                                (
+                                    size as usize,
+                                    TableProvingKey {
+                                        trees,
+                                        prover_channel: prover_channel.clone(),
```
Schaeff, line 144: Why do we need to save the prover channel here? If this is the Fiat Shamir transcript, isn't it determined by starting with an empty transcript and observing the commits? Or is this done here to save prover time?

Schaeff (commented):
Good but it seems it can be simplified

backend/src/stwo/circuit_builder.rs
```diff
@@ -80,14 +92,30 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
...
```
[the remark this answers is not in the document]
ShuangWu121: After making the witness to be folded bit reversed order, the next reference works, tests cases for next reference can pass. 
in stwo there is an[ assertion function](https://github.com/starkware-libs/stwo/blob/4d64300dc178d10d2eb7e9c43ac69cb73a634bb6/crates/prover/src/constraint_framework/assert.rs#L73), it is used for sanity check at the end of the proof, it evaluates constraints on trace directly AFAIK, I use this function to check why constraint fails, and figure out it reads the witness in a folded bit reversed order. Do a folded bit reverse order of witness can make this assertion pass. 
I can explain why witness can be in folded bit reversed order in theory, but I cannot explain it fully from the code yet, apart from using the assertion function to show directly.

backend/src/stwo/circuit_builder.rs
```diff
@@ -80,14 +92,30 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
...
+            })
+            .collect();
+        let constant_eval: BTreeMap<PolyID, <E as EvalAtRow>::F> = self
+            .constant_columns
+            .keys()
+            .map(|poly_id| {
+                (
+                    *poly_id,
+                    // PreprocessedColumn::Plonk(0) is unused argument in get_preprocessed_column,0 has no meaning
+                    eval.get_preprocessed_column(PreprocessedColumn::Plonk(0)),
```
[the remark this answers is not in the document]

backend/src/stwo/prover.rs
```diff
@@ -46,16 +52,111 @@ where
...
+                                    CommitmentSchemeProver::<'_, B, MC>::new(config, &twiddles);
+
+                                let domain = CanonicCoset::new(
+                                    fixed_columns
+                                        .keys()
+                                        .next()
+                                        .map(|&first_key| first_key.ilog2())
+                                        .unwrap_or(0),
+                                )
+                                .circle_domain();
```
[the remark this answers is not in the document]
ShuangWu121: backendFactory create function in mod.rs has
`if pil.degrees().len() > 1 {
            return Err(Error::NoVariableDegreeAvailable);
        }`
single size is checked here. 
unwarp_or() changed to unwarp()

backend/src/stwo/prover.rs
```diff
@@ -68,11 +169,47 @@ where
...
+        let (trees, mut prover_channel) = self
+            .proving_key
+            .borrow_mut()
+            .as_mut()
+            .and_then(|stark_proving_key| stark_proving_key.preprocessed.values_mut().next())
+            .and_then(|table_collection| table_collection.values_mut().next())
+            .map(|table_proving_key| {
+                (
+                    std::mem::take(&mut table_proving_key.trees),
+                    std::mem::take(&mut table_proving_key.prover_channel),
```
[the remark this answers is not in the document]
ShuangWu121: `RefCell` and` std::mem::take ` are removed. Proving_key now stores only the constant_trace evaluations in circle domain. 
I tried to store commitment_scheme in proving_key, it is defined as: 
`pub struct CommitmentSchemeProver<'a, B: BackendForChannel<MC>, MC: MerkleChannel> {
    pub trees: TreeVec<CommitmentTreeProver<B, MC>>,
    pub config: PcsConfig,
    twiddles: &'a TwiddleTree<B>,
}`
the field twiddles needs to specify 'a, it propagate to all the functions calling it, I couldn't make it work. I then store the field trees in proving_key only, it then has the problem of ownership, it doesn't implement clone trait. 
so now I store the constant_trace evaluations in proving_key, and do the commit in prove function.  
constant_trace still needs to be cloned to commit due to ownership, I didn't manage to avoid it. Is there any other way to avoid this clone?

backend/src/stwo/prover.rs
```diff
@@ -115,21 +252,13 @@ where
         );
 
         // Retrieve the expected column sizes in each commitment interaction, from the AIR.
-        // TODO: When constant columns are supported, there will be more than one sizes and proof.commitments
-        // size[0] is for constant columns, size[1] is for witness columns, size[2] is for lookup columns
-        // pass size[1] for witness columns now is not doable due to this branch is outdated for the new feature of constant columns
-        // it will throw errors.
```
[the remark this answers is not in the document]
ShuangWu121: Done

backend/src/stwo/prover.rs
```diff
@@ -1,95 +1,227 @@
...
+        let prover_channel = &mut <MC as MerkleChannel>::C::default();
+        let mut commitment_scheme =
+            CommitmentSchemeProver::<'_, B, MC>::new(config, twiddles_map.iter().next().unwrap().1);
 
-        // Setup protocol.
-        let mut prover_channel = <MC as MerkleChannel>::C::default();
-        let commitment_scheme = &mut CommitmentSchemeProver::<B, MC>::new(config, &twiddles);
+        let mut tree_builder = commitment_scheme.tree_builder();
+
+        // only the frist one is used, machines with varying sizes are not supported yet, and it is checked in backendfactory create function.
```
Schaeff, line 186:
```suggestion
        // only the first one is used, machines with varying sizes are not supported yet, and it is checked in backendfactory create function.
```

backend/src/stwo/proof.rs
```diff
@@ -0,0 +1,22 @@
...
+use stwo_prover::core::channel::MerkleChannel;
+use stwo_prover::core::fields::m31::BaseField;
+use stwo_prover::core::poly::circle::CircleEvaluation;
+use stwo_prover::core::poly::BitReversedOrder;
+use stwo_prover::core::ColumnVec;
+
+/// For each possible size, the commitment and prover data
+pub type TableProvingKeyCollection<B, MC> = BTreeMap<usize, TableProvingKey<B, MC>>;
+
+pub struct TableProvingKey<B: BackendForChannel<MC>, MC: MerkleChannel> {
```
Schaeff, line 12: One thing that is missing in this PR is serializing the proving key and setting it in the backend. Now `setup` is run when creating the backend. Is this something we'd do in a separate PR?

backend/src/stwo/proof.rs
```diff
@@ -0,0 +1,22 @@
...
+    pub constant_trace_circle_domain: ColumnVec<CircleEvaluation<B, BaseField, BitReversedOrder>>,
+    pub _marker: std::marker::PhantomData<MC>,
+}
+
+pub struct StarkProvingKey<B: BackendForChannel<MC>, MC: MerkleChannel> {
+    // for each table, the preprocessed data
+    pub preprocessed: Option<BTreeMap<String, TableProvingKeyCollection<B, MC>>>,
+}
+
+unsafe impl<B: BackendForChannel<MC>, MC: MerkleChannel> Send for TableProvingKey<B, MC> {}
```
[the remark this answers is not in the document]
Schaeff: I suspect it should rather be a `Send` bound on `B` wherever we use this?

backend/src/stwo/prover.rs
```diff
@@ -1,95 +1,227 @@
...
     _verifying_key: Option<()>,
     _channel_marker: PhantomData<C>,
-    _backend_marker: PhantomData<B>,
-    _merkle_channel_marker: PhantomData<MC>,
 }
 
 impl<'a, F: FieldElement, B, MC, C> StwoProver<F, B, MC, C>
 where
-    B: Backend + Send + BackendForChannel<MC>, // Ensure B implements BackendForChannel<MC>
+    B: Backend + Send + BackendForChannel<MC>,
```
Schaeff, line 53: Right, we have a `Send` bound here, so `TableProvingKey<B, MC>` should be automatically `Send` as well. Maybe there's a bound missing somewhere?

Schaeff: Reference: https://doc.rust-lang.org/nomicon/send-and-sync.html

backend/src/stwo/prover.rs
```diff
@@ -1,95 +1,227 @@
...
+                                                        ),
+                                                        values.len().ilog2(),
+                                                    ),
+                                                    v.try_into_i32().unwrap().into(),
+                                                );
+                                            });
+                                            CircleEvaluation::new(domain, column)
+                                        })
+                                    })
+                                    .collect();
```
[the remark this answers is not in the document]
Schaeff: generate_stwo_trace

backend/src/stwo/circuit_builder.rs
```diff
@@ -144,9 +185,10 @@ where
                     false => witness_eval[&poly_id][0].clone(),
                     true => witness_eval[&poly_id][1].clone(),
                 },
-                PolynomialType::Constant => {
-                    unimplemented!("Constant polynomials are not supported in stwo yet")
-                }
+                PolynomialType::Constant => match r.next {
+                    false => constant_eval[&poly_id].clone(),
+                    true => panic!("Next on constant polynomials is not supported"),
```
Schaeff, line 190: What is the plan to support this? Commit to the shifted column if it is used?

backend/src/stwo/prover.rs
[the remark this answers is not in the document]
Schaeff: I see. Let's keep it like this for now.

backend/src/stwo/proof.rs
ShuangWu121: Done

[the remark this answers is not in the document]
ShuangWu121: they are not needed, removed now

backend/src/stwo/proof.rs
```diff
@@ -0,0 +1,22 @@
...
+use stwo_prover::core::poly::circle::CircleEvaluation;
+use stwo_prover::core::poly::BitReversedOrder;
+use stwo_prover::core::ColumnVec;
+
+/// For each possible size, the commitment and prover data
+pub type TableProvingKeyCollection<B, MC> = BTreeMap<usize, TableProvingKey<B, MC>>;
+
+pub struct TableProvingKey<B: BackendForChannel<MC>, MC: MerkleChannel> {
+    pub constant_trace_circle_domain: ColumnVec<CircleEvaluation<B, BaseField, BitReversedOrder>>,
+    pub _marker: std::marker::PhantomData<MC>,
```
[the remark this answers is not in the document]
ShuangWu121: no need. removed

backend/src/stwo/prover.rs
```diff
@@ -1,95 +1,227 @@
...
     }
+    pub fn setup(&mut self) {
+        // machines with varying sizes are not supported yet, and it is checked in backendfactory create function.
+        //TODO: support machines with varying sizes
+        let preprocessed: BTreeMap<String, TableProvingKeyCollection<B, MC>> = self
+            .split
+            .iter()
+            .filter_map(|(namespace, pil)| {
+                // if we have neither fixed columns nor publics, we don't need to commit to anything
+                if pil.constant_count() + pil.publics_count() == 0 {
```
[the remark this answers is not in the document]
ShuangWu121: Done.

backend/src/stwo/prover.rs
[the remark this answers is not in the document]
ShuangWu121: generate_stwo_trace now changed to gen_stwo_circle_column, it works with slice

backend/src/stwo/proof.rs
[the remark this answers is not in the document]
ShuangWu121: no unsafe anymore.

backend/src/stwo/prover.rs
```diff
@@ -1,95 +1,227 @@
...
+                // Precompute twiddles for all sizes in the PIL
+                pil.committed_polys_in_source_order()
+                    .flat_map(|(s, _)| {
+                        s.degree.iter().flat_map(|range| {
+                            let min = range.min;
+                            let max = range.max;
+
+                            // Iterate over powers of 2 from min to max
+                            (min..=max)
+                                .filter(|&size| size.is_power_of_two()) // Only take powers of 2
```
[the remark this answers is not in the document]
ShuangWu121: Done.

backend/src/stwo/prover.rs
```diff
@@ -1,95 +1,315 @@
...
+                    CanonicCoset::new(size.ilog2()).circle_domain(),
+                )
+            })
+            .collect();
+
+        let preprocessed: BTreeMap<String, TableProvingKeyCollection<B>> = self
+            .split
+            .iter()
+            .filter_map(|(namespace, pil)| {
+                // if we no fixed columns, we don't need to commit to anything, publics is not supported yet.
```
Schaeff, line 133:
```suggestion
                // if we have no fixed columns, we don't need to commit to anything.
```

backend/src/stwo/prover.rs
```diff
@@ -1,95 +1,315 @@
...
+                if pil.constant_count() == 0 {
+                    None
+                } else {
+                    let fixed_columns = machine_fixed_columns(&self.fixed, pil);
+
+                    Some((
+                        namespace.to_string(),
+                        //why here it is committed_polys_in_source_order() instead of constant polys?
```
Schaeff, line 141: Good question actually, I think it does not matter because there is a runtime invariant that all columns inside a namespace have the same degree range, and that we always have at least one witness column. So we find the first witness column and use its degree range, we could do the same with the first constant column.

If this makes sense let's remove this line

backend/src/stwo/prover.rs
```diff
@@ -1,95 +1,315 @@
...
+            .map(|(_name, values)| {
+                gen_stwo_circle_column::<F, B, M31>(
+                    *domain_map.get(&(values.len().ilog2() as usize)).unwrap(),
+                    values,
+                )
+            })
+            .collect();
 
-        let trace = gen_stwo_circuit_trace::<F, B, M31>(witness);
+        //extend the witness trace with the constant polys that have next reference constraint
```
Schaeff, line 271: I have a doubt about this: what prevents the prover from setting these to arbitrary values? What I was rather expecting, would be to commit at setup phase to the shifted fixed columns, with an optimization being to only commit to what is being accessed in the constraints.

So basically having the proving key contain the non shifted and shifted fixed columns, instead of only the non-shifted fixed columns now.

backend/src/stwo/proof.rs
```diff
@@ -0,0 +1,106 @@
...
+                                            values.iter().enumerate().for_each(|(i, v)| {
+                                                column.set(i, *v);
+                                            });
+                                            CircleEvaluation::<B, BaseField, BitReversedOrder>::new(
+                                                CanonicCoset::new(size as u32).circle_domain(),
+                                                column,
+                                            )
+                                      })
+                                      .collect::<ColumnVec<_>>(),
+                                    }
```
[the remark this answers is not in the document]
ShuangWu121: introduced SerializableTableProvingKeyCollection, circle domain calculation is reused there

backend/src/stwo/prover.rs
ShuangWu121: got it. comment removed

backend/src/stwo/circuit_builder.rs
```diff
@@ -80,14 +106,46 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
...
+            .keys()
+            .enumerate()
+            .map(|(i, poly_id)| {
+                (
+                    *poly_id,
+                    eval.get_preprocessed_column(PreprocessedColumn::Plonk(
+                        i + constant_eval.len(),
+                    )),
+                )
+            })
```
Schaeff, line 148: Let's simplify this to have a single map like in the witness case

backend/src/stwo/circuit_builder.rs
```diff
@@ -96,7 +154,12 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
         {
             match id {
                 Identity::Polynomial(identity) => {
-                    let expr = to_stwo_expression(&identity.expression, &witness_eval);
+                    let expr = to_stwo_expression(
+                        &identity.expression,
+                        &witness_eval,
+                        &constant_shifted_eval,
+                        &constant_eval,
```
Schaeff, line 161: Then this only needs `witness_eval` and `constant_eval`

backend/src/stwo/circuit_builder.rs
```diff
@@ -191,3 +259,61 @@ where
...
+        if let Identity::Polynomial(identity) = id {
+            let mut constant_with_next: Vec<usize> = Vec::new();
+            constant_with_next_to_witness_col::<T>(&identity.expression, &mut constant_with_next);
+            all_constant_with_next.extend(constant_with_next)
+        }
+    }
+    all_constant_with_next.sort_unstable();
+    all_constant_with_next.dedup();
+    all_constant_with_next
+}
```
Schaeff, line 319: This is unused

backend/src/stwo/circuit_builder.rs
```diff
@@ -191,3 +259,61 @@ where
...
+        AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation { left, op: _, right }) => {
+            constant_with_next_to_witness_col::<T>(left, constant_with_next_list);
+            constant_with_next_to_witness_col::<T>(right, constant_with_next_list);
+        }
+        AlgebraicExpression::UnaryOperation(AlgebraicUnaryOperation { op: _, expr }) => {
+            constant_with_next_to_witness_col::<T>(expr, constant_with_next_list);
+        }
+        AlgebraicExpression::Challenge(_challenge) => {}
+    }
+}
```
Schaeff, line 304: This is unused

backend/src/stwo/circuit_builder.rs
Schaeff, line 319: This should use a `HashSet` or `BTreeSet` rather than a vector.

backend/src/stwo/circuit_builder.rs
Schaeff: The value can be `[Option<F>; 2]` so that we cover the case where we only commit to the current or next row. But for this PR it would be fine to always commit to both and use `[F; 2]`. That would simplify the rest of the code as well.

backend/src/stwo/prover.rs
```diff
@@ -1,95 +1,305 @@
...
+                                                    *domain_map
+                                                        .get(&(values.len().ilog2() as usize))
+                                                        .unwrap(),
+                                                    &rotated_values,
+                                                )
+                                            })
+                                    })
+                                    .collect();
+
+                                constant_trace.extend(constant_shifted_trace);
```
Schaeff, line 185: Could we keep these two separate, as different fields in `TableProvingKey`?

backend/src/stwo/circuit_builder.rs
Schaeff: Also it can be simplified by using `post_visit_expressions_in_identities_mut`.
Then `constant_with_next_to_witness_col` can be removed entirely.

ShuangWu121: cool function! the code is simplified now

backend/src/stwo/circuit_builder.rs
```diff
@@ -191,3 +259,61 @@ where
         }
     }
 }
+
+pub fn constant_with_next_to_witness_col<T: FieldElement>(
```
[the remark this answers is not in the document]
ShuangWu121: indeed. this function is removed now

backend/src/stwo/circuit_builder.rs
```diff
@@ -64,9 +74,28 @@ impl<T: FieldElement> PowdrEval<T> {
             .map(|(index, (_, id))| (id, index))
             .collect();
 
+        let constant_with_next_list = get_constant_with_next_list(&analyzed);
+
+        let constant_shifted: BTreeMap<PolyID, usize> = analyzed
+            .definitions_in_source_order(PolynomialType::Constant)
+            .flat_map(|(symbol, _)| symbol.array_elements())
+            .enumerate()
```
Schaeff, line 82: This enumerate can be removed

Schaeff (approved):

PR #2115, opened 2024-11-19
fix nightly tests

leonardoalt:


linker/src/lib.rs
```diff
@@ -879,7 +879,7 @@ namespace main_vm(64..128);
     fn permutation_instructions() {
         let expected = r#"namespace main(128);
     let _operation_id;
-    query |__i| std::prover::provide_if_unknown(_operation_id, __i, || 13);
+    query |__i| std::prover::provide_if_unknown(_operation_id, __i, || 9);
```
leonardoalt, line 882: these were caused by changing the asm test

pipeline/src/test_util.rs
```diff
@@ -265,7 +265,14 @@ pub fn gen_halo2_proof(pipeline: Pipeline<Bn254Field>, backend: BackendVariant)
...
     let setup_file_path = output_dir.join("params.bin");
-    let max_degree = pil.degrees().into_iter().max().unwrap();
+    let max_degree = pil
+        .degree_ranges()
+        .into_iter()
+        .map(|range| range.max)
+        .collect::<Vec<_>>()
+        .into_iter()
+        .max()
+        .unwrap();
```
leonardoalt, line 275: this function is called for both Composite and Mono proofs, and it assumed all degrees were unique ranges before. This now takes the max of all max degrees for the halo2 setup

test_data/asm/permutations/block_to_block.asm
```diff
@@ -43,9 +43,7 @@ machine Main with degree: 256 {
     Binary4 bin;
     Binary4x bin4;
 
-    // two permutations to machine bin
     instr or X, Y -> Z link ~> Z = bin.or(X, Y);
-    instr or_into_B X, Y link ~> B' = bin.or(X, Y);
```
leonardoalt, line 48: unfortunately Halo2 has a bug somewhere and we can't do this.
This is testes in other places (including RISCV tests) with witgen

test_data/asm/permutations/vm_to_block.asm
```diff
@@ -11,9 +11,7 @@ machine Main with degree: 128 {
 
     Binary4 bin;
 
-    // two permutations to machine bin
     instr or X, Y -> Z link ~> Z = bin.or(X, Y);
-    instr or_into_B X, Y link ~> B' = bin.or(X, Y);
```
leonardoalt, line 16: same here re halo2

test_data/asm/permutations/block_to_block.asm
pacheco: is it possible to disable halo2 for the test instead?

leonardoalt: we still want to run halo2 proofs for these tests to test permutations between machines

pipeline/src/test_util.rs
```diff
@@ -265,7 +265,14 @@ pub fn gen_halo2_proof(pipeline: Pipeline<Bn254Field>, backend: BackendVariant)
     // Setup
     let output_dir = pipeline.output_dir().clone().unwrap();
     let setup_file_path = output_dir.join("params.bin");
-    let max_degree = pil.degrees().into_iter().max().unwrap();
+    let max_degree = pil
+        .degree_ranges()
+        .into_iter()
+        .map(|range| range.max)
+        .collect::<Vec<_>>()
+        .into_iter()
```
chriseth, line 273:
```suggestion
```
Does this not work?

leonardoalt: you pasted the same code

test_data/asm/permutations/block_to_block.asm
chriseth: But do we test the previous functionality somewhere else?

pipeline/src/test_util.rs
leonardoalt: ah read it wrong, the suggestion is only removal, let me check

test_data/asm/permutations/block_to_block.asm
leonardoalt: language/compiler/witgen wise yes, the RISCV machine has this stuff and more

pipeline/src/test_util.rs
leonardoalt: it works, changed

chriseth (approved):

PR #2118, opened 2024-11-19
(WIP) hand-written witgen for publics and poseidon in memory

pacheco:


riscv/tests/common/mod.rs
```diff
@@ -37,30 +38,26 @@ pub fn verify_riscv_asm_string<T: FieldElement, S: serde::Serialize + Send + Syn
...
-                None,
-            );
-            pipeline.rollback_from_witness();
-            let executor_trace: Vec<_> = execution.trace.into_iter().collect();
-            let pipeline = pipeline.add_external_witness_values(executor_trace);
-            test_plonky3_pipeline::<T>(pipeline);
-        }
-    */
+    // verify executor generated witness
+    if executor_witgen {
```
pacheco, line 42: This new parameter was included so we can test the riscv witgen in the supported cases only (it won't work if unimplemented coprocessors poseidon2/keccak/arith are enabled, because they mess with the memory machine witness)

pacheco: it also only works for GL

riscv-executor/src/submachines.rs
```diff
@@ -462,79 +471,272 @@ impl SubmachineKind for SplitGlMachine {
...
+        trace: &mut SubmachineTrace<F>,
+        _op: &str,
+        args: &[Elem<F>],
+        selector: Option<u32>,
+        _submachines: &[&mut dyn Submachine<F>],
+    ) {
+        const STATE_SIZE: usize = 12;
+        const OUTPUT_SIZE: usize = 4;
+        // const FULL_ROUNDS: usize = 8;
+        // const PARTIAL_ROUNDS: usize = 22;
```
leonardoalt, line 601: leftover?

leonardoalt (approved):

PR #2131, opened 2024-11-22
riscv: add support for sra instruction

gballet:
While the `srai` instruction has been implemented, the `sra` (same thing with shift passed via a register) was not. This PR implements it.

### DIFF
--- a/riscv/src/large_field/code_gen.rs
+++ b/riscv/src/large_field/code_gen.rs
@@ -726,9 +726,9 @@ fn process_instruction<A: InstructionArgs>(
                     format!("affine {}, {}, 0xffffffff, 0;", tmp1.addr(), tmp1.addr()),
                     // Here, tmp1 is the full bit mask if rs is negative
                     // and zero otherwise.
-                    format!("xor {}, {}, 0, {};", tmp1.addr(), rs1.addr(), rd.addr()),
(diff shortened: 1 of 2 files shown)


riscv/src/large_field/code_gen.rs
```diff
@@ -708,6 +708,31 @@ fn process_instruction<A: InstructionArgs>(
...
+                        tmp1.addr(),
+                        tmp1.addr()
+                    ),
+                    format!("affine {}, {}, 0xffffffff, 0;", tmp1.addr(), tmp1.addr()),
+                    // Here, tmp1 is the full bit mask if rs is negative
+                    // and zero otherwise.
+                    format!("xor {}, {}, 0, {};", tmp1.addr(), rs1.addr(), tmp2.addr()),
+                    format!("and {}, 0, 0x1f, {};", rs2.addr(), tmp3.addr()),
+                    format!("shr {}, {}, 0, {};", tmp2.addr(), tmp3.addr(), rd.addr()),
+                    format!("xor {}, {}, 0, {};", tmp1.addr(), rd.addr(), rd.addr()),
```
gballet, line 732: the issue that I ran into, testing this PR, is that `rs2` and `rd` were the same register. To work around this, I had to introduce `tmp3`.

leonardoalt (approved):

PR #2132, opened 2024-11-22
Fix pipeline tests when no features are passed

Schaeff:
```console
cargo test -p powdr-pipeline
```

currently fails because it expects panics which cannot happen if the features are off.

This PR fixes it.


georgwiese (approved):
LGTM!

PR #2134, opened 2024-11-22
Goldilocks inverse system call

lvella:
This can help accelerate Plonky3 verification, as the bottleneck is calculating field element inverse.


riscv/src/large_field/code_gen.rs
```diff
@@ -443,6 +443,31 @@ fn preamble(field: KnownField, runtime: &Runtime, with_bootloader: bool) -> Stri
         tmp3_col = 1 - XXIsZero
     }
 
+    // ================= ground field arithmetic =================
+
+    // Inverts the Goldilocks field value in memory pointed by register X,
+    // returning the low and high limbs in registers Y and Z, respectively.
```
leonardoalt, line 449: comment is outdated

riscv/src/large_field/code_gen.rs
```diff
@@ -443,6 +443,31 @@ fn preamble(field: KnownField, runtime: &Runtime, with_bootloader: bool) -> Stri
...
+    //
+    // Unsolvable if X points to 0.
```
leonardoalt, line 451: this one too

riscv/src/large_field/code_gen.rs
```diff
@@ -443,6 +443,31 @@ fn preamble(field: KnownField, runtime: &Runtime, with_bootloader: bool) -> Stri
...
+    instr invert_gl X, Y
```
leonardoalt, line 452: should we always have the instruction though? for other things we require the coprocessor to be explicitly "requested".
I think it's fine like this, we should anyway add an optimization pass that removes unused instructions

lvella: I think this is fine because it is a single instruction, not a whole new machine.

leonardoalt: that's still at least a couple extra columns if unused, but fine

leonardoalt (dismissed):
Wait, why are changes to the executor not needed?

riscv-runtime/src/goldilocks.rs
```diff
@@ -20,11 +23,22 @@ impl Goldilocks {
         Self(value)
     }
 
+    /// Returns the inverse of the field element.
+    pub fn inverse(self) -> Self {
+        let mut low = self.0 as u32;
+        let mut high = (self.0 >> 32) as u32;
+        unsafe {
```
Schaeff, line 30: Would it make sense to only have this function on the riscv target?

leonardoalt: wdym by target, the Rust rv32 target, or our riscv asm machine?

riscv/src/large_field/code_gen.rs
lvella: fixed

riscv/src/large_field/code_gen.rs
lvella: fixed

lvella:
> Wait, why are changes to the executor not needed?

Good question! I forgot about it, but it worked anyway. I have no idea what is happening.

riscv-runtime/src/goldilocks.rs
Schaeff: My bad, thought this was in number

leonardoalt (approved):

Issue #1619, opened 2024-07-26
Rewrite the RISCV machine as template + renders


leonardoalt:
Currently the way we build the RISCV machine powdr-asm code is quite ugly imo, with a bunch of String concatenations that look quite messy. Ideally we should have a template file in powdr-asm that simply gets filled by the Rust code that collects the info needed about registers, instructions, continuations etc and renders the final machine.

Example:

```rust
use std::machines::memory::{{memory_machine}};
{{ binary_import }}
{{ shift_import }}
{{ split_import }}
{{ poseidon_import }}
{{ arith_import }}

// could actually always import all, don't think it hurts compilation

machine Main {
    reg val1;
    reg val2;
    
    {{ extra_registers }}

    col witness va1_col;
    col witness va2_col;
    col witness va3_col;
    col witness va3_col;

   // ... all base instructions
   
   {{ binary_instructions }}
   {{ shift_instructions }}
   {{ poseidon_instructions }}
   {{ arith_instructions }}
   
   function main {
     {{ code }}
   }
}

PR #2137, opened 2024-11-23
Extract RISCV machine from strings into template file

leonardoalt:
Fixes https://github.com/powdr-labs/powdr/issues/1619

### DIFF
--- a/riscv/src/large_field/code_gen.rs
+++ b/riscv/src/large_field/code_gen.rs
@@ -237,7 +237,7 @@ fn riscv_machine(
         .replace("{{MAX_DEGREE_LOG}}", &format!("{}", options.max_degree_log))
         // We're passing this as well because continuations requires
         // Main's max_degree to be a constant.
-        // We should fix that in the continuations code and remove this.
+        // TODO We should fix that in the continuations code and remove this.
         .replace(
             "{{MAIN_MAX_DEGREE}}",
             &format!("{}", 1 << options.max_degree_log),
--- a/riscv/src/large_field/code_gen.rs
+++ b/riscv/src/large_field/code_gen.rs
@@ -24,13 +24,7 @@ pub fn translate_program(program: impl RiscVProgram, options: CompilerOptions) -
     let (initial_mem, instructions) =
         translate_program_impl(program, options.field, &runtime, options.continuations);
 
-    riscv_machine(
-        options,
-        &runtime,
-        &preamble(options.field, &runtime, options.continuations),
-        initial_mem,
-        instructions,
-    )
+    riscv_machine(options, &runtime, initial_mem, instructions)
 }
 
 fn translate_program_impl(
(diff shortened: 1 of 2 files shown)


riscv/src/large_field/code_gen.rs
```diff
@@ -256,310 +202,56 @@ fn preamble(field: KnownField, runtime: &Runtime, with_bootloader: bool) -> Stri
...
+        .replace("{{IMPORTS}}", &runtime.submachines_import())
+        .replace(
+            "{{SUBMACHINE_EXTRA_REGISTERS}}",
+            &submachine_extra_registers,
+        )
+        .replace("{{MIN_DEGREE_LOG}}", &format!("{}", options.min_degree_log))
+        .replace("{{MAX_DEGREE_LOG}}", &format!("{}", options.max_degree_log))
+        // We're passing this as well because continuations requires
+        // Main's max_degree to be a constant.
+        // We should fix that in the continuations code and remove this.
```
pacheco, line 240: add the TODO keyword?

riscv/src/riscv_gl.asm
```diff
@@ -0,0 +1,345 @@
...
+
+    let initial_memory: (fe, fe)[] = [
+        {{INITIAL_MEMORY}}
+    ];
+
+    function main {
+        {{PROGRAM}}
+    }
+
+    // Extra columns we use to hold temporary values inside instructions.
```
pacheco, line 34: the `col` declarations are a bit scattered around the file, maybe they could be placed together? like the `tmpX_col` ones,  `Y_7bit` etc

riscv/src/riscv_gl.asm
```diff
@@ -0,0 +1,345 @@
...
+
+    std::machines::large_field::memory::Memory regs(byte2, MIN_DEGREE, LARGE_SUBMACHINES_MAX_DEGREE);
+    {{MEMORY_DECLARATION}}
+    {{SUBMACHINE_DECLARATIONS}}
+
+    let initial_memory: (fe, fe)[] = [
+        {{INITIAL_MEMORY}}
+    ];
+
+    function main {
```
pacheco, line 30: should `main` be the last thing in the file?

riscv/src/riscv_gl.asm
leonardoalt: I copied everything mostly as is from the original file, trying not to mess with the template too much to avoid diff hell, but looks like github didn't care anyway

riscv/src/riscv_gl.asm
leonardoalt: done

riscv/src/riscv_gl.asm
leonardoalt: done

pacheco (approved):

PR #2139, opened 2024-11-24
Remove unused asm instructions

leonardoalt:
I'm not sure this is the best way to do this, but wanted to get started.


ast/src/parsed/asm.rs
```diff
@@ -769,6 +770,77 @@ impl SourceReference for Param {
...
+            to_visit.push(main_machine);
+        }
+
+        while let Some(machine) = to_visit.pop() {
+            for statement in &machine.statements {
+                if let MachineStatement::FunctionDeclaration(_, _, _, body) = statement {
+                    for statement in body {
+                        match statement {
+                            // It's pretty cumbersome to get the name of the instruction in an
+                            // assignment.
```
leonardoalt, line 818: Are there any helpers for this?

ast/src/parsed/asm.rs
```diff
@@ -769,6 +770,77 @@ impl SourceReference for Param {
...
+                        _ => true,
+                    });
+                }
+            }
+        }
+
+        self
+    }
+
+    fn collect_used_instructions(&self) -> HashSet<String> {
```
leonardoalt, line 797: Is there an easier way to implement this with existing visitors?

Issue #682, opened 2023-10-09
powdr-asm optimizer: remove unused submachines, instructions, registers


leonardoalt:
https://github.com/powdr-labs/powdr/pull/678 adds the Poseidon submachine to the RISCV machine and a bunch of extra registers. This can be quite costly for cases that actually don't use Poseidon at all.

1. We could have a simple assembly pass that traverses all statements, and if certain instructions are not used, removes them.
2. When an instruction is removed, if it is delegation, we could potentially also remove the submachine instance (if all its delegations are removed)
3. We could also remove unused registers after the passes above

This would ensure that the RISCV machine is the same as before when Poseidon is not used

leonardoalt:
This could still be useful, I think @pacheco started on something similar.

PR #2143, opened 2024-11-25
powdr-asmopt: remove unused submachines, instructions, registers

gzanitti:
Solves #682


gzanitti:
If necessary, we could separate this PR into three smaller parts: 
1) The new ReferencedSymbols implementations.
2) The asmopt module
3) The integration of asmopt into the pipeline

asmopt/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+
+const MAIN_MACHINE_STR: &str = "::Main";
+const PC_REGISTER: &str = "pc";
+
+pub fn optimize(mut analyzed_asm: AnalysisASMFile) -> AnalysisASMFile {
+    // Optimizations assume the existence of a Main machine as an entry point.
+    // If it doesn't exist, return the ASM as-is to prevent removing all machines,
+    // which would break some examples from the book.
+    let main_machine_path = parse_absolute_path(MAIN_MACHINE_STR);
+    if analyzed_asm
```
gzanitti, line 21: Main seems to be the entry point for execution, but it is not mandatory to include it. This check avoids optimizing cases where the Main machine does not exist, thus avoiding that all machines are eliminated (cases like `pil_at_module_level.asm`).

asmopt/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+}
+
+fn machine_remove_unused_instructions(machine: &mut Machine, symbols: &HashSet<String>) {
+    machine
+        .instructions
+        .retain(|ins| symbols.contains(&ins.name));
+}
+
+fn collect_constrained_machines(asm_file: &AnalysisASMFile) -> impl Iterator<Item = String> + '_ {
+    asm_file
```
gzanitti, line 127: This function searches for constrained machines by looking for machines where columns are defined, but perhaps there is a more precise way?

asmopt/tests/optimizer.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    machine Unused with degree: 8 {
+        reg pc[@pc];
+        col witness w;
+        w = w * w;
+    }
+    "#;
+
+    let expectation = r#"machine Main with degree: 8 {
+    ::Required required
```
pacheco, line 220: Shouldn't `Required` be removed here? It is not used anymore

asmopt/tests/optimizer.rs
```diff
@@ -0,0 +1,311 @@
+use powdr_analysis::analyze;
+use powdr_asmopt::optimize;
+use powdr_parser::parse_asm;
+
```
pacheco, line 4: test suggestion: machines in `link`s (outside of instructions) should not be removed

ast/src/asm_analysis/mod.rs
```diff
@@ -882,6 +882,17 @@ impl Module {
         self.ordering.push(StatementReference::Module(name));
     }
 
+    pub fn retain_machines(&mut self, names: HashSet<String>) {
```
pacheco, line 885: is it clear from the context what `String` should be here?

pilopt/src/referenced_symbols.rs
```diff
@@ -149,8 +167,226 @@ fn symbols_in_expression(
     }
 }
 
-impl ReferencedSymbols for Type {
+fn symbols_in_expression_asm(
```
pacheco, line 170: should this be a `symbols()` method in `ExpressionAsm` objects? maybe implement the `ReferencedSymbols` trait?

test_data/asm/book/declarations.asm
```diff
@@ -5,9 +5,8 @@ mod utils {
         0 => 0,
         _ => arr[len - 1] + sum(len - 1, arr)
     };
-    // A simple function that returns the input incremented by one,
-    // as an expression.
-    let incremented: expr -> expr = |x| x + 1;
+    // A simple function that returns the same input as an expression.
```
pacheco, line 8: are these book changes related?

pipeline/src/pipeline.rs
```diff
@@ -773,6 +776,7 @@ impl<T: FieldElement> Pipeline<T> {
 
                 self.log("Run analysis");
                 let analyzed_asm = powdr_analysis::analyze(resolved)?;
+                self.maybe_write_pil(&analyzed_asm, "_asm")?;
```
pacheco, line 779: won't this generate a `foo_asm.pil`?

pipeline/src/pipeline.rs
```diff
@@ -787,14 +791,35 @@ impl<T: FieldElement> Pipeline<T> {
         Ok(self.artifact.analyzed_asm.as_ref().unwrap())
     }
 
+    pub fn compute_optimized_asm(&mut self) -> Result<AnalysisASMFile, Vec<String>> {
```
pacheco, line 794: i think this should return a ref to `AnalysisAsmFile` instead of a clone

pipeline/src/pipeline.rs
```diff
@@ -787,14 +791,35 @@ impl<T: FieldElement> Pipeline<T> {
...
+
     pub fn compute_constrained_machine_collection(
         &mut self,
     ) -> Result<&AnalysisASMFile, Vec<String>> {
         if self.artifact.constrained_machine_collection.is_none() {
             self.artifact.constrained_machine_collection = Some({
-                self.compute_analyzed_asm()?;
-                let analyzed_asm = self.artifact.analyzed_asm.take().unwrap();
-                powdr_asm_to_pil::compile::<T>(analyzed_asm)
+                self.compute_optimized_asm()?;
```
pacheco, line 820: this already returns the `optimized_asm`

asmopt/tests/optimizer.rs
gzanitti: In fact, it's being used, it just doesn't print (I'm going to open an issue on this). That's partly why I created the constrained_machines that are no longer needed (because they really weren't).

pilopt/src/referenced_symbols.rs
gzanitti: It's also implemented. But since symbols() uses all_children this function is used to get “first level” PublicReferences/References or nothing, without checking their children.

pipeline/src/pipeline.rs
gzanitti: Yes, but I understand that this way you avoid having to clone the file. It's similar to how it's done for the PIL optimization pass.

asmopt/tests/optimizer.rs
gzanitti: Added, thanks!

test_data/asm/book/declarations.asm
pacheco: minor, but why not keep the `incr` instruction and call it from `main`? it is a bit more "interesting" of an example

asmopt/tests/optimizer.rs
pacheco: mmm it's required because its a parameter to the submachine?

gzanitti: Oh, sorry, yes, exactly. Correctly printed, this should look like this:

```
::Required required
::ParamMachine sub(required)
```

pacheco (approved):
lgtm

PR #2163, opened 2024-11-27
Remove constant intermediates

Schaeff:
See [added test](https://github.com/powdr-labs/powdr/blob/48de0e832e14da0e5dd5377bc75c6eea55c1bcc6/pilopt/tests/optimizer.rs#L34)

This propagates knowledge about constant intermediates forward in source order. We could run this backwards as well but I think in practice intermediates are defined before usage.


pilopt/src/lib.rs
```diff
@@ -465,6 +466,52 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
+fn remove_constant_intermediate_columns<T: FieldElement>(pil_file: &mut Analyzed<T>) {
+    let constant_intermediate_polys = pil_file
+        .intermediate_polys_in_source_order()
+        .filter(|(symbol, _)| !symbol.is_array())
+        .fold(
+            BTreeMap::default(),
+            |mut values: BTreeMap<PolyID, (String, BigUint)>, (symbol, definitions)| {
+                let ((name, poly_id), mut definition) = symbol
+                    .array_elements()
+                    .zip_eq(definitions.iter().cloned())
```
gzanitti, line 481: Are we sure this is always valid?

pilopt/src/lib.rs
```diff
@@ -465,6 +466,52 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
+                        symbol.absolute_name
+                    );
+                    values.insert(poly_id, (name, value.to_arbitrary_integer()));
+                };
+
+                values
+            },
+        )
+        .into_iter()
+        .map(|(id, (name, e))| ((name, id), e))
```
gzanitti, line 509: Rearranging `values` so it's ordered in this way and avoiding the map is too complex, right?

pilopt/tests/optimizer.rs
```diff
@@ -30,6 +30,22 @@ fn replace_fixed() {
...
+fn replace_intermediate() {
+    let input = r#"namespace N(65536);
+    col witness X;
+    col intermediate = 1;
+    col other_intermediate = (intermediate - 1) * X;
+    X' = X + intermediate + other_intermediate;
+"#;
+    let expectation = r#"namespace N(65536);
+    col witness X;
+    N::X' = N::X + 1;
```
gzanitti, line 43: Nice :sunglasses:

pilopt/src/lib.rs
```diff
@@ -465,6 +466,52 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
+                    }
+                });
+
+                // simplify the expression
+                let definition = simplify_expression(definition.clone());
+
+                // update the known values if we ended up with a constant
+                if let AlgebraicExpression::Number(value) = definition {
+                    log::debug!(
+                        "Determined fixed column {} to be constant {value}. Removing.",
```
chriseth, line 499:
```suggestion
                        "Determined intermediate column {} to be constant {value}. Removing.",
```

pilopt/src/lib.rs
```diff
@@ -465,6 +466,52 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
+    let constant_intermediate_polys = pil_file
+        .intermediate_polys_in_source_order()
+        .filter(|(symbol, _)| !symbol.is_array())
+        .fold(
+            BTreeMap::default(),
+            |mut values: BTreeMap<PolyID, (String, BigUint)>, (symbol, definitions)| {
+                let ((name, poly_id), mut definition) = symbol
+                    .array_elements()
+                    .zip_eq(definitions.iter().cloned())
+                    .next()
```
chriseth, line 482: I think we should either properly iterate or check that it is not an array.

pilopt/src/lib.rs
```diff
@@ -465,6 +466,52 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
+                definition.post_visit_expressions_mut(&mut |e| {
+                    if let AlgebraicExpression::Reference(AlgebraicReference { poly_id, .. }) = e {
+                        if let Some((_, value)) = values.get(poly_id) {
+                            *e = AlgebraicExpression::Number(T::from(value.clone()));
+                        }
+                    }
+                });
+
+                // simplify the expression
+                let definition = simplify_expression(definition.clone());
```
chriseth, line 494: Should we really iteratively replace the values right away or rather keep the steps simple and run them multiple times? Do we have that many intermediate columns that reference other intermediate columns?

pilopt/src/lib.rs
Schaeff: It's checked line 475. I can add an assert also here for clarity?

pilopt/src/lib.rs
Schaeff: The use case here is that when we build the bus accumulator updates:

```rust
    col intermediate_fingerprint_0 = 0;
    col intermediate_fingerprint_1 = 0;
    col intermediate_fingerprint_0_1 = std::prelude::challenge(0, 1) * main::intermediate_fingerprint_0 + 11 * std::prelude::challenge(0, 2) * main::intermediate_fingerprint_1 + 0;
    col intermediate_fingerprint_1_1 = std::prelude::challenge(0, 2) * main::intermediate_fingerprint_0 + std::prelude::challenge(0, 1) * main::intermediate_fingerprint_1 + 0;
```

All of this should be removed as it's all 0.

We have a lot of intermediates which depend on each other (depending on the number of columns we're sending/receiving), but they typically don't end up being constant. This only happens in some special cases like above.

I *think* running it twice without memory would work, but writing the case with memory is not that much code. Up to you :)

pilopt/src/lib.rs
Schaeff: Actually, we could: we would add the name to the key of the constant values, and it's also available in the `AlgebraicExpression`. I'm not sure it makes sense though, it would just add a bunch of String hashing to get rid of one iterator map?

pilopt/src/lib.rs
```diff
@@ -466,6 +467,35 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
+            let mut symbols_and_definitions = symbol.array_elements().zip_eq(definitions);
+            match symbol.is_array() {
+                true => None,
+                false => {
+                    let ((name, poly_id), definition) = symbols_and_definitions.next().unwrap();
+                    match definition {
+                        AlgebraicExpression::Number(value) => {
+                            log::debug!(
+                                "Determined fixed column {} to be constant {value}. Removing.",
+                                name
```
chriseth, line 485:
```suggestion
                                "Determined intermediate column {name} to be constant {value}. Removing.",
```

Schaeff: oops, this got lost when reverting!

chriseth (approved):

PR #2166, opened 2024-11-27
Introduce workspace-level lints

Schaeff:
Reduce the cost of adding a lint from O(#workspace_crates) to 1.


leonardoalt (approved):

Schaeff:
Self review: in some places I did not add a comment about determinism, and in some others the comment is maybe a bit too vague

PR #2167, opened 2024-11-27
Disallow iteration over hash types

Schaeff:
Iterating over hash types introduces non-determinism.
Ban generally and require turning off the lint locally when it's fine to do so.
This [lint](https://rust-lang.github.io/rust-clippy/master/index.html#/iter_over_hash_type) only applies to loops.


Schaeff:
Self review: in some places I did not add a comment about determinism, and in some others the comment is maybe a bit too vague

pil-analyzer/src/call_graph.rs
ShuangWu121, line 1: The deterministic result is guaranteed by topo_sort_visit, right? I noticed you’ve added comments in other cases, so maybe it could be helpful to have one here too?

ast/src/analyzed/mod.rs
```diff
@@ -711,6 +711,7 @@ impl SolvedTraitImpls {
     /// This just updates the `index` fields.
     /// Assumes that `to_remove` is sorted.
     pub fn remove_trait_impls(&mut self, to_remove: &[usize]) {
+        #[allow(clippy::iter_over_hash_type)]
```
ShuangWu121, line 714: this is deterministic also because of no shared state?

Schaeff: agreed, added comment

Schaeff: actually, replaced by a BTreeMap

pil-analyzer/src/call_graph.rs
Schaeff: I'm actually not sure it is, since there is no single topo sort. Changing to BTreeMap

ShuangWu121 (approved):
LGTM

ShuangWu121 (approved):

PR #2171, opened 2024-11-28
Simplify empty permutations

leonardoalt:
Permutations of the form `0 $ [...] is S $ [...]` can be simplified to `S = 0`. If this were a lookup, it would be a tautology and the constraint could be removed entirely. But as a permutation we're saying the LHS (empty set) is a permutation of the RHS, therefore RHS must also be the empty set. This is enforced by the simplified constraint.

### DIFF
--- a/pilopt/src/lib.rs
+++ b/pilopt/src/lib.rs
@@ -28,6 +28,7 @@ pub fn optimize<T: FieldElement>(mut pil_file: Analyzed<T>) -> Analyzed<T> {
     deduplicate_fixed_columns(&mut pil_file);
     simplify_identities(&mut pil_file);
     extract_constant_lookups(&mut pil_file);
+    simplify_empty_permutations(&mut pil_file);
     remove_constant_witness_columns(&mut pil_file);
     simplify_identities(&mut pil_file);
     remove_trivial_identities(&mut pil_file);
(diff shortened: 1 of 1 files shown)


pilopt/src/lib.rs
```diff
@@ -439,6 +440,39 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+                if !matches!(&left.selector, AlgebraicExpression::Number(n) if n == &T::zero()) {
+                    continue;
+                }
+
+                let pol = PolynomialIdentity {
+                    id: *id,
+                    expression: right.selector.clone(),
+                    source: source.clone(),
+                };
+                *identity = Identity::Polynomial(pol);
```
chriseth, line 470: Can we just remove the phantom? Shouldn't we keep it so that the multiplicity can be set to zero?

chriseth: Ah, there is no multiplicity. Hm.

leonardoalt: I'm not too aware of that stuff, but if changing the multiplicity only makes more sense can do that too

chriseth: What I'm saying is: If we remove the phantom, we should also remove the columns and constraints generated from the phantom. We had a similar problem before, but not sure how we resolved it.

leonardoalt: Do you have a simple example of the other constraints that are generated?
I think it could be the case that they are still sound regardless, because we're not removing the phantom permutation, rather making them unsatisfiable.

leonardoalt:
I don't know how this is passing CI

leonardoalt:
@chriseth @georgwiese do you think this is fine then or not?

leonardoalt:
With https://github.com/powdr-labs/powdr/pull/2143 this might not be needed anymore. Closing for now.

PR #2178, opened 2024-11-29
Fix some tests

leonardoalt:


riscv/src/large_field/runtime.rs
```diff
@@ -331,7 +331,7 @@ impl Runtime {
             "std::machines::large_field::arith::Arith",
             None,
             "arith",
-            vec![],
+            vec!["MIN_DEGREE", "MAIN_MAX_DEGREE"],
```
leonardoalt, line 334: this was missing

riscv/tests/common/mod.rs
```diff
@@ -56,6 +44,18 @@ pub fn verify_riscv_asm_string<T: FieldElement, S: serde::Serialize + Send + Syn
         );
     }
 
+    // Compute the witness once for all tests that follow.
```
leonardoalt, line 47: moved here to have the fast executor always run before witgen, required by some tests

riscv/tests/riscv.rs
```diff
@@ -531,7 +531,7 @@ fn features_with_options<T: FieldElement>(options: CompilerOptions) {
         &from_elf,
         &[expected.into()],
         None,
-        true,
+        false,
```
leonardoalt, line 534: executor complains about these

riscv/tests/riscv.rs
```diff
@@ -615,12 +615,12 @@ fn many_chunks_memory() {
 }
 
 fn verify_riscv_crate(case: &str, inputs: &[u64], executor_witgen: bool) {
-    verify_riscv_crate_bb(case, inputs.iter().map(|&x| x.into()).collect());
     verify_riscv_crate_gl(
         case,
         inputs.iter().map(|&x| x.into()).collect(),
         executor_witgen,
     );
+    verify_riscv_crate_bb(case, inputs.iter().map(|&x| x.into()).collect());
```
leonardoalt, line 623: some tests rely on running GL before BB

riscv/src/small_field/runtime.rs
```diff
@@ -76,14 +76,14 @@ impl Runtime {
             [
                 r#"instr shl XL, YL, ZH, ZL, WL
                     link ~> (tmp1_h, tmp1_l) = regs.mload(0, XL, STEP)
-                    link ~> (tmp2_h, tmp2_l) = regs.mload(0, YL, STEP)
+                    link ~> (tmp2_h, tmp2_l) = regs.mload(0, YL, STEP + 1)
```
pacheco, line 79: just to understand: is it an issue having two reads in the same time step?

leonardoalt: For the same address actually yes (which happened earlier when I was testing the new Mock prover stuff). I also thought it wasn't a problem, and it actually didn't use to be when we used a lookup, but with a permutation it is an issue because the count doesn't match anymore.

riscv/tests/riscv.rs
pacheco: mmm that is weird, its a pretty simple program

riscv/tests/riscv.rs
leonardoalt: I think it was a test that expects a fail instruction from the executor, but since the executor isn't implemented for BB it failed differently

riscv/tests/riscv.rs
leonardoalt: yea it complains about a submachine with no columns, I feel like the fix would be simple but wanted to keep this PR going, can fix that separately

pacheco (approved):

PR #2187, opened 2024-12-02
update CI ubuntu

leonardoalt:


.github/workflows/nightly-tests.yml
```diff
@@ -43,22 +43,25 @@ jobs:
...
-          ~/.cargo/registry
-          ~/.cargo/git
-          target
-        key: ${{ runner.os }}-cargo-release-${{ hashFiles('**/Cargo.toml') }}
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+          target/
+          Cargo.lock
+        key: ${{ runner.os }}-cargo-nightly-tests
```
Schaeff, line 64: What is the reasoning for this change?

.github/workflows/nightly-tests.yml
```diff
@@ -74,18 +77,17 @@ jobs:
...
     - name: Install pilcom
       run: git clone https://github.com/0xPolygonHermez/pilcom.git  && cd pilcom && npm install
     - name: Check without Halo2
       run: cargo check --all --no-default-features
     - name: Build
       run: cargo build --all --release --all-features
     - name: Run tests
-      # Number threads is set to 1 because the runner does not have enough memeory for more.
-      run: PILCOM=$(pwd)/pilcom/ cargo test --all --release --verbose --all-features -- --include-ignored --nocapture --test-threads=1
+      # Number threads is set to 1 because the runner does not have enough memory for more.
```
Schaeff, line 90: Comment is outdated since test-threads is not 1 anymore

.github/workflows/nightly-tests.yml
leonardoalt: using the same cache as pr-tests, Lucas changed it from time to time so I assume this is more optimized

.github/workflows/nightly-tests.yml
leonardoalt: removed comment

PR #2188, opened 2024-12-03
Revert "Parallelize calls to prover functions (#2176)"

leonardoalt:
This reverts commit d980c68f89386160e3ffd6b2ce4e62f26c045cdf from https://github.com/powdr-labs/powdr/pull/2176


georgwiese (approved):
For context: This is due to a performance regression.

leonardoalt:
I ran some tests vs main and can confirm this brings back same witgen speed

PR #2189, opened 2024-12-03
Expose stdout/stderr in Session

leonardoalt:
This PR:
- Makes explicit the notion that 0=stdin, 1=stdout, 2=stdout in the QueryCallback's "FS"
- Exposes the outputs in Session
- Removes printing to stdout and stderr in the callback itself. This is now the responsibility of the host if needed.
- Adds the Fibonacci test with stdout to CI using the write mechanism

The idea is that after this we should also expose the proof's publics and make a stream mechanism for inputs and outputs

### DIFF
--- a/pipeline/src/lib.rs
+++ b/pipeline/src/lib.rs
@@ -34,6 +34,12 @@ impl HostContext {
         (ctx, cb)
     }
 
+    pub fn clear_outputs(&mut self) {
+        let mut fs = self.file_data.lock().unwrap();
+        fs.insert(1, vec![]);
+        fs.insert(2, vec![]);
+    }
+
     pub fn read<T: DeserializeOwned>(&self, fd: u32) -> Result<T, String> {
         let fs = self.file_data.lock().unwrap();
         if let Some(data) = fs.get(&fd) {
@@ -59,7 +65,7 @@ impl HostContext {
                         as char;
                     match fd {
                         // stdin, stdout and stderr are supported by the default callback
-                        0..=2 => return Err(format!("Unsupported file descriptor: {fd}")),
+                        0 => return Err(format!("Unsupported file descriptor: {fd}")),
                         _ => {
                             let mut map = fs.lock().unwrap();
                             map.entry(fd).or_default().push(byte as u8);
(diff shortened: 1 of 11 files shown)


pipeline/src/lib.rs
```diff
@@ -59,7 +65,7 @@ impl HostContext {
                         as char;
                     match fd {
                         // stdin, stdout and stderr are supported by the default callback
-                        0..=2 => return Err(format!("Unsupported file descriptor: {fd}")),
+                        0 => return Err(format!("Unsupported file descriptor: {fd}")),
```
pacheco, line 68: Comment right above needs to be updated i think.
Also looking at the code, who handles fds 1 and 2? is `HostContext` only used for riscv? 
`handle_simple_queries_callback` also seems to handle 1 and 2?

pipeline/src/lib.rs
```diff
@@ -34,6 +34,12 @@ impl HostContext {
         (ctx, cb)
     }
 
+    pub fn clear_outputs(&mut self) {
```
pacheco, line 37: shouldn't this be private and the pipeline then responsible for calling this when appropriate?

leonardoalt: but will pipeline be able to call this if it's private?

pacheco:
Also, what do we do about our `runtime::print`? it won't print anymore right?

pipeline/src/lib.rs
leonardoalt: IMO there should be no special treatment for 1 and 2 in the callback itself, the host should just read it afterwards from pipeline or session if needed

leonardoalt:
> Also, what do we do about our `runtime::print`? it won't print anymore right?

we can print it in the test

pacheco:
> > Also, what do we do about our `runtime::print`? it won't print anymore right?
> 
> we can print it in the test

i mean, won't it be confusing to users that `runtime::print` does nothing?

pipeline/src/lib.rs
pacheco: what I was thinking is if this should be "automatic" for pipeline users (as in, isn't it a kind of artifact too, that needs to be cleared if you run some part of the pipeline again)

leonardoalt:
ahhh runtime print, sorry I confused with the test. Yea true... let me think about it

pipeline/src/lib.rs
pacheco: but there is code in `handle_simple_queries_callback` to handle it... is the `HostContext` callback always present? which one handles it?

pipeline/src/lib.rs
leonardoalt: it's kind of part of the witness, but in a different data structure. But I'm not sure I understand what the suggestion is

pipeline/src/lib.rs
leonardoalt: I removed that though, check `pipeline/src/lib.rs` here

pipeline/src/lib.rs
```diff
@@ -169,22 +175,6 @@ pub fn handle_simple_queries_callback<'a, T: FieldElement>() -> impl QueryCallba
         let (id, data) = parse_query(query)?;
         match id {
             "None" => Ok(None),
-            "Output" => {
```
leonardoalt, line 172: @pacheco here

pacheco: oh right, my bad, i must have looked into the old version

pacheco (approved):

leonardoalt:
I removed it from the queue again because somehow the newly added Fibonacci RISCV test (as an example) is running in every job, instead of in one.

pacheco (approved):

PR #2196, opened 2024-12-04
don't generate memory machine traces executing in `Fast` mode

pacheco:
these calls were missing the `ExecMode::Trace` check

### DIFF
--- a/riscv-executor/src/lib.rs
+++ b/riscv-executor/src/lib.rs
@@ -699,28 +699,30 @@ mod builder {
             self.set_next_pc().and(Some(st_line))
         }
 
-        pub(crate) fn set_mem(&mut self, addr: u32, val: u32) {
+        pub(crate) fn set_mem(&mut self, addr: u32, val: u32, step: u32, selector: u32) {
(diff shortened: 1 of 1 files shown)


riscv-executor/src/lib.rs
```diff
@@ -1982,7 +1954,7 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
                 assert_eq!(input_ptr % 4, 0);
 
                 let inputs: [u64; 8] = (0..16)
-                    .map(|i| self.proc.get_mem(input_ptr + i * 4))
+                    .map(|i| self.proc.get_mem(input_ptr + i * 4, 0, 0)) // TODO: step/selector for poseidon2
```
leonardoalt, line 1957: TODO here or later?

leonardoalt (approved):

riscv-executor/src/lib.rs
pacheco: witness for poseidon2 is not being generated... so the value here doesn't matter

PR #2198, opened 2024-12-05
Run `cargo bench` in PR CI tests and report result to PR and Github Pages

Schaeff:


.github/workflows/pr-tests.yml
```diff
@@ -237,3 +237,56 @@ jobs:
...
+      uses: benchmark-action/github-action-benchmark@v1
+      with:
+        name: Benchmarks
+        tool: 'cargo'
+        output-file-path: output.txt
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        auto-push: true
+        alert-threshold: '120%'
+        comment-on-alert: true
+        fail-on-alert: true
```
pacheco, line 291: does this mean the regression will prevent merging? can the merge be forced anyways? maybe a regression is expected in some cases?

Schaeff: Yeah I wasn't sure about this, I can definitely make it not fail on alert!

pacheco: maybe we could start with only the alert and switch it back on once things are more stable

Schaeff: Done!

Schaeff:
Note: now https://docs.powdr.org/dev/bench/ gets deleted every time we merge to main. This PR changes that as well, so that the benchmarks are kept there.

PR #2199, opened 2024-12-05
Use arith-memory in RISCV

leonardoalt:


riscv-runtime/src/arith.rs
```diff
@@ -19,52 +29,70 @@ pub(crate) fn u32_to_be(from: &[u32; 8], to: &mut [u8; 32]) {
...
-    (a, b)
+    let mut res_u8: [u8; 64] = [0u8; 64];
+    u32x16_to_be(&res, &mut res_u8);
+    res_u8
 }
 
-/// Calculate `a*b + c = hi*2**256 + lo` for 256 bit values (as u8 little-endian arrays).
-/// Returns `(hi, lo)`.
-pub fn affine_256_u8_le(mut a: [u8; 32], mut b: [u8; 32], c: [u8; 32]) -> ([u8; 32], [u8; 32]) {
+/// Calculate `a*b + c = d` for 256 bit values (as u8 little-endian arrays).
```
Schaeff, line 69: `d` is 512 bits though right?

Schaeff (commented):
Looks fine, mostly just changing function signatures

riscv-runtime/src/arith.rs
leonardoalt: yes

riscv-executor/src/lib.rs
```diff
@@ -1979,96 +1979,149 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
             "mod_256" => {
-                assert!(args.is_empty());
-                // take input from registers
-                let y2 = (0..8)
-                    .map(|i| self.proc.get_reg(&register_by_idx(i)).into_fe())
+                // a mod b = c
+                let input_ptr_a = self.proc.get_reg_mem(args[0].u()).u();
+                assert_eq!(input_ptr_a % 4, 0);
+                let input_ptr_b = self.proc.get_reg_mem(args[1].u()).u();
+                assert_eq!(input_ptr_b % 4, 0);
```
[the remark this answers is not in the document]
leonardoalt: will do

riscv-runtime/src/arith.rs
leonardoalt: added a comment on the line after this one

riscv-executor/src/lib.rs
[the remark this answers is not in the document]
leonardoalt: changed most of these but not all in the file because some are older and would require more changes

PR #2205, opened 2024-12-07
Paged memory

chriseth:
Change memory machine to use pages instead of a "flat" map.


georgwiese (commented):
Cool!

I think we're wasting space because we don't use that fact that memory addresses of "normal" memory are guaranteed to be aligned. However, for register memory, we do use all slots (up to 32).

But I think we should fix this by changing the way we use the machine and divide any aligned address by 4 before accessing memory. That way, witgen doesn't need to be aware of any specific memory semantics, it would just be a key-value store where both keys and values are field elements.

georgwiese (approved):
Nice!

PR #2207, opened 2024-12-09
[WIP] RISCV executor trace

pacheco:


riscv-executor/src/lib.rs
```diff
@@ -699,28 +752,38 @@ mod builder {
...
         }
 
-        pub(crate) fn set_mem(&mut self, addr: u32, val: u32, step: u32, selector: u32) {
+        pub(crate) fn set_mem(&mut self, addr: u32, val: u32, step: u32, selector: u8) {
+            self.submachine_op(
+                MachineInstance::main_memory,
+                Some(selector),
+                &[1.into(), addr.into(), step.into(), val.into()],
+                &[],
+            );
```
leonardoalt, line 761: Shouldn't this be done only in trace mode below?

riscv-executor/src/lib.rs
```diff
@@ -699,28 +752,38 @@ mod builder {
...
 
-        pub(crate) fn get_mem(&mut self, addr: u32, step: u32, selector: u32) -> u32 {
+        pub(crate) fn get_mem(&mut self, addr: u32, step: u32, selector: u8) -> u32 {
             let val = *self.mem.get(&addr).unwrap_or(&0);
+            self.submachine_op(
+                MachineInstance::main_memory,
+                Some(selector),
+                &[0.into(), addr.into(), step.into(), val.into()],
+                &[],
+            );
```
leonardoalt, line 780: same here

riscv-executor/src/lib.rs
```diff
@@ -776,65 +835,68 @@ mod builder {
...
+                let m = match m {
+                    MachineInstance::main_memory => "memory",
+                    MachineInstance::main_regs => "regs",
+                    MachineInstance::main_binary => "binary",
+                    MachineInstance::main_shift => "shift",
+                    MachineInstance::main_publics => "publics",
+                    MachineInstance::main_split_gl => "split_gl",
+                    MachineInstance::main_poseidon_gl => "poseidon_gl",
+                };
+                for op in ops {
```
leonardoalt, line 850: Could we somehow do this in parallel per machine?

riscv-executor/src/lib.rs
```diff
@@ -776,65 +835,68 @@ mod builder {
...
+                        &op.extra,
+                    );
+                }
+            }
+
             // add submachine traces to main trace
             // ----------------------------
-            for mut machine in self.submachines.into_values().map(|m| m.into_inner()) {
-                // if the machine is not empty, we need to fill it up to the degree
+            for (name, mut machine) in self
```
leonardoalt, line 861: potentially parallel too, not a blocker

riscv-executor/src/lib.rs
```diff
@@ -1021,23 +1083,35 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
     /// read register value, updating the register memory machine
-    fn reg_read(&mut self, step_offset: u32, reg: u32, selector_idx: u32) -> Elem<F> {
+    fn reg_read(&mut self, step_offset: u32, reg: u32, selector: u8) -> Elem<F> {
         let val = self.proc.get_reg_mem(reg);
-        if let ExecMode::Trace = self.mode {
-            self.proc
-                .regs_machine
-                .read(self.step + step_offset, reg, val, selector_idx);
-        }
+        self.proc.submachine_op(
```
leonardoalt, line 1088: only trace mode or always needed?

riscv-executor/src/lib.rs
```diff
@@ -1021,23 +1083,35 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
 
     /// write value to register, updating the register memory machine
-    fn reg_write(&mut self, step_offset: u32, reg: u32, val: Elem<F>, selector_idx: u32) {
-        if let ExecMode::Trace = self.mode {
-            self.proc
-                .regs_machine
-                .write(self.step + step_offset, reg, val, selector_idx);
-        }
+    fn reg_write(&mut self, step_offset: u32, reg: u32, val: Elem<F>, selector: u8) {
+        self.proc.submachine_op(
```
leonardoalt, line 1104: same hree

riscv-executor/src/lib.rs
```diff
@@ -1098,12 +1185,14 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
                     set_col!(Y_free_value, val);
                 }
 
+                main_op!(set_reg,);
```
leonardoalt, line 1188: hm maybe a function would be nicer than a macro for this? The mandatory trailing comma looks kinda odd

riscv-executor/src/lib.rs
```diff
@@ -1725,15 +1841,13 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
-                if let ExecMode::Trace = self.mode {
-                    self.proc.submachine("split_gl").add_operation(
-                        "split_gl",
-                        &[hi.into(), lo.into()],
-                        Some(0),
-                        &[],
-                    );
-                }
-
+                let selector = 0;
```
leonardoalt, line 1844: why not inline this variable?

riscv-executor/src/lib.rs
```diff
@@ -1819,15 +1952,13 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
-                    self.proc.submachine("split_gl").add_operation(
-                        "split_gl",
-                        &[high_inv.into(), low_inv.into()],
-                        Some(0),
-                        &[],
-                    );
-                }
-
+                let selector = 0;
+                submachine_op!(
```
leonardoalt, line 1956: inline?

riscv-executor/src/lib.rs
```diff
@@ -1850,15 +1981,13 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
-                if let ExecMode::Trace = self.mode {
-                    self.proc.submachine("split_gl").add_operation(
-                        "split_gl",
-                        &[hi.into(), lo.into()],
-                        Some(0),
-                        &[],
-                    );
-                }
-
+                let selector = 0;
```
leonardoalt, line 1984: inline?

leonardoalt: also fine not inlining these btw

riscv-executor/src/submachines.rs
```diff
@@ -44,152 +44,157 @@ pub trait Submachine<F: FieldElement> {
...
-        let prev_len = self.len();
-        self.trace
-            .push_dummy_block(machine_max_degree, M::BLOCK_SIZE, M::SELECTORS);
-        let dummy_size = self.len() - prev_len;
-        M::dummy_block_fix(&mut self.trace, dummy_size);
-    }
-
-    fn take_cols(&mut self) -> Vec<(String, Vec<Elem<F>>)> {
+    fn finish(&mut self, degree: u32) -> Vec<(String, Vec<F>)> {
+        assert!(self.len() <= degree);
```
leonardoalt, line 97: we had a bug before related to this, where in the continuations dry run the executor is allowed to go to any amount of rows. Does this take that into account?

riscv-executor/src/submachines.rs
```diff
@@ -44,152 +44,157 @@ pub trait Submachine<F: FieldElement> {
...
-            .values_mut()
-            .for_each(|v| v.push(Elem::Field(0.into())));
+        self.cols.values_mut().for_each(|v| v.push(0.into()));
     }
 
     /// Push a dummy block to the trace.
     /// A dummy block is a copy of the first block, with the final row updates applied to it, and selectors set to 0.
-    fn push_dummy_block(&mut self, machine_max_degree: usize, size: u32, selectors: &'static str) {
+    fn push_dummy_block(&mut self, machine_max_degree: u32, size: u32, selectors: &'static str) {
+        // TODO: use Optional selector argument
```
leonardoalt, line 193: here or later?

riscv-executor/src/lib.rs
pacheco: yes, was actually trying it out, doing it in a separate PR

riscv-executor/src/lib.rs
pacheco: same

riscv-executor/src/lib.rs
pacheco: `submachine_op` already does the test so its not "wrong", but indeed, i'll move it inside the test

riscv-executor/src/lib.rs
pacheco: submachine_op is always only in trace mode, so the test is done inside it (cleaner code and less possibility of mistakes)

riscv-executor/src/lib.rs
pacheco: yeah, ill inline them

riscv-executor/src/submachines.rs
pacheco: removed the TODO changing the code here a bit

riscv-executor/src/lib.rs
pacheco: fixed the trailing comma

leonardoalt (approved):

PR #2208, opened 2024-12-09
remove poseidon operation id

pacheco:
In cases where the operation id was defined as a witness column, it seems the optimizer didn't remove them. So this can save a witness column (and lookup parameter) in some cases (e.g., poseidon machines)


leonardoalt (approved):

pacheco:
I don't know about all cases (e.g., fixed machines where opid = `[0]*`), but in the case of poseidon, where operation id was defined as a witness column, the optimized pil still contained the column and argument

pacheco:
ok, in the case of the fixed machines, it indeed was optimized away

PR #2211, opened 2024-12-09
Fix CI bench

Schaeff:
All cargo.toml must have this `bench = false`. This will be caught in pr tests in the future.

### DIFF
--- a/asmopt/Cargo.toml
+++ b/asmopt/Cargo.toml
@@ -12,3 +12,6 @@ powdr-analysis.workspace = true
 powdr-importer.workspace = true
 powdr-pilopt.workspace = true
 powdr-parser.workspace = true
+
+[lib]
+bench = false # See https://github.com/bheisler/criterion.rs/issues/458


asmopt/Cargo.toml
```diff
@@ -11,4 +11,7 @@ powdr-ast.workspace = true
 powdr-analysis.workspace = true
 powdr-importer.workspace = true
 powdr-pilopt.workspace = true
-powdr-parser.workspace = true
\ No newline at end of file
+powdr-parser.workspace = true
+
+[lib]
+bench = false # See https://github.com/bheisler/criterion.rs/issues/458
```
chriseth, line 17: What happens if we ever add a benchmark? Why is this not needed in all the other crates?

Schaeff: It's needed in all other crates, it was added in #2198 
This does not prevent having Criterion benchmarks, which is what we use. This disables `libtest` benchmarks, which we do not use. From the attached link:

>By default, Cargo implicitly adds a libtest benchmark harness to your crate when benchmarking, to handle any #[bench] functions, even if you have none. It compiles and runs this executable first, before any of the other benchmarks. Normally, this is fine - it detects that there are no libtest benchmarks to execute and exits, allowing Cargo to move on to the real benchmarks. Unfortunately, it checks the command-line arguments first, and panics when it finds one it doesn't understand. This causes Cargo to stop benchmarking early, and it never executes the Criterion.rs benchmarks.

chriseth (approved):

PR #2212, opened 2024-12-09
Expressions and solving routines.

chriseth:
This module is an equivalent of the existing affine_expression.rs, but for compile-time execution on symbolic values instead of run-time execution on concrete values.

Using the operators defined on that that type, you can build a SymbolicAffineExpression from a polynomial identity and then use `.solve()` to try to solve for one unknown variable. The result (instead of a concrete assignment as in affine_expression.rs) is a SymbolicExpression, i.e. a complex expression involving variables (assumed to have a concrete value known at run time), constants and certain operators on them.

The idea is that SymbolicAffineExpression is used on polynomial identities in turn and solving for one cell after the other in the trace. The resulting SymbolicExpression can be translated to rust or pil.


executor/src/witgen/jit/affine_symbolic_expression.rs
```diff
@@ -0,0 +1,430 @@
...
+                var.clone(),
+                masked.integer_div(&coeff.into()),
+            ));
+        }
+
+        if covered_bits >= T::modulus() {
+            return vec![];
+        }
+
+        // TODO we need to add an assertion that offset is covered by the masks.
```
chriseth, line 195: The problem is that `offset` is usually a run-time value. So we could do a run-time check and if the check fails, the system is conflicting. Or we just continue to solve and ignore the conflict...

lvella:
Please provide a minimal description.

executor/src/witgen/jit/affine_symbolic_expression.rs
```diff
@@ -0,0 +1,550 @@
...
+        for (var, coeff) in rhs.coefficients.iter() {
+            coefficients
+                .entry(var.clone())
+                .and_modify(|f| *f = &*f + coeff)
+                .or_insert_with(|| coeff.clone());
+        }
+        let coefficients = coefficients
+            .into_iter()
+            .filter(|(_, f)| !f.is_known_zero())
+            .collect();
```
lvella, line 321: Can use [`retain()`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.retain) instead.

But maybe it would be even better to use [`remove_entry()`](https://doc.rust-lang.org/std/collections/btree_map/struct.OccupiedEntry.html#method.remove_entry) on the previous loop, but probably wouldn't look so nice and idiomatic.

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -0,0 +1,314 @@
...
+#[derive(Debug, Clone)]
+pub enum SymbolicExpression<T: FieldElement, V> {
+    /// A concrete constant value known at compile time.
+    Concrete(T),
+    /// A symbolic value known at run-time, referencing either a cell or a local variable.
+    Variable(V, Option<RangeConstraint<T>>),
+    BinaryOperation(
+        Box<Self>,
+        BinaryOperator,
+        Box<Self>,
```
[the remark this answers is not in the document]
chriseth: Yes, but not thoroughly. We probably won't need thread-safety, so yeah, Rc is probably justified, especially given the many clones.

executor/src/witgen/jit/affine_symbolic_expression.rs
```diff
@@ -0,0 +1,550 @@
...
+}
+
+/// Represents an expression `a_1 * x_1 + ... + a_k * x_k + offset`,
+/// where the `a_i` and `offset` are symbolic expressions, i.e. values known at run-time,
+/// and the `x_i` are unknown variables.
+#[derive(Debug, Clone)]
+pub struct AffineSymbolicExpression<T: FieldElement, V> {
+    coefficients: BTreeMap<V, SymbolicExpression<T, V>>,
+    offset: SymbolicExpression<T, V>,
+}
```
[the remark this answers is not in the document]
chriseth: It is affine in the unknown variables. I mean in the end, the known variables or not really variables any more, once they are known...

affine is essentially linear + constant

executor/src/witgen/jit/affine_symbolic_expression.rs
```diff
@@ -0,0 +1,550 @@
...
+impl<T: FieldElement, V: Ord + Clone + Display> AffineSymbolicExpression<T, V> {
+    pub fn from_known_variable(var: V) -> Self {
+        SymbolicExpression::from_var(var).into()
+    }
+    pub fn from_unknown_variable(var: V) -> Self {
+        AffineSymbolicExpression {
+            coefficients: [(var, T::from(1).into())].into_iter().collect(),
+            offset: SymbolicExpression::from(T::from(0)),
+        }
+    }
```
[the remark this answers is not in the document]
chriseth: I wondered about that as well, but in the end, both will be `Cell`s, i.e. references to a cell in the trace. Some of them are known, some of them are not, and it will even change while we progress.

In the end, what I call `variable` here is just another name for a "terminal" in the symbolic system that is not a constant known value.

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -0,0 +1,314 @@
...
+            UnaryOperator::Neg => write!(f, "-"),
+        }
+    }
+}
+
+impl<T: FieldElement, V> SymbolicExpression<T, V> {
+    pub fn from_var(name: V) -> Self {
+        SymbolicExpression::Variable(name, None)
+    }
+}
```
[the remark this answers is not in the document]
chriseth: Moved.

executor/src/witgen/jit/symbolic_expression.rs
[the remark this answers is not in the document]
chriseth: Changed to Rc.

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -0,0 +1,314 @@
...
+
+impl<T: FieldElement, V: Clone> BitAnd for &SymbolicExpression<T, V> {
+    type Output = SymbolicExpression<T, V>;
+
+    fn bitand(self, rhs: Self) -> Self::Output {
+        if let (SymbolicExpression::Concrete(a), SymbolicExpression::Concrete(b)) = (self, rhs) {
+            SymbolicExpression::Concrete(T::from(a.to_integer() & b.to_integer()))
+        } else if self.is_known_zero() || rhs.is_known_zero() {
+            SymbolicExpression::Concrete(T::from(0))
+        } else {
```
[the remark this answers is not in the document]
chriseth: You mean check if `self` or `rhs` has "all bits set" and in that case, return the other value?

That is tricky because we are usually dealing with prime fields and if we have "all bits set", we are outside of the field.

executor/src/witgen/jit/affine_symbolic_expression.rs
```diff
@@ -0,0 +1,544 @@
...
+                let value = self.offset.field_div(&coeff.neg());
+                let assignment = Effect::Assignment(var.clone(), value);
+                if let Some(coeff) = coeff.try_to_number() {
+                    assert!(!coeff.is_zero(), "Zero coefficient has not been removed.");
+                    return Ok(vec![assignment]);
+                } else {
+                    return Ok(vec![
+                        Assertion::assert_is_nonzero(coeff.clone()),
+                        assignment,
+                    ]);
```
georgwiese, line 175: I've been thinking about this example:
`instr_add * (a + b - c) + instr_sub * (a - b - c) = 0`

What if both `instr_add` and `instr_sub` are zero?

Assuming the instruction flags and `a` and `b` are known, the affine expression would be:
`(-instr_add - instr_add) * c + (instr_add * (a + b) + instr_sub * (a - b)) = 0`

Solving for `c` yields:
`c := -(instr_add * (a + b) + instr_sub * (a - b)) / (-instr_add - instr_add)`

Which would be a division by zero, which triggers the assertion. I think what this means is that `0 / 0` should be OK, because it means that the constraint has effectively been disabled and the list of effects should be empty!

executor/src/witgen/jit/affine_symbolic_expression.rs
```diff
@@ -0,0 +1,544 @@
...
+                }
+            }
+            _ => {}
```
georgwiese, line 178: Let's move the rest of the function inside here :) That way you can remove a bunch of `return` statements.

executor/src/witgen/jit/affine_symbolic_expression.rs
georgwiese: > the list of effects should be empty

I guess not, because we don't know the value of the offset at compile time. Instead, we'd want to generate code like this, right?

```rs
let numerator = -(instr_add * (a + b) + instr_sub * (a - b));
let denominator = (-instr_add - instr_add);
if denominator != 0 {
  c = numerator / denominator;
} else {
  assert_eq!(numerator, 0);
}
```

chriseth: Yeah, just realized the same in a different PR. I would suggest to leave that case open until we hit it, because we actually cannot set `c` to `known` unconditionally.

executor/src/witgen/jit/affine_symbolic_expression.rs
```diff
@@ -0,0 +1,544 @@
...
+    type Error = ();
+
+    fn try_from(value: &AffineSymbolicExpression<T, V>) -> Result<Self, Self::Error> {
+        if value.coefficients.is_empty() {
+            Ok(value.offset.clone())
+        } else {
+            Err(())
+        }
+    }
+}
```
georgwiese, line 383: This doesn't appear to be used. In what situation would you need it?

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -0,0 +1,319 @@
...
+                }
+            }
+            SymbolicExpression::Variable(name, _) => write!(f, "{name}"),
+            SymbolicExpression::BinaryOperation(lhs, op, rhs, _) => {
+                write!(f, "({lhs} {op} {rhs})")
+            }
+            SymbolicExpression::UnaryOperation(op, expr, _) => write!(f, "{op}({expr})"),
+        }
+    }
+}
```
georgwiese, line 100: I think the use of parenthesis is inconsistent here: For `BinaryOperation`, you put them around the result, but not the operands, for `UnaryOperation`, you do the opposite, and for negative concrete numbers you don't put any.

So you'd get strings like this:
`-(((a * -5) + -(b)))`

I think it makes sense to always put parenthesis around the result? Then this would be:
`(-((a * (-5)) + (-b)))`

georgwiese: BTW, are these implementations for informational purposes, or are these supposed to be valid Rust?

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    pub fn range_constraint(&self) -> Option<RangeConstraint<T>> {
+        match self {
+            SymbolicExpression::Concrete(v) => Some(RangeConstraint::from_value(*v)),
+            SymbolicExpression::Variable(.., rc)
+            | SymbolicExpression::BinaryOperation(.., rc)
+            | SymbolicExpression::UnaryOperation(.., rc) => rc.clone(),
+        }
+    }
```
georgwiese, line 71: I don't understand yet what we need the range constraints for if we know the value at run-time! As far as I can see, we never set it for `Variable`, and it is only really used once in `AffineSymbolicExpression::transfer_constraints` to get the range constraint of the offset. In the test, the offset is a concrete value.

So, I'm not sure if we need it at all, but if we do, I'd expect that this function receives a `&impl RangeConstraintSet<T>` and recursively computes the constraint, rather than having to store it in the variables & operations!

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -0,0 +1,319 @@
...
+
+/// A value that is known at run-time, defined through a complex expression
+/// involving known cells or variables and compile-time constants.
+/// Each of the sub-expressions can have its own range constraint.
+#[derive(Debug, Clone)]
+pub enum SymbolicExpression<T: FieldElement, V> {
+    /// A concrete constant value known at compile time.
+    Concrete(T),
+    /// A symbolic value known at run-time, referencing either a cell or a local variable.
+    Variable(V, Option<RangeConstraint<T>>),
```
georgwiese, line 19: Maybe we call it something else? Like Lucas, I did take a while to understand why coefficients in `AffineSymbolicExpression` can include variables. Maybe `RuntimeValue`?

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -0,0 +1,319 @@
...
+use crate::witgen::range_constraints::RangeConstraint;
+
+/// A value that is known at run-time, defined through a complex expression
+/// involving known cells or variables and compile-time constants.
+/// Each of the sub-expressions can have its own range constraint.
+#[derive(Debug, Clone)]
+pub enum SymbolicExpression<T: FieldElement, V> {
+    /// A concrete constant value known at compile time.
+    Concrete(T),
+    /// A symbolic value known at run-time, referencing either a cell or a local variable.
```
georgwiese, line 18: What's a "local variable"?

In my current understanding, this could be:
- A cell
- A scalar public (kind of a global variable)
- A challenge (similar to scalar public, but known from the start)

Would "local variable" by an intermediate column?

georgwiese (approved):
Exciting stuff!

georgwiese (commented):
Opps, just meant to comment :P

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -0,0 +1,314 @@
...
+    pub fn is_known_zero(&self) -> bool {
+        self.try_to_number().map_or(false, |n| n.is_zero())
+    }
+
+    pub fn is_known_one(&self) -> bool {
+        self.try_to_number().map_or(false, |n| n.is_one())
+    }
+
+    pub fn is_known_minus_one(&self) -> bool {
+        self.try_to_number().map_or(false, |n| n == -T::from(1))
```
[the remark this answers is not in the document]
lvella: No.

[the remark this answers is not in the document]
lvella: No.

executor/src/witgen/jit/symbolic_expression.rs
chriseth: No, just informational, I'll add a comment.

chriseth: I'll put parentheses, but not around unary, since it's still unambiguous

executor/src/witgen/jit/symbolic_expression.rs
chriseth: Oh that's a good point, we'll need a constructor for variables where the range constraint is known. This will then be done from the outer interface that also knows the global range constraints etc.

I think it's more convenient to just compute it along.

executor/src/witgen/jit/symbolic_expression.rs
chriseth: I think I took this comment from the original pull request. There, I introduce local variables in the rust code to implement lookups, because it as more convenient than using the cells right away.

I mean this is a very generic data structure and variable is just a symbolic reference to something. We could also use `Symbol` instead of `Variable`, if you thin that captures it better that it might be known or might be unknown.

executor/src/witgen/jit/symbolic_expression.rs
chriseth: Oh now I know what you mean, I mixed the old impl with a new one and now it's confusing. let me fix.

executor/src/witgen/jit/symbolic_expression.rs
georgwiese: Makes sense that the code here is more generic and does not need to mention stuff like challenges. And yes, I think `Symbol` would be better!

executor/src/witgen/jit/affine_symbolic_expression.rs
```diff
@@ -0,0 +1,532 @@
...
+                    !coeff.is_known_zero(),
+                    "Zero coefficient has not been removed."
+                );
+                if coeff.is_known_nonzero() {
+                    let value = self.offset.field_div(&-coeff);
+                    vec![Effect::Assignment(var.clone(), value)]
+                } else {
+                    // We can only solve this if we know that the coefficient cannot be zero.
+                    // TODO We could do this by adding a runtime condition
+                    vec![]
```
georgwiese, line 177: Wait, so now (with the change of cad7f37b3b3e02686aab364bc7482b89310bf707), it wouldn't be able to solve `A * B = 5` anymore for a known `B`, right? Because `B` would not have a compile-time range constraint, so `coeff.is_known_nonzero()` would be false?

I think the way it was before is fine, just that the `Assertion::assert_is_nonzero(coeff.clone())` was overly strict (which I think we can fix later).

executor/src/witgen/range_constraints.rs
```diff
@@ -82,6 +82,18 @@ impl<T: FieldElement> RangeConstraint<T> {
...
+    /// Returns true if `v` is an allowed value for this range constraint.
+    pub fn allows_value(&self, v: T) -> bool {
+        let wrapping = self.min > self.max;
+        if (!wrapping && (v < self.min || v > self.max))
+            || (wrapping && (v < self.min && v > self.max))
+        {
+            false
+        } else {
+            v.to_integer() & self.mask == v.to_integer()
+        }
```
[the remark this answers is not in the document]
chriseth: yep, I'll change it but I will put the `min <= max` case first :)

executor/src/witgen/jit/affine_symbolic_expression.rs
georgwiese: Hmm, I think it is fine like this for now, but it doesn't cover all cases. I think [this example](https://github.com/powdr-labs/powdr/pull/2212#discussion_r1879801242) still wouldn't work, because neither the numerator nor denominator would be known to be (non)zero at compile time.

For context, the affine expression would be:
`(-instr_add - instr_sub) * c + (instr_add * (a + b) + instr_sub * (a - b)) = 0`

Assuming binary constraints on the instruction flags, what we have here is that:
- `coeff == 0` $\Rightarrow$ `instr_add == 0 && instr_sub == 0` $\Rightarrow$ `offset == 0`.
- But we'll only know the instruction flags at runtime, so at runtime we need to check the value of `coeff` and:
  - Do nothing if it zero
  - Otherwise assign `c := -offset / coeff`

But let's iterate once we get there...

executor/src/witgen/jit/symbolic_expression.rs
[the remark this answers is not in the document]
chriseth: Yeah, I think none of the solutions are particularly nice and they need a lot of code which is probably not worth it.

georgwiese (approved):
LGTM!
🚀         🌔

executor/src/witgen/jit/affine_symbolic_expression.rs
chriseth: Yep!

Issue #2210, opened 2024-12-09
Creating a new powdrVM project from command line has no .gitignore


MuhtasimTanmoy:
While bootstrapping a new project using `cargo-powdr new my-host` has missing `.gitignore` file.

As the build file is pretty huge need to manually add `.gitignore` which should have been part of the template.

PR #2213, opened 2024-12-09
cargo-powdr gitignore add on template

MuhtasimTanmoy:
https://github.com/powdr-labs/powdr/issues/2210


leonardoalt:
Thanks! I don't think we should add the Rust one for now, but we could at least add Cargo.lock for both host and guest.

leonardoalt:
Also, we recently fixed something in `main`, can you please merge `main` into your branch

MuhtasimTanmoy:
> we could at least add Cargo.lock for both host and guest.

What about `powdr-target`?

MuhtasimTanmoy:
> Can you please merge main into your branch.

Rebased.

Please check.

leonardoalt:
> What about `powdr-target`?

yea let's do that one too

MuhtasimTanmoy:
@leonardoalt Already done.

leonardoalt (approved):
Thanks for the contribution!

PR #2214, opened 2024-12-10
fix udeps and add to pr-tests

leonardoalt:
This PR fixes udeps again and finally adds it to the pr-tests so it fails in a PR already.

### DIFF
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -23,57 +23,57 @@ jobs:
     runs-on: warp-ubuntu-2404-x64-8x
 
     steps:
-    - uses: actions/checkout@v4
-      with:
-        submodules: recursive
-    - name: ⚡ Restore rust cache
-      id: cache
-      uses: WarpBuilds/cache/restore@v1
-      with:
-        path: |
-          ~/.cargo/registry/index/
-          ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-          Cargo.lock
-        key: ${{ runner.os }}-cargo-pr-tests
-    - name: Date of the restored cache
-      run: cat target/cache-build-date.txt
-      continue-on-error: true
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: ⚡ Restore rust cache
+        id: cache
+        uses: WarpBuilds/cache/restore@v1
+        with:
+          path: |
+            ~/.cargo/registry/index/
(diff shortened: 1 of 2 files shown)


asmopt/Cargo.toml
```diff
@@ -9,7 +9,6 @@ repository.workspace = true
 [dependencies]
 powdr-ast.workspace = true
 powdr-analysis.workspace = true
-powdr-importer.workspace = true
```
leonardoalt, line 12: this is the fix

.github/workflows/pr-tests.yml
```diff
@@ -247,45 +247,66 @@ jobs:
...
+          name: Benchmarks
+          tool: "cargo"
+          output-file-path: output.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          alert-threshold: "120%"
+          comment-on-alert: true
+          summary-always: true
+
+  udeps:
```
leonardoalt, line 293: This was added now.

.github/workflows/pr-tests.yml
```diff
@@ -23,220 +23,220 @@ jobs:
...
-          ~/.cargo/registry/index/
-          ~/.cargo/registry/cache/
-          ~/.cargo/git/db/
-          target/
-          Cargo.lock
-        key: ${{ runner.os }}-cargo-pr-tests
-    - name: Date of the restored cache
-      run: cat target/cache-build-date.txt
-      continue-on-error: true
+      - uses: actions/checkout@v4
```
leonardoalt, line 26: this was all fmt, no idea why

Schaeff (approved):

.github/workflows/pr-tests.yml
Schaeff: you mean `cargo fmt` ? Would it mean it looks at yml files as well?

Schaeff (approved):

.github/workflows/pr-tests.yml
```diff
@@ -247,45 +247,65 @@ jobs:
...
+          tool: "cargo"
+          output-file-path: output.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          alert-threshold: "120%"
+          comment-on-alert: true
+          summary-always: true
+
+  udeps:
+    needs: build
```
chriseth, line 294: Why does it need build?

chriseth (approved):

PR #2222, opened 2024-12-10
riscv-executor: handle machines that have been optimized away

pacheco:


leonardoalt (approved):

leonardoalt (changes requested):
instruction tests failed locally (@pacheco told me)

pacheco:
Yeah, the original problem you mentioned is solved by this PR (machines being optimized away). 
What fails is a different issue with the same origin: asm optimizer removes instructions, which messes with selector assignment, which breaks manual witgen. To solve this the executor needs to take into account the optimized PIL...

pacheco:
I'm closing this for now

PR #2223, opened 2024-12-10
pilopt: optimize until fixpoint

gzanitti:


gzanitti:
Replaced by #2225 , which can run bench tests.

PR #2224, opened 2024-12-11
pilopt: equal-constrained witness columns removal

gzanitti:


pilopt/src/lib.rs
```diff
@@ -120,33 +121,41 @@ fn remove_unreferenced_definitions<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
     #[allow(clippy::iter_over_hash_type)]
     for (name, (symbol, _)) in &pil_file.intermediate_columns {
-        symbol.array_elements().for_each(|(_, id)| {
-            poly_id_to_definition_name.insert(id, name);
-        });
+        symbol
+            .array_elements()
+            .enumerate()
+            .for_each(|(idx, (_, id))| {
+                let array_pos = if symbol.is_array() { Some(idx) } else { None };
```
chriseth, line 148:
```suggestion
                let array_pos = symbol.is_array().then_some(idx);
```

pilopt/src/lib.rs
```diff
@@ -688,3 +718,103 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+                reference.name = replacement_name.clone();
+            }
+        }
+    });
+}
+
+fn resolve_transitive_substitutions(
+    subs: Vec<((String, PolyID), (String, PolyID))>,
+) -> Vec<((String, PolyID), (String, PolyID))> {
+    let mut result = subs.clone();
```
chriseth, line 800: Why don't you make `subs` `mut` and use it directly?

pilopt/src/lib.rs
```diff
@@ -688,3 +718,103 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+    let mut changed = true;
+
+    while changed {
+        changed = false;
+        for i in 0..result.len() {
+            let (_, target1) = &result[i].1;
+            if let Some(j) = result
+                .iter()
+                .position(|((_, source2), _)| source2 == target1)
```
chriseth, line 809: Maybe turn it into a HashMap or a BTreeMap instead? I mean the substitution lists won't be that long, but still...

pilopt/src/lib.rs
```diff
@@ -688,3 +718,103 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+            {
+                let ((name1, source1), _) = &result[i];
+                let (_, (name3, target2)) = &result[j];
+                result[i] = ((name1.clone(), *source1), (name3.clone(), *target2));
```
chriseth, line 813: I know there is the `remove trivial` step which we assume runs before this, but I think it would still be good to avoid an infinite loop in case a column is replaced by itself.

pilopt/tests/optimizer.rs
```diff
@@ -375,3 +377,73 @@ fn handle_array_references_in_prover_functions() {
...
+        col witness a;
+        col witness b;
+        col witness c;
+        a = b;
+        b = c;
+        a + b + c = 5;
+    "#;
+    let expectation = r#"namespace N(65536);
+    col witness a;
+    N::a + N::a + N::a = 5;
```
chriseth, line 445: Nice!

pilopt/tests/optimizer.rs
```diff
@@ -375,3 +377,73 @@ fn handle_array_references_in_prover_functions() {
...
+    assert_eq!(optimized, expectation);
+}
+
+#[test]
+fn equal_constrained_array_elements() {
+    let input = r#"namespace N(65536);
+        col witness w[20];
+        w[4] = w[7];
+        w[3] = w[5];
+        w[7] + w[1] = 5;
```
chriseth, line 421: Can you add something like `col witness x; x = w[3];` here to see if the substitution at least works in one direction?

pilopt/src/lib.rs
```diff
@@ -121,33 +122,41 @@ fn remove_unreferenced_definitions<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
-        });
+        symbol
+            .array_elements()
+            .enumerate()
+            .for_each(|(idx, (_, id))| {
+                let array_pos = symbol.is_array().then_some(idx);
+                poly_id_to_definition_name.insert(id, (name, array_pos));
+            });
     }
     poly_id_to_definition_name
```
Schaeff, line 153: If this is already changing, we could turn it into something like

```rust
pil_file.definitions.iter().filter(|d| matches!(symbol.kind, SymbolKind::Poly(_))).chain(&pil_file.intermediate_columns).flat_map(|(name, (symbol, _))| symbol
            .array_elements()
            .enumerate()
            .map(|(idx, (_, id))| {
                let array_pos = symbol.is_array().then_some(idx);
                (id, (name, array_pos))
            })
   ).collect()
   ```
   
   to avoid side effects and code repetition

File: pilopt/src/lib.rs [hunk 8]
```diff
@@ -562,23 +571,44 @@ fn constrained_to_constant<T: FieldElement>(
...
+            Identity::Polynomial(PolynomialIdentity { expression, .. }) => match expression {
+                AlgebraicExpression::Number(n) => {
                     if *n == 0.into() {
-                        return Some(index);
+                        Some(index)
+                    } else {
+                        // Otherwise the constraint is not satisfiable,
+                        // but better to get the error elsewhere.
+                        None
                     }
```
Schaeff, line 589: `(*n == 0.into()).then_some(index)`

pilopt/src/lib.rs
```diff
@@ -562,23 +571,44 @@ fn constrained_to_constant<T: FieldElement>(
...
+                    ) = (left.as_ref(), right.as_ref())
+                    {
+                        if !left.next && !right.next && left == right {
+                            Some(index)
+                        } else {
+                            None
+                        }
+                    } else {
+                        None
+                    }
```
Schaeff, line 608: Why can't we just do `(left == right).then_some(index)`? Also, why don't we remove `x' = x'`?

Schaeff: Oh, is it because the other reduction turns cycles of equality into statements of the form `x = x`?

pilopt/src/lib.rs
```diff
@@ -717,3 +747,108 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+    match expression {
+        AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation {
+            left,
+            op: AlgebraicBinaryOperator::Sub,
+            right,
+        }) => match (left.as_ref(), right.as_ref()) {
+            (AlgebraicExpression::Reference(l), AlgebraicExpression::Reference(r)) => {
+                let is_valid = |x: &AlgebraicReference, left: bool| {
+                    x.is_witness()
+                        && !x.next
```
Schaeff, line 767: Could this be extended to `left.next == right.next`? Since `(x' == y') == (x == y)`?

pilopt/src/lib.rs
```diff
@@ -717,3 +747,108 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+                        && if left {
+                            poly_id_to_array_elem.get(&x.poly_id).unwrap().1.is_none()
```
Schaeff, line 769: So the lhs cannot be an array element? Why? These things should probably be explained in plain language

pilopt/src/lib.rs
gzanitti: You were right in your first comment. Fixed

pilopt/src/lib.rs
gzanitti: When we find identities like `x = y`, we want to remove one column and replace all its references with the other. However, if we allowed `x[0] = y`, removing x[0] would break the array x integrity since arrays need all their elements to "function properly". 

Added a small comment explaining this situation. Let me know if it's ok.

Schaeff (dismissed):
LGTM but maybe @pacheco wants to have a look at the riscv executor

riscv-executor/src/lib.rs
```diff
@@ -656,7 +656,7 @@ mod builder {
     }
 
     pub struct TraceBuilder<'b, F: FieldElement> {
-        trace: ExecutionTrace<F>,
+        pub trace: ExecutionTrace<F>,
```
pacheco, line 659: is this `pub` needed?

gzanitti: Sorry, didn't notice it. Fixed.

pacheco (approved):
lgtm too

leonardoalt:
test failing

PR #2225, opened 2024-12-11
pilopt: optimize until fixpoint

gzanitti:


backend/src/plonky3/stark.rs
```diff
@@ -465,28 +465,6 @@ mod tests {
         run_test(content);
     }
 
-    #[test]
-    fn two_tables() {
```
gzanitti, line 469: I am not 100% sure how to adapt this test :thinking:

pilopt/src/lib.rs
```diff
@@ -43,6 +48,43 @@ pub fn optimize<T: FieldElement>(mut pil_file: Analyzed<T>) -> Analyzed<T> {
     pil_file
 }
 
+fn hash_pil_state<T: FieldElement>(pil: &Analyzed<T>) -> u64 {
```
chriseth, line 51: Does it make sense to implement Hash for Analyzed? Would it be the correct implementation?

pilopt/src/lib.rs
```diff
@@ -22,16 +23,20 @@ use referenced_symbols::{ReferencedSymbols, SymbolReference};
...
+        pil_hash = hash_pil_state(&pil_file);
+        remove_constant_fixed_columns(&mut pil_file);
+        deduplicate_fixed_columns(&mut pil_file);
+        simplify_identities(&mut pil_file);
+        extract_constant_lookups(&mut pil_file);
+        remove_constant_witness_columns(&mut pil_file);
+        simplify_identities(&mut pil_file);
+        remove_trivial_identities(&mut pil_file);
+        remove_duplicate_identities(&mut pil_file);
+        remove_unreferenced_definitions(&mut pil_file);
```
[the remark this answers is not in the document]
gzanitti: The main idea behind making this first step out of the loop, was to remove everything that is trivially removable before even calculating the hash, because it will ultimately iterate over definitions, identities, etc.

pilopt/src/lib.rs
gzanitti: It wasn't my intention, because there are several fields in `Analyzed<T>` that don't interest me (because they will not be modified during the optimizations) and hashing them would not add anything to the solution. 

Perhaps if another use case for the Analyzed hash appears, it might be worth using it here directly as well, but otherwise, I don't see the advantage.

pilopt/src/lib.rs
```diff
@@ -43,6 +48,43 @@ pub fn optimize<T: FieldElement>(mut pil_file: Analyzed<T>) -> Analyzed<T> {
...
+fn hash_pil_state<T: FieldElement>(pil: &Analyzed<T>) -> u64 {
+    let mut hasher = DefaultHasher::new();
+
+    for identity in &pil.identities {
+        identity.hash(&mut hasher);
+    }
+
+    let mut keys: Vec<_> = pil.definitions.keys().collect();
+    keys.sort();
+    for key in keys {
```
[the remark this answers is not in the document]
chriseth: Yeah, whatever is easier :)

pilopt/src/lib.rs
chriseth: I'm a bit worried about us forgetting something here once we add it to Analyzed. The worry is almost as large when we implement Hash for Analyzed, but a bit smaller :)

What would be the fields that we don't need here?

pilopt/src/lib.rs
```diff
@@ -23,10 +23,9 @@ use referenced_symbols::{ReferencedSymbols, SymbolReference};
 
 pub fn optimize<T: FieldElement>(mut pil_file: Analyzed<T>) -> Analyzed<T> {
     let col_count_pre = (pil_file.commitment_count(), pil_file.constant_count());
-    let mut pil_hash = 0;
-    remove_unreferenced_definitions(&mut pil_file);
```
chriseth, line 25: You are right, probably a good idea to do that before computing the hash.

pilopt/src/lib.rs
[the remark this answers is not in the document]
chriseth: I mean you could also use `definitions_in_source_order`

pilopt/src/lib.rs
gzanitti: To be honest they are not a lot haha (`auto_added_symbols`, `source_order` & `solved_impls` right now).

Schaeff: If the only advantage is performance I'd hash the whole thing since optimizer performance is not our bottleneck now.

ast/src/analyzed/mod.rs
```diff
@@ -518,6 +518,57 @@ impl<T: FieldElement> Analyzed<T> {
     }
 }
 
+impl<T: FieldElement> Hash for Analyzed<T> {
```
Schaeff, line 521: Could this just be derived? If not, do we need the FieldElement bound?

backend/src/plonky3/stark.rs
```diff
@@ -414,7 +414,7 @@ mod tests {
             col witness x;
             col witness y;
             col witness z;
-            y - 1 = 0;
+            y - 1 = 1;
```
Schaeff, line 417: Why did this change?

pilopt/src/lib.rs
```diff
@@ -22,17 +23,32 @@ use referenced_symbols::{ReferencedSymbols, SymbolReference};
...
-    simplify_identities(&mut pil_file);
-    remove_trivial_identities(&mut pil_file);
-    remove_duplicate_identities(&mut pil_file);
-    remove_unreferenced_definitions(&mut pil_file);
+    loop {
+        let pil_hash = {
+            let mut hasher = DefaultHasher::new();
+            pil_file.hash(&mut hasher);
+            hasher.finish()
+        };
```
Schaeff, line 31: Let's have a `hash` function instead of a block since it's used twice

Schaeff:
I'm keen to have this but it's still unclear to me why we need a manual implementation of `Hash` on `Analyzed`? Also, this has a conflict

backend/src/plonky3/stark.rs
gzanitti: It no longer failed as it should. The final code with the new optimization is now (for the original code, without my modification):
```
namespace Add(8);
    col witness z;
    public outz = Add::z(7); 
```

which, if I understand correctly, is making public the 7th value of the Z column. Is this correct? In that case, vec![0] should be a valid value.

Anyway I changed the test: To make it fail and keep the expected panic, I changed vec![0] to vec![1]. This way we keep the code as it was and preserve the panic anyway.

Let me know if I'm assuming something incorrectly.

gzanitti:
> I'm keen to have this but it's still unclear to me why we need a manual implementation of `Hash` on `Analyzed`? Also, this has a conflict

Thank you, the conflict has been solved. As for the manual implementation, since Analyzed is composed of some HashMaps, it is not possible to derive Hash directly. So I think the only options are: Having an implementation for the whole structure, or modifying the struct so that it can be derived directly (changing HashMap for another structure that can derive Hash or wrapping it in one that implements Hash). Personally, I prefer option 1, but open to discussion. Same if you have any other idea :)

ast/src/analyzed/mod.rs
```diff
@@ -503,6 +503,57 @@ impl<T: FieldElement> Analyzed<T> {
     }
 }
 
+impl<T: Hash> Hash for Analyzed<T> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        for identity in &self.identities {
+            identity.hash(state);
+        }
```
Schaeff, line 510:
```suggestion
    self.identities.hash(state);
```
?

ast/src/analyzed/mod.rs
```diff
@@ -503,6 +503,57 @@ impl<T: FieldElement> Analyzed<T> {
...
+
+        for (key, (_, value)) in self.definitions.iter().sorted_by(|a, b| a.0.cmp(b.0)) {
+            key.hash(state);
+            if let Some(v) = value {
+                v.hash(state);
+            }
```
Schaeff, line 516:
```suggestion
    value.hash(state);
```
?

ast/src/analyzed/mod.rs
```diff
@@ -503,6 +503,57 @@ impl<T: FieldElement> Analyzed<T> {
     }
 }
 
+impl<T: Hash> Hash for Analyzed<T> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        for identity in &self.identities {
+            identity.hash(state);
+        }
+
+        for (key, (_, value)) in self.definitions.iter().sorted_by(|a, b| a.0.cmp(b.0)) {
```
Schaeff, line 512: Why is the `Symbol` ignored here?

ast/src/analyzed/mod.rs
```diff
@@ -503,6 +503,57 @@ impl<T: FieldElement> Analyzed<T> {
...
+
+        for (key, (_, value)) in self
+            .intermediate_columns
+            .iter()
+            .sorted_by(|a, b| a.0.cmp(b.0))
+        {
+            key.hash(state);
+            for v in value.iter() {
+                v.hash(state);
+            }
```
Schaeff, line 527:
```suggestion
    value.hash(state);
```
?

ast/src/analyzed/mod.rs
```diff
@@ -503,6 +503,57 @@ impl<T: FieldElement> Analyzed<T> {
...
+        }
+
+        for (key, (_, value)) in self.definitions.iter().sorted_by(|a, b| a.0.cmp(b.0)) {
+            key.hash(state);
+            if let Some(v) = value {
+                v.hash(state);
+            }
+        }
+
+        for (key, (_, value)) in self
```
Schaeff, line 519: Same, why is the symbol ignored here?

ast/src/analyzed/mod.rs
```diff
@@ -503,6 +503,57 @@ impl<T: FieldElement> Analyzed<T> {
...
+        {
+            key.hash(state);
+            for v in value.iter() {
+                v.hash(state);
+            }
+        }
+
+        for pf in &self.prover_functions {
+            pf.hash(state);
+        }
```
Schaeff, line 532:
```suggestion
    self.prover_functions.hash(state);
```
?

ast/src/analyzed/mod.rs
```diff
@@ -503,6 +503,57 @@ impl<T: FieldElement> Analyzed<T> {
...
+            .iter()
+            .sorted_by(|a, b| a.0.cmp(b.0))
+        {
+            key.hash(state);
+            value.hash(state);
+        }
+
+        for _impl in &self.trait_impls {
+            _impl.hash(state);
+        }
```
Schaeff, line 545:
```suggestion
    self.trait_impls.hash(state);
```
?

ast/src/analyzed/mod.rs
```diff
@@ -503,6 +503,57 @@ impl<T: FieldElement> Analyzed<T> {
...
+
+        for order in &self.source_order {
+            order.hash(state);
+        }
```
Schaeff, line 549:
```suggestion
    self.source_order.hash(state);
```
?

Schaeff (commented):
Some simplifications in the hash implementation. I'm slightly worried about unexpected collisions if we change Analysis, maybe we should have a warning on the Analysis struct?

backend/src/plonky3/stark.rs
Schaeff: The optimizer seems wrong here? In the original code:

```
y - 1 = 0
x = 0
x + y = z
```

therefore
```
y = 1
x = 0
z = x + y = 1
```

In the output you shared, the optimizer does not constrain `z` at all, but `z` should be 1. We should keep the test as is and fix the optimizer if it's wrong.

ast/src/analyzed/mod.rs
gzanitti: Sorry, I left them out when the hash function was specific to the optimization (because they didn't contribute anything to the uniqueness of the hash in that case) and forgot to add them now that it is the general derivation. Fixed.

chriseth:
I'm wondering, if we just iterate over `source_order` and hash everything it contains (of course dereferencing the IDs) - can we miss anything important?

backend/src/plonky3/stark.rs
gzanitti: You were right, there was an optimization that failed (and in fact I think in my explanation above I used one case to justify the other, sorry :sweat_smile: ). remove_constant_witness_columns removed cases where the column then declared a public as in this case. Added a small check to avoid it.

pilopt/src/lib.rs
```diff
@@ -44,6 +53,41 @@ pub fn optimize<T: FieldElement>(mut pil_file: Analyzed<T>) -> Analyzed<T> {
...
+
+    for so in &pil_file.source_order {
+        match so {
+            StatementIdentifier::Definition(d) => {
+                d.hash(&mut hasher);
+                if let Some(def) = pil_file.definitions.get(d) {
+                    def.hash(&mut hasher);
+                } else if let Some(def) = pil_file.intermediate_columns.get(d) {
+                    def.hash(&mut hasher);
+                }
```
Schaeff, line 67: else unreachable?

pilopt/src/lib.rs
```diff
@@ -457,9 +501,14 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
         .filter_map(constrained_to_constant)
         .collect::<Vec<((String, PolyID), _)>>();
     // We cannot remove arrays or array elements, so filter them out.
```
Schaeff, line 503: This comment should be updated accordingly

Schaeff (approved):

PR #2230, opened 2024-12-13
Use padded bitvec.

chriseth:


pacheco:
If I'm not mistaken, the `bit_vec` crate that we use already works like this (multiples of some unsigned type, and `u32` by default)

pacheco:
Mmm ok i guess its about the row alignment... but I don't see any "specialized" methods for setting multiple rows the same value, so what is the benefit here?

chriseth:
Assume we have jited code that sets all columns in a row of 8 column to known.
Then using the bitvec crate, that would be something like
```
for i in 0..column_count {
  known.set(row * column_count + i, 1);
}
```
where `set` is roughly implemented as (assuming we have enough space)
```
data[bit / 32] |= 1 << (bit % 32);
```

Now if we unroll the loop, there is no structure in the assignments, since `(row * column_count + i) / 32` for `i` from `0` to `column-count` can be any value.

For the padded bitvec, this is different. We call `known.set(row, i, 1)`, and it's implemented as
```
data[row * words_per_row + col / 32] |= 1 << (col % 32);
```

If you unroll the loop now, the index `row * words_per_row + col / 32` will be the same and it will be optimized to a single assignment, hopefully something like `data[row * words_per_row] |= (1 << column_count) - 1;`

pacheco:
this depends on the compiler doing this optimization right? doesn't it make sense to add a method to do it?

chriseth:
Setting the full row to known was just an example, it works with any other combination in the same row. In the end, we cannot transfer functions across the FFI boundary (at least it's complicated) so my idea is to just pass this as a `&mut [32]` instead anyway.

pacheco (approved):

PR #2232, opened 2024-12-13
make the RISCV executor look at the optimized PIL for link information

pacheco:
This PR modifies the RISCV executor to look into the optimized PIL for lookup/permutation definitions.
In particular, to figure out which selectors are assigned for each link, but also with the intent of later integration with witgen (which needs the identity ids for "submachine calls")


riscv-executor/src/memory.rs
```diff
@@ -51,16 +51,24 @@ impl<F: FieldElement> Submachine<F> for MemoryMachine<F> {
...
             panic!()
         };
+        // get the idx from the selector
+        let selector_idx = selector
+            .map(|s| {
+                let start = s.find('[').unwrap() + 1;
+                let end = s.find(']').unwrap();
+                s[start..end].parse::<u8>().unwrap()
+            })
+            .unwrap();
```
leonardoalt, line 65: This seems kinda costly if it has to run every time? Wouldn't it be better to have a map in the caller and just pass the number like before?

riscv-executor/src/submachines.rs
```diff
@@ -4,17 +4,20 @@ use powdr_number::{FieldElement, LargeInt};
 
 use std::collections::HashMap;
 
+fn only_column_name(name: &str) -> &str {
+    name.rfind("::").map(|i| &name[i + 2..]).unwrap_or(name)
```
leonardoalt, line 8: maybe leave a comment on why i + 2?

riscv-executor/src/submachines.rs
```diff
@@ -4,17 +4,20 @@ use powdr_number::{FieldElement, LargeInt};
...
     /// Which of the witness columns are selectors, if any
     const SELECTORS: &'static str;
     /// Row block size
     const BLOCK_SIZE: u32;
     /// Add an operation to the submachine trace
     fn add_operation<F: FieldElement>(
         trace: &mut SubmachineTrace<F>,
-        // pil lookup RHS values (including selector idx, if present)
-        selector_idx: Option<u8>,
+        selector: Option<&str>,
```
leonardoalt, line 20: same question here re number vs string, string sounds costly?

riscv-executor/src/pil.rs
```diff
@@ -0,0 +1,144 @@
...
+use powdr_ast::{
+    analyzed::{AlgebraicExpression, Analyzed, Identity, PermutationIdentity},
+    parsed::visitor::ExpressionVisitable,
+};
+use powdr_number::FieldElement;
+
+pub fn links_from_pil<F: FieldElement>(pil: &Analyzed<F>) -> Vec<Identity<F>> {
+    pil.identities
+        .iter()
+        .filter(|id| matches!(id, Identity::Permutation(_) | Identity::Lookup(_)))
```
leonardoalt, line 12: I guess later we'll also need to support the Phantom versions, and/or bus sends/receives. It might actually be simpler at that point.

riscv-executor/src/pil.rs
```diff
@@ -0,0 +1,144 @@
...
+                return Some(extract_selector(perm));
+            }
+        }
+        if let Identity::Lookup(lookup) = l {
+            if lookup.id == id {
+                return None;
+            }
+        }
+    }
+    panic!("identity {id} doesnt exist")
```
leonardoalt, line 50:
```suggestion
    panic!("identity {id} doesn't exist")
```

riscv-executor/src/pil.rs
pacheco: yeah, I didn't really look into it here, but should not be to bad I imagine

riscv-executor/src/pil.rs
```diff
@@ -0,0 +1,144 @@
...
+pub fn selector_for_link<F: FieldElement>(links: &[Identity<F>], id: u64) -> Option<String> {
+    for l in links.iter() {
+        if let Identity::Permutation(perm) = l {
+            if perm.id == id {
+                return Some(extract_selector(perm));
+            }
+        }
+        if let Identity::Lookup(lookup) = l {
+            if lookup.id == id {
+                return None;
```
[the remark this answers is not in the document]
pacheco: only Permutations have a selector

riscv-executor/src/memory.rs
pacheco: yes, this is ugly, i'm working on a separate PR for these details/performance (this here should be cached)

riscv-executor/src/lib.rs
```diff
@@ -2597,13 +2799,26 @@ fn execute_inner<F: FieldElement>(
...
+    let mut count = 0;
     loop {
         let stm = statements[curr_pc as usize];
 
         log::trace!("l {curr_pc}: {stm}",);
 
         e.step += 4;
 
+        count += 1;
+        if count % 10000 == 0 {
```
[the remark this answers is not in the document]
pacheco: the message is printed every second, this is just how frequently we sample the clock

riscv-executor/src/lib.rs
```diff
@@ -48,28 +50,123 @@ struct MainOp<F: FieldElement>(&'static str, u32, Vec<F>);
...
-    main_publics,
-    main_binary,
-    main_shift,
-    main_split_gl,
-    main_poseidon_gl,
-    // main_poseidon2_gl,
-    // main_keccakf,
-    // main_arith,
+/// Enum with asm machine RISCV instruction. Helps avoid using raw strings in the code.
+macro_rules! instructions {
```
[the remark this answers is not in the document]
pacheco: it won't be an enum then...

riscv-executor/src/lib.rs
```diff
@@ -944,17 +944,18 @@ mod builder {
...
+                // take each submachine and get its operations
+                .map(|(m, machine)| {
+                    let ops = self.trace.submachine_ops.remove(&m).unwrap_or_default();
+                    (m, machine.into_inner(), ops)
                 })
+                // handle submachines in parallel
                 .par_bridge()
                 .flat_map(|(m, mut machine, ops)| {
-                    // apply the operation to the submachine
+                    println!("machine: {:?}", m.name());
```
leonardoalt, line 957: leftover

riscv-executor/src/lib.rs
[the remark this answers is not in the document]
leonardoalt: ah ok

riscv-executor/src/lib.rs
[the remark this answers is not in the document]
leonardoalt: ok

leonardoalt (approved):

PR #2239, opened 2024-12-16
free input data

leonardoalt:
This PR allows for free runtime data besides free compile time data, based on `initial_memory`.

- Replaces `cbor` by `bincode` in prover queries (bincode is more efficient and cbor crashed with `u128`)
- Allows the prover to pass a runtime initial memory, only possible with continuations (from https://github.com/powdr-labs/powdr/pull/2251, was already merged into here, see commits list)
- Changes the `powdr` lib, which already always uses continuations, to always use this mechanism for prover data
- Provides a new stdin-stream-like function to read inputs in sequence, like other zkVMs.
- The function above is called `read_stdin` which I'm not super happy with, ideally it'd just be called `read` but the QueryCalldata function is already called `read`. I think we could just keep this as is and change later.


riscv/src/elf/mod.rs
```diff
@@ -190,6 +191,8 @@ fn load_elf(file_name: &Path) -> ElfProgram {
         text_labels: referenced_text_addrs,
         instructions: lifted_text_sections,
         entry_point: elf.entry as u32,
+        // TODO: properly load these from the ELF instead of hardcoding them.
+        prover_data_bounds: (0x10000000, 0x20000000),
```
leonardoalt, line 195: @lvella we still need to do this

riscv/Cargo.toml
```diff
@@ -56,6 +57,7 @@ serde = { version = "1.0", default-features = false, features = [
   "derive",
   "rc",
 ] }
+serde_cbor = "0.11.2"
```
leonardoalt, line 60: note to self: can probably be removed

powdr/src/lib.rs
```diff
@@ -125,7 +125,12 @@ impl Session {
         }
     }
 
-    pub fn write<S: serde::Serialize>(self, channel: u32, data: &S) -> Self {
+    pub fn write<S: serde::Serialize>(&mut self, data: &S) {
+        let bytes = bincode::serialize(&data).unwrap();
+        self.pipeline.add_prover_data(bytes);
+    }
+
+    pub fn write_fd<S: serde::Serialize>(self, channel: u32, data: &S) -> Self {
```
leonardoalt, line 133: Not too happy about these 2 functions being similar but ultimately quite different and based on different mechanisms...

powdr-test/examples/serialized-inputs/src/main.rs
```diff
@@ -0,0 +1,33 @@
...
+        .build();
+
+    session.write(&some_data);
+    session.write(&s);
+
+    // Fast dry run to test execution.
+    session.run();
+
+    // Uncomment to compute the proof.
+    //session.prove();
```
leonardoalt, line 32: I think it's fine to leave the proof out here to make this test faster, here we want to test inputs only anyway.

powdr/src/lib.rs
```diff
@@ -125,7 +125,12 @@ impl Session {
         }
     }
 
-    pub fn write<S: serde::Serialize>(self, channel: u32, data: &S) -> Self {
+    pub fn write<S: serde::Serialize>(&mut self, data: &S) {
```
leonardoalt, line 128: a bit annoying that this function doesn't take `self` and return `Self`, before this was nicer cus you could do
```rust
let session = Session::builder()
    ...
    .build()
    .write(...)
    .write(...);
```

With this we need to do
```rust
let mut session = Session::builder()
    ...
    .build();
session.write(...)
session.write(...);
```
We could keep the same interface, but then we'll need to `mem::take` `self.pipeline` here, change it, and make a new Session out of that. Kinda ugly too, we need to decide what's uglier.

Reviewer, lmk your opinion.

riscv/src/elf/mod.rs
pacheco: here?

riscv/Cargo.toml
pacheco: why not remove?

powdr/src/lib.rs
pacheco: why the need to mem::take the pipeline? should work fine if you take `mut self` no?

pipeline/src/pipeline.rs
```diff
@@ -322,8 +324,16 @@ impl<T: FieldElement> Pipeline<T> {
         self
     }
 
+    pub fn add_prover_data(&mut self, data: Vec<u8>) {
```
pacheco, line 327: I think we need to start adding documentation to the pipeline methods, most of them don't have it (in this PR, for the new and related methods)

pacheco: not sure about this name... its all prover data in the end right?

.github/workflows/pr-tests.yml
```diff
@@ -146,8 +146,10 @@ jobs:
         run: rustup target add riscv32imac-unknown-none-elf --toolchain nightly-2024-08-01-x86_64-unknown-linux-gnu
       - name: Install test dependencies
         run: sudo apt-get update && sudo apt-get install -y binutils-riscv64-unknown-elf lld
-      - name: Run examples
+      - name: Run examples with tagged powdr lib
```
pacheco, line 149: wdym tagged powdr lib?

pipeline/src/pipeline.rs
```diff
@@ -148,6 +148,7 @@ pub struct Pipeline<T: FieldElement> {
     arguments: Arguments<T>,
     /// The context for the host.
     host_context: HostContext,
+    initial_memory: Vec<Vec<u8>>,
```
pacheco, line 151: docstring?

powdr-test/examples/serialized-inputs/src/main.rs
pacheco: but proof with initial memory inputs tested somewhere else?

powdr/src/lib.rs
pacheco: we can't remove the old method because the new (free) one only works with continuations?

riscv/src/elf/mod.rs
lvella: Let me start a PR now. Can be merged before or after.

riscv-runtime/src/io.rs
```diff
@@ -7,6 +9,94 @@ use powdr_riscv_syscalls::Syscall;
...
+
+        unsafe {
+            // We skip the first page of the prover data, as it used as salt to
+            // randomize its merkle tree node.
+            let region_start: *const u32 = &__powdr_prover_data_start;
+            let data_start = region_start.byte_offset(POWDR_PAGE_SIZE);
+            let data_end: *const u32 = &__powdr_prover_data_end;
+
+            let prover_data_section =
+                slice::from_raw_parts(data_start, data_end.offset_from(data_start) as usize);
```
pacheco, line 51: i think only this statement is unsafe no?

riscv-runtime/src/io.rs
```diff
@@ -7,6 +9,94 @@ use powdr_riscv_syscalls::Syscall;
...
+        self.remaining_data = remaining;
+
+        // SAFETY: It is safe to cast an u32 slice to an u8 slice, which is the most general type.
+        unsafe {
+            let data_ptr = data.as_ptr() as *const u8;
+            Some(slice::from_raw_parts(data_ptr, len_bytes as usize))
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
```
pacheco, line 87: It should be cheap to turn this into an exact size iterator (either iterate over the sizes when creating the iterator, or write the object count as the second word (right after the total number of words). Maybe its worth it.

riscv-runtime/src/io.rs
lvella: No, `byte_offset` is also unsafe.

The unsafe block could encompass just the two lines, but when I wrote I felt that taking the pointers and trimming the slice also fit the feng shui of the unsafe block.

riscv-runtime/src/io.rs
pacheco: but then, we don't really expose this iterator to the user, only the read function, so I think it doesn't matter

riscv-runtime/src/io.rs
```diff
@@ -51,6 +141,16 @@ pub fn write_slice(fd: u32, data: &[u8]) {
 use serde::de::DeserializeOwned;
 use serde::Serialize;
 
+pub fn read_stdin<T: DeserializeOwned>() -> T {
```
pacheco, line 144: idk about stdin... i think id prefer something more explicit, specially because you could also read stdin from the other method, and it will not behave the same?

pacheco: add_memory_input / read_memory_input ?

riscv-runtime/src/io.rs
```diff
@@ -51,6 +141,16 @@ pub fn write_slice(fd: u32, data: &[u8]) {
...
+    let mut reader = get_prover_data_reader();
+    if let Some(ref mut reader) = *reader {
+        let slice = reader.next().unwrap();
+        bincode::deserialize(slice).unwrap()
+    } else {
+        panic!("powdr ProverDataReader not available");
```
pacheco, line 150: what happens if thsi is called outside of continuations?

riscv/src/continuations.rs
```diff
@@ -155,45 +156,127 @@ fn sanity_check(main_machine: &Machine, field: KnownField) {
...
+    // Actually fill the prover data
+
+    // Setup an iterator for the addresses to be filled. If the iterator runs out before
+    // we are done, it means that the prover data is too large to fit in the reserved space.
+    let mut word_addr_iter =
+        ((prover_data_start + PAGE_SIZE_BYTES) / 4..prover_data_end / 4).map(|i| i * 4);
+
+    // The first word is the total number of words that will follow in the user data.
+    // We save tha address to write it later, when we know it.
+    let total_word_count_addr = word_addr_iter.next().unwrap();
```
pacheco, line 245: maybe `next().unwrap()` => `next().expect(out_of_space_msg)`?

pacheco: or, this could be in its own function that returns an Option/Error (so you can use `?`) and here you just panic with a nicer msg?

riscv-runtime/src/io.rs
lvella: We don't? We do.

`ProverDataReader` is public and can be instantiated whenever the user wants. The stateful global instance is just to mimic read from stdin and allegedly be more user friendly.

That said, I am not sure complicating even more the binary format of the user data to make it an exact size iterator is worth it...

riscv-runtime/src/io.rs
lvella: Or use the `ProverDataReader` iterator directly...

I don't like this stateful interface.

leonardoalt: From the user's perspective, this is over complicated and just not good enough. In other zkVMs the user just calls "read", do we really want to force them to import some random type ProverDataReader just to read a T?

riscv-runtime/src/io.rs
leonardoalt: I think it will panic because the data hasn't been given, similar to the channel case

riscv/Cargo.toml
leonardoalt: will keep cbor for now because of cross platform

powdr/src/lib.rs
leonardoalt: decided in call:

Session will only expose `write` and `write_bytes` which depend on continuations and use the more efficient method via initial prover memory

powdr-test/examples/serialized-inputs/src/main.rs
leonardoalt: yea in the new riscv test added also here in this PR

powdr/src/lib.rs
leonardoalt: done

pipeline/src/pipeline.rs
leonardoalt: done

.github/workflows/pr-tests.yml
leonardoalt: the examples that use a tag version of powdr, like 0.1.2. Others use the "local" one in whatever branch it is run

pipeline/src/pipeline.rs
leonardoalt: done

.github/workflows/pr-tests.yml
pacheco: maybe change to "with latest powdr release"?

pacheco (dismissed):
lgtm

pacheco (approved):

PR #2247, opened 2024-12-17
Use truncate instead of pop.

chriseth:


leonardoalt:
why?

chriseth:
Ah sorry: `truncate` works on already finalized data, while `pop` does not. The reason is that `pop` returns the last (unfinalized) element and truncate just discards the last row. Since we don't need the element here, we can use truncate.

georgwiese (approved):
Cool!

PR #2249, opened 2024-12-17
Fix bit operations.

chriseth:
This fixes a bug in witjitgen where the bit operations used for masking in bit-decomposition would use field elements as types for the masks. The problem is that we sometimes mask using masks outside the field and the proper type for masks is FieldElement::Integer.


georgwiese (commented):
I think now we could have "field elements" > p, by ORing a field element with a value > p.

I checked where we even need bitwise or, and the only place I found is [this](https://github.com/powdr-labs/powdr/blob/40478ecd072719100ae5449e69f08ef4729b8d10/executor/src/witgen/jit/affine_symbolic_expression.rs#L286-L297). Here, the mask would never be > p. But then I think we should do one of:
- Only allow OR between two field elements (as before)
- Assert (at runtime or compile time) that the mask is smaller than p
- Wrap modulo p

chriseth:
The problem is that even if the mask is smaller than the modulus, the result can still be larger. The proper way here would be to do the whole operation in the integers instead. What about adding an operation that just checks that `x == x | y`? we could either make it an assertion or return 1 on true and 0 on false.

chriseth:
Oh wait no, it's much simpler!

georgwiese (approved):
Haha, nice!

PR #2250, opened 2024-12-17
default getenv

leonardoalt:


pacheco (commented):
lgtm for the time being

riscv-runtime/src/std_support.rs
```diff
@@ -58,7 +58,7 @@ extern "C" fn sys_getenv(
     _varname: *const u8,
     _varname_len: usize,
 ) -> usize {
-    todo!()
+    0
```
pacheco, line 61: a small comment here would be nice though, saying this is implementing the empty env

lvella: I'm fine after this fix.

leonardoalt: updated

pacheco (approved):

PR #2251, opened 2024-12-17
Reserved space for prover data in memory layout

lvella:


examples/fibonacci/guest/src/main.rs
```diff
@@ -10,8 +10,12 @@ fn fib(n: u32) -> u32 {
 }
 
 fn main() {
-    // Read input from stdin.
-    let n: u32 = read(0);
+    // Read input from prover data.
+    let mut reader = ProverDataReader::new();
+    // postcard is an overkill for this example, but it is serde compatible
+    // and can be used for structs, arrays and other complex types.
+    let n: u32 = postcard::from_bytes(reader.next().unwrap()).unwrap();
```
leonardoalt, line 17: I think we should wrap all this in some `powdr_riscv_runtime::io` helper. The main use case is for users to just do
```rust
let x: T = powdr_riscv_runtime::io::read();
```

riscv-runtime/src/io.rs
```diff
@@ -8,18 +9,81 @@ use powdr_riscv_syscalls::Syscall;
...
+            Some(slice::from_raw_parts(data_ptr, len_bytes as usize))
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        if self.remaining_data.is_empty() {
+            (0, Some(0))
+        } else {
+            // At the lowest, the next slice can contain all remaining words;
+            // At the highest, every remaining word might define a new slice with 0 elements.
```
leonardoalt, line 79: lowest and highest what?

riscv/Cargo.toml
```diff
@@ -9,7 +9,7 @@ homepage = { workspace = true }
 repository = { workspace = true }
 
 [features]
-default = []                                       # complex-tests is disabled by default
+default = [] # complex-tests is disabled by default
```
leonardoalt, line 12: this comment was removed in `main`

riscv/src/riscv_gl.asm
```diff
@@ -27,6 +27,11 @@ machine Main with min_degree: MIN_DEGREE, max_degree: {{MAIN_MAX_DEGREE}} {
         {{INITIAL_MEMORY}}
     ];
 
+    // Initial and final memory addresses of prover data.
+    // The data is to be filled in by the prover in between those ranges.
```
leonardoalt, line 31:
```suggestion
    // The data is to be filled in by the prover in this range.
```

riscv/src/continuations.rs
```diff
@@ -155,45 +156,127 @@ fn sanity_check(main_machine: &Machine, field: KnownField) {
...
+        MemoryState::default()
     };
 
-    let Expression::ArrayLiteral(_, array) = expr else {
-        panic!("initial_memory is not an array literal");
-    };
+    // Fill the first page with random data to be the salt.
+    // TODO: the random value should be the "hash" of the merkle tree leaf itself,
+    // and the preimage of such hash should be unknown. But to implement that it would
+    // require some inteligence on the MemoryState type.
```
leonardoalt, line 230: I guess this might also affect verification when we actually expose the publics in the bootloader?

examples/fibonacci/guest/src/main.rs
lvella: The lower level interface must exist anyway, so we can optimize heavy cases.

riscv-runtime/src/io.rs
lvella: fixed

riscv/src/continuations.rs
lvella: The idea is to better support secrecy of prover data in case it is not public. Different proofs with the same prover data will yield different merkle tree roots.

leonardoalt (approved):

PR #2252, opened 2024-12-18
Implement Neg for FieldElement.

chriseth:


executor/src/witgen/jit/compiler.rs
```diff
@@ -368,6 +368,17 @@ impl std::ops::Sub for FieldElement {{
         Self(IntType::try_from(((self.0 as DoubleIntType) + (MODULUS as DoubleIntType) - (b.0 as DoubleIntType)) % (MODULUS as DoubleIntType)).unwrap())
     }}
 }}
+impl std::ops::Neg for FieldElement {{
+    type Output = Self;
+    #[inline]
+    fn neg(self) -> Self {{
+        if self.0 == 0 {{
+            Self(0)
```
Schaeff, line 376: Why not `self`?

chriseth: good point :)

georgwiese (approved):

PR #2254, opened 2024-12-18
Rename `_start` to `_powdr_start`

### LINKED ISSUE #2220: Duplicate label: _start
When running `pil` on an assembly file that was generated and uses the `_start` symbol, 

```
> cargo run --bin powdr pil output/zeam-poc.asm -o output_pil -f
(...)
Duplicate label: _start                                                                          
```

This is because the code in `asm-to-pil/src/romgen.rs` will add a `_start` symbol. That doesn't matter if the code is generated with the rust runtime, but any random ELF generated from C or zig will have the `_start` symbol by default.

For reference, I'm attaching an example asm file that triggered the issue: [zeam-poc.zip](https://github.com/user-attachments/files/18080563/zeam-poc.zip)

Schaeff:
Should fix #2220 
cc @gballet

### DIFF
--- a/asm-to-pil/src/romgen.rs
+++ b/asm-to-pil/src/romgen.rs
@@ -93,7 +93,7 @@ pub fn generate_machine_rom<T: FieldElement>(mut machine: Machine) -> (Machine, 
         // add the beginning of the dispatcher
         rom.extend(vec![
             Batch::from(vec![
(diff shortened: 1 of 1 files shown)


leonardoalt (approved):

gballet:
I confirm that this works for me.

PR #2256, opened 2024-12-18
use can_process

chriseth:


executor/src/witgen/jit/block_machine_processor.rs
```diff
@@ -78,7 +79,11 @@ impl<'a, T: FieldElement> BlockMachineProcessor<'a, T> {
 
     /// Repeatedly processes all identities on all rows, until no progress is made.
     /// Fails iff there are incomplete machine calls in the latch row.
-    fn solve_block(&self, witgen: &mut WitgenInference<T, &Self>) -> Result<(), String> {
+    fn solve_block<CanProcess: CanProcessCall<T> + Clone>(
+        &self,
+        can_process: CanProcess,
```
chriseth, line 84: Not too happy about the names here.

executor/src/witgen/jit/witgen_inference.rs
```diff
@@ -152,58 +159,86 @@ impl<'a, T: FieldElement, FixedEval: FixedEvaluator<T>> WitgenInference<'a, T, F
...
-                        return ProcessResult::complete(effects);
-                    }
-                }
+        if let Some(effect) =
+            self.try_process_call(can_process, lookup_id, selector, arguments, offset)
+        {
+            ProcessResult::complete(vec![effect])
+        } else {
+            ProcessResult::empty()
+        }
```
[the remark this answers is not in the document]
chriseth: Yes because then the other function can use `?`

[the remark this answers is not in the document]
chriseth: Maybe I can wrap it in a lambda

executor/src/witgen/jit/witgen_inference.rs
```diff
@@ -152,58 +159,86 @@ impl<'a, T: FieldElement, FixedEval: FixedEvaluator<T>> WitgenInference<'a, T, F
...
+        &self,
+        can_process_call: CanProcess,
+        lookup_id: u64,
+        selector: &Expression<T>,
+        call_arguments: &[Expression<T>],
+        offset: i32,
+    ) -> Option<Effect<T, Variable>> {
+        if self
+            .evaluate(selector, offset)
+            .and_then(|s| s.try_to_known().map(|k| k.is_known_one()))
```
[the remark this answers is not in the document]
chriseth: Let's see how far we get!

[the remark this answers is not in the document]
chriseth: We can always branch...

executor/src/witgen/jit/witgen_inference.rs
```diff
@@ -152,58 +159,86 @@ impl<'a, T: FieldElement, FixedEval: FixedEvaluator<T>> WitgenInference<'a, T, F
...
+    ) -> Option<(BitVec, Vec<Option<RangeConstraint<T>>>)> {
+        let mut known = BitVec::new();
+        let mut range_constraints = Vec::new();
+        for e in expressions {
+            if let Some(e) = e.try_to_known() {
+                known.push(true);
+                range_constraints.push(e.range_constraint());
+            } else {
+                known.push(false);
+                range_constraints.push(None);
```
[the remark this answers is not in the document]
chriseth: We don't have one, we only get it in the known case. I mean we could derive one from AffineSymbolicExpression, but that would be a new function.

executor/src/witgen/jit/compiler.rs
```diff
@@ -417,6 +417,7 @@ fn util_code<T: FieldElement>(first_column_id: u64, column_count: usize) -> Resu
 #[cfg(test)]
 mod tests {
     use pretty_assertions::assert_eq;
+    use test_log::test;
```
georgwiese, line 420: Cool!

executor/src/witgen/jit/witgen_inference.rs
```diff
@@ -63,28 +66,32 @@ impl<'a, T: FieldElement, FixedEval: FixedEvaluator<T>> WitgenInference<'a, T, F
...
+                &left.selector,
+                &left.expressions,
+                row_offset,
+            ),
+            Identity::PhantomBusInteraction(PhantomBusInteractionIdentity {
+                id,
+                multiplicity,
+                tuple,
+                ..
+            }) => self.process_call(can_process, *id, multiplicity, &tuple.0, row_offset),
```
georgwiese, line 94: I think this would no work yet! This would call `MutableState::can_process_call_fully`, which would call `self.responsible_machine(identity_id).unwrap()`. But bus interactions are currently ignored when extracting machines, so the unwrap would panic. So I think we should keep the `// TODO(bus_interaction)` for now and do nothing.

georgwiese: Also, full bus interaction support would require branching:

As a first step, we'll probably just something like `bus_send(instr_foo, (OP_FOO, ...))` and `bus_receive(sel * LATCH, (OP_FOO, ...))`. In that case, the machine extractor can statically map the send to the receive and hence to the receiving machine.

But eventually, we'll want the receiver to be decided at runtime, and do something like `bus_send(1, (op_id, ...))`, where `op_id` is a witness. If I understand this right, the `CanProcess` interface already accounts for the statically-known op ID via the range constraints.

executor/src/witgen/jit/witgen_inference.rs
```diff
@@ -501,12 +542,12 @@ mod test {
         let mut complete = HashSet::new();
         let mut counter = 0;
         let expected_complete = expected_complete.unwrap_or(retained_identities.len() * rows.len());
-        while complete.len() != expected_complete {
+        while complete.len() < expected_complete {
```
georgwiese, line 545: This change is not necessary, is it? We can keep it like that, just wondering if it was necessary.

georgwiese (commented):
Nice! LGTM, except for the bus interaction stuff.

executor/src/witgen/jit/witgen_inference.rs
chriseth: We can also change the mutable state to detect the case where it is a bus interaction. The code here is complete wrt its responsibilities.

And yes, if multiple machines are receiving at the same bus, then the mutable state would call them in turn and ask them if they can answer the call. The first one that responds "yes" might be stored in some cache by the mutable state.

georgwiese: Not sure this is what I meant! I mean that only one machine can answer a given send, but which it is depends on runtime values. So until those are passed to `MutableState::can_process_call_fully`, `MutableState` won't be able to answer it. But that's not really the topic of this PR...

May point was that RN, the code would panic if there is a bus interactions in the JITed machine.

chriseth: But they are passed, as part of the range constraints.

georgwiese (approved):

PR #2263, opened 2024-12-19
Sponge construction for keccak syscall & Intruction::keccakf

gzanitti:


riscv-executor/src/lib.rs
```diff
@@ -2464,6 +2467,47 @@ impl<'a, 'b, F: FieldElement> Executor<'a, 'b, F> {
...
+                    let lo = *val as u32;
+                    let hi = (val >> 32) as u32;
+
+                    self.proc
+                        .set_mem(output_ptr.u() + i as u32 * 8 + 4, lo, self.step + 1, lid);
+                    self.proc
+                        .set_mem(output_ptr.u() + i as u32 * 8, hi, self.step + 1, lid);
+                }
+
+                let lid = self.instr_link_id(instr, "main_publics", 0);
```
leonardoalt, line 2506: why `main_publics`?

riscv-runtime/src/hash.rs
```diff
@@ -47,40 +47,72 @@ pub fn keccakf(input: &[u64; 25], output: &mut [u64; 25]) {
...
-            unsafe {
-                keccakf(mem::transmute(&b_input), mem::transmute(&mut b_output));
+pub struct Keccak {
+    state: [u64; 25],
+    buffer: [u8; 136],
+    buffer_pos: usize,
+    rate: usize,
+}
+
+impl Keccak {
```
leonardoalt, line 57: @lvella can you please check these changes too

riscv/src/large_field/runtime.rs
```diff
@@ -169,22 +169,22 @@ impl Runtime {
 
     fn with_keccak(mut self) -> Self {
         self.add_submachine(
-            "std::machines::hash::keccakf::KeccakF",
+            "std::machines::hash::keccakf32_memory::Keccakf32Memory",
             None,
             "keccakf",
-            vec!["memory"],
+            vec!["memory", "MIN_DEGREE", "LARGE_SUBMACHINES_MAX_DEGREE"],
```
leonardoalt, line 175: LARGE might be too much, we can use `MAIN_MAX_DEGREE` here I think

riscv/tests/riscv.rs
```diff
@@ -747,7 +752,11 @@ fn exported_csv_as_external_witness() {
     );
 
     // compile
-    let options = CompilerOptions::new(KnownField::GoldilocksField, RuntimeLibs::new(), false);
+    let options = CompilerOptions::new(
+        KnownField::GoldilocksField,
+        RuntimeLibs::new().with_keccak(),
```
leonardoalt, line 757: I think for this test it's intended that we use the software version of Keccak

riscv/tests/riscv_data/keccak/src/main.rs
```diff
@@ -2,24 +2,34 @@
 #![no_std]
 
 extern crate powdr_riscv_runtime;
-use tiny_keccak::{Hasher, Keccak};
+use powdr_riscv_runtime::hash::Keccak;
```
leonardoalt, line 5: please leave this test as is so we still test with tinykeccak via software, and just make a copy of the test that uses the syscall instead.

leonardoalt (commented):
Nice!!

riscv-runtime/src/hash.rs
```diff
@@ -47,40 +47,72 @@ pub fn keccakf(input: &[u64; 25], output: &mut [u64; 25]) {
...
+    pub fn v256() -> Self {
+        //RATE = 136
+
+        Self {
+            state: [0u64; 25],
+            buffer: [0u8; 136],
+            buffer_pos: 0,
+            rate: 136,
```
lvella, line 65: I think rate should be a `const`.

riscv-runtime/src/hash.rs
```diff
@@ -47,40 +47,72 @@ pub fn keccakf(input: &[u64; 25], output: &mut [u64; 25]) {
...
+    rate: usize,
+}
+
+impl Keccak {
+    pub fn v256() -> Self {
+        //RATE = 136
+
+        Self {
+            state: [0u64; 25],
+            buffer: [0u8; 136],
```
lvella, line 63: Being rate a `const`, it could be used here for the array size, instead of repeating the number 126.

riscv-runtime/src/hash.rs
```diff
@@ -47,40 +47,72 @@ pub fn keccakf(input: &[u64; 25], output: &mut [u64; 25]) {
...
-    b_input[rate - 1] ^= 0x80;
-    unsafe {
-        keccakf(mem::transmute(&b_input), mem::transmute(&mut b_output));
+    fn absorb_block(&mut self) {
+        let rate_words = self.rate / 8;
+        for i in 0..rate_words {
+            let chunk = &self.buffer[i * 8..(i + 1) * 8];
+            let val = u64::from_le_bytes(chunk.try_into().unwrap());
+            // Note: lo/hi positions are swapped to match the Keccak machine memory layout
+            let swapped = ((val << 32) & 0xFFFFFFFF00000000) | ((val >> 32) & 0x00000000FFFFFFFF);
```
lvella, line 93: Instead of swapping, it would probably be better to define the state as `[u32; _]` instead of `[u64; _]`, but I guess that would require changing the `keccakf` interface... which I don't know why it would use `[u64; _]` in the first place, instead of `[u32; _]` (which is the most natural choice, given the 4-byte alignment required of the machine). Maybe it was copied from Poseidon? But `poseidon_gl` has a reason to be `u64`, as it operates on field elements.

riscv-runtime/src/hash.rs
```diff
@@ -47,40 +47,72 @@ pub fn keccakf(input: &[u64; 25], output: &mut [u64; 25]) {
...
-        pt = (pt + 1) % rate;
-        if pt == 0 {
-            unsafe {
-                keccakf(mem::transmute(&b_input), mem::transmute(&mut b_output));
+pub struct Keccak {
+    state: [u64; 25],
+    buffer: [u8; 136],
+    buffer_pos: usize,
+    rate: usize,
+}
```
lvella, line 55: My overall suggestion is this: take the opportunity to change keccakf to `[u34; 50]`, then replace the state with:

```
const RATE: usize = 34; // Give it in u34 words

pub struct Keccak {
    state: [u32; 50],
    next_word: usize,

    input_buffer: u32,
    next_byte: usize,
}
```

When `next_byte == 4`, xor the `input_buffer` at `state[next_word]`. When `next_word == RATE`, call `keccakf`. This makes the struct much smaller, and saves a lot of data juggling/copying.

If you want to be fancy, you can write code very optimized for long input slices by using the unsafe method [`align_to()`](https://doc.rust-lang.org/std/primitive.slice.html#method.align_to) on the input slice.

lvella (changes requested):
I think there shouldn't be a buffer for the full state size. See my comment on the `Keccak` struct.

riscv-executor/src/lib.rs
```diff
@@ -127,7 +129,8 @@ instructions! {
     ec_add,
     ec_double,
     commit_public,
-    fail
+    fail,
+    keccakf
```
leonardoalt, line 133: I think this should not be uncommented here, otherwise we have to do manual witgen for keccak here too

gzanitti: I assume you are referring to the other, and not to the instruction. In that case, what about the submachine macro? I'm partially guessing with the links and stuff, so any comments would be welcome :)

riscv/tests/riscv_data/keccak/src/main.rs
gzanitti: I end up putting the new test in `keccak_powdr`, which we had said was no longer useful

riscv-executor/src/lib.rs
leonardoalt: Ah yes

riscv/tests/riscv.rs
```diff
@@ -185,7 +185,8 @@ fn memfuncs() {
 #[ignore = "Too slow"]
 fn keccak() {
     let case = "keccak";
-    verify_riscv_crate(case, Default::default(), false);
+    let options = CompilerOptions::new_gl().with_keccak();
+    verify_riscv_crate_gl_with_options(case, Default::default(), options, false);
```
leonardoalt, line 189: I think this has to be reverted so software keccak is still tested, and need to add a new test for the acc version

riscv/tests/riscv_data/keccak/src/main.rs
leonardoalt: Makes sense

riscv-runtime/src/hash.rs
```diff
@@ -40,79 +40,100 @@ pub fn poseidon2_gl(data: &[Goldilocks; 8]) -> [Goldilocks; 8] {
...
+                }
+            }
+
+            for &word in words {
+                if self.next_byte == 0 {
+                    self.xor_word_to_state(word);
+                } else {
+                    self.xor_word_to_state(self.input_buffer | (word << (8 * self.next_byte)));
+                    self.input_buffer = word >> (32 - 8 * self.next_byte);
+                }
```
lvella, line 106: Ah, pure poetry!

It is nice that you treated the special case `self.next_byte == 0` differently. I probably wouldn't, but I like it.

It would execute faster if the `if` was outside the `for`, but I am pretty confident LLVM can optimize it, and this way it is easier to read.

riscv-runtime/src/hash.rs
```diff
@@ -40,79 +40,100 @@ pub fn poseidon2_gl(data: &[Goldilocks; 8]) -> [Goldilocks; 8] {
...
+
+            for &byte in prefix {
+                self.input_buffer |= (byte as u32) << (8 * self.next_byte);
+                self.next_byte += 1;
+                if self.next_byte == 4 {
+                    self.xor_word_to_state(self.input_buffer);
+                    self.input_buffer = 0;
+                    self.next_byte = 0;
+                }
+            }
```
lvella, line 98: I recommend putting this block in a function, as it is identical for the suffix. Maybe call it `update_unaligned()`?

riscv-runtime/src/hash.rs
```diff
@@ -40,79 +40,100 @@ pub fn poseidon2_gl(data: &[Goldilocks; 8]) -> [Goldilocks; 8] {
...
+        }
+        self.xor_word_to_state(0x80000000);
 
         for i in 0..4 {
-            let swapped = ((self.state[i] << 32) & 0xFFFFFFFF00000000)
-                | ((self.state[i] >> 32) & 0x00000000FFFFFFFF);
-            output[i * 8..(i + 1) * 8].copy_from_slice(&swapped.to_le_bytes());
+            output[i * 8..(i * 8 + 4)].copy_from_slice(&self.state[i * 2 + 1].to_le_bytes());
+            output[(i * 8 + 4)..(i * 8 + 8)].copy_from_slice(&self.state[i * 2].to_le_bytes());
         }
```
lvella: This is very disturbing... basically, this code is to undo the crazy swapping done inside the keccak32_memory machine to make its interface big-endian. @leonardoalt do you know why the keccak32_memory interface is big-endian?

https://github.com/powdr-labs/powdr/blob/61510810630c5c195784f70ad68c83bd04674400/std/machines/hash/keccakf32_memory.asm#L64
https://github.com/powdr-labs/powdr/blob/61510810630c5c195784f70ad68c83bd04674400/std/machines/hash/keccakf32_memory.asm#L123

Maybe not in this PR, but if there are no other big consequences, I think keccak32_memory interface should be little-endian.

leonardoalt: I think the old version (no-memory) used to be little-endian, then we swapped to big-endian because it was easier to read/test, but I agree the memory version makes more sense as little-endian for memory compat.

gzanitti: Ok, I will change it in a future PR so as not to complicate this one. Thanks.

lvella (approved):
LGTM!

PR #2264, opened 2024-12-19
Update stwo dependency to the latest version

ShuangWu121:
This PR updates Powdr's stwo dependency to the latest version, which now uses the nightly-2024-12-17 Rust toolchain. As part of this update, it may also be necessary to align Powdr's toolchain with this new nightly version to ensure compatibility.

This is a draft PR, further adjustments or testing might be required based on the impacts of this toolchain update.

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -63,7 +63,7 @@ p3-commit = { git = "https://github.com/plonky3/Plonky3.git", rev = "2192432ddf2
 p3-matrix = { git = "https://github.com/plonky3/Plonky3.git", rev = "2192432ddf28e7359dd2c577447886463e6124f0", optional = true }
 p3-uni-stark = { git = "https://github.com/plonky3/Plonky3.git", rev = "2192432ddf28e7359dd2c577447886463e6124f0", optional = true }
 # TODO: Change this to main branch when the `andrew/dev/update-toolchain` branch is merged,the main branch is using "nightly-2024-01-04", not compatiable with plonky3
-stwo-prover = { git = "https://github.com/ShuangWu121/stwo.git", optional = true, rev = "564a4ddcde376ba0ae78da4d86ea5ad7338ef6fe",features = ["parallel"] }
+stwo-prover = { git = "https://github.com/starkware-libs/stwo.git", optional = true, rev = "f24cde6f8de3f1db75d04f87807f052aa889b077",features = ["parallel"] }
 
 strum = { version = "0.24.1", features = ["derive"] }
 log = "0.4.17"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly-2024-09-21"
+channel = "nightly-2024-12-17"


ShuangWu121:
fix clippy errors due to update to nightly 2024-12-17:
- empty line after doc comment
- explicit lifetimes could be elided: removing 'a, changing 'a to '_
- `map_or` can be simplified by using is_some_and
-  manually reimplementing `div_ceil`
fixed by applying the suggestions of clippy

backend/src/stwo/prover.rs
```diff
@@ -289,12 +291,13 @@ where
         let component = PowdrComponent::new(
             &mut TraceLocationAllocator::default(),
             PowdrEval::new(self.analyzed.clone()),
+            (SecureField::zero(), None),
```
ShuangWu121, line 294: this parameter is for logup, if logup is not used, this is the default value to be passed.

riscv-executor/src/lib.rs
```diff
@@ -577,7 +577,7 @@ mod builder {
         mode: ExecMode,
     }
 
-    impl<'a, 'b, F: FieldElement> TraceBuilder<'b, F> {
+    impl<'a, 'b: 'a, F: FieldElement> TraceBuilder<'a, F> {
```
Schaeff, line 580: Interesting, is this the same?

backend/src/stwo/prover.rs
ShuangWu121: comment added. 
Regarding SecureField::zero(), it is not a `const fn`, so I cannot use it directly in a const context. While I could create a function to call it, but this might not be necessary in this case?

Schaeff: Oh right, that's not necessary no!

.github/workflows/pr-tests.yml
```diff
@@ -41,12 +41,16 @@ jobs:
       - name: Date of the restored cache
         run: cat target/cache-build-date.txt
         continue-on-error: true
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y uuid-dev libgrpc++-dev
```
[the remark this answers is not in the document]
ShuangWu121: the build error without adding these dependencies is 
```
Running make test in zkevm-prover directory: /home/runner/work/powdr/powdr/target/pr-tests/build/pil-stark-prover-f4f78ef9ff32121d/out/zkevm-prover
  Package grpc++ was not found in the pkg-config search path.
  Perhaps you should add the directory containing `grpc++.pc'
  to the PKG_CONFIG_PATH environment variable
  Package 'grpc++', required by 'virtual:world', not found
  Package grpc++ was not found in the pkg-config search path.
  Perhaps you should add the directory containing `grpc++.pc'
  to the PKG_CONFIG_PATH environment variable
  Package 'grpc++', required by 'virtual:world', not found
  src/utils/utils.cpp:5:10: fatal error: uuid/uuid.h: No such file or directory
      5 | #include <uuid/uuid.h>
        |          ^~~~~~~~~~~~~
```
        
       it is related to zkevm-prover as a c++ external, but I didn't figure out what triggers the newer nightly rust to have this stricter checks for the presence of required libraries.
       
       I checked this [code](https://github.com/powdr-labs/powdr/blob/61510810630c5c195784f70ad68c83bd04674400/.github/runner/Dockerfile#L29), it installs uuid-dev for docker, related to github workflow running on a "self-hosted GitHub Actions runner that runs in a Docker container" and ChatGPT suggests me install libgrpc++-dev here,, I got confused to this point, how this is related?

riscv-executor/src/lib.rs
ShuangWu121: it is not the same. My mistake 😓

.github/workflows/pr-tests.yml
```diff
@@ -41,6 +41,10 @@ jobs:
       - name: Date of the restored cache
         run: cat target/cache-build-date.txt
         continue-on-error: true
+      - name: Install c++ dependencies for EStarkPolygon prover
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y uuid-dev libgrpc++-dev
```
Schaeff, line 47: This block can be merged with the one below since they're both installing things for EStark

PR #2269, opened 2024-12-19
improve executor handwritten witgen performance

pacheco:
- avoid hashing strings by using enums with known columns
- keep main trace as flat vec for memory access locality during execution
- keep pc column directly (instead of as "write events")
- other small changes


cli-rs/src/main.rs
```diff
@@ -456,33 +456,17 @@ fn execute<F: FieldElement>(
...
-        keys.sort();
-
-        let missing_cols = witness_cols
-            .iter()
-            .filter(|x| !keys.contains(x))
-            .collect::<Vec<_>>();
-
-        log::debug!("All witness column names: {:?}\n", witness_cols);
-        log::debug!("Executor provided columns: {:?}\n", keys);
-        log::debug!("Executor missing columns: {:?}", missing_cols);
```
leonardoalt, line 475: why remove?

riscv-executor/src/lib.rs
```diff
@@ -44,16 +46,13 @@ mod pil;
...
-
 #[derive(Debug)]
 struct SubmachineOp<F: FieldElement> {
     // pil identity id of the link
     identity_id: u64,
-    // these are the RHS values of the lookup (i.e., inside brackets in the PIL lookup)
-    lookup_args: Vec<F>,
+    // these are the RHS values of the lookup (i.e., inside brackets in the PIL lookup).
+    // This is a fixed size to avoid allocations in the common case.
+    lookup_args: [F; 4],
```
leonardoalt, line 55: does this mean we always have a max of 4 elements?

pacheco: for now yes, no calls with more than 4 elements, can be changed later if there's need

cli-rs/src/main.rs
pacheco: logic moved to the `finish` method in the executor

leonardoalt (approved):

PR #2280, opened 2024-12-23
 Extend stwo to Support Machines with Varying Degrees

ShuangWu121:
### Summary
This PR introduces support for proving machines with differing degrees in the `stwo` prover. It enables handling proving columns of varying sizes.

---

### Details
1. **Support for Proving Columns of Different Sizes**:
   - Machines with different column sizes require their own distinct components.
   - All components are stored in a single vector, from which a slice is created and passed into the `prove` function to handle the proving process.
   - Constant columns are indexed globally across all components to ensure consistency and correctness.

2. **Verification**:
   - The verification function requires the sizes of all columns (including constant columns and witness columns) as input. 

---

### Changes
- Updated the logic for managing machine-specific components and assembling them for the proving process.
- Implemented indexing of constant columns across components.

---

### Testing
- Added tests to validate proving functionality with machines of differing degrees for stwo prover.

---

### DIFF
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -41,6 +41,10 @@ jobs:
       - name: Date of the restored cache
         run: cat target/cache-build-date.txt
         continue-on-error: true
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y uuid-dev libgrpc++-dev
 
 
       ##### The block below is shared between cache build and PR build workflows #####
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -49,7 +49,7 @@ jobs:
 
       ##### The block below is shared between cache build and PR build workflows #####
       - name: Install EStarkPolygon prover dependencies
-        run: sudo apt-get update && sudo apt-get install -y nlohmann-json3-dev libpqxx-dev nasm
+        run: sudo apt-get update && sudo apt-get install -y nlohmann-json3-dev libpqxx-dev nasm libgrpc++-dev uuid-dev
       - name: Install Rust toolchain nightly-2024-12-17 (with clippy and rustfmt)
         run: rustup toolchain install nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu
       - name: Install Rust toolchain 1.81 (stable)
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -45,8 +45,8 @@ jobs:
       ##### The block below is shared between cache build and PR build workflows #####
       - name: Install EStarkPolygon prover dependencies
         run: sudo apt-get update && sudo apt-get install -y nlohmann-json3-dev libpqxx-dev nasm
-      - name: Install Rust toolchain nightly-2024-09-21 (with clippy and rustfmt)
-        run: rustup toolchain install nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu
+      - name: Install Rust toolchain nightly-2024-12-17 (with clippy and rustfmt)
+        run: rustup toolchain install nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu
       - name: Install Rust toolchain 1.81 (stable)
         run: rustup toolchain install 1.81-x86_64-unknown-linux-gnu
       - name: Set cargo to perform shallow clones
@@ -99,8 +99,8 @@ jobs:
           path: |
             ~/pilcom/node_modules
           key: ${{ runner.os }}-pilcom-node-modules
-      - name: Install Rust toolchain nightly-2024-09-21 (with clippy and rustfmt)
-        run: rustup toolchain install nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu
+      - name: Install Rust toolchain nightly-2024-12-17 (with clippy and rustfmt)
+        run: rustup toolchain install nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu
       - name: Install nightly-2024-08-01
         run: rustup toolchain install nightly-2024-08-01-x86_64-unknown-linux-gnu
       - name: Install std source
@@ -136,8 +136,8 @@ jobs:
             target/
             Cargo.lock
           key: ${{ runner.os }}-cargo-pr-tests
-      - name: Install Rust toolchain nightly-2024-09-21 (with clippy and rustfmt)
-        run: rustup toolchain install nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu
+      - name: Install Rust toolchain nightly-2024-12-17 (with clippy and rustfmt)
+        run: rustup toolchain install nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu
       - name: Install nightly
         run: rustup toolchain install nightly-2024-08-01-x86_64-unknown-linux-gnu
       - name: Install std source
@@ -166,8 +166,8 @@ jobs:
           path: |
             ~/pilcom/node_modules
           key: ${{ runner.os }}-pilcom-node-modules
-      - name: Install Rust toolchain nightly-2024-09-21(with clippy and rustfmt)
-        run: rustup toolchain install nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu
+      - name: Install Rust toolchain nightly-2024-12-17(with clippy and rustfmt)
+        run: rustup toolchain install nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu
       - name: Install nightly-2024-08-01
         run: rustup toolchain install nightly-2024-08-01-x86_64-unknown-linux-gnu
       - name: Install std source
@@ -216,8 +216,8 @@ jobs:
           path: |
             ~/pilcom/node_modules
           key: ${{ runner.os }}-pilcom-node-modules
-      - name: Install Rust toolchain nightly-2024-09-21 (with clippy and rustfmt)
-        run: rustup toolchain install nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-09-21-x86_64-unknown-linux-gnu
+      - name: Install Rust toolchain nightly-2024-12-17 (with clippy and rustfmt)
+        run: rustup toolchain install nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add clippy --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu && rustup component add rustfmt --toolchain nightly-2024-12-17-x86_64-unknown-linux-gnu
       - name: Install test dependencies
         run: sudo apt-get update && sudo apt-get install -y binutils-riscv64-unknown-elf lld
       - name: Install nightly-2024-08-01
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -41,6 +41,10 @@ jobs:
       - name: Date of the restored cache
         run: cat target/cache-build-date.txt
         continue-on-error: true
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y uuid-dev libgrpc++-dev
 
       ##### The block below is shared between cache build and PR build workflows #####
       - name: Install EStarkPolygon prover dependencies
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -16,7 +16,7 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   POWDR_GENERATE_PROOFS: "true"
-  POWDR_JIT_OPT_LEVEL: "1"
+  POWDR_JIT_OPT_LEVEL: "0"
   MAX_DEGREE_LOG: "20"
 
 jobs:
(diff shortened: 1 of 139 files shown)


backend/src/stwo/circuit_builder.rs
```diff
@@ -59,14 +59,18 @@ where
 }
 
 pub struct PowdrEval<T> {
-    analyzed: Arc<Analyzed<T>>,
+    log_degree: u32,
```
ShuangWu121, line 62: The pil.degrees() function does not provide a unique degree when there are multiple degrees. Now, the degree is determined upon the creation of PowdrEval, based on the witness length of the machine.

backend/src/stwo/prover.rs
```diff
@@ -206,106 +204,228 @@ where
...
+            .split
+            .iter()
+            .filter_map(|(machine, pil)| {
+                let witness_columns = machine_witness_columns(witness, pil, machine);
+                if witness_columns[0].1.is_empty() {
+                    //TODO: Empty machines can be removed entirely, but in verification it is not removed, need to be handled
+                    None
+                } else {
+                    let witness_by_machine = machine_witness_columns(witness, pil, machine);
+                    assert!(
```
ShuangWu121, line 252: Is this guaranteed somewhere else?

backend/src/stwo/circuit_builder.rs
```diff
@@ -156,8 +160,10 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
             .map(|(i, poly_id)| {
                 (
                     *poly_id,
-                    // PreprocessedColumn::Plonk(i) is unused argument in get_preprocessed_column
-                    eval.get_preprocessed_column(PreprocessedColumn::Plonk(i)),
+                    // PreprocessedColumn::Plonk(i), i is used for indexing the preprocessed columns
```
Schaeff, line 163: I find this comment confusing as it does not match the code below. It could potentially be removed entirely as it's clear the preprocessed columns are global.

backend/src/stwo/proof.rs
```diff
@@ -124,3 +126,15 @@ impl<B: Backend> From<StarkProvingKey<B>> for SerializableStarkProvingKey {
...
+#[derive(Serialize, Deserialize)]
+pub struct Proof<MC: MerkleChannel>
+where
+    MC::H: DeserializeOwned + Serialize,
+{
+    pub starkproof: StarkProof<MC::H>,
+    //sizes[0] stores the log_degree of all the constant columns
+    //sizes[1] stores the log_degree of all the witness columns
+    //sizes[2] stores the log_degree of all the machines
+    pub sizes: [Vec<u32>; 3],
```
Schaeff, line 139: Why not three struct fields?

backend/src/stwo/prover.rs
```diff
@@ -14,30 +14,29 @@ use crate::stwo::circuit_builder::{
...
 use stwo_prover::core::poly::circle::{CanonicCoset, CircleDomain, CircleEvaluation};
-use stwo_prover::core::poly::twiddles::TwiddleTree;
 use stwo_prover::core::poly::BitReversedOrder;
 use stwo_prover::core::ColumnVec;
 
 const FRI_LOG_BLOWUP: usize = 1;
 const FRI_NUM_QUERIES: usize = 100;
 const FRI_PROOF_OF_WORK_BITS: usize = 16;
 const LOG_LAST_LAYER_DEGREE_BOUND: usize = 0;
+const MACHINE_SIZES_IDX: usize = 2;
```
Schaeff, line 39: If this is needed, it would be good to have a comment explaining what it is

backend/src/stwo/prover.rs
```diff
@@ -114,19 +113,25 @@ where
...
-            .analyzed
-            .degrees()
+        let mut domain_degree_range = DegreeRange {
+            min: self.analyzed.degree_ranges().iter().next().unwrap().min,
+            max: self.analyzed.degree_ranges().iter().next().unwrap().max,
+        };
+        self.analyzed.degree_ranges().iter().for_each(|range| {
+            domain_degree_range.min = domain_degree_range.min.min(range.min);
+            domain_degree_range.max = domain_degree_range.max.max(range.max);
+        });
```
Schaeff, line 123: Something like this should work:

```
let domain_degree_range = self.analyzed.degree_ranges.iter().reduce(|mut acc, range| {
    acc.min = acc.min.min(range.min);
    acc.max = acc.max.max(range.max);
    acc
}).unwrap();
```

Schaeff: or

```
let range_min = self.analyzed.degree_ranges().iter().map(|range| range.min).min();
```

etc

backend/src/stwo/prover.rs
```diff
@@ -206,106 +204,179 @@ where
...
+
+        let domain_map: BTreeMap<usize, CircleDomain> = domain_degree_range
             .iter()
+            .filter(|&size| size.is_power_of_two())
```
Schaeff, line 219: These are already guaranteed to be powers of two.

backend/src/stwo/prover.rs
```diff
@@ -206,106 +204,179 @@ where
...
-                                            .half_coset,
-                                    );
-                                    (size as usize, twiddles)
+                                .find_map(|(namespace, table_provingkey)| {
+                                    if namespace == machine {
+                                        Some(table_provingkey)
+                                    } else {
+                                        None
+                                    }
                                 })
```
Schaeff, line 271: `preprocessed` is a `BTreeMap`, isn't this just `get(machine)`?

backend/src/stwo/prover.rs
```diff
@@ -206,106 +204,179 @@ where
...
+                                        if *size == machine_length {
+                                            Some(
+                                                table_provingkey
+                                                    .constant_trace_circle_domain
+                                                    .clone(),
+                                            )
+                                        } else {
+                                            None
+                                        }
+                                    })
```
Schaeff, line 286: Same here for `table_proving_key`?

backend/src/stwo/prover.rs
```diff
@@ -206,106 +204,179 @@ where
...
+                if witness_columns[0].1.is_empty() {
+                    //TODO: Empty machines can be removed entirely, but in verification it is not removed, need to be handled
+                    None
+                } else {
+                    let witness_by_machine = machine_witness_columns(witness, pil, machine);
+                    let machine_length = witness_by_machine[0].1.len();
+                    assert!(
+                        witness_by_machine
+                            .iter()
+                            .all(|(_name, vec)| vec.len() == machine_length),
```
Schaeff, line 257:
```suggestion
                            .all(|(_, vec)| vec.len() == machine_length),
```

backend/src/stwo/prover.rs
Schaeff: In general it seems like these `if let` could be replaced with a chain of `Option` adapters, for example https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then

backend/src/stwo/prover.rs
```diff
@@ -206,106 +204,179 @@ where
...
+                .collect::<Vec<_>>(),
+        );
+        sizes[ORIGINAL_TRACE_IDX].extend(
+            witness_by_machine
+                .iter()
+                .map(|eval| eval.len().ilog2())
+                .collect::<Vec<_>>(),
+        );
+
+        let twiddles_maxdegree = B::precompute_twiddles(
```
Schaeff, line 341:
```suggestion
        let twiddles_max_degree = B::precompute_twiddles(
```

backend/src/stwo/prover.rs
```diff
@@ -206,106 +204,179 @@ where
...
 
         let proof_result = stwo_prover::core::prover::prove::<B, MC>(
-            &[&component],
+            components_slice,
             prover_channel,
             commitment_scheme,
         );
 
-        let proof = match proof_result {
+        let starkproof = match proof_result {
```
Schaeff, line 374:
```suggestion
        let stark_proof = match proof_result {
```

backend/src/stwo/prover.rs
```diff
@@ -317,36 +388,63 @@ where
...
-        );
 
-        // Retrieve the expected column sizes in each commitment interaction, from the AIR.
-        // the sizes include the degrees of the constant, witness, native lookups. Native lookups are not used yet.
-        let sizes = component.trace_log_degree_bounds();
+        let tree_span_provider = &mut TraceLocationAllocator::default();
+        let mut components = Vec::new();
+
+        let mut constant_cols_offset_acc = 0;
+        let mut index_machine = 0;
```
Schaeff, line 404:
```suggestion
        let mut machine_index = 0;
```

backend/src/stwo/prover.rs
```diff
@@ -317,36 +388,63 @@ where
...
+
+        self.split.iter().for_each(|(_, pil)| {
```
Schaeff, line 406: This could be a map instead of a `for_each`, returning `components` instead of using side effects.

```
let components = self.split.iter().map(|...| ...).collect();
```

Schaeff: Also, instead of incrementing `index_machines` by hand, we could use `self.split.iter().zip_eq(proof.sizes[MACHINE_SIZES_IDS])`

pipeline/tests/pil.rs
```diff
@@ -288,6 +288,12 @@ fn stwo_incremental_one() {
     test_stwo(f, Default::default());
 }
 
+#[test]
+fn stwo_different_degrees() {
+    let f = "pil/different_degrees.pil";
+    test_stwo(f, Default::default());
+}
```
Schaeff, line 295: We already have a test using this file, why not add stwo there? I realise we already have a lot of these specific stwo tests, we should merge them with others. This can be a different PR though.

Schaeff (commented):
First pass on this, nice change! Some of the code could be more idiomatic, it currently uses a lot of side effects.

ShuangWu121:
> First pass on this, nice change! Some of the code could be more idiomatic, it currently uses a lot of side effects.

Thank you for the quick review!  I'll take a closer look at making the code more idiomatic and reducing side effects.

.github/workflows/pr-tests.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         run: cat target/cache-build-date.txt
         continue-on-error: true
 
+
```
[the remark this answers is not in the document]
ShuangWu121: Done

backend/src/stwo/circuit_builder.rs
ShuangWu121: Removed.

backend/src/stwo/proof.rs
ShuangWu121: It changed to three struct fields now.

backend/src/stwo/prover.rs
ShuangWu121: Nice! now it is simplified using the second option.

backend/src/stwo/prover.rs
ShuangWu121: removed

backend/src/stwo/prover.rs
ShuangWu121: Nice tips! These are quite useful functions, thanks! now the code is simplified.

backend/src/stwo/prover.rs
```diff
@@ -253,37 +267,20 @@ where
...
-                                        } else {
-                                            None
-                                        }
-                                    })
-                            {
-                                // Extend constant_cols only if the constant_trace exists
-                                constant_cols.extend(constant_trace);
-                            }
-                        }
+                    if let Some(constant_trace) = self
```
Schaeff, line 270: I suspect this can be removed as well, by calling `extend` directly on the option

pipeline/tests/pil.rs
ShuangWu121: make sense. I removed this code and opened another PR to fix the tests.

backend/src/stwo/prover.rs
```diff
@@ -267,21 +267,19 @@ where
...
+                            .preprocessed
+                            .as_ref()
+                            .and_then(|preprocessed| preprocessed.get(machine))
+                            .and_then(|table_provingkey| table_provingkey.get(&machine_length))
+                            .map(|table_provingkey_machine_size| {
+                                table_provingkey_machine_size
+                                    .constant_trace_circle_domain
+                                    .clone()
+                            })
+                            .unwrap(),
```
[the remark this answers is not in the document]
ShuangWu121: Indeed. 
I try to make it to this
```
self
    .proving_key
    .preprocessed
    .as_ref()
    .and_then(|preprocessed| preprocessed.get(machine))
    .and_then(|table_provingkey| table_provingkey.get(&machine_length))
    .map(|table_provingkey_machine_size| {
        table_provingkey_machine_size.constant_trace_circle_domain.clone()
    })
    .map(|constant_trace| constant_cols.extend(constant_trace));

```

then clippy gives error

```
called `map(f)` on an `Option` value where `f` is a closure that returns the unit type `()``
```
and suggests: 

```
if let Some(constant_trace) = self
                        .proving_key
                        .preproccessed
                        .as_ref()
                        .and_then(|preprocessed| preprocessed.get(machine))
                        .and_then(|table_provingkey| table_provingkey.get(&machine_length))
                        .map(|table_provingkey_machine_size| {
                            table_provingkey_machine_size
                                .constant_trace_circle_domain
                                .clone()
                        })
                    {
                        constant_cols.extend(constant_trace)
                    }
```
so I get back to the previous solution.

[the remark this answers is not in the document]
Schaeff:
```rust
                    constant_cols.extend(
                        self.proving_key
                            .preprocessed
                            .as_ref()
                            .and_then(|preprocessed| preprocessed.get(machine))
                            .and_then(|table_provingkey| table_provingkey.get(&machine_length))
                            .map(|table_provingkey_machine_size| {
                                &table_provingkey_machine_size.constant_trace_circle_domain
                            })
                            .into_iter()
                            .flatten(),
                    );
```

this seems to work, and also saves a clone. It relies on the fact that option implements `into_iter`.
Your version looks better though. The clone could still be avoided in yours I think!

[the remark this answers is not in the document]
ShuangWu121: The later function call `tree_builder.extend_evals(constant_cols)` needs to take ownership of constant_cols. Borrowing by reference doesn't work here, and `Copy` is not implemented for the `constant_cols` type in stwo. 👀 Shall we keep it like this, or do you have other suggestions?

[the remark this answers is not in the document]
Schaeff: Oh all good then, my bad!

backend/src/stwo/prover.rs
```diff
@@ -114,19 +113,32 @@ where
...
             .iter()
             .map(|size| {
                 (
-                    (size.ilog2() as usize),
+                    size.ilog2() as usize,
                     CanonicCoset::new(size.ilog2()).circle_domain(),
                 )
             })
-            .collect();
+            .collect::<BTreeMap<_, _>>();
```
Schaeff, line 141: This can be reverted

backend/src/stwo/proof.rs
```diff
@@ -124,3 +126,14 @@ impl<B: Backend> From<StarkProvingKey<B>> for SerializableStarkProvingKey {
         Self { preprocessed }
     }
 }
+
+#[derive(Serialize, Deserialize)]
+pub struct Proof<MC: MerkleChannel>
+where
+    MC::H: DeserializeOwned + Serialize,
+{
+    pub starkproof: StarkProof<MC::H>,
```
Schaeff, line 135:
```suggestion
    pub stark_proof: StarkProof<MC::H>,
```

backend/src/stwo/prover.rs

backend/src/stwo/proof.rs
ShuangWu121: updated.

backend/src/stwo/prover.rs
```diff
@@ -206,106 +211,165 @@ where
...
-                                            .half_coset,
-                                    );
-                                    (size as usize, twiddles)
-                                })
-                                .collect::<Vec<_>>()
+            .iter()
+            .filter_map(|(machine, pil)| {
+                let witness_columns = machine_witness_columns(witness, pil, machine);
+                if witness_columns[0].1.is_empty() {
+                    //TODO: Empty machines can be removed entirely, but in verification it is not removed, need to be handled
```
leonardoalt, line 258: Is this gonna be handled in this PR?

ShuangWu121: in another PR maybe? It would be nice to get a test case that has empty machine, to see what I can do on the verification side.

leonardoalt: The RISCV instruction tests have that effect, but yes we can leave it for another PR.

ShuangWu121: I will check there then, thanks!

backend/src/stwo/proof.rs
```diff
@@ -124,3 +126,14 @@ impl<B: Backend> From<StarkProvingKey<B>> for SerializableStarkProvingKey {
...
+
+#[derive(Serialize, Deserialize)]
+pub struct Proof<MC: MerkleChannel>
+where
+    MC::H: DeserializeOwned + Serialize,
+{
+    pub stark_proof: StarkProof<MC::H>,
+    pub constant_col_log_sizes: Vec<u32>,
+    pub witness_col_log_sizes: Vec<u32>,
+    pub machine_log_sizes: Vec<u32>,
```
Schaeff, line 138: Maybe it's clearer to have `BTreeMap`s here instead of vectors, indexed by machine name and column name?

Schaeff: Can be another PR

Schaeff (approved):

PR #2293, opened 2024-12-30
Loading the limits of prover data from ELF

lvella:


riscv-runtime/powdr.x
```diff
@@ -1,40 +1,70 @@
 # Powdr linker script.
 #
-# If you are using powdr-riscv-runtime, it expects the following
+# A segment of type LOOS+0xda (0x600000da) is understood by powdr, and it
+# delimits an unconstrained memory region that can be used to pass data
+# to the ZK program. If present, this region must be bigger than
```
leonardoalt, line 5:
```suggestion
# to the guest program. If present, this region must be bigger than
```

riscv-runtime/powdr.x
```diff
@@ -1,40 +1,70 @@
...
+# defines the bounds of the prover data segment.
 #
 # This linker script provides usable definitions to these
 # symbols, with a stack of almost 256 MB. If you are not building
 # via powdr-rs, you must manually specify "-C link-arg=-Tpowdr.x"
 # in rustc to use this linker script (e.g. via RUSTFLAGS).
 
+PHDRS {
+  # The powdr specific p_type 0x600000da indicates to
+  # powdr ELF parser where is the prover data.
```
leonardoalt, line 28:
```suggestion
  # powdr ELF parser where the prover data is.
```

riscv-runtime/powdr.x
lvella: I think it is fine the way it is.

leonardoalt: I think `guest` is more accurate in the sense of a VM, but fine

leonardoalt (approved):

PR #2299, opened 2025-01-02
Remove test function.

chriseth:
Removes `MutableState::get_machine` since we already have the machine before we construct the MutableState.


georgwiese (approved):
nice

PR #2301, opened 2025-01-02
Interpreter for witgen effects

pacheco:
Implementation of an interpreter for jit witgen effects.
Performance seems to be around 0.20x of the compiled jit code (poseidon benchmark).
- Disabled by default, can be enabled setting `POWDR_JIT_INTERPRETER=1`.
- Also introduces a way to disable JIT in general, with `POWDR_JIT_DISABLE=1`.


executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+        Self {
+            first_column_id,
+            column_count,
+            vars,
+            actions,
+        }
+    }
+
+    // Execute the machine effects for the given the parameters
+    pub fn call<Q: QueryCallback<T>>(&self, params: WitgenFunctionParams<'_, T>) {
```
chriseth, line 153: Is it possible to use a higher-level interface than WitgenFunctionParams? They require unsafe code to be used and are tailored towards an FFI.

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+        let mut map_var_idx = |var: &Variable| -> usize {
+            let idx = *var_idx.entry(var.clone()).or_insert_with(|| {
+                let idx = vars;
+                vars += 1;
+                idx
+            });
+            idx
+        };
+
+        // load known inputs
```
chriseth, line 58: Sounds like this code would be perfect inside a function called `load_known_inputs` ;)

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -283,13 +297,71 @@ impl<T: FieldElement, V: Clone> SymbolicExpression<T, V> {
...
+    }
+
+    /// Convert to an RPNExpression
+    pub fn to_rpn(&self) -> RPNExpression<T, V> {
+        let mut elems = Vec::new();
+        self.to_rpn_inner(&mut elems);
+        RPNExpression { elems }
+    }
+
+    fn to_rpn_inner(&self, elems: &mut Vec<RPNExpressionElem<T, V>>) {
```
chriseth, line 342: Are you worried about machine stack size or is the RPN transformation a performance optimization?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+        for var in known_inputs.iter() {
```
Schaeff, line 59:
```suggestion
        for var in &known_inputs {
```

Schaeff, line 59: This loop could look better as `actions.extend(known_inputs.iter().map(|var| ...))`

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+                Variable::Param(i) => {
+                    let idx = map_var_idx(var);
+                    actions.push(InterpreterAction::ReadParam(idx, *i));
+                }
+                Variable::MachineCallReturnValue(_) => unreachable!(),
+            }
+        }
+
+        // process effect
+        for effect in effects.iter() {
```
Schaeff, line 74: Same here since each effect leads to one action

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+                    actions.push(InterpreterAction::Assertion(
+                        lhs.map_variables(&mut map_var_idx).to_rpn(),
+                        rhs.map_variables(&mut map_var_idx).to_rpn(),
+                        *expected_equal,
+                    ));
+                }
+                Effect::MachineCall(id, arguments) => {
+                    let mut result_vars = vec![];
+
+                    arguments.iter().for_each(|a| match a {
```
Schaeff, line 100: This is `let result_vars = arguments.iter().filter_map(|a| ...).collect();

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+                                }
+                            })
+                            .collect(),
+                    ));
+                }
+            }
+        }
+
+        // write variables back
+        let vars_known: Vec<_> = effects.iter().flat_map(written_vars_in_effect).collect();
```
Schaeff, line 126:
```suggestion
        let known_vars: Vec<_> = effects.iter().flat_map(written_vars_in_effect).collect();
```

?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+        vars_known.iter().for_each(|var| {
```
Schaeff, line 127: Also extend

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+                    };
+                    stack.push(result);
+                }
+            }
+        }
+        stack.pop().unwrap()
+    }
+}
+
+// the following functions come from the interface.rs file also included in the compiled jit code
```
Schaeff, line 302: Does this mean it's duplicated? Any way it could be reused?

Schaeff (commented):
Just some mostly non-blocking comments

executor/src/witgen/jit/symbolic_expression.rs
pacheco: from some profiling, the "evaluate expression" code was a hot spot... RPN was quite a bit faster.
I'm not sure if its just due to avoiding recursion or because the whole expression is kept in a vec (avoiding pointers).

executor/src/witgen/jit/interpreter.rs
chriseth: I think `CompactDataRef<'_, T>,` would be the right thing to pass

executor/src/witgen/jit/symbolic_expression.rs
chriseth: Yeah, thought, about the same, locality might be a thing.

Could you move this function to the interpreter, though? I think SymbolicExpression should not care that there is a thing called RPNExpressionElem.

executor/src/witgen/jit/function_cache.rs
```diff
@@ -103,13 +103,26 @@ impl<'a, T: FieldElement> FunctionCache<'a, T> {
...
 
-                compile_effects(
-                    self.column_layout.first_column_id,
-                    self.column_layout.column_count,
-                    &known_inputs,
-                    &code,
-                )
-                .unwrap()
+                let compiled_jit =
+                    !matches!(std::env::var("POWDR_JIT_INTERPRETER"), Ok(val) if val == "1");
```
chriseth, line 107: I don't particularly like the use of environment variables. Could we move them as far up as possible? In my opinion, it should just be a configuration setting in the constructor of FuncitonCache.

chriseth:
While this is a great piece of software, I think we should actually just not at all integrate it for now. The FunctionCache is still specialized for the BlockMachineProcessor (i.e. it does not support dynamic machines), and it will be even harder to disentangle if we also have to deal with the interpreter as well.

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+impl<T: FieldElement> EffectsInterpreter<T> {
+    pub fn new(
+        first_column_id: u64,
+        column_count: usize,
+        known_inputs: &[Variable],
+        effects: &[Effect<T, Variable>],
+    ) -> Self {
+        let mut actions = vec![];
+        let mut var_idx = HashMap::new();
+        let mut vars = 0;
```
chriseth, line 46: `var_counter`?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+        }
+
+        // process effect
+        for effect in effects.iter() {
+            match effect {
+                Effect::Assignment(var, e) => {
+                    let idx = map_var_idx(var);
+                    actions.push(InterpreterAction::AssignExpression(
+                        idx,
+                        e.map_variables(&mut map_var_idx).to_rpn(),
```
chriseth, line 80: Would it be possible to perform variable mapping and rpn conversion in a single struct function where the variable mapping is internal state of the struct?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+        let pparams = params.params.into_mut_slice();
+
+        let mut vars = vec![None; self.vars];
+
+        let mut eval_stack = vec![];
+        for action in &self.actions {
+            match action {
+                InterpreterAction::AssignExpression(idx, e) => {
+                    let val = self.evaluate_expression(&mut eval_stack, &vars, e);
+                    assert!(vars[*idx].replace(val).is_none());
```
chriseth, line 165: Instead of this runtime assertion, we can actually check in the conversion inside the `new` function if each variable is written to exactly once (and only read after it has been written).

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+                    );
+                }
+                InterpreterAction::MachineCall(result_vars, id, arguments) => {
+                    let mutable_state =
+                        unsafe { &*(params.mutable_state as *const MutableState<T, Q>) };
+
+                    let mut arg_values: Vec<_> = arguments
+                        .iter()
+                        .map(|a| match a {
+                            MachineCallArgument::Unknown(v) => (Some(v), Default::default()),
```
chriseth, line 214: Is this `Some(v)` used anywhere?

Do you think there is a way to directly store a `&mut` of the variable inside `args` instead of having to construct this second `arg_values` vector?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+                        assert_eq!(lhs_value, rhs_value, "Assertion failed");
+                    } else {
+                        assert_ne!(lhs_value, rhs_value, "Assertion failed");
+                    }
+                }
+            }
+        }
+    }
+
+    // evaluates an expression using the provided variable values
```
chriseth, line 256:
```suggestion
    /// Evaluates an expression using the provided variable values
```

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+    fn evaluate_expression(
+        &self,
```
chriseth, line 258: `self` is unused as far as I can see. can this be a function on `RPNExpresssion`?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,391 @@
...
+                RPNExpressionElem::BinaryOperation(op) => {
+                    let right = stack.pop().unwrap();
+                    let left = stack.pop().unwrap();
+                    let result = match op {
+                        BinaryOperator::Add => left + right,
+                        BinaryOperator::Sub => left - right,
+                        BinaryOperator::Mul => left * right,
+                        BinaryOperator::Div => left / right,
+                        BinaryOperator::IntegerDiv => T::from(
+                            left.to_integer().try_into_u64().unwrap()
```
chriseth, line 276: I think we have to turn this into arbitrary integer. Or (maybe better) we require division on `FieldElement::Integer`.

executor/src/witgen/jit/interpreter.rs
chriseth: I think we should just use `CompactDataRef`

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -1,40 +1,54 @@
...
+    sync::Arc,
 };
 
 use num_traits::Zero;
 use powdr_number::FieldElement;
 
 use crate::witgen::range_constraints::RangeConstraint;
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RPNExpressionElem<T: FieldElement, S> {
```
chriseth, line 13: Can we move this to the interpreter?

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -1,40 +1,54 @@
...
+/// An expression in Reverse Polish Notation.
+pub struct RPNExpression<T: FieldElement, S> {
+    pub elems: Vec<RPNExpressionElem<T, S>>,
+}
+
 /// A value that is known at run-time, defined through a complex expression
 /// involving known cells or variables and compile-time constants.
 /// Each of the sub-expressions can have its own range constraint.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
chriseth, line 29: `PartialEq` is a recursive operation - where do you need it? Maybe it can be done based on pointers?

chriseth: Although maybe the PartialEq operation of Arc shortcuts if the pointers are the same...

pacheco: mm I don't recall why I added it, ill see if its needed

pacheco: removed

executor/src/witgen/jit/interpreter.rs
pacheco: done

executor/src/witgen/jit/interpreter.rs
pacheco: logic moved to struct

executor/src/witgen/jit/interpreter.rs
pacheco: I'm not sure where's the right place to put the `Div` bound on `FieldElement::Integer`, if that's what you mean

executor/src/witgen/jit/interpreter.rs
pacheco: changed things around a bit here: the knowns/unknowns are transformed into variable indexes (known arguments generate a previous assignment action)

for the second bit, i managed to do it, but the compiler didn't like it (needed a bit of unsafe)

executor/src/witgen/jit/symbolic_expression.rs
```diff
@@ -20,21 +20,21 @@ pub enum SymbolicExpression<T: FieldElement, S> {
...
-        Rc<Self>,
+        Arc<Self>,
         BitOperator,
         T::Integer,
         Option<RangeConstraint<T>>,
     ),
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
pacheco: i think this is present in main now

executor/src/witgen/jit/function_cache.rs
pacheco: i removed the other env var (for fully disabling jit), can we just keep this one as is for the moment (while interpreter is fully integrated) for easy testing?

chriseth: To be honest, I would prefer to remove the interpreter from function_cache for the time being...

executor/src/witgen/jit/symbolic_expression.rs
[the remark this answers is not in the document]
chriseth: haha, ok, I added them and I know why :)

It's needed to detect if two branches generate the same code and thus can be combined.

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+            }
+        });
+        stack.pop().unwrap()
+    }
+}
+
+// the following functions come from the interface.rs file also included in the compiled jit code
+
+#[inline]
+fn index(
```
chriseth, line 468: I think it would be better to implement these functions on CompactDataRef

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+                    }
+                });
+            }
+            _ => {}
+        }
+        set
+    }
+}
+
+/// Helper struct to map variables to unique indices, so they can be kept in
```
chriseth, line 331:
```suggestion
/// Helper struct to map variables to contiguous indices, so they can be kept in
```

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+    // get the index of a variable if it was previously mapped
+    pub fn get_var(&mut self, var: &Variable) -> Option<usize> {
+        self.var_idx.get(var).copied()
+    }
+
+    pub fn map_expr_to_rpn<T: FieldElement>(
+        &mut self,
+        expr: &SymbolicExpression<T, Variable>,
+    ) -> RPNExpression<T, usize> {
+        RPNExpression::from(&expr.map_variables(&mut |var| self.map_var(var)))
```
chriseth, line 374: Does it make sense to pass (a mut ref to) the var mapper to the rpn conversion routine so that we only have to map the var for the symbol itself instead of creating a full copy of the expression?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+        ) {
+            match expr {
+                SymbolicExpression::Concrete(n) => {
+                    elems.push(RPNExpressionElem::Concrete(*n));
+                }
+                SymbolicExpression::Symbol(s, _) => {
+                    elems.push(RPNExpressionElem::Symbol(s.clone()));
+                }
+                SymbolicExpression::BinaryOperation(lhs, op, rhs, _) => {
+                    from_inner(lhs, elems);
```
chriseth, line 406: Returning an iterator instead of pushing to a `&mut` does not work (or would require a `Box<dyn Iterator`) because of these recursive calls, right?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+                            MachineCallArgumentIdx::Known(idx) => {
+                                let var = &vars[*idx] as *const T;
+                                LookupCell::Input(unsafe { var.as_ref().unwrap() })
+                            }
+                        })
+                        .collect::<Vec<_>>();
+                    mutable_state.call_direct(*id, &mut args[..]).unwrap();
+                }
+                InterpreterAction::Assertion(e1, e2, expected_equal) => {
+                    let lhs_value = e1.evaluate(&mut eval_stack, &vars);
```
chriseth, line 256: Maybe assert that `eval_stack` is empty? Or create a public and a private version of `evaluate` where the public one has the assert?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+            }
+        }
+        let mut elems = Vec::new();
+        from_inner(expr, &mut elems);
+        RPNExpression { elems }
+    }
+}
+
+impl<T: FieldElement> RPNExpression<T, usize> {
+    /// Evaluate the expression using the provided variables
```
chriseth, line 427: Please also document `stack`

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+            RPNExpressionElem::BinaryOperation(op) => {
+                let right = stack.pop().unwrap();
+                let left = stack.pop().unwrap();
+                let result = match op {
+                    BinaryOperator::Add => left + right,
+                    BinaryOperator::Sub => left - right,
+                    BinaryOperator::Mul => left * right,
+                    BinaryOperator::Div => left / right,
+                    BinaryOperator::IntegerDiv => {
+                        T::from(left.to_arbitrary_integer() / right.to_arbitrary_integer())
```
chriseth, line 441: I think it would be a big performance gain to implement division on FieldElement::Integer instead of bigint.

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+    pub fn new(
+        first_column_id: u64,
+        column_count: usize,
+        known_inputs: &[Variable],
+        effects: &[Effect<T, Variable>],
+    ) -> Self {
+        let mut actions = vec![];
+        let mut var_mapper = VariableMapper::new();
+
+        Self::load_known_inputs(&mut var_mapper, &mut actions, known_inputs);
```
chriseth, line 54: Maybe these three should be put somewhere else, since they perform data type conversion instead of actual interpreting?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+                        actions.push(InterpreterAction::WriteParam(idx, *i));
+                    }
+                    Variable::MachineCallReturnValue(_) => {
+                        // This is just an internal variable.
+                    }
+                }
+            });
+    }
+
+    // Check that actions are valid (e.g., variables writen to only once, and only read after being written to)
```
chriseth, line 165:
```suggestion
    /// Check that actions are valid (e.g., variables written to only once, and only read after being written to)
```

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+    fn is_valid(&self) -> bool {
```
chriseth, line 166: Could be a function on a data structure that also performs conversion, contains the InterpreterActions, but not the code for actually interpreting.

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+                        self.first_column_id,
+                        self.column_count,
+                        known,
+                        row_offset,
+                        c.row_offset,
+                        c.id,
+                    );
+                }
+                InterpreterAction::MachineCall(id, arguments) => {
+                    // we know it's safe to escape the references here, but the compiler doesn't, so we use unsafe
```
chriseth, line 239: What do you mean by "escape"? Is the unsafe part here that we need to know that the indices are pairwise disjoint?

chriseth: Especially after `https://github.com/powdr-labs/powdr/pull/2315`, each machine call has unique variables. This means you could store them directly in the arguments, which of course makes referencing them more complicated.

Another way to avoid the unsafe would be to mutably iterate over the variables, filter and map into LookupCells.

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+        match self {
+            InterpreterAction::ReadCell(idx, _) => {
+                set.insert(*idx);
+            }
+            InterpreterAction::ReadParam(idx, _) => {
+                set.insert(*idx);
+            }
+            InterpreterAction::AssignExpression(idx, _) => {
+                set.insert(*idx);
+            }
```
chriseth, line 282:
```suggestion
            InterpreterAction::ReadCell(idx, _) | InterpreterAction::ReadParam(idx, _) | InterpreterAction::AssignExpression(idx, _) => {
                set.insert(*idx);
            }
```

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+    }
+
+    // Execute the machine effects for the given the parameters
+    pub fn call<Q: QueryCallback<T>>(
+        &self,
+        mutable_state: &MutableState<'_, T, Q>,
+        params: &mut [LookupCell<T>],
+        mut data: CompactDataRef<'_, T>,
+    ) {
+        let row_offset = data.row_offset().try_into().unwrap();
```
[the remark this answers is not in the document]
pacheco: using `CompactData` directly does make sense, ill just make the `CompactDataRef` members public

executor/src/witgen/jit/interpreter.rs
pacheco: not sure, i think this way is clearer, we're building the vec we're using anyways and its internal

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,546 @@
...
+                SymbolicExpression::Concrete(n) => {
+                    elems.push(RPNExpressionElem::Concrete(*n));
+                }
+                SymbolicExpression::Symbol(s, _) => {
+                    elems.push(RPNExpressionElem::Symbol(s.clone()));
+                }
+                SymbolicExpression::BinaryOperation(lhs, op, rhs, _) => {
+                    from_inner(lhs, elems);
+                    from_inner(rhs, elems);
+                    elems.push(RPNExpressionElem::BinaryOperation(op.clone()));
```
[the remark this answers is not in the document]
pacheco: clone() is ok here by me

executor/src/witgen/jit/interpreter.rs
pacheco: added an assert at the end of the `call` function instead

executor/src/witgen/jit/interpreter.rs
pacheco: I'd like to leave this as is for this PR, i have another branch where i did some optimizations and have a "Builder" struct for the interpreter

executor/src/witgen/jit/interpreter.rs
pacheco: i'd do it in a separate PR (see my previous comment)

executor/src/witgen/jit/interpreter.rs
pacheco: The compiler doesn't like that we collect mut refs to the variables inside the map (e.g., would allow multiple mutable refs to the same var). There can be a _lot_ of variables, so I think I prefer the little bit of unsafe here

executor/src/witgen/jit/function_cache.rs
pacheco: ok, removed it completely

executor/src/witgen/jit/compiler.rs
```diff
@@ -76,10 +76,8 @@ pub fn compile_effects<T: FieldElement>(
     let code = format!("{utils}\n//-------------------------------\n{witgen_code}");
 
     record_start("JIT-compilation");
-    let start = std::time::Instant::now();
     log::trace!("Calling cargo...");
     let r = powdr_jit_compiler::call_cargo(&code);
-    log::trace!("Done compiling, took {:.2}s", start.elapsed().as_secs_f32());
```
chriseth, line 82: Why did you remove this?

executor/src/witgen/jit/compiler.rs
```diff
@@ -227,7 +225,7 @@ extern "C" fn witgen(
 /// Returns an iterator over all variables written to in the effect.
 /// The flag indicates if the variable is the return value of a machine call and thus needs
 /// to be declared mutable.
-fn written_vars_in_effect<T: FieldElement>(
+pub fn written_vars_in_effect<T: FieldElement>(
```
chriseth, line 228: If this is going to be pub, then maybe turn it into a function of `Effect`?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,441 @@
...
+use crate::witgen::data_structures::mutable_state::MutableState;
+use crate::witgen::jit::effect::MachineCallArgument;
+use crate::witgen::machines::LookupCell;
+use crate::witgen::QueryCallback;
+use powdr_number::FieldElement;
+
+use std::collections::{BTreeSet, HashMap};
+
+/// Witgen effects compiled into interpreter instructions.
+pub struct EffectsInterpreter<T: FieldElement> {
```
chriseth, line 18: InterpreterEffects? InterpreterCode?

I am actually thinking about splitting up `Effects` into `Effects` (used at inference time) and `Code` (used for the final code), since they do not really overlap that much any more. Just as a side note...

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,441 @@
...
+    var_count: usize,
+    actions: Vec<InterpreterAction<T>>,
+}
+
+/// Witgen effects compiled into "instructions".
```
chriseth, line 23:
```suggestion
/// Witgen effects compiled into instructions for a stack machine.
```

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    /// Execute the machine effects for the given the parameters
+    pub fn call<Q: QueryCallback<T>>(
+        &self,
+        mutable_state: &MutableState<'_, T, Q>,
+        params: &mut [LookupCell<T>],
+        data: CompactDataRef<'_, T>,
+    ) {
+        let mut vars = vec![T::zero(); self.var_count];
```
chriseth, line 181: We could actually keep that one in the state - but can be done in another PR. There is also no need to reset it to zero between calls, since we checked that nothing is read before it is written.

executor/src/witgen/jit/interpreter.rs
```diff
@@ -0,0 +1,441 @@
...
+                }
+                InterpreterAction::ReadCell(idx, c) => {
+                    let cell_offset: usize = c.row_offset.try_into().unwrap();
+                    vars[*idx] = data.data.get(data.row_offset + cell_offset, c.id).0;
+                }
+                InterpreterAction::ReadParam(idx, i) => {
+                    vars[*idx] = get_param(params, *i);
+                }
+                InterpreterAction::WriteCell(idx, c) => {
+                    let cell_offset: usize = c.row_offset.try_into().unwrap();
```
chriseth, line 198: Same here.

executor/src/witgen/jit/interpreter.rs
chriseth: Can you 1) explain a bit more why it needs unsafe and why it is safe and 2) add code in `is_valid` that checks that the indices in a machine call do not overlap?

executor/src/witgen/jit/interpreter.rs
chriseth: Yes, exactly.

chriseth:
Only some cosmetic comments left. It looks great!
Could you add some tests to the interpreter? Something like running the block machine processor on `"../test_data/pil/binary.pil"` and executing one block?

chriseth:
Or poseidon for that matter.

executor/src/witgen/jit/compiler.rs
pacheco: oh, must've been an oversight

executor/src/witgen/jit/interpreter.rs
pacheco: yep, i've done exactly this in the other branch I have

executor/src/witgen/jit/effect.rs
```diff
@@ -60,6 +60,27 @@ pub struct Assertion<T: FieldElement, V> {
     pub expected_equal: bool,
 }
 
+impl<T: FieldElement> Effect<T, Variable> {
```
chriseth, line 63: This way it is between `struct Assertion` and its impl. Can you move it up to effect?

executor/src/witgen/jit/interpreter.rs
chriseth: ah sorry, this is the actual interpreter. all good :)

executor/src/witgen/jit/interpreter.rs
```diff
@@ -439,3 +435,95 @@ fn set_param<T: FieldElement>(params: &mut [LookupCell<T>], i: usize, value: T)
...
+        let poly_ids = analyzed
+            .committed_polys_in_source_order()
+            .flat_map(|p| p.0.array_elements().map(|e| e.1))
+            .collect_vec();
+
+        let mut data = CompactData::new(&poly_ids);
+        data.append_new_rows(31);
+        let data_ref = CompactDataRef::new(&mut data, 0);
+        interpreter.call(&mutable_state, &mut param_lookups, data_ref);
+    }
```
chriseth, line 528: Can you add an assertion about the outputs?

chriseth (approved):

PR #2308, opened 2025-01-07
Expose profiler session

leonardoalt:


powdr/src/lib.rs
```diff
@@ -282,6 +286,20 @@ pub fn pipeline_from_guest(
 }
 
 pub fn run(pipeline: &mut Pipeline<GoldilocksField>) {
+    run_internal(pipeline, None)
+}
+
+pub fn run_with_profiler(
```
pacheco, line 292: what if we don't expose the full profiler options in the session?
could always generate both callgrind and flamegraph with some default name, and maybe just take the output dir here?

leonardoalt: it's slower though, it would be good to be able to not generate it as well

leonardoalt: update after chat: the idea is to not expose the options to the user who just calls `run_with_profiler()` and the object gets created internally in `Session`.

PR #2311, opened 2025-01-07
Simplify stwo tests

ShuangWu121:
This PR simplify the tests for stwo


leonardoalt:
linter failing

ShuangWu121:
this simplification would make more sense when stwo works with publics, then some tests can be combined. 
working with publics now.

PR #2316, opened 2025-01-08
bugfix in identifying a chunk's accessed memory

pacheco:


leonardoalt (commented):
lgtm, but test is failing here

leonardoalt (approved):

PR #2318, opened 2025-01-08
Keccak endianness (BE -> LE)

gzanitti:
Changes the endianness of the Keccak machine (only Keccakf32Memory for now) from BE to LE. As a result, the sponge construction can now avoid the step that required inverting the u32 of each u64 word and, directly use full u64 directly.


gzanitti:
> I think you'll need to change the test file as well to swap the arguments

I thought the same thing, but after making the change, it continues to fail. The strange thing is that the original values of the assert are the same as using keccakf from tinykeccak... but they should be inverted, right?

leonardoalt:
I dont mean the Rust test, I mean the asm test in test_data std

leonardoalt (commented):
Nice!

leonardoalt (approved):

PR #2328, opened 2025-01-13
Simplify stwo proof

Schaeff:


backend/src/stwo/prover.rs
```diff
@@ -394,20 +383,35 @@ where
...
+            .map(
+                |((machine_name, pil), (proof_machine_name, &machine_log_size))| {
+                    assert_eq!(machine_name, proof_machine_name);
+                    let machine_component = PowdrComponent::new(
+                        tree_span_provider,
+                        PowdrEval::new((*pil).clone(), constant_cols_offset_acc, machine_log_size),
+                        (SecureField::zero(), None),
+                    );
+
+                    constant_cols_offset_acc += pil.constant_count();
```
ShuangWu121, line 402: Offset updates need to occur after PowdrEval. Otherwise, the preprocess column index will not start from 0. It is detected when verifying a Keccak proof. ^^

ShuangWu121:
I removed stwo from the cli and cli-rs Cargo default features because it caused cargo check to fail. The check uses +1.81-x86_64-unknown-linux-gnu against nightly code. I remember trying this before but couldn’t get it to work. 😐 Could we handle this in a separate PR? It might require some CI adjustments.

ShuangWu121 (approved):

PR #2331, opened 2025-01-13
Keep completed assignments.

chriseth:
If we want to propagate range constraints across completed assignments, we cannot delete them.


georgwiese (approved):
LGTM

PR #2333, opened 2025-01-13
add an ExecMode::Witness to the riscv executor

pacheco:
ExecMode::Trace should only produce the memory and registers trace


riscv-executor/src/lib.rs
```diff
@@ -2854,6 +2877,7 @@ pub fn execute<F: FieldElement>(
     prover_ctx: &Callback<F>,
     bootloader_inputs: &[F],
     max_steps_to_execute: Option<usize>,
+    full_witness: bool,
```
leonardoalt, line 2880: Shouldn't this just be ExecMode?

pacheco: but then you could also pass `Fast`, which is invalid

pacheco: exec mode is internal to the crate

leonardoalt: yea, just don't really like `bool` parameters..but ok

leonardoalt (approved):

leonardoalt (approved):

PR #2336, opened 2025-01-14
Fix: PILAnalyzer::Children visits Trait functions twice

### LINKED ISSUE #1964: PILAnalyzer::Children visits Trait functions twice
I think PILAnalyzer::Children visits trait functions twice, once via the trait declaration and once via the individual trait functions.
The other problem here is that we store full copies, we should think about using Arcs instead.

gzanitti:
Fix #1964

### DIFF
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -792,7 +792,7 @@ impl Children<Expression> for FunctionValueDefinition {
             }
             FunctionValueDefinition::TypeConstructor(_, variant) => variant.children_mut(),
             FunctionValueDefinition::TraitDeclaration(trait_decl) => {
-                Arc::get_mut(trait_decl).unwrap().children_mut()
+                Arc::make_mut(trait_decl).children_mut()
             }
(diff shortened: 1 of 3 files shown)


ast/src/analyzed/mod.rs
```diff
@@ -761,7 +761,7 @@ pub enum FunctionValueDefinition {
     Expression(TypedExpression),
     TypeDeclaration(TypeDeclaration),
     TypeConstructor(Arc<EnumDeclaration>, EnumVariant),
-    TraitDeclaration(TraitDeclaration),
+    TraitDeclaration(Arc<TraitDeclaration>),
```
chriseth, line 764: Do you nee the arc?

chriseth: In that case, please make sure you don't create a new Arc when constructing TraitFunctions, but clone the one form the declaration.

chriseth: Maybe it would be good to create  a function that creates a TraitFunction from a TraitDeclaration.

chriseth: Yes, great idea!

chriseth (approved):

PR #2346, opened 2025-01-15
Evaluaton of fixed columns that are constant outside the first and last row.

chriseth:
Allows fixed columns to be evaluated in the "single step jit processor" by pre-computing if a fixed column has a constant value apart from the first and last row.


georgwiese (dismissed):
Nice! Needs merge though.

georgwiese (approved):

PR #2359, opened 2025-01-17
Single receive in bus linker

Schaeff:
- separate the native and bus linker as two implementations of a `LinkerBackend` trait
- refactor the object data structures to simplify linker implementation
- change the bus linker to use the bus directly and use a single receive per operation, with the `interaction_id` being hash based
- pull the permutation selector array treatment into the linker, since it should be linker specific (per link for native, per operation for bus)


ast/src/object/mod.rs
```diff
@@ -41,15 +41,14 @@ impl Location {
 
 #[derive(Clone)]
 pub struct MachineInstanceGraph {
-    pub main: Machine,
-    pub entry_points: Vec<Operation>,
```
Schaeff, line 45: This is not needed anymore since `Object` now exposes its operations

Schaeff:
Status: all native linker tests pass, bus linker fails because witgen is not implemented

Schaeff:
This is now basically feature complete

linker/src/bus.rs
```diff
@@ -0,0 +1,354 @@
...
+        .into();
+
+        self.pil.push(PilStatement::Expression(
+            SourceRef::unknown(),
+            Expression::FunctionCall(
+                SourceRef::unknown(),
+                FunctionCall {
+                    function: Box::new(Expression::Reference(
+                        SourceRef::unknown(),
+                        SymbolPath::from_str("std::protocols::bus::bus_send")
```
leonardoalt, line 121: Where is the corresponding receive added, in the operation?

linker/src/bus.rs
```diff
@@ -0,0 +1,354 @@
...
+                        )
+                    }))
+                    .collect(),
+            }
+            .into();
+
+            let latch = namespaced_reference(namespace.clone(), object.latch.clone().unwrap());
+
+            let (function, arguments) = match selector_index {
+                // a selector index of None means this operation is called via lookup
```
leonardoalt, line 236: Is this really always an invariant? I mean, makes sense but just wondering.

leonardoalt (commented):
Nice!

linker/src/native.rs
```diff
@@ -0,0 +1,1005 @@
...
+            .objects
+            .iter()
+            // go through all permutation calls
+            .flat_map(|(location, object)| {
+                object.links.iter().filter_map(|link| {
+                    link.is_permutation
+                        .then_some((location.clone(), link.clone()))
+                })
+            })
+            // note: many calls to the same machine instance will generate many call selectors
```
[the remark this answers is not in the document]
Schaeff: no, this is the current behaviour where each call reduces to a permutation, and each permutation has a different selector.

linker/src/bus.rs
```diff
@@ -0,0 +1,354 @@
...
+    let result = hasher.finalize();
+    let mut bytes = [0u8; 4];
+    bytes.copy_from_slice(&result[..4]);
+    u32::from_le_bytes(bytes)
+}
+
+pub struct BusLinker {
+    /// the pil statements
+    pil: Vec<PilStatement>,
+    /// for each machine instance, the size of the selector array, based on the number of permutation links pointing to it
```
Schaeff, line 34:
```suggestion
    /// for each machine instance, the size of the selector array, based on the number of used operations accessed via permutation
```

note for self

linker/src/bus.rs
```diff
@@ -0,0 +1,354 @@
...
+    selector_array_size_by_instance: BTreeMap<Location, usize>,
+    /// for each link destination, the index in the selector array that the link points to. For lookups, this is None.
```
Schaeff, line 36:
```suggestion
    /// for each used operation, the index in the selector array. For operations accessed via lookups, this is None.
```

note for self

linker/src/bus.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+
+    fn into_pil(self) -> Vec<PilStatement> {
+        self.pil
+    }
+}
+
+impl BusLinker {
+    /// Process an operation, which means receive from the bus with the correct [InteractionType]
```
Schaeff, line 193:
```suggestion
    /// Process an operation, which means receive from the bus through lookup or permutation
```

linker/src/bus.rs
Schaeff: yes, when we construct `selector_array_index_by_operation` we insert None for lookups and Some(index) for permutations.

Note that `self.selector_array_index_by_operation.get` returns `Option<Option<index>>>`, the outer one to spot unused operations (for which we don't receive), the inner one to distinguish operations accessed via lookup vs permutation.

linker/src/bus.rs
```diff
@@ -0,0 +1,354 @@
...
+                assert!(
+                    existing_value,
+                    "Operation should not be called both as a lookup and a permutation"
+                );
+                Some(link)
+            })
+            .fold(
+                (BTreeMap::new(), BTreeMap::new()),
+                |(mut indices, mut sizes), link| {
+                    let to_insert = link.is_permutation.then(|| {
```
Schaeff, line 69: @leonardoalt this is where lookups yield None

linker/src/bus.rs
Schaeff: This could be cleaner if we had an optimizer pass on the object graph removing unused operations. We could then unwrap the `get` call.

linker/src/native.rs
```diff
@@ -0,0 +1,1005 @@
...
+    fn try_new(graph: &MachineInstanceGraph, degree_mode: DegreeMode) -> Result<Self, Vec<String>> {
+        let max_degree = match degree_mode {
+            DegreeMode::Monolithic => Some(graph
+                .objects
+                .iter()
+                .filter_map(|(_, object)| object.degree.max.clone()).map(|e| match e {
+                    Expression::Number(_, n) => n,
+                    _ => unimplemented!("Only constant max degrees are supported when using monolithic degree mode"),
+                }).max().unwrap()),
+            DegreeMode::Vadcop => None,
```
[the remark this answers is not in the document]
Schaeff: This is the current behavior. The semantics are that if max_degree is Some, we align all namespaces to that. That's monolithic. For vadcop we keep the machine's degree.

linker/src/native.rs
```diff
@@ -0,0 +1,1005 @@
...
+
+        let to_namespace = to.machine.clone().to_string();
+        let to_machine = &objects[&to.machine];
+        let operation = &to_machine.operations[&to.operation];
+
+        let op_id = operation.id.clone().map(|operation_id| operation_id.into());
+
+        // lhs is `flag { operation_id, inputs, outputs }`
+        let lhs = selected(
+            combine_flags(from.instr_flag.clone(), from.link_flag.clone()),
```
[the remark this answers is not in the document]
Schaeff: No, this optimization is not permutation specific. The call selector are only for the permutation rhs, see line 128 where it is accessed.

linker/src/native.rs
```diff
@@ -0,0 +1,1005 @@
...
+
+    fn process_object(&mut self, location: &Location, objects: &BTreeMap<Location, Object>) {
+        let object = objects[location].clone();
+
+        let namespace_degree = match &self.degree_mode {
+            DegreeMode::Monolithic => {
+                Expression::Number(SourceRef::unknown(), self.max_degree.clone().unwrap()).into()
+            }
+            DegreeMode::Vadcop => try_into_namespace_degree(object.degree)
+                .unwrap_or_else(|| panic!("machine at {location} must have an explicit degree")),
```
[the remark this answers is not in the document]
Schaeff: i think it's unreachable, by construction max_degree is Some if degree_mode is monolithic, line 32. The one in this arm is reachable, as in vadcop mode you could still have an undefined degree, and we don't have defaults anymore.

I guess this entire match could be on max_degree, which is a richer version of degree_mode (it has the degree you should align all namespaces to)

linker/src/native.rs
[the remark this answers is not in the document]
leonardoalt: but don't they get grouped later?

linker/src/native.rs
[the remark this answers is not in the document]
leonardoalt: ok

linker/src/bus.rs
leonardoalt: ok

leonardoalt (approved):

linker/src/native.rs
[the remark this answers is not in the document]
Schaeff: answered on comment below

PR #2364, opened 2025-01-20
Bootloader fix

pacheco:
Currently, our bootloader setup doesn't take into acount the cost of running the bootloader while setting up the inputs.
It may happen, in memory heavy workloads, that too many pages get paged in by the bootloader and there is no rows left in the chunk for running any computation.
This PR takes into account the cost of the bootloader when building its inputs.


riscv/src/continuations/bootloader.rs
```diff
@@ -35,6 +37,55 @@ pub const PAGE_INPUTS_OFFSET: usize = NUM_PAGES_INDEX + 1;
 // Ensure we have enough addresses for the scratch space.
 const_assert!(PAGE_SIZE_BYTES > 384);
 
+/// Computes the size of the bootloader given the number of input pages.
+pub fn bootloader_size(accessed_pages: &BTreeSet<u32>) -> usize {
```
leonardoalt, line 41: crazy stuff

leonardoalt (approved):

PR #2365, opened 2025-01-20
Using JAL instead of CALL, allowing any link register.

lvella:
Apparently this is needed by newer versions of LLVM.


leonardoalt (approved):

qwang98:
Thanks for the fix! FYI would need this merged so that I can link patched crypto-bigint to git dependencies. @lvella

qwang98 (approved):
Can't comment on the specifics of the fix but this works for the AUPIC bug I encountered :)

PR #2367, opened 2025-01-21
Support Not sound challenges in Stwo

ShuangWu121:
# Support for NOT Sound Challenges in Stwo

## PR Description

This PR introduces support for NOT sound challenges in the Stwo prover.

### Changes
- **Support for proving in two stages**:
  - Added assertions in `mod.rs` to enforce a maximum of two stages.
- **Implementation of dummy challenges**:
  - Added deterministic challenges for testing and validation.
  - The `challenge_channel`, used as a hash function, now generates deterministic values as challenges.

### TODO:
- [x] Add support for proving with two stages, with validation in `mod.rs`.
- [x] Implement dummy challenges for testing and ensure `challenge_channel` provides deterministic output.


backend/src/stwo/prover.rs
```diff
@@ -282,41 +285,113 @@ where
...
+            identity.pre_visit_expressions(&mut |expr| {
+                if let AlgebraicExpression::Challenge(challenge) = expr {
+                    challenges_stage0.insert(challenge.id);
+                }
+            });
+        }
+
+        //challenge_channel is used to draw random bytes for challenges
+        let challenge_channel = &mut <MC as MerkleChannel>::C::default();
+        let challenge_single_value = challenge_channel.draw_random_bytes();
```
ShuangWu121, line 343: stwo doesn't have function to draw from BaseField, it has function to draw from extension field, which consist of 4 BaseField elements, but then it doesn't have a straight function to transfer it to generic type F, which is required by the witness call back next_stage_witness function, so here I draw directly to bytes.

georgwiese: I had a quick look at how `SecureField` is defined (what's returned from `draw_felt()`), and looks like it's:

```rs
pub struct CM31(pub M31, pub M31);
pub struct QM31(pub CM31, pub CM31);
pub type SecureField = QM31;
```

But then, you could just do `ext_fe.0.0` to access the base field elements?

backend/src/stwo/prover.rs
```diff
@@ -282,41 +285,113 @@ where
...
+                    .iter()
+                    .map(|(machine_name, machine_witness)| {
+                        let new_witness = witgen_callback.next_stage_witness(
+                            &self.split[machine_name],
+                            machine_witness,
+                            stage0_challenges.clone(),
+                            1,
+                        );
+                        machine_log_sizes
+                            .insert(machine_name.clone(), new_witness[0].1.len().ilog2());
```
georgwiese, line 370: It should always be the same as the stage 0 witness. You could add an insertion, I don't think you need to update the map?

backend/src/stwo/prover.rs
```diff
@@ -386,16 +489,45 @@ where
         let mut constant_col_log_sizes = vec![];
         let mut witness_col_log_sizes = vec![];
 
+        //TODO: make the challenge sound, now the challenge is built the same way in prover.
+        let identities = self.analyzed.identities.clone();
```
georgwiese, line 493: Looks like this block could be extracted as a function and used by the prover & verifier.

backend/src/stwo/circuit_builder.rs
```diff
@@ -180,12 +190,19 @@ impl<T: FieldElement> FrameworkEval for PowdrEval<T> {
                 )
             })
             .collect();
+        let challenges = self
+            .challenges
+            .clone()
+            .into_iter()
+            .map(|(k, v)| (k, E::F::from(v.to_integer().try_into_u32().unwrap().into())))
+            .collect();
```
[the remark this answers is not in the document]
georgwiese: Also, I feel like it would be helpful to have a conversion function from our field to Stwo's :)

backend/src/stwo/prover.rs

backend/src/stwo/prover.rs
```diff
@@ -282,41 +285,113 @@ where
...
+                    })
+                    .collect();
+
+            witness_cols_circle_domain_eval = witness_by_machine_stage1
+                .into_iter()
+                .flat_map(|(_machine_name, witness_cols)| {
+                    witness_cols.into_iter().map(|(witness_name, vec)| {
+                        //no need to transfer the witness columns to circle domain if it is already calculated in stage 0
+                        if let Some(index) = witness_col_circle_domain_index.get(&witness_name) {
+                            witness_cols_circle_domain_eval[*index].clone()
```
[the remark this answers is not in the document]
ShuangWu121: yes, now this clone is avoided by using Options.^^

backend/src/stwo/prover.rs
```diff
@@ -282,41 +285,113 @@ where
...
+        let challenge_single_value = challenge_channel.draw_random_bytes();
+
+        //challenge_single_value is a vector more than 4  bytes, we know F is mersenne31, it needs 4 bytes to transfer to u32 and then to F
+        //TODO: make the challenge sound
+        let stage0_challenges = challenges_stage0
+            .iter()
+            .map(|&index| {
+                (
+                    index,
+                    F::from_bytes_le(challenge_single_value.get(0..4).unwrap()),
```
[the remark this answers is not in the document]
ShuangWu121: yes, use byte is not a good idea, if it wraps, it won't be uniform.
now draw_felt is used in the way you said, nice suggestion!

backend/src/stwo/prover.rs

backend/src/stwo/circuit_builder.rs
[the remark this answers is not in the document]

backend/src/stwo/circuit_builder.rs
```diff
@@ -20,14 +20,15 @@ use stwo_prover::core::poly::circle::{CircleDomain, CircleEvaluation};
 use stwo_prover::core::poly::BitReversedOrder;
 use stwo_prover::core::utils::{bit_reverse_index, coset_index_to_circle_domain_index};
 
-pub type PowdrComponent<'a, F> = FrameworkComponent<PowdrEval<F>>;
+use crate::stwo::params::BaseFieldElementMap;
 
-pub fn gen_stwo_circle_column<T, B, F>(
+pub type PowdrComponent<'a> = FrameworkComponent<PowdrEval>;
```
Schaeff, line 25:
```suggestion
pub type PowdrComponent = FrameworkComponent<PowdrEval>;
```

backend/src/stwo/prover.rs
```diff
@@ -165,7 +173,7 @@ where
                                 > = fixed_columns
                                     .iter()
                                     .map(|(_, vec)| {
-                                        gen_stwo_circle_column::<F, B, M31>(
+                                        gen_stwo_circle_column::<B, M31>(
```
Schaeff, line 176:
```suggestion
                                        gen_stwo_circle_column::<_, M31>(
```

`B` is correctly inferred here, and eventually we shouldn't need the other type annotation either, but that can be done separately

backend/src/stwo/prover.rs
```diff
@@ -240,23 +252,17 @@ where
...
 
-        let tree_span_provider = &mut TraceLocationAllocator::default();
-        //Each column size in machines needs its own component, the components from different machines are stored in this vector
-        let mut components = Vec::new();
-
-        //The preprocessed columns needs to be indexed in the whole execution instead of each machine, so we need to keep track of the offset
-        let mut constant_cols_offset_acc = 0;
-        let mut machine_log_sizes = BTreeMap::new();
-
+        //Generate witness for stage 0,build constant columns in circle domain at the same time
```
Schaeff, line 255:
```suggestion
        // Generate witness for stage 0, build constant columns in circle domain at the same time
```

test_data/pil/add_and_equal.pil
```diff
@@ -1,10 +0,0 @@
-namespace Add(4);
-
-  let a;
-  std::prelude::set_hint(a, |i| if i==0||i==2 {std::prelude::Query::Hint(2147483646)}else{std::prelude::Query::Hint(1)});
-  let b;
-  let c;
-  b = c;
-  a+a'=0;
```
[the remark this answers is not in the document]
ShuangWu121: Yes, there is a pil file "fixed_with_incremental.pil" tests the next references on both witness column and  fixed column

backend/src/stwo/circuit_builder.rs
```diff
@@ -20,14 +20,15 @@ use stwo_prover::core::poly::circle::{CircleDomain, CircleEvaluation};
...
+
+pub fn gen_stwo_circle_column<B, F>(
```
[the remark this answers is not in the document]
ShuangWu121: Stwo only works with M31, so I removed the generic parameter `T` here. Are there any specific reasons to keep it generic?

backend/src/stwo/prover.rs
```diff
@@ -453,3 +562,43 @@ fn get_config() -> PcsConfig {
...
+        .iter()
+        .flat_map(|identity| {
+            identity.all_children().filter_map(|expr| match expr {
+                AlgebraicExpression::Challenge(challenge) => Some(challenge.id),
+                _ => None,
+            })
+        })
+        .collect::<BTreeSet<_>>();
+
+    let number_of_qm31_challenges = ((challenges_stage0.len() + 4) & !3) >> 2;
```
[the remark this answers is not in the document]
ShuangWu121: Done

ShuangWu121:
> Cool! So, what's missing to make it sound is to draw the challenges after the stage 0 witness is committed, not before, right?

yes!

backend/src/stwo/params.rs
```diff
@@ -0,0 +1,18 @@
+use powdr_number::{FieldElement, LargeInt, Mersenne31Field};
+use stwo_prover::core::fields::m31::M31;
+
+pub trait BaseFieldElementMap: FieldElement {
+    fn into_stwo_m31(self) -> M31;
```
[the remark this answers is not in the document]
ShuangWu121: Done

backend/src/stwo/prover.rs
```diff
@@ -453,3 +557,46 @@ fn get_config() -> PcsConfig {
...
+                qm31_challenge.1 .0,
+                qm31_challenge.1 .1,
+            ]
+        })
+        .collect();
+
+    challenges_stage0
+        .into_iter()
+        .zip(
+            draw_challenges
```
Schaeff, line 597: Maybe we could remove `number_of_qm31_challenges` entirely, since `zip` stops when the shortest iterator stops. So we could do 
```rust
let draw_challenges = std::iter::repeat_with(|| {
   let qm31_challenge = challenge_channel.draw_felt();
           [
                qm31_challenge.0 .0,
                qm31_challenge.0 .1,
                qm31_challenge.1 .0,
                qm31_challenge.1 .1,
            ]
}).flatten();
```

This iterator will draw lazily, and by zipping it with `challenge_stage0`, we get the right number

backend/src/stwo/prover.rs
```diff
@@ -453,3 +562,43 @@ fn get_config() -> PcsConfig {
...
+
+    let number_of_qm31_challenges = ((challenges_stage0.len() + 4) & !3) >> 2;
+
+    let challenge_channel = &mut <MC as MerkleChannel>::C::default();
+
+    let draw_challenges: Vec<_> = (0..number_of_qm31_challenges)
+        .flat_map(|_| {
+            let qm31_challenge = challenge_channel.draw_felt();
+            [
+                qm31_challenge.0 .0,
```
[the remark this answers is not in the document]
Schaeff: formatting still looks odd here?

pipeline/tests/pil.rs
```diff
@@ -265,8 +265,8 @@ fn add() {
 }
 
 #[test]
-fn stwo_add_and_equal() {
-    let f = "pil/add_and_equal.pil";
+fn stwo_challenges() {
```
Schaeff, line 268: Do we already test challenges with other backends? Maybe we can reuse the test file instead of adding a new one?

ShuangWu121: Plonky3 already has this test, but it's located in the Plonky3 prover file. I extracted the PIL content from there and added it to the pipeline test file here. Once public is implemented in the Stwo backend, several tests from Stwo can be combined with other backends. 
I didn't find other test for challenges in pil though, maybe it is more related to the bus tests? 🙈

backend/src/stwo/prover.rs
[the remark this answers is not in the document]
ShuangWu121: it is restricted by cargo fmt 😅

pipeline/tests/pil.rs
Schaeff: nice, then let's do this later

backend/src/stwo/prover.rs
ShuangWu121: Great trick! I updated it.

backend/src/stwo/prover.rs
```diff
@@ -282,41 +288,102 @@ where
...
+                        .iter()
+                        .filter(|(name, _)| !keys_in_stage0_machine.contains(name))
+                        .cloned()
+                        .collect::<Vec<_>>();
+
+                    machine_witness.extend(new_witness_in_stage1);
+                    // The updated witness cols need to have the same order with the witness cols generated by stage 1
+                    machine_witness.sort_by_key(|(k, _)| {
+                        new_witness.iter().position(|(name, _)| name == k).unwrap()
+                    });
```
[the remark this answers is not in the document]
georgwiese: OK, I might be overlooking something, but I think the entire block could be this:
```rs

            // build witness columns for stage 1 using the callback function, with the generated challenges
            witness_cols_circle_domain_eval = witness_by_machine
                .into_iter()
                .flat_map(|(machine_name, machine_witness)| {
                    witgen_callback.next_stage_witness(
                        &self.split[&machine_name],
                        &machine_witness,
                        stage0_challenges.clone(),
                        1,
                    )
                })
                .map(|(witness_name, vec)| {
                    if let Some(index) = witness_col_circle_domain_index.get(&witness_name) {
                        witness_cols_circle_domain_eval[*index].take()
                    } else {
                        Some(gen_stwo_circle_column::<B, M31>(
                            *domain_map
                                .get(&(vec.len().ilog2() as usize))
                                .expect("Domain not found for given size"),
                            &vec,
                        ))
                    }
                })
                .collect();
        }
```

backend/src/stwo/prover.rs
```diff
@@ -334,9 +401,37 @@ where
...
+        tree_builder.extend_evals(witness_cols_circle_domain_eval.into_iter().flatten());
         tree_builder.commit(prover_channel);
 
+        let tree_span_provider = &mut TraceLocationAllocator::default();
+        // Each column size in machines needs its own component, the components from different machines are stored in this vector
+        let mut components = Vec::new();
+
+        // Build the circuit. The circuit includes constraints of all the machines in both stage 0 and stage 1
+        let mut constant_cols_offset_acc = 0;
+        self.split.iter().zip_eq(machine_log_sizes.iter()).for_each(
```
georgwiese, line 413:
```suggestion
        let components = self.split.iter().zip_eq(machine_log_sizes.iter()).map(
```

I think this functional style of writing code has a few advantages. For example, written like this it is obvious that there will be a component for each machine (you don't have to track the `push` calls), and also that they are independent of each other (computing one machine's component doesn't change a previous machine's component).

georgwiese: With this, because `components` is no longer `mut`, the `components.iter_mut()` call won't work, but there is no reason not to use `iter()` AFAICS. Also, the `let components_slice = components_slice.as_mut_slice();` is not needed I think, you could just apss `&components_slice` to `prove`.

pipeline/tests/pil.rs
georgwiese: We do have the `challenges_asm` test, which is very similar and should work with stwo as well?

backend/src/stwo/prover.rs
ShuangWu121: nice! code updated.

backend/src/stwo/prover.rs
```diff
@@ -282,41 +288,102 @@ where
...
+                        .cloned()
+                        .collect::<Vec<_>>();
+
+                    machine_witness.extend(new_witness_in_stage1);
+                    // The updated witness cols need to have the same order with the witness cols generated by stage 1
+                    machine_witness.sort_by_key(|(k, _)| {
+                        new_witness.iter().position(|(name, _)| name == k).unwrap()
+                    });
+
+                    machine_log_sizes.insert(machine_name.clone(), new_witness[0].1.len().ilog2());
```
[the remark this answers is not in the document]
ShuangWu121: Removed. I initially thought there might be new machines in stage 1, but that was a misunderstanding. There shouldn't be any

backend/src/stwo/prover.rs
```diff
@@ -282,41 +288,102 @@ where
...
+        // remember the witness cols that are already transferred to circle domain, so that we don't need to transfer them again in stage 1
+        let mut witness_col_circle_domain_index = BTreeMap::new();
+        let mut index_acc = 0;
+
+        // Get witness columns in circle domain for stage 0
+        let mut witness_cols_circle_domain_eval: ColumnVec<
+            Option<CircleEvaluation<B, BaseField, BitReversedOrder>>,
+        > = witness_by_machine
+            .clone()
+            .into_values()
```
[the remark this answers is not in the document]

ShuangWu121:
https://github.com/powdr-labs/powdr/pull/2367#discussion_r1932736337  
nice, I didn't check asm.rs file before, now the challenges test is combined there

backend/src/stwo/prover.rs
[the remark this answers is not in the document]
ShuangWu121: yes. this is much nicer. code updated

pipeline/tests/pil.rs
ShuangWu121: Yes, I combined the test in asm.rs file, I didn't check the tests there before. - -!

pipeline/src/test_util.rs
```diff
@@ -662,3 +677,8 @@ pub fn assert_proofs_fail_for_invalid_witnesses_stwo(
 
 #[cfg(not(feature = "stwo"))]
 pub fn test_stwo(_file_name: &str, _inputs: Vec<u32>) {}
+
+#[cfg(not(feature = "stwo"))]
+use powdr_number::Mersenne31Field;
```
georgwiese, line 682: Now, we import it in both cases, so we should just import it unconditionally, together with the other fields.

backend/src/stwo/prover.rs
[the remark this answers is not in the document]
Schaeff: wow!

backend/src/stwo/prover.rs
```diff
@@ -38,6 +45,7 @@ const FRI_LOG_BLOWUP: usize = 1;
 const FRI_NUM_QUERIES: usize = 100;
 const FRI_PROOF_OF_WORK_BITS: usize = 16;
 const LOG_LAST_LAYER_DEGREE_BOUND: usize = 0;
+pub type T = Mersenne31Field;
```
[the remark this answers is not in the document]
Schaeff: I suggested this to reduce the size of the diff and because MersenneField31 is long. M31 sounds like a good solution. Sorry about the back and forth @ShuangWu121 !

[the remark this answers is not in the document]
Schaeff: and have another alias for the stwo M31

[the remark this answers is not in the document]
ShuangWu121: No worries. Now I use M31 for Mersenne31Field, stwo M31 is replaced by BaseField, which is an alias for its M31.

pipeline/src/test_util.rs
ShuangWu121: Done

georgwiese (approved):
LGTM! But maybe @Schaeff can also give a 👍 cause he had some comments too.

PR #2373, opened 2025-01-22
Simplify interface.

chriseth:
Uses the `CanProcess` trait instead of the full `MutableState` struct in some function and also some grammatical simplifications (use `: impl CanProcess`).


georgwiese (approved):
Cool

PR #2378, opened 2025-01-23
Simplify constructor of Processor.

chriseth:
Removes optional parameters from the constructor of Processor and introduces builder functions for them.


executor/src/witgen/jit/processor.rs
```diff
@@ -60,15 +58,28 @@ impl<'a, T: FieldElement, FixedEval: FixedEvaluator<T>> Processor<'a, T, FixedEv
...
+
+    /// Activates the check to see if the code for two subsequently generated
+    /// blocks conflicts.
+    pub fn with_block_shape_check(mut self) -> Self {
+        self.check_block_shape = true;
+        self
+    }
+
     pub fn generate_code(
-        &self,
```
chriseth, line 71: This ensures that we cannot modify the parameters after the call to `generate_code`, which is the only public function that is not a builder function.

leonardoalt (approved):

PR #2384, opened 2025-01-24
Handle error inside ingest_effects.

chriseth:


executor/src/witgen/jit/witgen_inference.rs
```diff
@@ -424,9 +424,9 @@ impl<'a, T: FieldElement, FixedEval: FixedEvaluator<T>> WitgenInference<'a, T, F
         }
         if !updated_variables.is_empty() {
             // TODO we could have an occurrence map for the assignments as well.
-            updated_variables.extend(self.process_assignments().unwrap());
```
chriseth, line 427: This unwrap was triggered in one of the tests for an upcoming pr

leonardoalt (approved):

PR #2385, opened 2025-01-24
Simple solution for one-row block machines.

chriseth:
If no identities in a block machine contains a next reference, it must be a single-row block and thus it is fine to evaluate all identities just on (relative) row zero.

### DIFF
--- a/executor/src/witgen/jit/block_machine_processor.rs
+++ b/executor/src/witgen/jit/block_machine_processor.rs
@@ -75,12 +75,33 @@ impl<'a, T: FieldElement> BlockMachineProcessor<'a, T> {
             witgen.assign_variable(expr, self.latch_row as i32, Variable::Param(index));
         }
 
-        let identities = self.row_range().flat_map(move |row| {
+        // Compute the identity-row-pairs we consider.
+        let identities = if self
+            .machine_parts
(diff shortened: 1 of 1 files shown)


leonardoalt (dismissed):

executor/src/witgen/jit/block_machine_processor.rs
```diff
@@ -75,12 +75,28 @@ impl<'a, T: FieldElement> BlockMachineProcessor<'a, T> {
...
+        let start_row = if !have_next_ref {
+            // No identity contains a next reference - we do not need to consider row -1,
+            // and the block has to be rectangular-shaped.
+            0
+        } else {
+            // A machine that might have a non-rectangular shape.
+            // We iterate over all rows of the block +/- one row.
+            -1
+        };
+        let identities = (start_row..self.block_size as i32).flat_map(move |row| {
```
leonardoalt, line 93: Would it make sense to keep the `block_size == 1` assert in the if?

chriseth: There is a test that has a latch of `[0,1]*` and triggers the assertion.

leonardoalt (approved):

PR #2388, opened 2025-01-25
add div instr test to pr tests

leonardoalt:
This should break the CI PR tests. The idea is to fix it in this PR.

Edit: it is fixed now


georgwiese (approved):
Nice!

PR #2389, opened 2025-01-27
Make the binary machine single row

leonardoalt:


leonardoalt:
The jitwitgen issues in this PR were fixed by https://github.com/powdr-labs/powdr/pull/2392, but it still fails tests because manual witgen would need to be changed in riscv-executor.

std/machines/large_field/binary.asm
```diff
@@ -18,27 +18,31 @@ machine Binary(byte_binary: ByteBinary) with
...
+    let A_byte2;
+    let A_byte3;
+    let B_byte0;
+    let B_byte1;
+    let B_byte2;
+    let B_byte3;
+    let C_byte0;
+    let C_byte1;
+    let C_byte2;
+    let C_byte3;
```
qwang98, line 35: I think instead the comment could be something like: "We do not range check the bytes, because they are constrained to be 8 bits by the `ByteBinary` machine".

std/machines/large_field/binary.asm
```diff
@@ -18,27 +18,31 @@ machine Binary(byte_binary: ByteBinary) with
     operation xor<2> A, B -> C;
```
qwang98, line 18: We can add a comment here at the top: "We do not range check the inputs, because they are constrained to be 32 bites by the byte columns."

qwang98:
LGTM pending the executor fix. The overall effect seems to be more columns, fewer rows, and also fewer # of cells. Is fewer # of cells the key indicator to optimize for post VadCop?

qwang98:
I'm also happy to take a stab at the manual witgen in the executor as a learning opportunity :)

Doesn't look like anyone is taking it yet from the PRs but let me know if otherwise.

qwang98:
I also wonder why have we been doing manual witgen for this submachine, because given the inputs A and B and operation, everything else should be entirely deterministic for auto witgen?

qwang98:
I studied the executor code a bit and fixed it via #2429 as the fix wasn't too complicated, which is PR'ed against and depends on this branch being merged. Is it possible to force merge this with failing test and then #2429 immediately after? Also feel free to just paste over changes to this PR if it's too complicated.

qwang98 (dismissed):

std/machines/large_field/binary.asm
leonardoalt: done

leonardoalt:
Merged main after merging with the executor fix PR.

leonardoalt:
> LGTM pending the executor fix. The overall effect seems to be more columns, fewer rows, and also fewer # of cells. Is fewer # of cells the key indicator to optimize for post VadCop?

Correct!

leonardoalt:
> I also wonder why have we been doing manual witgen for this submachine, because given the inputs A and B and operation, everything else should be entirely deterministic for auto witgen?

Yea auto witgen works but it's very slow compared to proof time, so we wrote a manual version for a couple machines to speed up RISCV proofs.

qwang98 (approved):
Already reviewed this so LGTM. (maybe consider we reviewed each other's parts? lol)

PR #2394, opened 2025-01-27
make stwo default in cli and cli-rs

Schaeff:


leonardoalt:
looks good, but can't be compiled with stable Rust

leonardoalt (approved):

PR #2395, opened 2025-01-27
Combine "else if" in formatter.

chriseth:


executor/src/witgen/jit/effect.rs
```diff
@@ -165,3 +176,96 @@ fn format_condition<T: FieldElement>(condition: &BranchCondition<T, Variable>) -
...
+                },
+                vec![Effect::Assignment(
+                    var(1),
+                    SymbolicExpression::from(T::from(2)),
+                )],
+                vec![Effect::Branch(
+                    BranchCondition {
+                        variable: var(1),
+                        first_branch: RangeConstraint::from_range(T::from(5), T::from(6)),
+                        second_branch: RangeConstraint::from_range(T::from(7), T::from(8)),
```
leonardoalt, line 217: why does [7,8] not show in the generated code?

executor/src/witgen/jit/effect.rs
```diff
@@ -165,3 +176,96 @@ fn format_condition<T: FieldElement>(condition: &BranchCondition<T, Variable>) -
...
+
+    #[test]
+    fn combine_if_else() {
+        let effects = vec![
+            Effect::Assignment(var(0), SymbolicExpression::from(T::from(1))),
+            Effect::Branch(
+                BranchCondition {
+                    variable: var(0),
+                    first_branch: RangeConstraint::from_range(T::from(1), T::from(2)),
+                    second_branch: RangeConstraint::from_range(T::from(3), T::from(4)),
```
leonardoalt, line 207: same for [3,4]

executor/src/witgen/jit/effect.rs
chriseth: The else case is implicit, it should be the negation of the first branch with respect to the current range constraints.

But let me check if we can simplify that.

chriseth:
It is indeed not needed. Removed in https://github.com/powdr-labs/powdr/pull/2397

leonardoalt:
ok nice, approved the other PR

leonardoalt (approved):

PR #2397, opened 2025-01-28
Do not store second branch condition.

chriseth:


executor/src/witgen/jit/effect.rs
```diff
@@ -109,8 +109,7 @@ impl<T: FieldElement, V> Assertion<T, V> {
 #[derive(Clone, PartialEq, Eq)]
 pub struct BranchCondition<T: FieldElement, V> {
     pub variable: V,
-    pub first_branch: RangeConstraint<T>,
-    pub second_branch: RangeConstraint<T>,
```
chriseth, line 113: This PR removes the condition for the "else" branch, because it was never used.

leonardoalt (approved):

PR #2399, opened 2025-01-28
Restore counters before returning the error.

chriseth:
This restores `self.symbol_counters` before the error is propagated upwards. Without this, we get a panic at the `unwrap()` because the same thing is probably done further up in the call chain.


gzanitti (approved):
I took a look at the modifications we made to handle errors and, in my opinion, I don't see any other cases similar to this one. Maybe the only “little thing” would be to assign the namespace inside `handle_namespace` before evaluating the degree for consistency (so the namespace will be the correct one in case the degree evaluation fails), but it's more a detail than a bug.

So I will approve the PR.

PR #2400, opened 2025-01-28
Compute from prover function

chriseth:


chriseth:
It took me a while but now I understand what went wrong: The new function uses `try_eval` for everything. The problem with that is that the runtime processor marks a prover function as complete if it evaluated without error.

georgwiese (approved):
Haha, glad it didn't work on the first try then!

PR #2409, opened 2025-01-29
Fix `contains_next_ref`

Schaeff:


ast/src/analyzed/mod.rs
```diff
@@ -1057,8 +1059,12 @@ pub enum Identity<T> {
...
 
 impl<T> Identity<T> {
-    pub fn contains_next_ref(&self) -> bool {
-        self.children().any(|e| e.contains_next_ref())
+    pub fn contains_next_ref(
+        &self,
+        intermediate_definitions: &BTreeMap<AlgebraicReferenceThin, AlgebraicExpression<T>>,
+    ) -> bool {
+        self.children()
+            .any(|e| e.contains_next_ref(intermediate_definitions))
```
Schaeff, line 1067: todo: maybe we could have a single evaluator here, for better caching

executor/src/witgen/jit/block_machine_processor.rs
```diff
@@ -90,15 +92,21 @@ impl<'a, T: FieldElement> BlockMachineProcessor<'a, T> {
...
-                }
-            })
+        let identities = (start_row..self.block_size as i32).flat_map(|row| {
+            self.machine_parts
+                .identities
+                .iter()
+                .filter_map(|id| {
+                    // Filter out identities with next references on the last row.
+                    if row as usize == self.block_size - 1
+                        && id.contains_next_ref(&intermediate_definitions)
```
Schaeff, line 102: note: we could filter out identities with next ref once before instead of doing it on each row

chriseth:
I'm not super convinced by this abuse :)

The main thing this saves over a "manual" implementation with a `BTreeMap<AlgebraicReferenceThin, bool>`-cache for intermediates is that you don't need to do the lookup and recursive call yourself.

The downside is that if you use the evaluator, it will actually expand exponentiations.

If we keep this, can you rename the file to `algebraic_expression_evaluator` to avoid confusion?

Schaeff:
@chriseth thanks for the review, some arguments here:
- Dereferencing intermediates should be the default when visiting identities, it's a big advantage of using the evaluator rather than a manual cache
- I spent time thinking about the exponentiations, and I think it can be fixed cheaply
   - we currently abuse field elements by allowing them as exponents, where we really mean an integer constant
   - therefore, it does not work to add a `Pow` bound on the returned expression of the evaluator: the p3 and stwo expressions do not implement Pow, and they shouldn't because of the previous point
   - instead, we should pass a lambda when constructing the evaluator, specifying how to treat Pow. For backends, we'd expand (btw, using double and add would be better than now, but it probably doesn't matter since constrains have low degree anyway), and for these more exotic evaluators, we wouldn't
   - Doing this would even remove the `FieldElement` bound on `T` in the evaluator, and therefore the one I added on `contains_next_ref`
- We also need the same approach for identity degree detection, and I find it clearer to do something like this there as well, rather than a manual cache
- Another attempt from @georgwiese used more traits, see #2312. I find this PR nicer.

Can you explain which confusion your rename is avoiding?

chriseth:
I'm still not very convinced, the whole evaluator seems to be tailored towards witgen and execution. There is probably less boilerplate to implement an evaluator by simple recursive calls and `match` expressions over the operators.

This could be an implementation to find a next reference, which does not use the evaulator:
```
    pub fn contains_next_ref_with_intermediates(
        &self,
        intermediate_definitions: &BTreeMap<AlgebraicReferenceThin, AlgebraicExpression<T>>,
        intermediates_cache: &mut BTreeMap<AlgebraicReferenceThin, bool>,
    ) -> bool {
        self.all_children()
            .filter_map(|e| {
                if let AlgebraicExpression::Reference(reference) = e {
                    Some(reference)
                } else {
                    None
                }
            })
            .any(|reference| {
                if reference.next {
                    true
                } else if reference.poly_id.ptype == PolynomialType::Intermediate {
                    let reference = reference.to_thin();
                    intermediates_cache
                        .get(&reference)
                        .cloned()
                        .unwrap_or_else(|| {
                            let result = intermediate_definitions[&reference]
                                .contains_next_ref_with_intermediates(
                                    intermediate_definitions,
                                    intermediates_cache,
                                );
                            intermediates_cache.insert(reference, result);
                            result
                        })
                } else {
                    false
                }
            })
    }
```

About the naming: It's an evaluator for algebraic expressions, not for expressions (pil-ast-expressions). analyzed/mod.rs makes a distinction between those because it has to know about both. Most use-cases in witgen only consider AlgebraicExpressions and that's why they are imported as `Expression` in some modules.

ast/src/analyzed/mod.rs
```diff
@@ -1686,6 +1667,7 @@ impl Display for PolynomialType {
 
 #[cfg(test)]
 mod tests {
+
```
Schaeff, line 1670:
```suggestion
```

Schaeff:
Changed to custom implementation as trait.

ast/src/parsed/mod.rs
```diff
@@ -929,6 +929,7 @@ impl<R> Expression<R> {
     }
 
     /// Returns true if the expression contains a reference to a next value
+    // TODO: Is it fine that this does not check references to intermediate polynomials?
```
chriseth, line 932: If it doesn't follow function calls, I think it's fair that it also does not follow references to intermediate columns.

chriseth: The only place this is used is in airgen. There is a comment "// TODO: links with next references can't be merged due to a witgen limitation." - maybe this is not the case any more?

PR #2415, opened 2025-01-30
Log JIT code derivation.

chriseth:


executor/src/witgen/machines/profiling.rs
```diff
@@ -81,7 +81,6 @@ pub fn reset_and_print_profile_summary() {
             // Update the call stack.
             match event {
                 Event::Start => {
-                    assert!(current_machine_id != id, "Unexpected recursive call!");
```
chriseth, line 84: This happens if witjitgen calls into another machine to witjitgen that one. Actually let me solve this in a different way.

leonardoalt (dismissed):

leonardoalt (approved):

PR #2416, opened 2025-01-30
Allow blocks with single expressions when detecting prover functions

chriseth:


executor/src/witgen/jit/prover_function_heuristics.rs
```diff
@@ -172,23 +172,38 @@ fn try_as_lambda_expression(
...
+    match unpack(e) {
         Expression::ArrayLiteral(_, ArrayLiteral { items }) => Some(items.as_slice()),
         _ => None,
     }
 }
 
+/// If `e` is a block with just a single expression, returns this inner expression,
+/// otherwise returns `e` again.
+fn unpack(e: &Expression) -> &Expression {
+    match e {
```
pacheco, line 195: should unpack happen recursively here?

chriseth: Oh right!

pacheco (approved):

PR #2417, opened 2025-01-30
Add prover function call as new effect.

chriseth:


executor/src/witgen/jit/effect.rs
```diff
@@ -23,6 +23,8 @@ pub enum Effect<T: FieldElement, V> {
     Assertion(Assertion<T, V>),
     /// A call to a different machine, with identity ID, known inputs and argument variables.
     MachineCall(u64, BitVec, Vec<V>),
+    /// Compute one variable by executing a prover function (given by index) on the value of other variables.
```
pacheco, line 26: maybe not for this PR, but this enum might be a good candidate for using named fields

executor/src/witgen/jit/effect.rs
```diff
@@ -144,6 +149,12 @@ pub fn format_code<T: FieldElement>(effects: &[Effect<T, Variable>]) -> String {
             Effect::RangeConstraint(..) => {
                 panic!("Range constraints should not be part of the code.")
             }
+            Effect::ProverFunctionCall(v, index, args) => {
+                format!(
+                    "{v} = prover_function_{index}([{}]);",
```
pacheco, line 154: so the generated code for these takes in an array of arguments? just to know, why is that?

chriseth: In the end, it calls a PIL function that takes an array. It would be better to take a fixed-size slice or something, but then we can't do it in a generic way. In the end, we do pass a slice and the code that destructures the array again is in closures that are local to a function. If we are lucky, the rust optimizer can disentangle everything. I'll take a look at the assembly. Even if not, I think it's OK, performance-wise.

pacheco (approved):

PR #2418, opened 2025-01-30
Extract IdentityQueue.

chriseth:


executor/src/witgen/jit/identity_queue.rs
```diff
@@ -0,0 +1,186 @@
...
+                .iter()
+                .map(|(id, row)| QueueItem(id, *row))
+                .collect(),
+            occurrences,
+        }
+    }
+
+    /// Returns the next identity to be processed and its row and
+    /// removes it from the queue.
+    pub fn next(&mut self) -> Option<(&'a Identity<T>, i32)> {
```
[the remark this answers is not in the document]
chriseth: Yeah, considered that as well, but I don't want to give the impression that it's a stack, but rather an iterator.

executor/src/witgen/jit/processor.rs
```diff
@@ -216,13 +209,23 @@ impl<'a, T: FieldElement, FixedEval: FixedEvaluator<T>> Processor<'a, T, FixedEv
...
+        let first_branch_result = self.generate_code_for_branch(
+            can_process.clone(),
+            first_branch,
+            identity_queue.clone(),
+            branch_depth + 1,
+        );
+        let second_branch_result = self.generate_code_for_branch(
+            can_process,
+            second_branch,
+            identity_queue.clone(),
```
pacheco, line 226: this one doesn't need to be a clone i think

pacheco, line 226: this one doesn't need to be a clone i think

pacheco (approved):

PR #2419, opened 2025-01-30
Pilopt: Optimize associative ADD operations

### LINKED ISSUE #2413: PIL OPT: optimize ((x + 1) + 3) to (x + 4)
The pil optimizer can evaluated operations on literals like `1 + 3` -> `4`, but it cannot make use of associativity and commutativity. So it does not optimize `x + 1 + 3` to `x + 4`, because it is parsed as `(x + 1) + 3`.

We should implement the following rules (`Li` is a literal, `X` is any expression) - in post-visit-order:

- `(X + L1) + L2` -> `X + (L1 + L2)`
- `(L1 + X) + L2` -> `X + (L1 + L2)`
- `(L1 + X) + Y` -> `(X + Y) + L1` (this is needed to simplify e.g. `((X + 3) + Y) + 9)`
- `(X + L1) + Y` -> `(X + Y) + L1`
- ...

gzanitti:
Close #2413

### DIFF
--- a/pilopt/src/lib.rs
+++ b/pilopt/src/lib.rs
@@ -435,9 +435,10 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
     }
 }
 
+/// Uses associative properties to simplify expressions by regrouping constant terms.
 fn try_simplify_associative_operation<T: FieldElement>(
-    left: &AlgebraicExpression<T>,
-    right: &AlgebraicExpression<T>,
+    left: &mut AlgebraicExpression<T>,
+    right: &mut AlgebraicExpression<T>,
     op: AlgebraicBinaryOperator,
 ) -> Option<AlgebraicExpression<T>> {
     if op != AlgebraicBinaryOperator::Add {
@@ -471,13 +472,14 @@ fn try_simplify_associative_operation<T: FieldElement>(
     // (X + C1) -> (X, C1)
     // (C1 + X) -> (X, C1)
     let (x, c1_val) = if let AlgebraicExpression::Number(val) = x1.as_ref() {
-        (x2, val)
+        (x2.as_mut(), val)
     } else if let AlgebraicExpression::Number(val) = x2.as_ref() {
-        (x1, val)
+        (x1.as_mut(), val)
     } else {
         return None;
     };
 
+    let x = std::mem::replace(x, AlgebraicExpression::Number(0.into()));
     match other_expr {
         // Case 1: Combining with a constant
         // (X + C1) + C2 -> X + (C1 + C2)
@@ -485,7 +487,7 @@ fn try_simplify_associative_operation<T: FieldElement>(
             let result = *c1_val + *c2;
             Some(AlgebraicExpression::BinaryOperation(
                 AlgebraicBinaryOperation {
-                    left: x.clone(),
+                    left: Box::new(x),
                     op: AlgebraicBinaryOperator::Add,
                     right: Box::new(AlgebraicExpression::Number(result)),
                 },
@@ -494,19 +496,22 @@ fn try_simplify_associative_operation<T: FieldElement>(
 
         // Case 2: Combining with any non-numeric expression
         // (X + C1) + Y -> (X + Y) + C1
-        y => Some(AlgebraicExpression::BinaryOperation(
-            AlgebraicBinaryOperation {
-                left: Box::new(AlgebraicExpression::BinaryOperation(
-                    AlgebraicBinaryOperation {
-                        left: x.clone(),
-                        op: AlgebraicBinaryOperator::Add,
-                        right: Box::new(y.clone()),
-                    },
-                )),
-                op: AlgebraicBinaryOperator::Add,
-                right: Box::new(AlgebraicExpression::Number(*c1_val)),
-            },
-        )),
+        y => {
+            let y = std::mem::replace(y, AlgebraicExpression::Number(0.into()));
+            Some(AlgebraicExpression::BinaryOperation(
+                AlgebraicBinaryOperation {
+                    left: Box::new(AlgebraicExpression::BinaryOperation(
+                        AlgebraicBinaryOperation {
+                            left: Box::new(x),
+                            op: AlgebraicBinaryOperator::Add,
+                            right: Box::new(y),
+                        },
+                    )),
+                    op: AlgebraicBinaryOperator::Add,
+                    right: Box::new(AlgebraicExpression::Number(*c1_val)),
+                },
+            ))
+        }
     }
 }
 
--- a/pilopt/src/lib.rs
+++ b/pilopt/src/lib.rs
@@ -436,8 +436,8 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
 }
 
 fn try_simplify_associative_operation<T: FieldElement>(
-    left: &Box<AlgebraicExpression<T>>,
-    right: &Box<AlgebraicExpression<T>>,
+    left: &AlgebraicExpression<T>,
+    right: &AlgebraicExpression<T>,
     op: AlgebraicBinaryOperator,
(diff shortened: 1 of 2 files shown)


pilopt/tests/optimizer.rs
```diff
@@ -465,3 +465,39 @@ fn equal_constrained_transitive() {
...
+        ((-x + 3) + c3) + 12 = z;
+    "#;
+
+    let expectation = r#"namespace N(150);
+    col witness x;
+    col witness y;
+    col witness z;
+    N::x + 3 = N::y;
+    N::x + 5 = N::y;
+    N::x - 2 + 1 = N::y;
```
chriseth, line 494: Nice! I think as a follow-up PR, we should also try to simplify this and the next line.

pilopt/src/lib.rs
```diff
@@ -427,6 +435,81 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
     }
 }
 
+fn try_simplify_associative_operation<T: FieldElement>(
+    left: &AlgebraicExpression<T>,
+    right: &AlgebraicExpression<T>,
+    op: AlgebraicBinaryOperator,
+) -> Option<AlgebraicExpression<T>> {
```
chriseth, line 442: Is it possibly to avoid the cloning? The expressions might be quite large, and we might do it recursively for all sub-expressions.

pilopt/src/lib.rs
```diff
@@ -427,6 +435,81 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
     }
 }
 
+fn try_simplify_associative_operation<T: FieldElement>(
```
chriseth, line 438: Can you document this function?

pilopt/src/lib.rs
```diff
@@ -427,6 +435,86 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
...
+    left: &mut AlgebraicExpression<T>,
+    right: &mut AlgebraicExpression<T>,
+    op: AlgebraicBinaryOperator,
+) -> Option<AlgebraicExpression<T>> {
+    if op != AlgebraicBinaryOperator::Add {
+        return None;
+    }
+
+    // Find binary operation and other expression, handling both orderings:
+    // (X + C1) + Other
```
Schaeff, line 449: I find the use of `C1` here a bit confusing, since it sounds like it's a constant but at this point it is not. Maybe `Y`?

gzanitti: Improved the comments to make them clearer and to better follow actual variables in the code. Thanks!

pilopt/src/lib.rs
```diff
@@ -471,21 +472,22 @@ fn try_simplify_associative_operation<T: FieldElement>(
...
-        (x2, val)
+        (x2.as_mut(), val)
     } else if let AlgebraicExpression::Number(val) = x2.as_ref() {
-        (x1, val)
+        (x1.as_mut(), val)
     } else {
         return None;
     };
 
+    let x = std::mem::replace(x, AlgebraicExpression::Number(0.into()));
```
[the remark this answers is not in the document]
gzanitti: `std::mem::take` needs to implement `Default`, so it would be similar to this case, but forcing us to choose a default for `AlgebraicExpression` in their impl. Do you think it's better that way?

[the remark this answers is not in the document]
Schaeff: Right, not necessarily, thanks!

pilopt/src/lib.rs
```diff
@@ -427,6 +435,86 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
     }
 }
 
+/// Uses associative properties to simplify expressions by regrouping constant terms.
+fn try_simplify_associative_operation<T: FieldElement>(
```
chriseth, line 439: The signature still looks a bit weird to me - how can you take two mutable references and return a new value?
I think it would be cleaner to just take a mutable reference to an `Expression` and return nothing (or a bool).

Otherwise you would have to explain what happens to `left` and `right` after the function call.

chriseth: Or, if that works with the returned value, take a mutable ref to a BinaryOperation.

pilopt/src/lib.rs
```diff
@@ -427,6 +435,86 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
...
+            let y = std::mem::replace(y, AlgebraicExpression::Number(0.into()));
+            Some(AlgebraicExpression::BinaryOperation(
+                AlgebraicBinaryOperation {
+                    left: Box::new(AlgebraicExpression::BinaryOperation(
+                        AlgebraicBinaryOperation {
+                            left: Box::new(x),
+                            op: AlgebraicBinaryOperator::Add,
+                            right: Box::new(y),
+                        },
+                    )),
```
chriseth, line 509: Does this work?
```suggestion
                    left: x + y,
```

gzanitti: Yes, thanks!

pilopt/src/lib.rs
```diff
@@ -427,6 +435,76 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
...
+    let (x, c1_val) = if let AlgebraicExpression::Number(val) = x1.as_ref() {
+        (x2.as_mut(), val)
+    } else if let AlgebraicExpression::Number(val) = x2.as_ref() {
+        (x1.as_mut(), val)
+    } else {
+        return None;
+    };
+
+    let x = std::mem::replace(x, AlgebraicExpression::Number(0.into()));
+    match other_expr {
```
chriseth, line 480: And now you could assign `binary_op = ...` here.

pilopt/src/lib.rs
```diff
@@ -427,6 +435,76 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
...
+            other,
+            AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation {
+                left: x1,
+                right: x2,
+                op: AlgebraicBinaryOperator::Add,
+            }),
+        ) => (x1, x2, other),
+        _ => return None,
+    };
+
```
chriseth, line 467: Can you add a comment like:
`Now we have "binary_op = x1 + x2 + other_expr"`.

pilopt/src/lib.rs
```diff
@@ -427,6 +435,76 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
...
+    // (X1 + C1) -> (X1, C1) if X2 is a constant
+    // (C1 + X2) -> (X2, C1) if X1 is a constant
+    let (x, c1_val) = if let AlgebraicExpression::Number(val) = x1.as_ref() {
+        (x2.as_mut(), val)
+    } else if let AlgebraicExpression::Number(val) = x2.as_ref() {
+        (x1.as_mut(), val)
+    } else {
+        return None;
+    };
+
```
chriseth, line 478: Can you add a comment like
`Now we have "binary_op = x + c1_val + other"`

pilopt/src/lib.rs
```diff
@@ -427,6 +435,76 @@ fn simplify_expression_single<T: FieldElement>(e: &mut AlgebraicExpression<T>) {
...
+    let x = std::mem::replace(x, AlgebraicExpression::Number(0.into()));
+    match other_expr {
+        // Case 1: Combining with a constant
+        // (X + C1) + Other -> X + (C1 + Other)
+        AlgebraicExpression::Number(c2) => {
+            let result = *c1_val + *c2;
+            Some(AlgebraicExpression::BinaryOperation(
+                AlgebraicBinaryOperation {
```
chriseth, line 486: Can you use `x + AlgebraicExpression(*c1_val + *c2)` here as well?

chriseth (approved):

PR #2421, opened 2025-01-31
Stwo challenges sound

ShuangWu121:
# Make Challenge Function in `stwo` Sound

## Summary
This PR adds soundness to stwo challenge function

## Main Changes
- **Separation of Stage 0 and Stage 1 Witness Columns**  
 separate stage 0 and stage 1 witness columns in circuit building phase (PowdrEval) and proving phase 
- **Make the challenge generation based on commitments of constant cols and stage 0 witness cols**  
- **Add more stwo tests to existing tests**


ast/src/analyzed/mod.rs
```diff
@@ -101,6 +101,12 @@ impl<T> Analyzed<T> {
     pub fn commitment_count(&self) -> usize {
         self.declaration_type_count(PolynomialType::Committed)
     }
+
+    /// @returns the number of committed polynomials (with multiplicities for arrays) in a specific stage
+    pub fn stage_commitment_count(&self, stage: Option<u32>) -> usize {
```
Schaeff, line 106: Could we change stage to u32 here? It defaults to 0 in the called function anyway

ast/src/analyzed/mod.rs
```diff
@@ -200,6 +206,21 @@ impl<T> Analyzed<T> {
             .sum()
     }
 
+    fn stage_declaration_type_count(&self, poly_type: PolynomialType, stage: Option<u32>) -> usize {
```
Schaeff, line 209: Same here

ast/src/analyzed/mod.rs
```diff
@@ -200,6 +206,21 @@ impl<T> Analyzed<T> {
...
+        self.definitions
+            .iter()
```
Schaeff, line 211:
```suggestion
            .values()
```
the name is never used

backend/src/stwo/circuit_builder.rs
```diff
@@ -56,7 +58,8 @@ pub struct PowdrEval {
     // the pre-processed are indexed in the whole proof, instead of in each component.
     // this offset represents the index of the first pre-processed column in this component
     preprocess_col_offset: usize,
-    witness_columns: BTreeMap<PolyID, usize>,
+    witness_columns_stage0: BTreeMap<PolyID, usize>,
```
Schaeff, line 61:
```suggestion
    stage0_witness_columns: BTreeMap<PolyID, usize>,
```
? same below

backend/src/stwo/circuit_builder.rs
```diff
@@ -70,13 +73,26 @@ impl PowdrEval {
         log_degree: u32,
         challenges: BTreeMap<u64, M31>,
     ) -> Self {
-        let witness_columns: BTreeMap<PolyID, usize> = analyzed
+        let witness_columns_stage0: BTreeMap<PolyID, usize> = analyzed
             .definitions_in_source_order(PolynomialType::Committed)
+            .filter(|(symbol, _)| symbol.stage.is_none() || symbol.stage == Some(0))
```
Schaeff, line 78: same here, `unwrap_or(0)`

backend/src/stwo/circuit_builder.rs
```diff
@@ -70,13 +73,26 @@ impl PowdrEval {
...
             .flat_map(|(symbol, _)| symbol.array_elements())
             .enumerate()
             .map(|(index, (_, id))| (id, index))
             .collect();
 
+        let witness_columns_stage1: BTreeMap<PolyID, usize> = analyzed
+            .definitions_in_source_order(PolynomialType::Committed)
+            .filter(|(symbol, _)| symbol.stage == Some(1))
```
Schaeff, line 86: Same here

backend/src/stwo/prover.rs
```diff
@@ -453,12 +456,45 @@ where
...
+                    constant_col_log_sizes.extend(
+                        repeat(machine_log_size)
+                            .take(pil.constant_count() + get_constant_with_next_list(pil).len()),
+                    );
+                    stage0_witness_col_log_sizes
+                        .extend(repeat(machine_log_size).take(pil.stage_commitment_count(None)));
+                    stage1_witness_col_log_sizes
+                        .extend(repeat(machine_log_size).take(pil.stage_commitment_count(Some(1))));
+                },
+            );
```
Schaeff, line 479: Could we avoid `mut` and `for_each` here? Basically when a closure accesses variables declared outside of itself, it makes the code harder to read. Instead we could do:
```rust
// create the iterator once since we use it on all three vectors
let iter = self.split().iter().zip_eq(proof.machine_log_sizes);
// building the arrays separately is nicer because we can use `collect` directly
let constant_col_log_sizes = iter.clone().flat_map(|...| ...).collect();
let stage0_witness_col_log_sizes = iter.clone().flat_map(|...| ...).collect();
// etc
```

Alternatively we could use 
```rust
fold((a, b, c), |(mut a, mut b, mut c), ...| {
   a.extend(...);
   b.extend(...);
   c.extend(...);
   (a, b, c)
})
```

but it's more verbose, and iterating three times instead of one is fine

Schaeff (commented):
Super nice! A few comments to make the code nicer

ast/src/analyzed/mod.rs
ShuangWu121: Done.

ast/src/analyzed/mod.rs
ShuangWu121: Done

backend/src/stwo/circuit_builder.rs

backend/src/stwo/circuit_builder.rs

backend/src/stwo/circuit_builder.rs

backend/src/stwo/prover.rs
ShuangWu121: Got it. code updated

backend/src/stwo/circuit_builder.rs
```diff
@@ -223,11 +223,11 @@ impl FrameworkEval for PowdrEval {
             .map(|(k, v)| (*k, E::F::from(into_stwo_field(v))))
             .collect();
         // concatenate stage0 and stage1 witness columns
-        witness_eval_stage0.extend(witness_eval_stage1);
+        stage0_witness_eval.extend(stage1_witness_eval);
 
         let intermediate_definitions = self.analyzed.intermediate_definitions();
         let data = Data {
-            witness_eval: &witness_eval_stage0,
+            witness_eval: &stage0_witness_eval,
```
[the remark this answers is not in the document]
Schaeff: I think p3 has something like this, but I may be missing reasons why the approach doesn't fit here

backend/src/stwo/prover.rs
```diff
@@ -451,14 +454,56 @@ where
...
         let mut constant_cols_offset_acc = 0;
+        let iter = self.split.iter().zip_eq(proof.machine_log_sizes.iter());
 
-        let mut constant_col_log_sizes = vec![];
-        let mut witness_col_log_sizes = vec![];
+        let constant_col_log_sizes = iter
+            .clone()
+            .flat_map(
+                |((machine_name, pil), (proof_machine_name, &machine_log_size))| {
+                    assert_eq!(machine_name, proof_machine_name);
```
Schaeff, line 463: These assertions can be put on the initial iterator

backend/src/stwo/circuit_builder.rs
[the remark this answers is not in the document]
ShuangWu121: yes. code updated

backend/src/stwo/prover.rs
```diff
@@ -451,14 +454,56 @@ where
...
+                    repeat(machine_log_size)
+                        .take(pil.constant_count() + get_constant_with_next_list(pil).len())
+                },
+            )
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
ShuangWu121: good idea, done.

backend/src/stwo/circuit_builder.rs
```diff
@@ -190,6 +193,13 @@ impl FrameworkEval for PowdrEval {
             })
             .collect();
 
+        let poly_stage_map: BTreeMap<PolyID, usize> = self
+            .stage0_witness_columns
+            .keys()
+            .map(|k| (*k, 0))
+            .chain(self.stage1_witness_columns.keys().map(|k| (*k, 1)))
+            .collect();
```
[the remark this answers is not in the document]
ShuangWu121: Make sense. I moved the `poly_stage_map` to PowdrEval, created by PowdrEval::new function, this function is called per machine for creating circuit of the machine, would this be good enough? I saw `ConstraintSystem` is created for every split pil.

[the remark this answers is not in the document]
Schaeff: Yes that's fine! One good next step if performance is disappointing would be to profile a nontrivial stwo prover run and make sure we're not spending too much time outside of stwo. This can be a different PR.

backend/src/stwo/prover.rs
```diff
@@ -336,77 +322,94 @@ where
...
 
         // Generate challenges for stage 1 based on stage 0 traces.
-        // Stwo supports a maximum of 2 stages, and challenges are created only for stage 0.
+        // Stwo supports a maximum of 2 stages, and challenges are created only after stage 0.
+        let stage0_challenges = get_challenges::<MC>(&self.analyzed, prover_channel);
 
-        let stage0_challenges = get_dummy_challenges::<MC>(&self.analyzed);
-
-        if self.analyzed.stage_count() > 1 {
+        let stage1_witness_cols_circle_domain_eval = if self.analyzed.stage_count() > 1 {
```
Schaeff, line 340: We have this `self.analyzed.stage_count() > 1` condition twice, using options and iterating over them. We can simplify this.

```rust
   let stage1_witness_cols_circle_domain_eval = ...;
   tree_builder.extend_evals(...);
   tree_builder.commit(...);
}
```

Let's make sure this prover code is as simple as possible

backend/src/stwo/prover.rs
```diff
@@ -336,77 +322,94 @@ where
...
+        };
+        if self.analyzed.stage_count() > 1 {
+            let mut tree_builder = commitment_scheme.tree_builder();
+            tree_builder.extend_evals(stage1_witness_cols_circle_domain_eval.into_iter().flatten());
+            tree_builder.commit(prover_channel);
+        };
 
         let tree_span_provider = &mut TraceLocationAllocator::default();
+        // Each column size in machines needs its own component, the components from different machines are stored in this vector
+        let mut components = Vec::new();
```
Schaeff, line 388: This should not change, the (almost, with the exception of the side effect on constant_cols_offset_acc) functional approach was clearer.

backend/src/stwo/prover.rs
```diff
@@ -337,79 +337,76 @@ where
...
+                                        .get(&(vec.len().ilog2() as usize))
+                                        .expect("Domain not found for given size"),
+                                    vec,
+                                ))
+                            }
+                        })
+                        .collect_vec()
+                });
+
             let mut tree_builder = commitment_scheme.tree_builder();
```
Schaeff, line 375: can `let mut tree_builder = commitment_scheme.tree_builder();` be removed? `tree_builder` is already defined before

backend/src/stwo/prover.rs
```diff
@@ -499,31 +499,25 @@ where
...
-                        (SecureField::zero(), None),
-                    );
-
-                    constant_cols_offset_acc += pil.constant_count();
-
-                    constant_cols_offset_acc += get_constant_with_next_list(pil).len();
-                    machine_component
-                },
-            )
+        let components = iter
```
Schaeff, line 502: nice!

backend/src/stwo/prover.rs
ShuangWu121: this tree_builder is to build one merkle tree and one merkle tree root, constant cols, witness for stage 0, witness for stage 1, all have their own merkle tree and merkle tree root, they need to be separated. So we cannot reuse here ^^

backend/src/stwo/prover.rs
ShuangWu121: I realized this change comes from a bad merge of this branch with the stwo-challenges-NotSound branch, --; Sorry

backend/src/stwo/prover.rs

Schaeff (approved):

backend/src/stwo/prover.rs
ShuangWu121: Now it changed back.

PR #2425, opened 2025-02-01
Linear constraint remover

Schaeff:
# what

Replace certain multilinear constraints* with intermediate polynomials in pilopt.
*with no next references, not touching public inputs

# why

[This comment](https://github.com/powdr-labs/powdr/issues/2337#issuecomment-2621634278) highlights a use-case where materialising a column which is only constrained to a non-shifted multilinear polynomial is wasteful. Fixing this when we create the code is tricky, since we cannot know the degree of the constraint from pil. This is however something pilopt can do, which is what we attempt to do here

# unknowns

- [ ] Is it sound to do this?
- [ ] What are the consequences on witgen? It seems ok since the colun still "exists" as intermediate, it's just not materialized
- [ ] We have similar optimizers for `x = 42` and `x = y` which are both multilinear expressions. They do no yield intermediate polynomials, since in that case it is fine to "inline" them. This implementation currently makes sure all three optimizer apply to distinct cases, but maybe it would be better to have a single optimizer which handles all three cases, only introducing intermediate polynomials in the non trivial cases (`42` and `y`)

### DIFF
--- a/ast/src/parsed/mod.rs
+++ b/ast/src/parsed/mod.rs
@@ -15,13 +15,13 @@ use std::{
 
 use auto_enums::auto_enum;
 use derive_more::Display;
-use powdr_number::{BigInt, BigUint, DegreeType};
+use powdr_number::{BigInt, BigUint, DegreeType, FieldElement};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 use powdr_parser_util::SourceRef;
 
-use crate::analyzed::Reference;
+use crate::analyzed::{AlgebraicExpression, PolynomialReference, Reference};
 
 use self::{
     asm::{Part, SymbolPath},
@@ -738,6 +738,30 @@ impl<Ref> From<u32> for Expression<Ref> {
         BigUint::from(value).into()
     }
 }
+
+impl<T: FieldElement> From<AlgebraicExpression<T>> for Expression<Reference> {
+    fn from(e: AlgebraicExpression<T>) -> Self {
+        match e {
+            AlgebraicExpression::Reference(algebraic_reference) => Expression::Reference(
+                SourceRef::unknown(),
+                Reference::Poly(PolynomialReference {
+                    name: algebraic_reference.name,
+                    type_args: None,
+                }),
+            ),
+            AlgebraicExpression::PublicReference(_) => todo!(),
+            AlgebraicExpression::Challenge(_) => todo!(),
+            AlgebraicExpression::Number(n) => Number {
+                value: n.to_arbitrary_integer(),
+                type_: Some(Type::Expr),
+            }
+            .into(),
+            AlgebraicExpression::BinaryOperation(_) => todo!(),
+            AlgebraicExpression::UnaryOperation(_) => todo!(),
+        }
+    }
+}
+
 pub type ExpressionPrecedence = u64;
 
 impl<Ref> Expression<Ref> {
--- a/pilopt/src/lib.rs
+++ b/pilopt/src/lib.rs
@@ -7,12 +7,12 @@ use std::hash::{DefaultHasher, Hash, Hasher};
 use itertools::Itertools;
 use powdr_ast::analyzed::{
     AlgebraicBinaryOperation, AlgebraicBinaryOperator, AlgebraicExpression, AlgebraicReference,
-    AlgebraicUnaryOperation, AlgebraicUnaryOperator, Analyzed, ConnectIdentity, Expression,
-    FunctionValueDefinition, Identity, LookupIdentity, PermutationIdentity, PhantomLookupIdentity,
-    PhantomPermutationIdentity, PolyID, PolynomialIdentity, PolynomialReference, PolynomialType,
-    Reference, StatementIdentifier, Symbol, SymbolKind,
+    AlgebraicReferenceThin, AlgebraicUnaryOperation, AlgebraicUnaryOperator, Analyzed,
+    ConnectIdentity, Expression, FunctionValueDefinition, Identity, LookupIdentity,
+    PermutationIdentity, PhantomLookupIdentity, PhantomPermutationIdentity, PolyID,
+    PolynomialIdentity, PolynomialReference, PolynomialType, Reference, StatementIdentifier,
+    Symbol, SymbolKind,
 };
-use powdr_ast::parsed::types::Type;
 use powdr_ast::parsed::visitor::{AllChildren, Children, ExpressionVisitable};
 use powdr_ast::parsed::Number;
 use powdr_number::{BigUint, FieldElement};
@@ -30,6 +30,9 @@ pub fn optimize<T: FieldElement>(mut pil_file: Analyzed<T>) -> Analyzed<T> {
         deduplicate_fixed_columns(&mut pil_file);
         simplify_identities(&mut pil_file);
         extract_constant_lookups(&mut pil_file);
+        println!("before {pil_file}");
+        replace_linear_witness_columns(&mut pil_file);
+        println!("after {pil_file}");
         remove_constant_witness_columns(&mut pil_file);
         remove_constant_intermediate_columns(&mut pil_file);
         simplify_identities(&mut pil_file);
@@ -237,7 +240,10 @@ fn remove_constant_fixed_columns<T: FieldElement>(pil_file: &mut Analyzed<T>) {
                 "Determined fixed column {} to be constant {value}. Removing.",
                 poly.absolute_name
             );
-            Some(((poly.absolute_name.clone(), poly.into()), value))
+            Some((
+                (poly.absolute_name.clone(), poly.into()),
+                T::from(value).into(),
+            ))
         })
         .collect::<Vec<((String, PolyID), _)>>();
 
@@ -492,6 +498,76 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
     }
 }
 
+/// Identifies witness columns that are constrained to a (multi)linear expression, replaces the witness column by an intermediate polynomial
+fn replace_linear_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>) {
+    let intermediate_definitions = pil_file.intermediate_definitions();
+
+    let mut linear_polys = pil_file
+        .identities
+        .iter()
+        .enumerate()
+        .filter_map(|(index, id)| {
+            if let Identity::Polynomial(PolynomialIdentity { expression: e, .. }) = id {
+                Some((index, e))
+            } else {
+                None
+            }
+        })
+        .filter_map(|(index, identity)| {
+            constrained_to_linear(identity, &intermediate_definitions).map(|e| (index, e))
+        })
+        .collect::<Vec<(usize, ((String, PolyID), _))>>();
+
+    let in_publics: HashSet<_> = pil_file
+        .public_declarations
+        .values()
+        .map(|pubd| pubd.polynomial.name.clone())
+        .collect();
+    // We cannot remove arrays or array elements, so filter them out.
+    // Also, we filter out columns that are used in public declarations.
+    let columns = pil_file
+        .committed_polys_in_source_order()
+        .filter(|&(s, _)| !s.is_array() && !in_publics.contains(&s.absolute_name))
+        .map(|(s, _)| s.into())
+        .collect::<HashSet<PolyID>>();
+    linear_polys.retain(|(_, ((_, id), _))| columns.contains(id));
+
+    let (identities_to_remove, intermediate_polys): (BTreeSet<_>, Vec<_>) = linear_polys
+        .iter()
+        .filter_map(|(index, ((name, poly_id), expression))| {
+            // Remove the definition. If this fails, we have already replaced it by an intermediate column.
+            pil_file
+                .definitions
+                .remove(name)
+                .map(|(mut new_symbol, _)| {
+                    let new_poly_id = PolyID {
+                        ptype: PolynomialType::Intermediate,
+                        ..*poly_id
+                    };
+                    new_symbol.kind = SymbolKind::Poly(PolynomialType::Intermediate);
+                    // Add the definition to the intermediate columns
+                    pil_file
+                        .intermediate_columns
+                        .insert(name.clone(), (new_symbol, vec![expression.clone()]));
+                    // Note: the `statement_order` does not need to be updated, since we are still declaring the same symbol
+                    let r = AlgebraicReference {
+                        name: name.clone(),
+                        poly_id: new_poly_id,
+                        next: false,
+                    };
+                    (
+                        index,
+                        ((name.clone(), *poly_id), AlgebraicExpression::Reference(r)),
+                    )
+                })
+        })
+        .unzip();
+    // we remove the identities by hand here, because they do *not* become trivial by applying this optimization
+    pil_file.remove_identities(&identities_to_remove);
+    // we substitute the references to the witnesses by references to the intermediate columns
+    substitute_polynomial_references(pil_file, intermediate_polys);
+}
+
 /// Identifies witness columns that are constrained to a single value, replaces every
 /// reference to this column by the value and deletes the column.
 fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>) {
@@ -506,6 +582,7 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>) 
             }
         })
         .filter_map(constrained_to_constant)
+        .map(|(k, v)| (k, T::from(v).into()))
         .collect::<Vec<((String, PolyID), _)>>();
 
     let in_publics: HashSet<_> = pil_file
@@ -541,7 +618,7 @@ fn remove_constant_intermediate_columns<T: FieldElement>(pil_file: &mut Analyzed
                             log::debug!(
                                 "Determined intermediate column {name} to be constant {value}. Removing.",
                             );
-                            Some(((name.clone(), poly_id), value.to_arbitrary_integer()))
+                            Some(((name.clone(), poly_id), (*value).into()))
                         }
                         _ => None,
                     }
@@ -553,10 +630,10 @@ fn remove_constant_intermediate_columns<T: FieldElement>(pil_file: &mut Analyzed
     substitute_polynomial_references(pil_file, intermediate_polys);
 }
 
-/// Substitutes all references to certain polynomials by the given field elements.
+/// Substitutes all references to certain polynomials by the given non-shifted expressions.
 fn substitute_polynomial_references<T: FieldElement>(
     pil_file: &mut Analyzed<T>,
-    substitutions: Vec<((String, PolyID), BigUint)>,
+    substitutions: Vec<((String, PolyID), AlgebraicExpression<T>)>,
 ) {
     let substitutions_by_id = substitutions
         .iter()
@@ -573,18 +650,19 @@ fn substitute_polynomial_references<T: FieldElement>(
         ) = e
         {
             if let Some(value) = substitutions_by_name.get(name) {
-                *e = Number {
-                    value: (*value).clone(),
-                    type_: Some(Type::Expr),
-                }
-                .into();
+                *e = value.clone().into();
             }
         }
     });
     pil_file.post_visit_expressions_in_identities_mut(&mut |e: &mut AlgebraicExpression<_>| {
-        if let AlgebraicExpression::Reference(AlgebraicReference { poly_id, .. }) = e {
+        if let AlgebraicExpression::Reference(AlgebraicReference { poly_id, next, .. }) = e {
             if let Some(value) = substitutions_by_id.get(poly_id) {
-                *e = AlgebraicExpression::Number(T::checked_from((*value).clone()).unwrap());
+                let value = if *next {
+                    value.clone().next().unwrap()
+                } else {
+                    value.clone()
+                };
+                *e = value;
             }
         }
     });
@@ -618,6 +696,67 @@ fn constrained_to_constant<T: FieldElement>(
         _ => {}
     }
     None
+}
+
+/// Identifies witness columns that are constrained to a (multi)linear expression.
+fn constrained_to_linear<T: FieldElement>(
+    identity: &AlgebraicExpression<T>,
+    intermediate_columns: &BTreeMap<AlgebraicReferenceThin, AlgebraicExpression<T>>,
+) -> Option<((String, PolyID), AlgebraicExpression<T>)> {
+    // We require the constraint to be of the form `left = right`, represented as `left - right`
+    let (left, right) = if let AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation {
+        left,
+        op: AlgebraicBinaryOperator::Sub,
+        right,
+    }) = identity
+    {
+        (left, right)
+    } else {
+        return None;
+    };
+
+    // We require `left` to be a single, non-shifted, witness column `w`
+    let w = if let AlgebraicExpression::Reference(
+        r @ AlgebraicReference {
+            poly_id:
+                PolyID {
+                    ptype: PolynomialType::Committed,
+                    ..
+                },
+            next: false,
+            ..
+        },
+    ) = left.as_ref()
+    {
+        r
+    } else {
+        return None;
+    };
+
+    // we require `y` to be a multi-linear expression over non-shifted columns
+    if right.contains_next_ref()
+        || right.degree_with_cache(intermediate_columns, &mut Default::default()) != 1
+    {
+        return None;
+    }
+
+    // we require `y` not to be a single, non-shifted, witness column, because this case is already covered and does not require an intermediate column
+    // TODO: maybe we should avoid this edge case by removing the other optimizer, see xxx
+    if matches!(
+        right.as_ref(),
+        AlgebraicExpression::Reference(AlgebraicReference {
+            poly_id: PolyID {
+                ptype: PolynomialType::Committed,
+                ..
+            },
+            next: false,
+            ..
+        })
+    ) {
+        return None;
+    }
+
+    Some(((w.name.clone(), w.poly_id), *right.clone()))
 }
 
 /// Removes identities that evaluate to zero (including constraints of the form "X = X") and lookups with empty columns.
--- a/pilopt/tests/optimizer.rs
+++ b/pilopt/tests/optimizer.rs
@@ -15,14 +15,14 @@ fn replace_fixed() {
         let _ = one;
     };
     X * one = X * zero - zero + Y;
-    one * Y = zero * Y + 7 * X;
+    one * Y = zero * Y + 7 * X * X;
 "#;
     let expectation = r#"namespace N(65536);
     col witness X;
     query |i| {
         let _: expr = 1_expr;
     };
-    N::X = 7 * N::X;
+    N::X = 7 * N::X * N::X;
 "#;
     let optimized = optimize(analyze_string::<GoldilocksField>(input).unwrap()).to_string();
     assert_eq!(optimized, expectation);
@@ -235,7 +235,7 @@ fn remove_unreferenced_keep_enums() {
         // Y is not mentioned anywhere.
         let f: col = |i| if i == 0 { t([]) } else { (|x| 1)(Y::F([])) };
         let x;
-        x = f;
+        x = f * f;
     "#;
     let expectation = r#"namespace N(65536);
     enum X {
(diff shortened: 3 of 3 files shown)


ast/src/parsed/mod.rs
```diff
@@ -738,6 +738,30 @@ impl<Ref> From<u32> for Expression<Ref> {
         BigUint::from(value).into()
     }
 }
+
+impl<T: FieldElement> From<AlgebraicExpression<T>> for Expression<Reference> {
```
chriseth, line 742: There is `pil-analyzer/src/epressionizer.rs` that does this on a large scale. Maybe call that from pilopt instead of adding this here?

Schaeff: I knew we had this somewhere! Thanks! I added this just to make it compile

pilopt/tests/optimizer.rs
```diff
@@ -465,3 +465,46 @@ fn equal_constrained_transitive() {
...
+        constrained_twice = w + f;
+    "#;
+    let expectation = r#"namespace N(65536);
+    col witness w;
+    col fixed f = [1_fe, 0_fe]*;
+    col can_be_replaced = 2 * N::w + 3 * N::f + 5;
+    N::can_be_replaced + N::w = 5;
+    col witness linear_with_next_ref;
+    N::linear_with_next_ref = 2 * N::w + 3 * N::f' + 5;
+    N::linear_with_next_ref + N::w = 5;
```
qwang98, line 501: Does it mean here that `col` is an expression that's not materialized whereas `col witness` is a materialized witness column?

pilopt/src/lib.rs
```diff
@@ -492,6 +497,105 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+        .public_declarations
+        .values()
+        .map(|pubd| pubd.polynomial.name.clone())
+        .collect();
+
+    let keep = pil_file
+        .committed_polys_in_source_order()
+        .filter(|&(s, _)| s.is_array() || in_publics.contains(&s.absolute_name))
+        .map(|(s, _)| s.into())
+        .collect::<HashSet<PolyID>>();
```
qwang98, line 517: Note that in the latest PR that processes arbitrary number of bus interactions with the same accumulator, folded columns, which could have degree 1 and thus be optimized away, are defined in an array: https://github.com/powdr-labs/powdr/blob/166808b6ab2ec11e969f8b6b445feb32c6fbcebf/std/protocols/bus.asm#L82-L83

Does it mean that because we can't remove array elements, pilopt (at least of now) won't be able to remove such folded columns?

pilopt/src/lib.rs
```diff
@@ -620,6 +730,67 @@ fn constrained_to_constant<T: FieldElement>(
...
+            poly_id: PolyID {
+                ptype: PolynomialType::Committed,
+                ..
+            },
+            next: false,
+            ..
+        })
+    ) {
+        return None;
+    }
```
qwang98, line 789: I'm a bit confused by this chunk. It seems that we have opposite requirements for left and right being a single, non-shifted, witness column AND also no requirements for left being a multi-linear expression. In practice, shouldn't we have both left and right be single and non shifted multilinear expression OR witness column?

pilopt/src/lib.rs
```diff
@@ -492,6 +497,105 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+
+    let (identities_to_remove, intermediate_polys): (BTreeSet<_>, Vec<_>) = linear_polys
+        .iter()
+        .filter_map(|(index, ((name, old_poly_id), expression))| {
+            // Remove the definition. If this fails, we have already replaced it by an intermediate column.
+            pil_file.definitions.remove(name).map(|(symbol, value)| {
+                // sanity checks that we are looking at a single witness column
+                assert!(symbol.kind == SymbolKind::Poly(PolynomialType::Committed));
+                assert!(value.is_none());
+                assert!(symbol.length.is_none());
```
qwang98, line 544: For my learning, these should always be true because LHS of `linear_polys` are already constrained to be non-array witness with no value definition?

pilopt/src/lib.rs
```diff
@@ -492,6 +497,105 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+                    .map(|(_, poly_id)| poly_id.id)
+                    .max()
+                    .unwrap_or(0)
+                    + 1;
+
+                // the symbol is not an array, so the poly_id uses the same id as the symbol
+                let poly_id = PolyID {
+                    ptype: PolynomialType::Intermediate,
+                    id,
+                };
```
qwang98, line 560: Will this style of defining the poly_id (using the symbol id) cause some clash in poly_id? Symbol id has the following definition: `/// Unique ID of the symbol within the set of symbols of this kind.`

I'm also a bit confused by the difference between symbol_id and poly_id. Sounds like symbol_id can be for an array which contains multiple polynomials with different poly_id, but what would be some other use cases?

pilopt/src/lib.rs
```diff
@@ -492,6 +497,105 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+                // Note: the `statement_order` does not need to be updated, since we are still declaring the same symbol
+                let r = AlgebraicReference {
+                    name: name.clone(),
+                    poly_id,
+                    next: false,
+                };
+                (
+                    index,
+                    (
+                        (name.clone(), *old_poly_id),
```
qwang98, line 586: Because these are used to substitute polynomials, does it mean that neither name nor poly_id uniquely defines a polynomial and we have to use both? What was the reason we didn't define a unique poly_id initially?

pilopt/src/lib.rs
```diff
@@ -573,18 +682,19 @@ fn substitute_polynomial_references<T: FieldElement>(
...
             if let Some(value) = substitutions_by_id.get(poly_id) {
-                *e = AlgebraicExpression::Number(T::checked_from((*value).clone()).unwrap());
+                let value = if *next {
+                    value.clone().next().unwrap()
+                } else {
+                    value.clone()
+                };
+                *e = value;
             }
         }
```
qwang98, line 699: Mostly for my learning purpose, why are `definitions` and `prover_functions` in `pil_file` have expressions defined on the parser level (higher level) whereas `identities` and `intermediate_columns` in `pil_file` have algebraic expressions defined on the analyzed pil level (lower level)? This also helps understand why `try_algrebraic_expression_to_expression` is needed, as it seems to convert backwards from a lower level to a higher level.

test_data/pil/simple_div.pil
```diff
@@ -16,8 +17,8 @@ namespace SimpleDiv(N);
 
     X * Z + R = Y;
     Z - R - 1 = Y_b1 + Y_b2 * 0x100;
-    X = X_b1 + X_b2 * 0x100;
-    R = R_b1 + R_b2 * 0x100;
+    X = one * (X_b1 + X_b2 * 0x100);
+    R = one * (R_b1 + R_b2 * 0x100);
```
qwang98, line 21: Is there a way to force materializing a witness column instead of multiplying by `one` here? If `X` and `R` are exposed as an operation here, will they still be optimized away?

backend/src/mock/mod.rs
```diff
@@ -86,8 +87,18 @@ impl<F: FieldElement> Backend<F> for MockBackend<F> {
...
+                    .iter()
+                    .flat_map(|identity| identity.all_children())
+                    .chain(
+                        // Note: we iterate on a `HashMap` so the ordering is not guaranteed, but this is ok since we're building another map.
+                        pil.intermediate_columns
+                            .values()
+                            .flat_map(|(_, def)| def.iter().flat_map(|d| d.all_children())),
+                    )
+                    .collect_vec()
+            })
```
qwang98, line 101: Is this edit because we now move some identities to intermediates, so we have to use both as the preimage of challenge derivation, in order to guarantee sufficient randomness?

backend/src/stwo/circuit_builder.rs
```diff
@@ -272,21 +274,55 @@ impl FrameworkEval for PowdrEval {
...
+                        intermediates_cache.insert(reference, result.clone());
+                        result
+                    })
+            }
+            PolynomialType::Committed => std::iter::empty().collect(),
+        })
+        .fold(BTreeSet::new(), |mut acc, set| {
+            acc.extend(set);
+            acc
+        })
```
qwang98, line 327: Are these edits needed, because previously intermediates can't contain next references to constants as children but now they can and these intermediates are always prior witness columns downgraded to intermediates by the optimizer?

qwang98: So basically no next reference to constants for intermediates is a user interface level requirement but not a hard requirement in lower level?

plonky3/src/circuit_builder.rs
```diff
@@ -97,12 +97,21 @@ impl<T: FieldElement> From<&Analyzed<T>> for ConstraintSystem<T> {
...
+                    }
+                });
+            }
+            for definition in analyzed.intermediate_definitions().values() {
+                definition.pre_visit_expressions(&mut |expr| {
+                    if let AlgebraicExpression::Challenge(challenge) = expr {
+                        challenges_by_stage[challenge.stage as usize].insert(challenge.id);
+                    }
+                });
+            }
```
qwang98, line 114: This is because some challenge in witness stage can be moved to intermediates now?

pilopt/tests/optimizer.rs
Schaeff: Exactly!

pilopt/src/lib.rs
Schaeff: No, here I'm referring to column arrays like `col witness w[3]`. In the bus, we create columns with `prover::new_column_with_stage("w", 1)` which yields `col witness w`, which is compatible with this optimizer.

pilopt/src/lib.rs
Schaeff: The current pattern is:
```
w = lin
```

where `w` is a non shifted witness column and lin is a multilinear expression over non-shifted columns.
Additionally, we require `lin` NOT to be a single non shifted column, because we already have other optimizer for this case that we don't want to interfere with here.

Are you suggesting that we could also support `lin = w` with the same requirements?

pilopt/src/lib.rs
Schaeff: The reason is legacy code from polygon where the ids are per column type.

pilopt/src/lib.rs
Schaeff: I *think* that's because identities and intermediate columns are defined in terms of algebraic expressions, whereas definitions and prover functions have arbitrary expressions. Algebraic expressions are the language of the proof system (more restricted), while arbitrary expressions are the language of witness generation (less restricted).

plonky3/src/circuit_builder.rs
Schaeff: No, this is fixing an existing bug: when collecting all challenges used in the program, we have to check the intermediate definitions as well, not just the identities.

test_data/pil/simple_div.pil
Schaeff: Both having a way to block the optimizer and #2464 seem overkill just to fix a test. In practice the quotient and remainder do turn out to be single witness columns, it's the simplicity of this example which makes it fail due to inlining, so the `one` trick seems acceptable to me, similar to other tests here.

qwang98: Could we add a comment here to explain that we don't want to be optimized away? Otherwise I think this example could be confusing to first time readers.

backend/src/stwo/circuit_builder.rs
Schaeff: This is needed because we need to find all constant columns which are referenced in the current and next row, respectively, and this requires recursively looking at intermediate columns:

```
col fixed F = ...;
col a = F;
col a_next = F';
col witness b;
b = a_next; // this means that `F'` is accessed, by definition of `a_next`
b = a; // this means that `F` is accessed, by definition of `a`
b = a'; // this means that `F'` is accessed, by definition of `a`
```

backend/src/mock/mod.rs
Schaeff: No it's simpler than that:

```
col a = challenge(0, 0); // the challenge is declared in the intermediate definition
a = a * a; // in the identity, the challenge does not appear
```

Here we simply look through all expressions as opposed to starting with identities and visiting the definitions of the intermediates they access, which is the same assuming that unused intermediates have been optimized away. If not, we may end up introducing challenges which are not needed.

test_data/pil/simple_div.pil
Schaeff: Added

ast/src/analyzed/mod.rs
```diff
@@ -89,6 +89,24 @@ impl<T> Analyzed<T> {
...
+    pub fn challenges(&self) -> BTreeSet<&Challenge> {
+        self.identities
+            .iter()
+            .flat_map(|identity| identity.all_children())
+            .chain(
+                // Note: we iterate on a `HashMap` so the ordering is not guaranteed, but this is ok since we're building another map.
+                self.intermediate_columns
+                    .values()
+                    .flat_map(|(_, def)| def.iter().flat_map(|d| d.all_children())),
+            )
```
chriseth, line 102:
```suggestion
        self.all_children()
```

backend/src/stwo/circuit_builder.rs
```diff
@@ -310,21 +312,55 @@ impl FrameworkEval for PowdrEval {
...
-            name,
-            poly_id,
-            next,
-        }) = e
-        {
-            if matches!(poly_id.ptype, PolynomialType::Constant) && *next {
-                // add the name of the constant polynomial to the list
-                constant_with_next_list.insert(name);
+// This function creates a list of the names of the constant polynomials that have next references
+// Note that the anaylsis should also dereference next references to intermediate polynomials
```
chriseth, line 316:
```suggestion
/// This function creates a list of the names of the constant polynomials that have next references
/// Note that the anaylsis should also dereference next references to intermediate polynomials
```

pil-analyzer/src/expressionizer.rs
```diff
@@ -58,6 +58,14 @@ pub fn try_value_to_expression<T: FieldElement>(
     Expressionizer { poly_id_to_name }.try_value_to_expression(value)
 }
 
+/// Tries to convert an algebraic expression to an expression.
```
chriseth, line 61:
```suggestion
/// Tries to convert an algebraic expression to an expression that evaluates to the same value.
```

pilopt/src/lib.rs
```diff
@@ -558,6 +563,178 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
     }
 }
 
+/// Identifies witness columns that are constrained to a non-shifted (multi)linear expression, replaces the witness column by an intermediate polynomial
```
chriseth, line 566: Can you extend the comment a little? What is the intermediate polynomial and why are we doing this optimization?

pilopt/src/lib.rs
```diff
@@ -558,6 +563,178 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+    // Also, we filter out columns that are boolean constrained.
+
+    let in_publics: HashSet<_> = pil_file
+        .public_declarations
+        .values()
+        .map(|pubd| pubd.polynomial.name.clone())
+        .collect();
+
+    // pattern match identities looking for `w * (1 - w) = 0` and `(1 - w) * w = 0` constraints
+    let boolean_constrained_witnesses = pil_file
```
chriseth, line 583: Can this be a helper function?

pilopt/src/lib.rs
```diff
@@ -686,6 +886,67 @@ fn constrained_to_constant<T: FieldElement>(
     None
 }
 
+/// Identifies witness columns that are constrained to a (multi)linear expression.
```
chriseth, line 889: Can you specify the semantics a bit more concretely? What does it return?

pilopt/src/lib.rs
```diff
@@ -686,6 +886,67 @@ fn constrained_to_constant<T: FieldElement>(
...
+        op: AlgebraicBinaryOperator::Sub,
+        right,
+    }) = identity
+    {
+        (left, right)
+    } else {
+        return None;
+    };
+
+    // We require `left` to be a single, non-shifted, witness column `w`
```
chriseth, line 906: Should we also try with left and right swapped?

pilopt/src/lib.rs
Schaeff: done

pilopt/src/lib.rs
Schaeff: done

pilopt/src/lib.rs
Schaeff: Done

ast/src/analyzed/mod.rs
Schaeff: done

pilopt/src/lib.rs
```diff
@@ -891,6 +1118,65 @@ fn equal_constrained<T: FieldElement>(
     }
 }
 
+/// Tries to extract a boolean constrained witness column from a polynomial identity.
+/// The pattern used is `x * (1 - x) = 0` or `(1 - x) * x = 0` where `x` is a witness column.
```
ShuangWu121, line 1122: The pattern for boolean constraints is strictly limited to  x * (1-x) or (1 - x) * x, There will never be expressions like x * (x - 1) or (x - 1) * x, This is guaranteed by the way PIL generates constraints. Is my understanding correct?

Schaeff: Correct! It aligns with how witgen detects boolean constraints.

backend/src/stwo/circuit_builder.rs
```diff
@@ -310,21 +312,55 @@ impl FrameworkEval for PowdrEval {
...
-            name,
-            poly_id,
-            next,
-        }) = e
-        {
-            if matches!(poly_id.ptype, PolynomialType::Constant) && *next {
-                // add the name of the constant polynomial to the list
-                constant_with_next_list.insert(name);
+/// This function creates a list of the names of the constant polynomials that have next references
+/// Note that the anaylsis should also dereference next references to intermediate polynomials
```
chriseth, line 316:
```suggestion
/// Note that the analysis should also dereference next references to intermediate polynomials
```

pilopt/src/lib.rs
```diff
@@ -558,6 +563,129 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
     }
 }
 
+/// Identifies witness columns that are constrained to a non-shifted (multi)linear expression, replaces the witness column by an intermediate polynomial.
```
chriseth, line 566:
```suggestion
/// Identifies witness columns that are constrained to a non-shifted (multi)linear expression, replaces the witness column by an intermediate polynomial defined to be that expression.
```
(obvious, but this is what I was missing to understand it :) )

chriseth: Ah and maybe also use `intermediate column` instead of `polynomial`?

pilopt/src/lib.rs
```diff
@@ -558,6 +563,129 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+            try_to_linearly_constrained(identity, &intermediate_definitions).map(|e| (index, e))
+        })
+        .filter(|(_, ((_, poly_id), _))| !keep.contains(poly_id))
+        .collect::<Vec<(usize, ((String, PolyID), _))>>();
+
+    let (identities_to_remove, intermediate_polys): (BTreeSet<_>, Vec<_>) = linear_polys
+        .iter()
+        .filter_map(|(index, ((name, old_poly_id), expression))| {
+            // Remove the definition. If this fails, we have already replaced it by an intermediate column.
+            pil_file.definitions.remove(name).map(|(symbol, value)| {
```
chriseth, line 630: Can you turn this into a function inside Analyzed? I think we should avoid modifying it directly, especially when we have to deal with creating new IDs and so on.

pilopt/src/lib.rs
gzanitti: but both are possible cases, right? I don't see why we should consider `w = lin` and not do it for `lin = w`, could you elaborate on that?

pilopt/src/lib.rs
```diff
@@ -686,6 +841,82 @@ fn constrained_to_constant<T: FieldElement>(
     None
 }
 
+/// Tries to extract a witness column which is constrained to a non-shifted multi-linear expression.
+/// Inputs:
+/// - `identity`: a polynomial identity
+/// - `intermediate_columns`: the intermediate column definitions
+///
+/// Outputs:
+/// - `None` if the identity does not match the pattern
```
ShuangWu121, line 850:
```suggestion
///     - `expression` is the (multi)linear expression that the witness column is constrained by
```

pilopt/src/lib.rs
```diff
@@ -686,6 +841,82 @@ fn constrained_to_constant<T: FieldElement>(
...
+/// - `Some(((name, poly_id), expression))` if the identity matches the pattern, where:
+///     - `name` is the name of the witness column
+///     - `poly_id` is the id of the witness column
+///     - `expression` is the (multi)linear expression that the witness column is constrained to
```
ShuangWu121, line 854:
```suggestion
///     - `expression` is the (multi)linear expression that the witness column is constrained by
```

pilopt/src/lib.rs
Schaeff: I had a look into this and it's a bit tricky:

- extracting it requires going from a valid pil to a valid pil
- here we're replacing a witness column by an intermediate column with a given definition
- the result is a valid pil only after we replace all references to the original witness by references to the intermediate. This is because poly_id is *not* global but local to each polynomial type, so we cannot reuse the one from the witness :(
- the function which does that is `pilopt::substitute_polynomial_references`, which relies on `try_algebraic_expression_to_expression`, which is in `pil-analyzer`
- therefore, `substitute_polynomial_references` cannot be turned into a method on `Analyzed` without causing a circular dep

This could be simplified a lot if poly_ids were global, then we could simply replace
```
col witness a;
a = lin
```

by

```
col a = lin
```

reusing the poly_id

Another option would be:
- go through the columns to remove and rename them to some dummy names, do the same for their references (this could be a `rename` method on `Analyzed`)
- introduce the new intermediates with the correct names, which is valid since the names are now unused
- replace the references to the dummy names by references to the correct names
- remove the identities involving the dummy witnesses

I think these four steps are valid, although they involve a lot of back and forth. @chriseth do you have any thoughts on these?

pilopt/src/lib.rs
Schaeff: I checked:

> Quick tip:
> Use "by" when you're pointing to the reason.
> Use "to" when you're pointing to the range or boundary.

So I think "to" is correct

pilopt/src/lib.rs
chriseth: At least inserting a new intermediate polynomial could be a function of Analyzed. Then we don't have to deal with allocating a new ID here.

pilopt/src/lib.rs
Schaeff: Yes good point, this PR now supports `w = lin`

pilopt/src/lib.rs
Schaeff: Yes exactly, this would work if the names are different, that was my second suggestion. If we don't use dummy names, it's hard to make it a small atomic transformation.

Schaeff:
@chriseth can we merge this?

pilopt/src/lib.rs
```diff
@@ -554,6 +562,128 @@ fn extract_constant_lookups<T: FieldElement>(pil_file: &mut Analyzed<T>) {
...
+        .filter_map(|(index, ((name, old_poly_id), expression))| {
+            // Remove the definition. If this fails, we have already replaced it by an intermediate column.
+            pil_file.definitions.remove(name).map(|(symbol, value)| {
+                // sanity checks that we are looking at a single witness column
+                assert!(symbol.kind == SymbolKind::Poly(PolynomialType::Committed));
+                assert!(value.is_none());
+                assert!(symbol.length.is_none());
+
+                // we introduce a new intermediate symbol, so we need a new unique id
+                let id = pil_file
```
chriseth, line 635: This at least could be a function in Analyzed: return a new ID of a certain kind.

pilopt/src/lib.rs
```diff
@@ -586,25 +717,28 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
                 false => {
-                    let ((name, poly_id), definition) = symbols_and_definitions.next().unwrap();
-                    match definition {
-                        AlgebraicExpression::Number(value) => {
+                    let ((name, poly_id), value) = symbols_and_definitions.next().unwrap();
+                    match value {
+                        AlgebraicExpression::BinaryOperation(_) | AlgebraicExpression::UnaryOperation(_) => {
+                            None
+                        }
+                        _ =>{
```
chriseth, line 736: Could you list all the alternatives here? If we ever introduce a new operation (not likely, but still...) we won't find it here.

pilopt/src/lib.rs
```diff
@@ -614,11 +748,32 @@ fn remove_constant_intermediate_columns<T: FieldElement>(pil_file: &mut Analyzed
...
 }
 
-/// Substitutes all references to certain polynomials by the given field elements.
+/// Substitutes all references to certain polynomials by the given non-shifted expressions.
 fn substitute_polynomial_references<T: FieldElement>(
     pil_file: &mut Analyzed<T>,
-    substitutions: Vec<((String, PolyID), BigUint)>,
+    substitutions: Vec<((String, PolyID), AlgebraicExpression<T>)>,
 ) {
+    let poly_id_to_name = pil_file
```
chriseth, line 756: Can you re-use `build_poly_id_to_definition_name_lookup`? Also should we move that to `Analyzed` next to `name_to_poly_id`?

pilopt/src/lib.rs
gzanitti: Done.

pilopt/src/lib.rs
gzanitti: Done.

pilopt/src/lib.rs
gzanitti: Moved.

ast/src/analyzed/mod.rs
```diff
@@ -107,6 +107,34 @@ impl<T> Analyzed<T> {
             + 1
     }
 
+    pub fn next_id_for_kind(&self, poly_type: PolynomialType) -> u64 {
+        match poly_type {
+            PolynomialType::Committed => {
+                self.committed_polys_in_source_order()
+                    .map(|(symbol, _)| symbol.id)
```
chriseth, line 114: should also use array_elements, shouldn't it?

ast/src/analyzed/mod.rs
```diff
@@ -107,6 +107,34 @@ impl<T> Analyzed<T> {
...
+                    .max()
+                    .unwrap_or(0)
+                    + 1
+            }
+            PolynomialType::Constant => {
+                self.constant_polys_in_source_order()
+                    .map(|(symbol, _)| symbol.id)
+                    .max()
```
chriseth, line 122: Can the `.max().unwrap_or(0) + 1` be shared across the branches, or at least part of it?

ast/src/analyzed/mod.rs
```diff
@@ -146,6 +174,32 @@ impl<T> Analyzed<T> {
             .collect()
     }
 
+    /// Builds a lookup-table that can be used to turn all poly ids into the names of the symbols that define them.
+    /// For array elements, this contains the array name and the index of the element in the array.
+    pub fn build_poly_id_to_definition_name_lookup(
+        &self,
+    ) -> BTreeMap<(PolynomialType, u64), (&String, Option<usize>)> {
```
chriseth, line 181: Why don't you return PolyID directly?

ast/src/analyzed/mod.rs
gzanitti: Done

ast/src/analyzed/mod.rs
gzanitti: True, fixed.

ast/src/analyzed/mod.rs
```diff
@@ -107,6 +107,34 @@ impl<T> Analyzed<T> {
             + 1
     }
 
+    pub fn next_id_for_kind(&self, poly_type: PolynomialType) -> u64 {
```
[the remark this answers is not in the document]
gzanitti: Done.

chriseth (approved):

PR #2429, opened 2025-02-03
Single row binary executor

qwang98:
Fix required by #2389. Can only be merged after #2389.


leonardoalt:
Nice! I'll merge it into my branch so we don't need to force merge a failing PR into main.

PR #2430, opened 2025-02-03
Combine util functions for jit.

chriseth:
The JIT compiler used for computing fixed column and the new code for witgen used two different "util" code snippets. Since the field implementation for the new witgen code is more advanced, it is now moved to the generic jit compiler and used by both. It also adds some types that will later be used by prover functions (PilVec, Callable, etc). Those types were present already in the fixed column jit code.

### DIFF
--- a/jit-compiler/src/util_code.rs
+++ b/jit-compiler/src/util_code.rs
@@ -18,15 +18,10 @@ pub fn util_code<T: FieldElement>() -> Result<String, String> {
             include_str!("includes/field_goldilocks.rs").to_string()
         }
         _ => {
-            let int_type = if mem::size_of::<T>() == 8 {
-                "u64"
+            let (int_type, double_int_type) = if mem::size_of::<T>() == 8 {
+                ("u64", "u128")
             } else {
-                "u32"
-            };
-            let double_int_type = if mem::size_of::<T>() == 8 {
-                "u128"
-            } else {
-                "u64"
+                ("u32", "u64")
             };
             let modulus = T::modulus();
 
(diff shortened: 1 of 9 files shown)


jit-compiler/src/util_code.rs
```diff
@@ -0,0 +1,54 @@
...
+use powdr_number::{FieldElement, KnownField};
+
+/// Returns the rust code containing utility functions for PIL types.
+pub fn util_code<T: FieldElement>() -> Result<String, String> {
+    if !(T::has_direct_repr() && (mem::size_of::<T>() == 8 || mem::size_of::<T>() == 4)) {
+        return Err(format!(
+            "Field {}not supported",
+            T::known_field()
+                .map(|f| format!("{f} "))
+                .unwrap_or_default()
```
leonardoalt, line 12:
```suggestion
            "Field {} not supported",
            T::known_field()
                .map(|f| format!("{f}"))
                .unwrap_or_default()
```

jit-compiler/src/util_code.rs
```diff
@@ -0,0 +1,54 @@
...
+            let int_type = if mem::size_of::<T>() == 8 {
+                "u64"
+            } else {
+                "u32"
+            };
+            let double_int_type = if mem::size_of::<T>() == 8 {
+                "u128"
+            } else {
+                "u64"
+            };
```
leonardoalt, line 30:
```suggestion
            let (int_type, double_int_type) = if mem::size_of::<T>() == 8 {
                ("u64", "u128")
            } else {
                ("u32", "u64")
            };
```

leonardoalt (commented):
small comments

jit-compiler/src/util_code.rs
chriseth: The idea is that if the field is not a known field, the error message says `"Field not supported"`

leonardoalt: ah ok

PR #2434, opened 2025-02-03
Compute multiple cell values conditionally from other cell values.

chriseth:
I think this is the last part for prover functions, which is needed for the arithmetic machine:

Compute the values of multiple cells at the same time from other known cells, but only if a condition is true. The biggest challenge here is that we need to be able to evaluate the condition at compile time.


pacheco (commented):
just for context, is there some specific purpose for the refactor into the new `compute_*` query functions?

chriseth:
The idea  behind this refactoring is that we can detect calls to these new functions and statically determine when these prover functions will succeed. We also do not need to implement std::prover::eval for that, because it is part of the utility function from prover.asm whose semantics witgen will know (and assume, but I think that's ok because it's part of the stdlib).

pacheco (approved):

PR #2436, opened 2025-02-04
machine option to disable pc update rule

pacheco:
with continuations, we need to be able to install a custom pc update rule that is able to jump to the shutdown routine


leonardoalt:
Closing for now.

PR #2438, opened 2025-02-04
Stwo publics

ShuangWu121:
# Extend `stwo` to Support Public Input

## Description
This PR extends `stwo` to support public input.

While `stwo` does not natively implement the public input feature, it is feasible to achieve this by implementing it in the `stwo` backend.

## Implementation Details
The public values are formed using a selector to pick specific values from a witness column, which are then made public to the verifier. 

This logic is already implemented in the `Plonky3` backend, where public values are passed directly as inputs to the `prove` and `verify` functions, simplifying the process.

## ✅ To-Do List
To achieve similar functionality in `stwo`, the following steps need to be implemented:

- [x] Add selector columns in setup.
- [x] Add constraints linking the public columns, the selector, and the corresponding witness columns.
- [x] pass public values to verification.


backend/src/stwo/prover.rs
```diff
@@ -291,6 +318,30 @@ where
             })
             .collect::<BTreeMap<_, _>>();
 
+        // get publics of stage0
+        // TODO:when publics are supplied, the order of the witness might cause problem
```
Schaeff, line 322: Are these leftovers?

backend/src/stwo/prover.rs
```diff
@@ -338,22 +389,47 @@ where
 
         if self.analyzed.stage_count() > 1 {
             // Build witness columns for stage 1 using the callback function, with the generated challenges
-            let stage1_witness_cols_circle_domain_eval = witness_by_machine
-                .into_iter()
+
+            let stage0_witness_names_stage1_witness_cols = witness_by_machine
```
Schaeff, line 393: The variable name is confusing here, is it correct?

backend/src/stwo/prover.rs
```diff
@@ -430,12 +508,15 @@ where
...
-            "Expected _instances slice to be empty, but it has {} elements.",
-            _instances.len()
-        );
+    pub fn verify(&self, proof: &[u8], instances: &[M31]) -> Result<(), String> {
+        // get public values
+        let public_values: BTreeMap<String, M31> = self
+            .analyzed
+            .get_publics()
+            .iter()
+            .zip_eq(instances.iter())
```
Schaeff, line 517: This panics if the passed instances have the wrong size, instead we should error out cleanly before this iteration.

backend/src/stwo/prover.rs
ShuangWu121: Oops, I missed that. I've cleaned them up now.

backend/src/stwo/prover.rs
ShuangWu121: It is indeed confusing. 
It contains only the witness column **names** from stage 0 and the witness columns from stage 1. They are used later by a function that transfers columns to the circle domain, filtering out those already transferred in stage 0. Additionally, they are used in public value generation in stage 1. I've now separated them for better clarity.

ShuangWu121:
Adding more tests before making it ready

pipeline/tests/pil.rs
```diff
@@ -277,12 +277,33 @@ fn stwo_fixed_columns() {
...
 fn stwo_stage1_publics() {
     let f = "pil/stage1_publics.pil";
-    // witness col y[7]=1191445910
-    // witness col x[7]=8
-    test_stwo_stage1_public(
-        f,
-        Default::default(),
-        vec![Mersenne31Field::from(1191445910), Mersenne31Field::from(8)],
+
+    let first_result = std::panic::catch_unwind(|| {
```
Schaeff, line 281: Can we instead have two tests and use `should_panic` on the one which has the wrong public inputs?

ShuangWu121: Done

Schaeff (dismissed):

Schaeff (approved):

PR #2440, opened 2025-02-04
Specialize functions for operation id.

chriseth:
If one item in the RHS of the connection has `operation_id` in its name and has a known concrete value, create a specialized function for this value.

### DIFF
--- a/executor/src/witgen/jit/function_cache.rs
+++ b/executor/src/witgen/jit/function_cache.rs
@@ -26,8 +26,11 @@ use super::{
 };
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
-struct CacheKey {
+struct CacheKey<T: FieldElement> {
     identity_id: u64,
+    /// If `Some((index, value))`, then this function is used only if the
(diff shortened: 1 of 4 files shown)


executor/src/witgen/jit/function_cache.rs
```diff
@@ -108,21 +114,30 @@ impl<'a, T: FieldElement> FunctionCache<'a, T> {
...
         log::debug!(
-            "Compiling JIT function for\n  Machine: {}\n  Connection: {}\n   Inputs: {:?}",
+            "Compiling JIT function for\n  Machine: {}\n  Connection: {}\n   Inputs: {:?}{}",
             self.machine_name,
             self.parts.connections[&cache_key.identity_id],
-            cache_key.known_args
+            cache_key.known_args,
+            cache_key
+                .known_concrete
+                .map(|(i, v)| format!("\n input {i} = {v}"))
```
chriseth, line 126:
```suggestion
                .map(|(i, v)| format!("\n   Input {i} = {v}"))
```

leonardoalt (approved):

PR #2442, opened 2025-02-05
Expect a machine to be responsible.

chriseth:
I think this was a leftover, at least I don't see how the call can fail, except for recursive machine dependencies.


georgwiese (approved):
LGTM!

PR #2444, opened 2025-02-05
Test

chriseth:


chriseth:
This still has the problem that at the end, this identity is not complete:
```
--------------[ identity 96 on row 30: ]--------------
((main_arith::x1[15]' - main_arith::x1[15] ) * (1 - main_arith::CLK32[31])) = 0
        <known>                 ???                           0             =  
  [0, 65535] & 0xffff   [0, 65535] & 0xffff                                 =  
(main_arith::x1[15]' - main_arith::x1[15] ) = 0
       <known>                 ???          =  
 [0, 65535] & 0xffff   [0, 65535] & 0xffff  =  
```

Note that it can easily be processed - so this hints at a problem with the identity queue.

chriseth:
The problem above was fixed by pulling out the assignments into the processor.

executor/src/witgen/machines/mod.rs
```diff
@@ -54,24 +54,26 @@ pub trait Machine<'a, T: FieldElement>: Send + Sync {
...
         &mut self,
         _can_process: impl CanProcessCall<T>,
         _identity_id: u64,
         _known_arguments: &BitVec,
-        _range_constraints: &[RangeConstraint<T>],
-    ) -> Option<Vec<RangeConstraint<T>>> {
-        None
+        range_constraints: &[RangeConstraint<T>],
+    ) -> (bool, Vec<RangeConstraint<T>>) {
+        (false, range_constraints.to_vec())
```
pacheco, line 76: it looks a bit weird that every place where (false, constraints) is returned, the constraints argument needs to be `to_vec()` and returned... or is there some case where something different might be returned? Otherwise, maybe this should be handled at the calling site?

chriseth: Yes, the fixed lookup actually returns different constraints.

chriseth: Here: https://github.com/powdr-labs/powdr/pull/2444/files#diff-20b382dd5fa786b248177294f7c7497094f1ce43dbc78af3a2328024a6066e93R379

pacheco: hmm i see... maybe `can_process_call_fully` could take a `Vec` directly and just return it then

chriseth: good idea!

pacheco (approved):

PR #2454, opened 2025-02-06
Move assignments to processor

chriseth:
This moves the handling of assignments from witgen_inference to the processor and also uses the queue to schedule assignments.


executor/src/witgen/jit/witgen_inference.rs
```diff
@@ -677,6 +624,8 @@ fn is_known_zero<T: FieldElement>(x: &Option<AffineSymbolicExpression<T, Variabl
         .unwrap_or(false)
 }
 
+// TODO  we could split theassignmnet into variable and constant now.
```
chriseth, line 627: done in another pr

leonardoalt (approved):

PR #2455, opened 2025-02-07
Bus multi interaction (arbitrary number of columns)

qwang98:
WIP. Broken up from #2449 so that this works end to end for arbitrary number of columns (both even and odd).

Instead of using and `Option` for helper columns, which currently bugs out, this version uses an array and forces the case of odd number of bus interactions to materialize a `helper_last =  multiplicity_last / payloads_last`. This has the advantage of passing tests end to end, so we can debug in another PR.

Another major change is that instead of removing the single bus interaction API entirely, I kept it as a user friendly choice (so that the user can just provide raw inputs in the case of a single bus interaction, instead of providing a vector of tuple of inputs). The single version still calls the multi version (and falls to the "odd" case) to ensure maximum code reuse and best optimization.

Fails a CI test at proving stage for `block_to_block_with_bus_composite()`. I think this is likely caused by me changing the single bus API (the `static_bus.asm` test might be too simple to fully test the new single bus API compared to this failing test). Another hypothesis is that some element from lookup/permutation using the new bus API might be bugging out, so I'm a bit stuck here. @georgwiese


ast/src/analyzed/display.rs
```diff
@@ -431,7 +431,7 @@ impl<T: Display> Display for PhantomBusInteractionIdentity<T> {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result {
         write!(
             f,
-            "Constr::PhantomBusInteraction({}, {}, [{}], {}, [{}], [{}]);",
+            "Constr::PhantomBusInteraction({}, {}, [{}], {}, [{}], [{}], [{}]);",
```
qwang98, line 434: At this point, helper columns are NOT `Option` yet, but always array.

PR #2458, opened 2025-02-07
Bus multi interaction optional helper

qwang98:
Depends on #2457. Bus logic, bus interaction type, display, and witgen changes to support helper column being an Option::None in case of odd number of bus interactions or Option::Some in case of even number of bus interactions.

### DIFF
--- a/executor/src/witgen/bus_accumulator/mod.rs
+++ b/executor/src/witgen/bus_accumulator/mod.rs
@@ -215,11 +215,16 @@ impl<'a, T: FieldElement, Ext: ExtensionField<T> + Sync> BusAccumulatorGenerator
                 .collect::<Vec<_>>();
             let folded = self.beta - self.fingerprint(&tuple);
 
-            let helper = folded.inverse() * multiplicity;
+            let to_add = folded.inverse() * multiplicity;
+
+            let helper = match bus_interaction.helper_columns {
+                Some(_) => to_add,
+                None => Ext::zero(),
+            };
 
             let new_acc = match multiplicity.is_zero() {
                 true => current_acc,
-                false => current_acc + helper,
+                false => current_acc + to_add,
             };
 
             folded_list[i] = folded;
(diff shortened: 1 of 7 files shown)


executor/src/witgen/bus_accumulator/mod.rs
```diff
@@ -299,10 +304,16 @@ fn collect_acc_columns<T>(
 fn collect_helper_columns<T>(
     bus_interaction: &PhantomBusInteractionIdentity<T>,
     helper: Vec<Vec<T>>,
-) -> impl Iterator<Item = (PolyID, Vec<T>)> + '_ {
-    bus_interaction
-        .helper_columns
-        .iter()
-        .zip_eq(helper)
-        .map(|(column_reference, column)| (column_reference.poly_id, column))
+) -> impl Iterator<Item = (PolyID, Vec<T>)> {
```
Schaeff, line 307: The implementation can be simplified since option inplements IntoIter, no need to go through Vec

qwang98: Note that here we have `Option<Vec<T>>`, not `Vec<Option<T>>`.

This is actually an intentional workaround. If we don't collect results from the `Some` case into a `Vec`, the type it returns can't unify with the type we return from the `None` case, whether it be an empty `Vec::new().into_iter()` or an empty `std::iter::empty()`. This is because Rust not only requires that either case return a type that implements `Iterator<Item = (PolyID, Vec<T>)>`, but also that both return the same CONCRETE type. Neither empty vector nor `std::iter::empty` is the same concrete type as what return from the `Some` case. Another workaround is to use `Either`, but I think it involves more overhead than simply collecting it into a `Vec` and then convert to an iterator. Let me know if this doesn't make sense.

Schaeff: Oh right I was wrong here !

PR #2461, opened 2025-02-07
Prover functions in queue

chriseth:
This change also adds prover functions as a third kind of "queue item". The biggest change, though, is a refactoring concerning the code that computes which variables are contained in an identity. The whole mechanism is changed to handle queue items uniformly.


executor/src/witgen/jit/identity_queue.rs
```diff
@@ -102,151 +130,145 @@ impl<T: FieldElement> PartialEq for QueueItem<'_, T> {
...
-                            |index| {
+impl<'a, T: FieldElement> ReferencesComputer<'a, T> {
+    pub fn new(fixed_data: &'a FixedData<'a, T>) -> Self {
+        Self {
+            fixed_data,
+            intermediate_cache: HashMap::new(),
+            references_per_identity: HashMap::new(),
+        }
+    }
+    pub fn references(&mut self, item: &QueueItem<'a, T>) -> Vec<Variable> {
```
pacheco, line 151: maybe this can return an iterator, usage seems to immediately call `iter()`.

chriseth: I tried to make this return an iterator, but I got some errors wrt either references to `&mut self` or temporaries in lambda functions.

pacheco (approved):

PR #2464, opened 2025-02-07
Support sums in div witgen

Schaeff:
Division witgen expects a pattern like `dividend = divisor * quotient + remainder` (where the two terms on the rhs can also be swapped).
This works when `quotient` and `remainder` are witness columns, but fails when they are intermediate columns.

Extend division witgen to support a pattern of the form:
```
dividend = sum_n(divisor * (limb_size ** i) * quotient_i) + sum_n((limb_size ** i) * remainder_i)
```

This enables in particular solving:
```
12 = 3 * SimpleDiv::X_b1 + 768 * SimpleDiv::X_b2 + SimpleDiv::R_b1 + 256 * SimpleDiv::R_b2
```

### DIFF
--- a/executor/src/witgen/affine_expression.rs
+++ b/executor/src/witgen/affine_expression.rs
@@ -214,13 +214,27 @@ where
         let negated = -self.clone();
 
         // Try to find a division-with-remainder pattern and solve it.
-        if let Some(result) = self.try_solve_division(known_constraints).transpose()? {
+        println!("try_solve_division {self}");
+        if let Some(result) = self
+            .try_solve_division(known_constraints)
+            .map_err(|e| println!("{e}"))
+            .ok()
+            .transpose()?
+        {
             if !result.is_empty() {
+                for (k, constraint) in &result.constraints {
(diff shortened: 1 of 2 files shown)


executor/src/witgen/affine_expression.rs
```diff
@@ -273,59 +287,147 @@ where
...
         }
 
-        let quotient_value =
-            (dividend.to_arbitrary_integer() / divisor.to_arbitrary_integer()).into();
-        let remainder_value =
-            (dividend.to_arbitrary_integer() % divisor.to_arbitrary_integer()).into();
-        Some(
+        let quotient_value = dividend.to_arbitrary_integer() / divisor.to_arbitrary_integer();
+        println!("Quotient: {quotient_value}");
+        let limb_size_bits = limb_size_bits.ok_or("No limb size")?;
```
Schaeff, line 384: The original case where there's a single limb is not implemented yet

Schaeff:
Closing for now as we can instead amend the div test in #2425

PR #2468, opened 2025-02-11
Bus multi interaction linker

qwang98:
Ready for review. Depends on #2459, which depends on #2457.

1. Added `lookup_multi_receive` and `permutation_multi_receive` APIs for the linker and test files to use.
2. Modified linker to use the multi version by default. This in practice reduces the number of bus interactions to 2 or 3 (depending if we only have permutation, only have lookup, or have both). This can be furthur optimized by directly using `bus_multi_receive` function in the linker, but requires more processing of lower level expressions, which I'd leave to another PR. In practice, this would reduce the number of bus interactions from 2 or 3 to always 2. This is currently being implemented in #2469.
3. Added test cases to the pipeline.

PR #2469, opened 2025-02-11
Bus multi interaction batch lookup and permutation receives

qwang98:
Ready for review. Depends on #2468. Implements bullet point number 2 mentioned in #2468.

The biggest challenge would be adding the multiplicity columns and an additional constraint `(1 - selector) * multiplicity = 0` for the lookup case, and this can be achieve by two methods:
1. Adding the array of column and array of constraints at the linker level.
2. Create another PIL function that batch permutation receives and lookup receives to the same bus multi interaction.

I'm opting for method 2 because method 1 likely requires lots of processing of numerous lower level ast `Expression` types, which would likely look a lot messier than method 2.


std/protocols/bus.asm
```diff
@@ -299,6 +299,28 @@ let bus_multi_receive: (expr, expr[], expr, expr)[] -> () = constr |bus_inputs|
     bus_multi_interaction(ids, payloads, negated_multiplicities, latches);
 };
 
+/// Exposed to the linker and not intended for end user because it requires expert knowledge of Powdr's bus protocol.
+/// Inputs in the order of: id, selector, payload, is_permutation for both lookup and permutation.
+let bus_multi_receive_batch_lookup_permutation: (expr, expr, expr[], int)[] -> () = constr |inputs| {
```
Schaeff, line 304: The last argument feels a bit magic, maybe we could use a boolean or an enum instead?

linker/src/bus.rs
```diff
@@ -364,7 +354,7 @@ namespace main__rom(4);
     pol constant p_instr_return = [0]*;
     pol constant operation_id = [0]*;
     pol constant latch = [1]*;
-    std::protocols::lookup_via_bus::lookup_multi_receive([(12064, main__rom::latch, [main__rom::operation_id, main__rom::p_line, main__rom::p_instr__jump_to_operation, main__rom::p_instr__reset, main__rom::p_instr__loop, main__rom::p_instr_return])]);
+    std::protocols::bus::bus_multi_receive_batch_lookup_permutation([(12064, main__rom::latch, [main__rom::operation_id, main__rom::p_line, main__rom::p_instr__jump_to_operation, main__rom::p_instr__reset, main__rom::p_instr__loop, main__rom::p_instr_return], 0)]);
```
Schaeff, line 357: This example is a bit simplistic, do we have a test somewhere which uses a mix of incoming permutations and lookups?

Schaeff (approved):
LGTM but two suggestions

std/protocols/bus.asm
qwang98: Exactly! I used a boolean before but because this function is used by the linker and I couldn't figure out which enum in asm::Expression this corresponds to (there's Number enum but no bool related enum), so I used a number instead. Merging for now but will fix in the next PR.

linker/src/bus.rs
qwang98: Agree. Will add in the next PR :)

PR #2472, opened 2025-02-11
Full FE in memory word

lvella:
This helps making successive Poseidon2GL calls cheaper (e.g. in merkle tree and sponge).

My plan for follow-up PRs:
- use Poseidon2GL in the bootloader, hopefully making its trace smaller;
- remove all variants of PoseidonGL, as it would have been superseded;
- make another patch to the plonky3 verifier, so it benefits from this change (instead of suffering a regression).

### DIFF
--- a/riscv-executor/src/lib.rs
+++ b/riscv-executor/src/lib.rs
@@ -3308,3 +3308,7 @@ pub fn write_executor_csv<F: FieldElement, P: AsRef<Path>>(
 fn is_multiple_of_4(n: u32) -> bool {
     n % 4 == 0
 }
+
+pub fn hash_map_to_memory_state<F: FieldElement>(map: HashMap<u32, u32>) -> MemoryState<F> {
+    map.into_iter().map(|(k, v)| (k, v.into())).collect()
+}
--- a/riscv-executor/src/lib.rs
+++ b/riscv-executor/src/lib.rs
@@ -2672,20 +2672,21 @@ impl<F: FieldElement> Executor<'_, '_, F> {
                     .collect::<Vec<_>>();
 
                 result.into_iter().enumerate().for_each(|(i, v)| {
-                    self.proc.set_mem(output_ptr + i as u32 * 4, v, 0, 0);
+                    self.proc.set_mem(output_ptr + i as u32 * 4, v.into(), 1, 0);
                 });
 
                 None
             }
             Instruction::merge_gl => {
-                let lo_and_output_reg = args[0].u();
-                let lo = self.proc.get_reg_mem(lo_and_output_reg).u();
-                let hi = self.proc.get_reg_mem(args[1].u()).u();
+                let lo = self.proc.get_reg_mem(args[0].u()).into_fe();
+                let hi = self.proc.get_reg_mem(args[1].u()).into_fe();
 
-                // Write the result to the original lo register.
-                let result = F::from((hi as u64) << 32 | lo as u64);
-                self.proc
-                    .set_reg_mem(lo_and_output_reg, Elem::Field(result));
+                let output_ptr = self.proc.get_reg_mem(args[2].u()).u();
+                assert!(is_multiple_of_4(output_ptr));
+
+                // Write the result to memory
+                let result = hi * F::from(0x1_0000_0000_u64) + lo;
+                self.proc.set_mem(output_ptr, Elem::Field(result), 2, 0);
 
                 None
             }
--- a/riscv-executor/src/lib.rs
+++ b/riscv-executor/src/lib.rs
@@ -135,7 +135,9 @@ instructions! {
     ec_double,
     commit_public,
     fail,
-    keccakf
+    keccakf,
+    split_gl_vec,
+    merge_gl
 }
 
 /// Enum with columns directly accessed by the executor (as to avoid matching on strings)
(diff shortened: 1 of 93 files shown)


std/machines/split/split_gl_vec.asm
```diff
@@ -0,0 +1,78 @@
+use std::array;
+use std::machines::large_field::memory::Memory;
+use super::split_gl::SplitGL;
+
+machine SplitGLVec8(mem: Memory, split_gl: SplitGL) with
```
lvella, line 5: This is a question for someone with a better intuition on execution cost (@georgwiese): if instead of a fixed 8 element machine like this, would it be better if it took an N argument, and executed one split per row?

leonardoalt: I think it should be about the same

test_data/std/split_gl_vec_test.asm
```diff
@@ -0,0 +1,81 @@
...
+
+        assert_eq 240, 0x00000004;
+        assert_eq 244, 0x00000003;
+
+        assert_eq 248, 0x00000006;
+        assert_eq 252, 0x00000005;
+
+        assert_eq 256, 0x00000008;
+        assert_eq 260, 0x00000007;
+
```
leonardoalt, line 78: Can you also add a repeated test case but where input_addr = output_addr to verify that the memory addresses and timestamps are working as intended

riscv/src/lib.rs
```diff
@@ -56,6 +58,13 @@ impl RuntimeLibs {
             ..self
         }
     }
+
+    pub fn with_splt_vec(self) -> Self {
```
leonardoalt, line 62: I think we can add this always to the main machine

riscv/src/large_field/runtime.rs
```diff
@@ -311,6 +309,60 @@ impl Runtime {
         self
     }
 
+    fn with_split_vec(mut self) -> Self {
```
leonardoalt, line 312: I think we can add it to the main machine instead of an extra configurable one

riscv-runtime/src/goldilocks.rs
```diff
@@ -1,13 +1,61 @@
-use core::arch::asm;
+use core::{arch::asm, mem::MaybeUninit};
 use powdr_riscv_syscalls::Syscall;
 
 pub const PRIME: u64 = 0xffffffff00000001;
 
+/// Goldilocks field element.
+///
+/// Not a legal value in RISC-V, as it uses a supposedly
+/// 32-bit memory word to store a full field element.
```
leonardoalt, line 9: Shouldn't we use usize then?

riscv-executor/src/lib.rs
```diff
@@ -1560,7 +1568,7 @@ impl<F: FieldElement> Executor<'_, '_, F> {
 
                 let addr = addr as u32;
                 let lid = self.instr_link_id(instr, MachineInstance::memory, 0);
-                self.proc.set_mem(addr, value.u(), self.step + 3, lid);
```
leonardoalt, line 1563: Do we have any `mload/mstore` used directly on FE values? I would assume not, in which case it would be good to assert here that value is u32

leonardoalt (commented):
Nice!

Does the bootloader not change at all? Or is that next PR?

riscv-runtime/src/goldilocks.rs
lvella: Probably not. The rust type can be anything that has the size and alignment of our implementation of one memory word. `usize` is identical, in those regards, to `u32`.

But suppose we are implementing RISCV-64: would we still use our 32-bit memory word implementation? Probably yes, in the Goldilocks field. So the correct type would still be `u32`, but not `usize`.

So, `u32` seems better.

riscv-executor/src/lib.rs
lvella: As it stands now, in the `riscv-runtime`, the compiler is free to move around and even `memcpy()` `OpaqueGoldilocks` values.

lvella:
> Does the bootloader not change at all? Or is that next PR?

Next PR.

riscv-runtime/src/goldilocks.rs
leonardoalt: Ok.
But regarding RV64, why would we implement it as 32-bit memory words?

riscv-executor/src/lib.rs
leonardoalt: right

riscv-runtime/src/goldilocks.rs
lvella: It is the largest power of two that fits Goldilocks. These memory machine words are more a restriction of the field than the implemented VM, right?

leonardoalt: Well if we implemented RV64 with GL we'd use two limbs, like we do with BB for RV32

lvella: Why BB was done like that instead of using words of 16 bits? To avoid too many changes in the RV32 machine?

test_data/std/split_gl_vec_test.asm
lvella: Ok.

riscv/src/large_field/runtime.rs
lvella: I have placed it together with poseidon2, since you can not use one without the other.

riscv/src/lib.rs
lvella: Is enabled with poseidon2 now.

riscv-runtime/src/goldilocks.rs
leonardoalt: Not sure I get the question, the BB RISCV memory machine was written exactly like that: 2 FE limbs of 16 bits

leonardoalt (approved):

PR #2465, opened 2025-02-10
Bus Interaction Enum

qwang98:
Ready for review. Solves #2460. Depends on #2469. Prior bug fixed (due to an error in the `bus.asm` code update when unwrapping the new `BusInteraction` enum type.

PR #2474, opened 2025-02-12
30-bit bus interaction ID

qwang98:
Solves #2470. Depends on #2465. Approval will only works after merging #2465.

The following is a response to #2470 and motivation of this PR: 

Chance of collission:
For 24-bits:
100 interactions: 0.000295000053831429
1000 interactions: 0.02933425835911685
10000 interactions: 0.9492338975723106

For 30-bits (theoretically the largest we could get to without changing pil code due to field prime of bb and m31):
100 interactions: 4.610036260510597e-06
1000 interactions: 0.0004650875835883195
10000 interactions: 0.04549425469529611

Neither look ideal, but I think after https://github.com/powdr-labs/powdr/pull/2469, bus linker mode should always have 2 bus interactions only. For user defined bus interactions, they can intentionally input IDs that don't clash, so I think just using 30-bit interaction id should be good enough for 29 bits of security (under 2 interactions linker case).


linker/src/bus.rs
```diff
@@ -354,7 +353,7 @@ namespace main__rom(4);
     pol constant p_instr_return = [0]*;
     pol constant operation_id = [0]*;
     pol constant latch = [1]*;
-    std::protocols::bus::bus_multi_receive_batch_lookup_permutation([(12064, main__rom::latch, [main__rom::operation_id, main__rom::p_line, main__rom::p_instr__jump_to_operation, main__rom::p_instr__reset, main__rom::p_instr__loop, main__rom::p_instr_return], 0)]);
+    std::protocols::bus::bus_multi_receive_batch_lookup_permutation([(742731552, main__rom::latch, [main__rom::operation_id, main__rom::p_line, main__rom::p_instr__jump_to_operation, main__rom::p_instr__reset, main__rom::p_instr__loop, main__rom::p_instr_return], 0)]);
```
qwang98, line 356: 30 bit number indeed.

linker/src/bus.rs
```diff
@@ -19,13 +19,17 @@ use std::{
...
-    let mut bytes = [0u8; 2];
-    bytes.copy_from_slice(&result[..2]);
-    u16::from_le_bytes(bytes)
+    let mut bytes = [0u8; 4];
+    bytes.copy_from_slice(&result[..4]);
+    // Mask the most significant two bits of the most significant limb
+    // to ensure that interaction_id is at most 30-bits, in order to
+    // fill in a single field element for BabyBear and M31.
+    bytes[3] &= 0b0011_1111;
+    u32::from_le_bytes(bytes)
```
Schaeff, line 32:
```suggestion
    u32::from_le_bytes(bytes) >> 2
```
?

Schaeff (dismissed):

linker/src/bus.rs
qwang98: Done.

Schaeff (approved):

PR #2475, opened 2025-02-12
Try zero, but only for simple send params.

chriseth:


pacheco:
could you explain the reasoning for it?

chriseth:
If a column `x` is not used (for example because it is only relevant for a different operation id), it might still have fixed lookups for range constraints: `[ x ] in [ BYTE ]`. Since the lookup is not de-selected, we need to execute it. This code performs this execution.

pacheco (approved):

PR #2478, opened 2025-02-12
Support compiling prover functions.

chriseth:


leonardoalt (approved):

chriseth:
Actually it turns out that `compute_multi` is not properly taken into account yet

chriseth:
TODO: Always log error in the witjitgen case, and only log if we are in debug for the fixed column computation instead.

leonardoalt (approved):

PR #2481, opened 2025-02-13
jit interpreter branch handling

pacheco:


executor/src/witgen/jit/interpreter.rs
```diff
@@ -203,63 +229,94 @@ impl<T: FieldElement> EffectsInterpreter<T> {
...
-                }
-                InterpreterAction::MachineCall(id, arguments) => {
-                    // we know it's safe to escape the references here, but the compiler doesn't, so we use unsafe
-                    let mut args = arguments
-                        .iter()
-                        .map(|a| match a {
-                            MachineCallArgumentIdx::Unknown(idx) => {
-                                let var = &mut vars[*idx] as *mut T;
-                                LookupCell::Output(unsafe { var.as_mut().unwrap() })
+        let mut block_stack = vec![self.actions.iter()];
```
pacheco, line 232: handling of the existing actions didn't change, its only code for handling the `Branch` action that is new

executor/src/witgen/jit/compiler.rs
```diff
@@ -29,10 +29,15 @@ use super::{
     variable::Variable,
 };
 
-pub struct WitgenFunction<T> {
+pub enum WitgenFunction<T: FieldElement> {
```
chriseth, line 32: I think it would be better to keep `compiler.rs` only concerned about compiling effects and transforming parameters to fit the loaded function.

What do you think about putting the enum in function_cache? Then we also don't need a `#[cfg(test)]`-function for the unit tests here.

executor/src/witgen/jit/function_cache.rs
```diff
@@ -194,15 +194,22 @@ impl<'a, T: FieldElement> FunctionCache<'a, T> {
...
-        log::trace!("Compiling effects...");
-        let function = compile_effects(
-            self.fixed_data.analyzed,
-            self.column_layout.clone(),
-            &known_inputs,
-            &code,
-            prover_functions,
-        )
-        .unwrap();
+        let compiled_jit = !matches!(std::env::var("POWDR_JIT_INTERPRETER"), Ok(val) if val == "1");
```
chriseth, line 197: As you know, I'm not so happy about changing the behaviour of a function by something that is not an argument to the function.

How do you use this environment variable?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -203,63 +229,94 @@ impl<T: FieldElement> EffectsInterpreter<T> {
...
-                    if *expected_equal {
-                        assert_eq!(lhs_value, rhs_value, "Assertion failed");
-                    } else {
-                        assert_ne!(lhs_value, rhs_value, "Assertion failed");
+                        };
+                        if condition {
+                            block_stack.push(if_branch.iter());
+                        } else {
+                            block_stack.push(else_branch.iter());
+                        }
```
chriseth, line 317:
```suggestion
                        block_stack.push(if condition {
                            if_branch.iter()
                        } else {
                            else_branch.iter()
                        });
```

executor/src/witgen/jit/interpreter.rs
```diff
@@ -31,6 +32,44 @@ enum InterpreterAction<T: FieldElement> {
...
+        BranchTest<T>,
+        Vec<InterpreterAction<T>>,
+        Vec<InterpreterAction<T>>,
+    ),
+}
+
+#[derive(Debug)]
+enum BranchTest<T: FieldElement> {
+    Equal { var: usize, value: T },
+    Less { var: usize, min: T, max: T },
```
chriseth, line 45: I think in this case, we could maybe call them `Inside` and `Outside`.

executor/src/witgen/jit/interpreter.rs
```diff
@@ -283,7 +367,15 @@ impl<T: FieldElement> InterpreterAction<T> {
                     set.insert(*v);
                 }
             }),
-            _ => {}
+            InterpreterAction::Branch(_branch_test, if_actions, else_actions) => {
+                if_actions.iter().chain(else_actions).for_each(|a| {
+                    set.extend(a.writes());
+                });
```
chriseth, line 373:
```suggestion
                set.extend(if_actions.iter().chain(else_actions).flat_map(|a| a.writes()));
```

executor/src/witgen/jit/interpreter.rs
```diff
@@ -486,22 +580,196 @@ mod test {
...
 
+    use pretty_assertions::assert_eq;
+    use test_log::test;
+
     use bit_vec::BitVec;
     use itertools::Itertools;
     use powdr_number::GoldilocksField;
 
+    #[test]
+    fn branching() {
```
chriseth, line 599: Very nice test!

executor/src/witgen/jit/function_cache.rs
pacheco: this seemed like the smallest change to have a way of enabling the interpreter, at least for now:
interpreter is disabled by default, and only enabled when `POWDR_JIT_INTERPRETER=1`

chriseth: You could enable it if the field is not supported by the compiler, for example. Then we test it on a lot of inputs and in the future, we can think about enabling it if the degree is small.

executor/src/witgen/jit/function_cache.rs
```diff
@@ -44,8 +45,39 @@ pub struct FunctionCache<'a, T: FieldElement> {
     parts: MachineParts<'a, T>,
 }
 
+enum WitgenFunction<T: FieldElement> {
+    // TODO We might want to pass arguments as direct function parameters
```
chriseth, line 49: Can you move this comment back to `CompiledFunction`? It mainly concerns the WitgenFunctionParameters, i.e. could also be moved there.

pacheco: right!

executor/src/witgen/jit/interpreter.rs
```diff
@@ -81,7 +81,25 @@ pub enum MachineCallArgumentIdx {
...
+        // TODO: interpreter doesn't support prover functions yet
+        fn has_prover_fn<T: FieldElement>(effect: &Effect<T, Variable>) -> bool {
+            match effect {
+                Effect::ProverFunctionCall(..) => true,
+                Effect::Branch(_, if_branch, else_branch) => {
+                    if if_branch.iter().any(has_prover_fn) || else_branch.iter().any(has_prover_fn)
+                    {
+                        return true;
+                    }
+                    false
```
chriseth, line 94:
```suggestion
                    if_branch.iter().any(has_prover_fn) || else_branch.iter().any(has_prover_fn)
```

chriseth:
Are the test failures an error in autowitjitgen?

pacheco:
could we merge this if I remove the integration? tests seemed do work fine with GL

executor/src/witgen/jit/interpreter.rs
chriseth: What about this one?

executor/src/witgen/jit/interpreter.rs
```diff
@@ -208,63 +252,94 @@ impl<T: FieldElement> EffectsInterpreter<T> {
...
+                        } else {
+                            assert_ne!(lhs_value, rhs_value, "Assertion failed");
+                        }
+                    }
+                    InterpreterAction::Branch(test, if_branch, else_branch) => {
+                        // push the current block on the stack to continue execution once the branch is done
+                        block_stack.push(iter);
+                        // test the condition
+                        let condition = match test {
+                            BranchTest::Equal { var, value } => vars[*var] == *value,
```
chriseth, line 328: This could be a function of BranchTest. That way, we could encapsulate the logic of what the enum alternatives mean inside BrachTest and only provide functions `var()` and `test(value_to_test)`

chriseth (approved):

PR #2482, opened 2025-02-13
Fix bus interaction `Children` implementation

Schaeff:
When adding new fields to `PhantomBusInteraction`, we forgot to update the `Children` implementation.
- change the definition of `PhantomBusInteraction` to use expressions, in order for these fields to be visited by the `Children` implementation. Move the expectation that these fields are simple references to witgen.
- update the `Children` implementation

cc @qwang98

### DIFF
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -1037,15 +1037,8 @@ pub struct PhantomBusInteractionIdentity<T> {
     pub payload: ExpressionList<T>,
     pub latch: AlgebraicExpression<T>,
     pub folded_expressions: ExpressionList<T>,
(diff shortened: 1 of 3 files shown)


qwang98:
Took a detailed look an am good with all edits. Just one question: how is `Children` trait implementation useful for `BusInteraction` (and really all identities)? The furthest I can *directly* track it up is here: https://github.com/powdr-labs/powdr/blob/main/ast/src/analyzed/mod.rs#L1154-L1165. Ctrl + F `children_mut()` also returns hundreds of references in the codebase, so I'm a bit lost.

qwang98 (approved):
lgtm

PR #2486, opened 2025-02-14
Bus interaction linker single accumulator

qwang98:
Depends on #2484. Implements task number 1 in issue #2485. Create one bus interaction for each object (machine).


std/protocols/bus.asm
```diff
@@ -259,23 +259,30 @@ let bus_multi: BusInteraction[] -> () = constr |bus_inputs| {
 };
 
 /// Exposed to the linker and not intended for end user because it requires expert knowledge of Powdr's bus protocol.
-/// Inputs in the order of: id, selector, payload, is_permutation for both lookup and permutation.
-let bus_multi_receive_batch_lookup_permutation: (expr, expr, expr[], int)[] -> () = constr |inputs| {
+/// Inputs in the order of: id, selector, payload, type for both lookup and permutation.
+/// where type == 0 corresponds to bus send, type == 1 lookup bus receive, and type == 2 permutation bus receive.
```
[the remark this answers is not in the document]
qwang98: sgtm. will do on Monday :)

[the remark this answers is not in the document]
qwang98: Done and added `BusLinkerType` enum in `bus.asm`.

std/protocols/bus.asm
```diff
@@ -258,24 +258,35 @@ let bus_multi: BusInteraction[] -> () = constr |bus_inputs| {
...
+        match type {
+            BusLinkerType::Send => acc + [BusInteraction::Send(id, payload, selector)],
+            BusLinkerType::LookupReceive => {
+                let multiplicity;
+                (1 - selector) * multiplicity = 0;
+                acc + [BusInteraction::Receive(id, payload, multiplicity, selector)]
+            },
+            BusLinkerType::PermutationReceive => acc + [BusInteraction::Receive(id, payload, selector, selector)],
+            // Practically this can never be reached because `bus_multi_linker` should be used by the linker only.
+            _ => std::check::panic("`bus_multi_linker`: type param not valid")
```
Schaeff, line 288:
```suggestion
```

this can be removed (in Rust it would be flagged as unreachable match arm)

test_data/std/bus_multi_linker.asm
```diff
@@ -41,19 +42,20 @@ machine Main with degree: 8 {
...
-    // Multi receive (last argument `is_permutation` is 1 for permutation and 0 for lookup)
-    bus_multi_receive_batch_lookup_permutation([
-      (LOOKUP_ID_0, lookup_latch, [LOOKUP_X, LOOKUP_Y], 0), // selector is lookup_latch, multiplicity is a witness column (not an input here)
-      (LOOKUP_ID_1, lookup_latch, [LOOKUP_X, LOOKUP_Z], 0), // selector is lookup_latch, multiplicity is a witness column (not an input here)
-      (PERMUTATION_ID_0, sub_sel_0 * perm_latch, [PERM_X, PERM_A], 1), // selector is sub_sel_0 * perm_latch, multiplicity is `-selector`
-      (PERMUTATION_ID_1, sub_sel_1 * perm_latch, [PERM_X, PERM_B], 1) // selector is sub_sel_1 * perm_latch, multiplicity is `-selector`
+    // Batch all sends, lookup receives, and permutation receives
+    // Input format: id, selector, payload, type
+    bus_multi_linker([
+      // Bus send (type param == 0)
```
Schaeff, line 48:
```suggestion
      // Bus send
```

test_data/std/bus_multi_linker.asm
```diff
@@ -41,19 +42,20 @@ machine Main with degree: 8 {
...
+      // Latch is always the same as multiplicity for sends (for both lookup and permutation)
+      (LOOKUP_ID_0, lookup_latch, [x, y], BusLinkerType::Send),
+      (LOOKUP_ID_1, lookup_latch, [x, z], BusLinkerType::Send),
+      (PERMUTATION_ID_0, sel, [x, a], BusLinkerType::Send),
+      (PERMUTATION_ID_1, sel, [x, b], BusLinkerType::Send),
+      // Lookup receive (type param == 1)
```
Schaeff, line 54:
```suggestion
      // Lookup receive
```

test_data/std/bus_multi_linker.asm
```diff
@@ -41,19 +42,20 @@ machine Main with degree: 8 {
...
+      (LOOKUP_ID_0, lookup_latch, [LOOKUP_X, LOOKUP_Y], BusLinkerType::LookupReceive), // selector is lookup_latch, multiplicity is a witness column (not an input here)
+      (LOOKUP_ID_1, lookup_latch, [LOOKUP_X, LOOKUP_Z], BusLinkerType::LookupReceive), // selector is lookup_latch, multiplicity is a witness column (not an input here)
+      // Permutation receive (type param == 2)
```
Schaeff, line 57:
```suggestion
      // Permutation receive
```

linker/src/bus.rs
```diff
@@ -176,32 +168,26 @@ impl LinkerBackend for BusLinker {
...
-            (&mut self.bus_multi_args, "std::protocols::bus::bus_multi"),
-            (
-                &mut self.bus_multi_receive_batch_lookup_permutation_args,
-                "std::protocols::bus::bus_multi_receive_batch_lookup_permutation",
-            ),
-        ] {
-            if !args.items.is_empty() {
-                self.pil.push(PilStatement::Expression(
+        // add pil for bus_multi_linker
+        if !self.bus_multi_linker_args.items.is_empty() {
```
Schaeff, line 172: What if we remove this `if` for the sake of simplicity? Everything should be optimized away anyway?

std/protocols/bus.asm
qwang98: Nice :) You mean in our compiler stack?

test_data/std/bus_multi_linker.asm
qwang98: Done

test_data/std/bus_multi_linker.asm
qwang98: Done

std/protocols/bus.asm
Schaeff: No, PIL does not flag these afaik

linker/src/bus.rs
qwang98: Do you mean that we know for a fact that each machine/object, at this stage of the compiler stack, should at least have either a link or an operation?

std/protocols/bus.asm
qwang98: Then wdym in Rust it would be flagged as unreachable match arm? Do you mean that at some stage where our PIL match statement is parsed, anything that's not a match case will be flagged in Rust to be unreachable, so it will panic in Rust and therefore no PIL std::check::panic is required?

linker/src/bus.rs
qwang98: If yes then I think this statement can be removed.

std/protocols/bus.asm
Schaeff: No no, you can ignore the comment about Rust, my only point is that this match arm is unreachable and should therefore be removed.

linker/src/bus.rs
Schaeff: No, rather that it's fine to have no links or operations, then calling the function on an empty array of interactions should just yield nothing.

std/protocols/bus.asm
qwang98: Ok! :)

linker/src/bus.rs
qwang98: Ok!

qwang98: Looks like removing this statement fails the `static_bus.asm` test in pipeline: https://github.com/powdr-labs/powdr/actions/runs/13372098179/job/37343855741?pr=2486.

I put it back and rerun the test locally and it passes.

I can investigate it more tomorrow but seems that Pilopt doesn't remove such function call with empty args? I'm not particularly familiar with Pilopt but if you find this reasonable, feel free to approve this PR.

Schaeff (approved):

PR #2490, opened 2025-02-17
Add native bus linker mode to Stwo

ShuangWu121:
This PR add native bus linker mode to Stwo

more description will be added when the implementation is more clear.
- Add constraints for lookup identity in PowdrEval evaluate function, need to check the maximum numbers of the columns that are combined by random linear combination, to put in either left or right side of the logup
- gen_interaction_trace function
- provide the related witness columns to gen_interaction_trace function
- buid multiplicity


ShuangWu121:
use another branch

PR #2492, opened 2025-02-17
Make memory arith JIT ready

Schaeff:
Blocking #2425 , adapted from `arith256.asm`

### DIFF
--- a/std/machines/large_field/arith256_memory.asm
+++ b/std/machines/large_field/arith256_memory.asm
@@ -53,6 +53,7 @@ machine Arith256Memory(mem: Memory) with
     // Get an intermediate column that indicates that we're in an
     // actual block, not a default block. Its value is constant
     // within the block.
+    // TODO: Witgen fails if this is an intermediate column.
     col witness used;
     used = array::sum(sel);
     array::map(sel, |s| unchanged_until(s, CLK32[31]));
--- a/std/machines/large_field/arith256_memory.asm
+++ b/std/machines/large_field/arith256_memory.asm
@@ -202,10 +202,6 @@ machine Arith256Memory(mem: Memory) with
     let limbs_to_ints: fe[] -> int[] = |l| array::new(array::len(l) / 16, |i| limbs_to_int(array::sub_array(l, i * 16, 16)));
     let int_to_limbs: int -> fe[] = |x| array::new(16, |i| fe(select_limb(x, i)));
 
-    let provide_values = query |column_arr, row, value| {
-        let _ = array::map_enumerated(column_arr, |j, column| std::prover::provide_value(column, row, fe(select_limb(value, j))));
-    };
-
     let get_operation = query || match eval(operation_id) {
         1 => "affine_256",
         2 => "mod_256",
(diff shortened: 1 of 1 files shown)


std/machines/large_field/arith256_memory.asm
```diff
@@ -53,7 +54,6 @@ machine Arith256Memory(mem: Memory) with
     // Get an intermediate column that indicates that we're in an
     // actual block, not a default block. Its value is constant
     // within the block.
-    // TODO: Witgen fails if this is an intermediate column.
```
chriseth, line 56: Does it not fail any more?

Schaeff: oops, it does, I reverted but didn't put back the comment

std/machines/large_field/arith256_memory.asm
```diff
@@ -218,61 +212,85 @@ machine Arith256Memory(mem: Memory) with
...
-            let x3_val = (s_val * s_val - x1 - x2 + 2 * secp_modulus) % secp_modulus;
-            provide_values(x3, i, x3_val);
-            let y3_val = (s_val * ((x1 - x3_val) + secp_modulus) - y1 + secp_modulus) % secp_modulus;
-            provide_values(y3, i, y3_val);
-
-            provide_values(q1, i, -(s_val * s_val - x1 - x2 - x3_val) / secp_modulus + (1 << 258));
-            provide_values(q2, i, -(s_val * x1 - s_val * x3_val - y1 - y3_val) / secp_modulus + (1 << 258));
+    // Prover function for mod_256
+    query |i| std::prover::compute_from_multi_if(
+        operation_id = 2,
```
chriseth, line 232: We don't have that in the other file - did I miss it?

std/machines/large_field/arith256_memory.asm
```diff
@@ -32,7 +32,8 @@ machine Arith256Memory(mem: Memory) with
     let operation_selectors = [is_affine, is_mod, is_ec_add, is_ec_double];
     array::map(operation_selectors, |s| force_bool(s));
     array::map(operation_selectors, fixed_inside_32_block);
-    let operation_id = sum(4, |i| 2 ** i * operation_selectors[i]);
+    col witness operation_id;
```
Schaeff, line 35: I had to change this because if I understand correctly, the jit prover function heuristic expects an algebraic expression in the condition of `compute_from_multi_if`, but this is an `expr`

Schaeff: If I use `col operation_id: inter`, then I get a [panic in the optimizer](https://github.com/powdr-labs/powdr/blob/2614c0402de171eaf2c6a4cc1967af0cedda64c2/pilopt/src/lib.rs#L132-L133) with `n.type_args = Some([])`

std/machines/large_field/arith256_memory.asm
```diff
@@ -218,61 +212,85 @@ machine Arith256Memory(mem: Memory) with
         _ => panic("Unknown operation")
     };
 
+    // Prover function for affine_256
+    query |i| std::prover::compute_from_multi_if(
+        operation_id = 1,
```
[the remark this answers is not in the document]
Schaeff: Oh, this would solve [this](https://github.com/powdr-labs/powdr/pull/2492#discussion_r1958652863)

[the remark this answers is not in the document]
Schaeff: Done

std/machines/large_field/arith256_memory.asm
Schaeff: In arith, affine_256 and mod_256 share the same operation id, so maybe you shared the prover function there as well? In this machine, the operation ids are different.

chriseth (approved):

PR #2493, opened 2025-02-18
Mark zero-selector calls as complete.

chriseth:


lvella (approved):
I don't know the context, but LGTM.

PR #2498, opened 2025-02-18
Always use first parameter as operation ID.

chriseth:


leonardoalt (approved):

chriseth:
The benches are right to reject this one.

chriseth:
We first need to try to solve the generic version and then try the specific if the generic does not work.

leonardoalt (approved):

PR #2502, opened 2025-02-19
Refactor publics: remove dependency on trace

qwang98:
WIP but this represents a "complete-ish" PR feature wise. With this PR, public reference values computed in #1756 is passed down to Plonky3 prover, so that public values are no longer obtained from witness trace via public declaration. An end result is that the following test passes: `cargo run pil test_data/pil/fibonacci_with_publics.pil --prove-with plonky3 -f`.

A few major questions remain:
1. I still haven't edited Plonky3 prover to remove our current way of "exposing publics", which basically creates selector columns for the specific row index of a witness column here https://github.com/powdr-labs/powdr/blob/main/backend/src/plonky3/stark.rs#L132-L161 and then constrain public inputs using witness columns in stage 0: https://github.com/powdr-labs/powdr/blob/main/plonky3/src/circuit_builder.rs#L357-L374. A few Plonky3 specific questions for @Schaeff here:

- Are these "witness columns" newly created or refer to existing witness column that the public exposes?
- Why is it specifically stage 0?
- What are these "public constraints" and how are they different from the selector/witness constraint above: https://github.com/powdr-labs/powdr/blob/main/plonky3/src/circuit_builder.rs#L311-L319
- Would you mind pointing me to the "native API" for publics in Plonky3?
- Which stage will the scalar publics using the native API be?

2. Currently public reference and public declaration still don't have a good way of matching each other except by the local variable name. `AlgebraicExpression::PublicReference(name)` has a local variable `name` like "out" whereas `PublicDeclaration` contains the absolute name like "Fibonacci::out" and the witness polynomial it attempts to expose. We discussed in chat that a globally unique id for public reference is needed, but it still doesn't match a public reference to its corresponding public declaration. @georgwiese @chriseth 

3. Am I correct that *both* public declaration and public references will be needed eventually? Currently Plonky3 prover still uses public declaration to get the stage information of a public input.

----------------Historical Notes Below This Line-------------------
WIP and this doesn't even compile, but I'd like to get a brief temp check so that I don't go in the wrong direction. @georgwiese 

This PR is motivated by "Third Task" of #2485, which attempts to push forward the progress of #1633, whose progress is as far as #1756. As of now, we have both public declaration and public references. Our public declaration refers to a trace cell in witgen, which is extracted for use in our provers. #1756 calculates public references using the witgen, but its results are not passed to the prover yet. In the eventual view of #1633, public reference is the only route to public values in provers and we would remove the public declaration -> witgen trace cell -> prover route. My understanding is that public declarations will still be needed together with public references.

To break this down a bit, I envision a series of the following PRs:
1. Remove provers' dependency on witness for extracting public declaration values as a trace cell, which this PR is in progress of accomplishing. More specifically, I will attempt to pass public reference values computed in #1756 down the pipeline to the prover. These values will be a `BTreeMap<String, T>` across *all* machines where `String` is the *local* name rather than absolute path, which obviously isn't ideal (think of scenario where two machines expose two publics both called `out`). I'm only editing Plonky3 so far.
5. Make public references machine specific. I think this will require passing user level information, especially the absolute path, to `parser::Expression::PublicReference`, which might require editing the `lalrpop`? The absolute path will be needed to split `PublicReference` to the correct machine, similar to how `PublicDeclaration` is split.
6. Remove redundant APIs for public declaration, such as the row index used to extract witgen trace cell. Remove such user syntax as well.
7. Add checkers for the consistency among the same public reference calculated by different machines.

Some specific questions: 
1. What do you envision as the best "key" for public reference? Currently in AlgebraicExpression::PublicReference it's just the *local* name of the variable.
2. In this PR, I attempt to pass publics as a separate input param to the `Backend::prove` function alongside witness. I believe this is needed, right?
3. Any thoughts on the order/division of PRs listed above?
4. Why is witgen for public reference only for dynamic machine and block machine? My initial thought is that it's basically the main machine plus user defined machines that's not specific ones like doublesortedwitness.

### DIFF
--- a/backend/src/plonky3/stark.rs
+++ b/backend/src/plonky3/stark.rs
@@ -209,6 +209,7 @@ where
         public: &BTreeMap<String, T>,
         witgen_callback: WitgenCallback<T>,
     ) -> Result<Vec<u8>, String> {
+        println!("Plonky3Prover::prove public: {:?}", public);
         let mut witness_by_machine = self
             .split
             .iter()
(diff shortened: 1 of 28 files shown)


backend/src/plonky3/stark.rs
```diff
@@ -206,8 +206,10 @@ where
     pub fn prove(
         &self,
         witness: &[(String, Vec<T>)],
+        public: &BTreeMap<String, T>,
```
qwang98, line 209: Map from public reference *local* variable name to value, created by witgen.

backend/src/plonky3/stark.rs
```diff
@@ -235,14 +237,15 @@ where
...
+            public,
             &mut challenger,
         );
 
         let mut challenger = T::get_challenger();
 
         let verifying_key = self.verifying_key.as_ref();
 
-        let public_values = circuit.public_values_so_far(&witness_by_machine);
+        let public_values = circuit.public_values_so_far(&public);
```
qwang98, line 248: This removes dependence on witness values. Previously, public values are obtained by locating a trace cell from witgen. Now, it's directly looked up from the map of public reference name to value.

executor/src/witgen/data_structures/mutable_state.rs
```diff
@@ -100,25 +102,41 @@ impl<'a, T: FieldElement, Q: QueryCallback<T>> MutableState<'a, T, Q> {
...
-            .iter()
-            .flat_map(|machine| {
-                let mut machine = machine
-                    .try_borrow_mut()
-                    .map_err(|_| {
-                        panic!("Recursive machine dependencies while finishing machines.");
-                    })
-                    .unwrap();
-                let columns = machine.take_witness_col_values(&self).into_iter();
+        let (witness_columns, public_values, _processed) = self.machines.iter().fold(
```
qwang98, line 109: Need to return _processed here to make sure it lives long enough. This is for:
```
// We keep the already processed machines mutably borrowed so that
// "later" machines do not try to create new rows in already processed
// machines.
```

executor/src/witgen/machines/dynamic_machine.rs
```diff
@@ -55,10 +55,11 @@ impl<'a, T: FieldElement> Machine<'a, T> for DynamicMachine<'a, T> {
...
         assert!(self.data.is_empty());
         let first_row = self.compute_partial_first_row(mutable_state);
-        self.data = self
+        let process_result = self
             .process(first_row, 0, mutable_state, None, true)
-            .updated_data
-            .block;
+            .updated_data;
+        self.data = process_result.block;
+        self.publics.extend(process_result.publics); // at least for the fibonacci example, this is required here to pass the publics to witgen results (i think it might be that the main machine is the dynamic machine and it's not called via a lookup, so we have to also extend publics here as well if it's called via run(), which means the first machine called i think, typically the main)
```
qwang98, line 62: This seems to be a patch to PR #1756, which passes the generated public reference values in `process_plookup` but not here in `run`. 

At least for the fibonacci example, passing out the public reference values to `run` is required so that `WitnessGenerator::generate()` can return it. It's likely because the main machine (the only machine in the fibonacci_with_publics example) is a dynamic machine that's not called via a lookup, so we have to also extend publics here if it's called via `run`.

executor/src/witgen/machines/mod.rs
```diff
@@ -140,6 +140,13 @@ pub trait Machine<'a, T: FieldElement>: Send + Sync {
         mutable_state: &'b MutableState<'a, T, Q>,
     ) -> HashMap<String, Vec<T>>;
 
+    fn take_public_values(&mut self) -> BTreeMap<String, T> {
+        unimplemented!(
+            "Public values are not supported for machine {}",
+            self.name()
+        );
+    }
```
qwang98, line 148: Defaults to unimplemented, and is only implemented by dynamic machine and block machine. I guess these are the only "normal" machines that are not specific ones like the memory machine and etc., which shouldn't contain publics.

This is at least the case for #1756 so I followed here. Let me know if otherwise.

executor/src/witgen/mod.rs
```diff
@@ -241,24 +242,21 @@ impl<'a, 'b, T: FieldElement> WitnessGenerator<'a, 'b, T> {
         let machines = MachineExtractor::new(&fixed).split_out_machines();
 
         // Run main machine and extract columns from all machines.
-        let columns = MutableState::new(machines.into_iter(), &self.query_callback).run();
+        let (columns, publics) =
+            MutableState::new(machines.into_iter(), &self.query_callback).run();
 
-        let publics = extract_publics(&columns, self.analyzed);
```
qwang98, line 246: Publics were obtained from `extract_publics` which uses trace cell of `columns`, i.e. witness columns.

Now, it's directly generated as a map from public reference name to value, returned by witgen as `publics` here.

executor/src/witgen/mod.rs
```diff
@@ -241,24 +242,21 @@ impl<'a, 'b, T: FieldElement> WitnessGenerator<'a, 'b, T> {
...
             log::debug!("Publics:");
         }
         for (name, value) in publics.iter() {
-            log::debug!(
-                "  {name:>30}: {}",
-                value
-                    .map(|value| value.to_string())
-                    .unwrap_or_else(|| "Not yet known at this stage".to_string())
-            );
+            log::debug!("  {name:>30}: {}", value.to_string());
```
qwang98, line 254: Because public values were previously obtained from witgen, which seems to be in multiple stages and therefore sometimes public value might not be available for a later stage witgen. However, now public values are separately calculated for all public references in witgen in #1756, so I'm not sure if this still holds. @georgwiese

executor/src/witgen/multiplicity_column_generator.rs
```diff
@@ -70,11 +70,7 @@ impl<'a, T: FieldElement> MultiplicityColumnGenerator<'a, T> {
             .collect::<BTreeMap<_, _>>();
         let terminal_values = OwnedTerminalValues {
             trace: all_columns,
-            public_values: publics
-                .into_iter()
-                // Publics might be unavailable if they are later-stage publics.
-                .filter_map(|(k, v)| v.map(|v| (k, v)))
-                .collect(),
+            public_values: publics.clone(),
```
qwang98, line 73: Similar question as before.

plonky3/src/circuit_builder.rs
```diff
@@ -159,24 +159,31 @@ where
...
                     .map(|publics| {
                         publics
                             .iter()
-                            .map(|(_, name, _, row)| witness.get(name).map(|column| column[*row]))
+                            .map(|(name, _, _, _)| {
+                                println!("public_values_so_far publics: {:?}", public);
+                                println!("public_values_so_far public: {:?}", name);
+                                public
+                                    .get(name.rsplit("::").next().unwrap_or(name))
+                                    .cloned()
```
qwang98, line 185: `public` contains local name from public references. `name` are from public declarations which are absolute, so we need to split the local name here.

plonky3/src/circuit_builder.rs
```diff
@@ -258,7 +265,7 @@ struct Data<'a, T, AB: MultistageAirBuilder> {
     constraint_system: &'a ConstraintSystem<T>,
     traces_by_stage: &'a [AB::M],
     fixed: &'a AB::M,
-    publics: &'a BTreeMap<&'a str, <AB as MultistageAirBuilder>::PublicVar>,
+    publics: &'a BTreeMap<&'a str, <AB as MultistageAirBuilder>::PublicVar>, // currently these are public declarations
```
qwang98, line 268: Note that these are from public declarations.

plonky3/src/circuit_builder.rs
```diff
@@ -285,11 +292,17 @@ impl<T, AB: MultistageAirBuilder> TerminalAccess<AB::Expr> for &Data<'_, T, AB>
...
+        println!("get_public all publics: {:?}", self.publics); // these are from public declarations
+        println!("get_public public: {:?}", public); // these are from public references and are local names
+                                                     // need a way to match the local names
+
         (*self
             .publics
-            .get(public)
+            .iter()
+            .find(|(abs_name, _)| abs_name.rsplit("::").next().unwrap_or(abs_name) == public)
+            .map(|(_, value)| value)
```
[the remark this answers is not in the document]
qwang98: Only the first matching entry is returned by find here.

plonky3/src/circuit_builder.rs
```diff
@@ -308,7 +321,7 @@ where
             .map(|i| builder.stage_public_values(i))
             .collect_vec();
 
-        // public constraints
```
qwang98, line 311: These are initially from ConstraintSystem, which comes from split pil, which contain public declarations rather than public references

plonky3/src/prover.rs
```diff
@@ -440,14 +443,19 @@ where
...
-                                    .unwrap()
-                                    .iter()
-                                    .find_map(|(n, v)| (n == column_name).then(|| v[*row]))
-                                    .unwrap()
-                                    .into_p3_field()
+                            .map(|(name, column_name, _, _)| {
+                                println!("name: {:?}", name); // absolute name, need to get local name
+                                println!("column_name: {:?}", column_name);
+                                let local_name = name.rsplit("::").next().unwrap_or(name);
+                                println!("local_name: {:?}", local_name);
```
qwang98, line 450: removes dependence on witness

Schaeff:
> A few Plonky3 specific questions for @Schaeff here:
> Are these "witness columns" newly created or refer to existing witness column that the public exposes?
for `public p = w[42];` we introduce `F42 = i == 42 ? 1 : 0` and p3 public `q` with constraint `F42 * (w - q) = 0`
So `w` is the original column, `F42` and `q` are added in p3, as well as the constraint.
> Why is it specifically stage 0?
It is not stage 0, there is no injected witness, see above.
> What are these "public constraints" and how are they different from the selector/witness constraint above: https://github.com/powdr-labs/powdr/blob/main/plonky3/src/circuit_builder.rs#L311-L319
This is inside `eval` which is a visitor of the program. This bit gets the value of the public input, in whatever type this program visit uses. You'd have the same map for scalar publics.
> Would you mind pointing me to the "native API" for publics in Plonky3?
[here](https://github.com/powdr-labs/powdr/blob/d81f8bd347813003af256a4b53203dd8a0507965/plonky3/src/circuit_builder.rs#L405) we define how many public inputs there are for each stage (which can be determined from the program) then [here](https://github.com/powdr-labs/powdr/blob/d81f8bd347813003af256a4b53203dd8a0507965/plonky3/src/circuit_builder.rs#L307) we get values for these public inputs. Then in this same `eval` function, we can constrain these public inputs.
> Which stage will the scalar publics using the native API be?
We need them at all stages: stage 0 publics expose program inputs and outputs, stage 1 publics expose the logup accumulators

PR #2503, opened 2025-02-19
Handle handle_query prover functions.

chriseth:
Parses query / prover functions of the form
```
 query |<i>| std::prover::handle_query(<Y>, <i>, match std::prover::eval(<pc>) {
   <value1> => std::prelude::Query::Output(std::convert::int::<fe>(std::prover::eval(<arg1>)), std::prover::eval(<arg2>)),
   <value2> => std::prelude::Query::Input(std::convert::int::<fe>(std::prover::eval(<arg1>)), std::convert::int::<fe>(std::prover::eval(<arg2>))),
   ...
   _ => std::prelude::Query::None,
```
i.e. the ones used for "free input" columns.

Also calls the query function from the generated code accordingly.

### DIFF
--- a/executor/src/witgen/jit/compiler.rs
+++ b/executor/src/witgen/jit/compiler.rs
@@ -835,6 +835,25 @@ extern \"C\" fn witgen(
         }
     }
 
+    fn witgen_fun_params_with_params<'a>(
+        data: &mut [GoldilocksField],
+        known: &mut [u32],
+        params: &'a mut [LookupCell<'a, GoldilocksField>],
+    ) -> WitgenFunctionParams<'a, GoldilocksField> {
+        WitgenFunctionParams {
+            data: data.into(),
+            known: known.as_mut_ptr(),
+            row_offset: 0,
+            params: params.into(),
+            mutable_state: std::ptr::null(),
+            call_machine: no_call_machine,
+            fixed_data: null(),
+            get_fixed_value: get_fixed_data_test,
+            input_from_channel: input_from_channel_test,
+            output_to_channel: output_to_channel_test,
+        }
+    }
+
     #[test]
     fn load_code() {
         let x = cell("x", 0, 0);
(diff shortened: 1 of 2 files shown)


executor/src/witgen/jit/prover_function_heuristics.rs
```diff
@@ -36,12 +38,23 @@ pub enum ProverFunctionComputation<'a> {
     /// The expression `f` in `query |i| std::prover::compute_from(Y, i, [X, ...], f)`,
     /// where f: (fe[] -> fe)
     ComputeFrom(&'a Expression),
+    /// Represents a call to `handle_query` (see `try_decode_handle_query`), where the
+    /// keys are the pc values (the first input column). The two two arguments to both input
```
pacheco, line 42: "two two arguments"

pacheco: I think this comment is missing some context here (not sure what "two arguments" or "input columns" are referring to)... maybe it should be elsewhere or have a bit more context?

executor/src/witgen/jit/prover_function_heuristics.rs
```diff
@@ -59,33 +72,41 @@ fn decode_prover_function<T: FieldElement>(
     function: &Expression,
     try_symbol_by_name: impl TrySymbolByName,
 ) -> Result<ProverFunction<'_, T>, String> {
-    if let Some(provide_if_unknown) =
-        try_decode_provide_if_unknown(index, function, try_symbol_by_name)
-    {
-        Ok(provide_if_unknown)
+    let results = [
+        try_decode_provide_if_unknown(index, function, try_symbol_by_name),
```
pacheco, line 76: very minor, but this way its always evaluating all three, instead of stopping on the first success... if the functions have the same signature, maybe its possible to do something like `[try_func1, try_func2, try_func3].iter().map(|f| f(...))`?

chriseth: It will fail rather quickly and it makes it easy to access all the errors.

executor/src/witgen/jit/compiler.rs
```diff
@@ -553,6 +587,28 @@ fn prover_function_code<T: FieldElement, D: DefinitionFetcher>(
             assert!(!f.compute_multi);
             format!("({}).call()", codegen.generate_code_for_expression(code)?)
         }
+        ProverFunctionComputation::HandleQueryInputOutput(branches) => {
+            let indent = "        ";
+            // We assign zero in the "no match" case. The correct behaviour would be to
+            // not assign anything, but it should work for all our use-casse.
```
Schaeff, line 593:
```suggestion
            // not assign anything, but it should work for all our use-cases.
```

PR #2504, opened 2025-02-20
Skip dispatcher for main VM

Schaeff:
To simplify witgen, avoid using a dispatcher for the main VM


asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -202,29 +177,22 @@ impl<T: FieldElement> VMConverter<T> {
                         use RegisterTy::*;
                         match reg.ty {
                             // Force pc to zero on first row.
-                            Pc => {
+                            Pc | Write => {
```
chriseth, line 180: Is adding `Write` here needed?

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -128,19 +122,10 @@ struct VMConverter<T> {
     line_lookup: Vec<(String, String)>,
     /// Names of fixed columns that contain the rom.
     rom_constant_names: Vec<String>,
-    /// the maximum number of inputs in all functions
-    output_count: usize,
```
chriseth, line 132: Why don't we need this any more?

asm-to-pil/src/vm_to_constrained.rs
chriseth: Ah, because previously, reset would reset the write registers.

asm-to-pil/src/vm_to_constrained.rs
Schaeff: It got moved somewhere else iirc, cleaner now

chriseth:
Ok it turns out that the bug is that witgen uses externally provided values in the following way: if we go past the number of provided values, we wrap. The riscv-executor only generates 4096 values, but later, witgen extends this pretty much. Now it wraps `pc` and starts the program from scratch in row 4096. This used to work, but it does not work any more, since the transition to `pc = 0` is not valid, we have to continue looping on the same `pc`.

asm-to-pil/src/vm_to_constrained.rs
```diff
@@ -202,29 +177,22 @@ impl<T: FieldElement> VMConverter<T> {
                         use RegisterTy::*;
                         match reg.ty {
                             // Force pc to zero on first row.
```
Schaeff, line 179:
```suggestion
                            // Force the pc and the write registers to zero on the first row.
```

riscv-executor/src/lib.rs
```diff
@@ -3094,15 +3062,17 @@ fn execute_inner<F: FieldElement>(
...
     let mut curr_pc = 0u32;
-    e.proc.push_row(PC_INITIAL_VAL as u32);
+    e.proc.push_row(curr_pc);
     let mut last = Instant::now();
     let mut count = 0;
     loop {
         let stm = statements[curr_pc as usize];
 
-        log::trace!("l {curr_pc}: {stm}",);
+        // step is updated by 4 cause we have instructions that need that many memory accesses,
```
Schaeff, line 3071:
```suggestion
        // step is updated by 4 because we have instructions that need that many memory accesses,
```

riscv-executor/src/lib.rs
```diff
@@ -3094,15 +3062,17 @@ fn execute_inner<F: FieldElement>(
...
+        // except on a DebugDirective which is a noop
+        let mut step_update = 4;
 
-        e.step += 4;
+        log::trace!("l {curr_pc}: {stm}",);
```
Schaeff, line 3075: Do we still want this?

chriseth (approved):

PR #2505, opened 2025-02-24
Bus using stwo logup

ShuangWu121:
This PR implements the bus via stwo native logup.

stwo provides a evaluate function that accepts extension field element.  for bus constraints the following relations are to be built:
```
((acc'-acc)*payload -multiplicity)*selector_not_first
public=acc*is_last
```

in the verification side, verify the sum of the publics with the same bus_id is 0.

currently these relation information is extract from `PhantomBusInteraction`, acc is built in the backend. The function I used from stwo to build acc in extension field has SIMD features, it cannot accept witness with length less than 16 (the parellized computation unit size)

the `shift_small_test.asm, bus_lookup.asm (modified to size 16), bus_multi_linker.asm, bus_multi_lookup.asm, memory_small_test.asm ` can be proved successfully. With current bus interface, this means the backend stwo proves  bus relations built by powdr pil and the extra stwo native logup relations extracted from PhantomBusInteraction.

would be good to have a compiled pil file, with PhantomBusInteraction identity but without stage 1 witness.

### DIFF
--- a/backend/src/stwo/bus_acc_gen.rs
+++ b/backend/src/stwo/bus_acc_gen.rs
@@ -101,11 +101,7 @@ impl PowdrLogupColGenerator<'_> {
 
         self.gen.trace.push(self.numerator.clone());
 
-        // println!(
-        //     "\n fraction trace now in logup generator is {:?} ",
-        //     self.gen.trace
-        // );
-        // println!("length of gen.trace after fraction {}", self.gen.trace.len());
+        
 
         let mut last_value = QM31::zero();
 
@@ -132,7 +128,7 @@ impl PowdrLogupColGenerator<'_> {
         }
         self.gen.trace.push(self.numerator);
 
-        //println!("\n length of gen.trace after accumulator compute {}", self.gen.trace.len());
+   
 
         
 
@@ -162,62 +158,9 @@ impl PowdrLogupColGenerator<'_> {
                 })
             })
             .collect_vec();
-        // println!(
-        //     "\n the gen.trace trace in the bus acc gen is {:?}",
-        //     self.gen.trace
-        // );
-
-        // println!(
-        //     "\n the trace in the bus acc gen is {:?}",
-        //     trace
-        // );
+        
        trace[trace.len().saturating_sub(4)..].to_vec()
        
     }
 
-    // pub fn compute_acc_col(mut self)->
-    //     ColumnVec<CircleEvaluation<SimdBackend, BaseField, BitReversedOrder>>
-    //     {
-
-    //     // let mut last_col_coords = self.trace.pop().unwrap().columns;
-
-    //     // // Compute cumsum_shift.
-    //     // let coordinate_sums = last_col_coords.each_ref().map(|c| {
-    //     //     c.data
-    //     //         .iter()
-    //     //         .copied()
-    //     //         .sum::<PackedBaseField>()
-    //     //         .pointwise_sum()
-    //     // });
-
-    //     let mut acc=PackedQM31::zero();
-
-    //     #[allow(clippy::needless_range_loop)]
-    //     for vec_row in 0..(1 << (self.gen.log_size - LOG_N_LANES)) {
-    //         unsafe {
-    //             let value = self.gen.trace.last().unwrap().packed_at(vec_row);
-    //             acc=acc+value;
-    //             let prev_value = self
-    //                 .gen
-    //                 .trace
-    //                 .last()
-    //                 .map(|col| col.packed_at(vec_row))
-    //                 .unwrap_or_else(PackedSecureField::zero);
-    //             self.numerator.set_packed(vec_row, acc)
-    //         };
-    //     }
-    //     self.gen.trace.push(self.numerator);
-
-    //     let trace = self.gen
-    //     .trace.clone()
-    //     .into_iter()
-    //     .flat_map(|eval| {
-    //         eval.columns.map(|col| {
-    //             CircleEvaluation::new(CanonicCoset::new(self.log_size).circle_domain(), col)
-    //         })
-    //     })
-    //     .collect_vec();
-    //     trace
-
-    // }
 }
--- a/backend/src/stwo/bus_acc_gen.rs
+++ b/backend/src/stwo/bus_acc_gen.rs
@@ -1,0 +1,223 @@
+use itertools::Itertools;
+use num_traits::Zero;
+use stwo_prover::core::backend::simd::column::BaseColumn;
+use stwo_prover::core::backend::simd::column::SecureColumn;
+use stwo_prover::core::backend::simd::conversion::Pack;
+use stwo_prover::core::backend::simd::m31::PackedBaseField;
+use stwo_prover::core::backend::simd::m31::LOG_N_LANES;
+use stwo_prover::core::backend::simd::qm31::{PackedQM31, PackedSecureField};
+use stwo_prover::core::backend::simd::SimdBackend;
+use stwo_prover::core::backend::Column;
+use stwo_prover::core::backend::ColumnOps;
+use stwo_prover::core::fields::m31::BaseField;
+use stwo_prover::core::fields::qm31::{SecureField, QM31};
+use stwo_prover::core::fields::secure_column::SecureColumnByCoords;
+use stwo_prover::core::fields::FieldExpOps;
+use stwo_prover::core::poly::circle::CanonicCoset;
+use stwo_prover::core::poly::circle::CircleEvaluation;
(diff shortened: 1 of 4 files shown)


backend/src/stwo/prover.rs
```diff
@@ -217,15 +240,44 @@ where
...
+                                }
+                                col.set(0, BaseField::zero());
+                                let selector = CircleEvaluation::<
+                                    SimdBackend,
+                                    BaseField,
+                                    BitReversedOrder,
+                                >::new(
+                                    *domain_map.get(&(log_size as usize)).unwrap(),
+                                    col,
+                                );
```
qwang98, line 269: Is this a column with 1 everywhere except 0 in the first row?

Another note is that currently we haven't exposed the bus accumulator in other backends, because our publics infrastructure requires exposing a specific row of a witness column, but we can't know which row in the accumulator column to expose. I'm actually working on this right now so eventually we can expose bus accumulator.

backend/src/stwo/prover.rs
```diff
@@ -470,6 +549,156 @@ where
...
+                    // println!(
+                    //     "bus is found in machine: {}, with bus id {}",
+                    //     machine_name, id
+                    // );
+                    index += 1;
+                    bus_id_to_interaction_map
+                        .insert(index, (id.bus_id.clone(), machine_name.clone(), id.id));
+                }
+            }
+        });
```
qwang98, line 571: This `index` here is called `interaction_id` below and isn't used. I wonder what is it intended for and why does it start from 2?

backend/src/stwo/prover.rs
```diff
@@ -470,6 +549,156 @@ where
...
+            .values()
+            .flat_map(|v| v.iter().cloned()) // Flatten and clone values
+            .collect();
+        all_witness.extend(all_fixed_columns.clone());
+
+        for (interaction_id, (bus_id, machine_name, identity_id)) in &bus_id_to_interaction_map {
+            let log_size_payload = machine_log_sizes.get(machine_name).unwrap().clone();
+
+            let mut payload_cols = Vec::<_>::new();
+            // the bus id.id can change if it is accessed from analyze or split pil
```
qwang98, line 604: What do you mean here? Do you mean the unique id for identities in a machine can change? I think bus_id should remain the same?

backend/src/stwo/prover.rs
```diff
@@ -470,6 +549,156 @@ where
...
+                                let mut col_gen = logup_gen.new_col();
+                                for vec_row in 0..(1 << (log_size_payload - LOG_N_LANES)) {
+                                    let p = bitreverse_multiplicity_col.data[vec_row];
+                                    let payload_denom_values: Vec<PackedM31> = payload_cols
+                                        .iter()
+                                        .map(|col| col.data[vec_row]) // Extract data at vec_row
+                                        .collect();
+                                 
+                                    let q: PackedSecureField =
+                                        lookup_elements.combine(&payload_denom_values);
```
qwang98, line 683: Is this basically doing linear combination with challenge \alpha (lookup_elements) like \sum_i {\alpha^i * payload_i}?

I think `PhantomBusInteractionIdentity` has a `folded_expressions` field, which should have done this exact same calculation. Am I understanding correctly that we need to do this calculation here manually in order to use Stwo's native LogUp?

backend/src/stwo/bus_acc_gen.rs
```diff
@@ -0,0 +1,166 @@
...
+use stwo_prover::core::backend::Col;
+use stwo_prover::core::utils::coset_index_to_circle_domain_index;
+
+pub struct PowdrLogupTraceGenerator {
+    log_size: u32,
+    /// Current allocated interaction columns.
+    pub trace: Vec<SecureColumnByCoords<SimdBackend>>,
+    /// Denominator expressions (z + sum_i alpha^i * x_i) being generated for the current lookup.
+    denom: SecureColumn,
+}
```
qwang98, line 31: `denom` is set to the trace generator while numerator the column generator. Is it because many trace columns share the same denominator?

backend/src/stwo/bus_acc_gen.rs
```diff
@@ -0,0 +1,166 @@
...
+                    .scan(last_value, |state, &x| {
+                        *state += x;
+                        Some(*state)
+                    })
+                    .collect::<Vec<QM31>>() // Collect into a Vec<QM31>
+                    .try_into() // Convert Vec<QM31> into [QM31; 16]
+                    .expect("Expected exactly 16 elements");
+                last_value=acc_pack_array[15];
+                
+                let acc_pack = PackedQM31::from_array(acc_pack_array);
```
qwang98, line 124: Is this the sum of all accumulators (seems to be so from the scan)? Why we expect exactly 16 elements? Is it multiplicity_0 / folded_payload_0 + multiplicity_1 / folded_payload_1, so 4 sets of 4 columns for each extension field = 16 elements?

backend/src/stwo/bus_acc_gen.rs
```diff
@@ -0,0 +1,166 @@
...
+                                ),
+                                self.log_size,
+                            ),
+                            nature_order_col[index],
+                        );
+                    }
+                    CircleEvaluation::new(CanonicCoset::new(self.log_size).circle_domain(), bit_reversed_col)
+                })
+            })
+            .collect_vec();
```
qwang98, line 160: What does this chunk do?

backend/src/stwo/circuit_builder.rs
```diff
@@ -280,6 +319,29 @@ impl FrameworkEval for PowdrEval {
...
+            .iter()
+            .filter(|(_, (_, machine_name, _))| *machine_name == self.machine_name)
+            .map(|(interaction_index, (bus_id, machine_name, identity_id))| {
+                println!("in evaluate function, machine name is {}, bus identity id is {}, interaction index is {}", machine_name, identity_id, interaction_index);
+                (
+                    identity_id,
+                    eval.next_extension_interaction_mask(*interaction_index as usize, [-1, 0]),
+                )
+            })
+            .collect();
```
qwang98, line 343: I assume the purpose here is to add an "expose public" style constraint for bus accumulators. Again, there's no way to know which row of the bus accumulator to expose given our current "expose a row in witness trace" style API for publics. Would need to wait till publics are decoupled from trace rows implemented.

qwang98, line 343: I assume the purpose here is to add an "expose public" style constraint for bus accumulators. Again, there's no way to know which row of the bus accumulator to expose given our current "expose a row in witness trace" style API for publics. Would need to wait till publics are decoupled from trace rows implemented.

backend/src/stwo/circuit_builder.rs
```diff
@@ -300,9 +364,50 @@ impl FrameworkEval for PowdrEval {
...
+                    let denominator: <E as EvalAtRow>::EF = self.lookup_elements.combine(&payload);
+                   
+
+                    eval.add_constraint::<<E as EvalAtRow>::EF>(
+                        <E as EvalAtRow>::EF::from(selector_not_first.clone())
+                            * ((bus_accumulator_eval.get(&id.id).unwrap()[1].clone()
+                                - bus_accumulator_eval.get(&id.id).unwrap()[0].clone())
+                                * denominator
+                                - multiplicity),
+                    );
```
qwang98, line 405: This seems to be the main bus constraint. Is this the "native log up" we are talking about? I'd also note that in our current PIL function call for adding constraints (the non native log up way), we batch two bus interactions for each accumulator column as `acc * selector_not_first + mult_0/denom_0 + mult_1/denom_1 = acc'`. We technically can batch as many `mult/denom` as possible, but currently enforce a degree 3 constraint due to the limit from Plonky3 backend. I think your backend log up native implementation here with one mult/denom per accumulator column is good as a first step, but as some future optimization, we can consider batching multiple mult/denom per accumulator :) Relatedly, do we have a maximum degree for our set up of Stwo?

backend/src/stwo/bus_acc_gen.rs
qwang98: I think my guess isn't correct from a further read of the code.

backend/src/stwo/prover.rs
ShuangWu121: I noticed the identity id changed depends on if I access it from analyzed for all the machines, or I accessed it after split, namely the pil for individual machine.

leonardoalt:
@georgwiese @Schaeff do we still want to merge this soon?

ShuangWu121:
close this one as using stwo logup (none GKR version) for bus constraint is essentially the same as how we implement bus in Powdr

PR #2508, opened 2025-02-25
Native bus interaction

qwang98:
WIP. For Stwo native logup use.

### DIFF
--- a/ast/src/analyzed/mod.rs
+++ b/ast/src/analyzed/mod.rs
@@ -1028,6 +1028,39 @@ impl<T> Children<AlgebraicExpression<T>> for ExpressionList<T> {
     Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema, Hash,
 )]
 /// For documentation, see the equivalent `Constr` variant in std/prelude.asm.
+pub struct BusInteractionIdentity<T> {
+    // The ID is globally unique among identities.
+    pub id: u64,
+    pub source: SourceRef,
+    pub multiplicity: AlgebraicExpression<T>,
+    pub bus_id: AlgebraicExpression<T>,
+    pub payload: ExpressionList<T>,
+    pub latch: AlgebraicExpression<T>,
+}
+
+impl<T> Children<AlgebraicExpression<T>> for BusInteractionIdentity<T> {
+    fn children_mut(&mut self) -> Box<dyn Iterator<Item = &mut AlgebraicExpression<T>> + '_> {
+        Box::new(
+            once(&mut self.multiplicity)
+                .chain(once(&mut self.bus_id))
+                .chain(self.payload.children_mut())
+                .chain(once(&mut self.latch)),
+        )
+    }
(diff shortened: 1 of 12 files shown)


ast/src/analyzed/mod.rs
```diff
@@ -1024,6 +1024,39 @@ impl<T> Children<AlgebraicExpression<T>> for ExpressionList<T> {
...
+)]
+/// For documentation, see the equivalent `Constr` variant in std/prelude.asm.
+pub struct BusInteractionIdentity<T> {
+    // The ID is globally unique among identities.
+    pub id: u64,
+    pub source: SourceRef,
+    pub multiplicity: AlgebraicExpression<T>,
+    pub bus_id: AlgebraicExpression<T>,
+    pub payload: ExpressionList<T>,
+    pub latch: AlgebraicExpression<T>,
```
leonardoalt, line 1038: do we need the latch to be part of this struct?

plonky3/src/circuit_builder.rs
```diff
@@ -159,13 +159,14 @@ where
     /// For stages in which there are no public values, return an empty vector
     pub fn public_values_so_far(
         &self,
-        witness_by_machine: &BTreeMap<String, Vec<(String, Vec<T>)>>,
+        // witness_by_machine: &BTreeMap<String, Vec<(String, Vec<T>)>>,
+        public: &BTreeMap<String, T>,
```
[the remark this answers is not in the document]
qwang98: Yes. Removed.

pilopt/src/lib.rs
```diff
@@ -834,6 +838,7 @@ fn remove_duplicate_identities<T: FieldElement>(pil_file: &mut Analyzed<T>) {
         .enumerate()
         .filter_map(|(index, identity)| match identity {
             // Duplicate bus interactions should not be removed, because that changes the statement.
+            Identity::BusInteraction(_) => None,
```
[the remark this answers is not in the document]
qwang98: Yes, this is exactly where the exception happens. I'm not entirely sure in which case would "duplicate bus interaction" occur, but I think the same shall apply to BusInteraction as PhantomBusInteraction. Let me know why this exception is needed!

ast/src/analyzed/mod.rs
qwang98: I think so, because latch is needed as a separate column from multiplicity in lookup receives to constrain that `(1 - latch) * multiplicity = 0`, i.e. multiplicity must be zero when latch is zero.

qwang98: See: https://github.com/powdr-labs/powdr/blob/main/std/protocols/bus.asm#L281-L285

pilopt/src/lib.rs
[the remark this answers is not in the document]
Schaeff: Right makes sense. The reason is that `send(a), send(a) != send(a)` but `(a in b), (a in b) == (a in b)`.

ast/src/analyzed/mod.rs
leonardoalt: that's kind of annoying...

Schaeff (approved):

PR #2509, opened 2025-02-26
Autoprecompiles: BusInteractionIdentity to SymbolicBusInteraction

gzanitti:


autoprecompiles/src/powdr.rs
```diff
@@ -1247,3 +1251,25 @@ fn optimize_precompile(mut machine: Machine) -> Machine {
 
     machine
 }
+
+fn powdr_interaction_to_symbolic<T: FieldElement>(
+    powdr_interaction: BusInteractionIdentity<T>,
+) -> SymbolicBusInteraction<T> {
+    let kind = match powdr_interaction.multiplicity {
+        AlgebraicExpression::Number(_) => BusInteractionKind::Send,
+        AlgebraicExpression::UnaryOperation(_) => BusInteractionKind::Receive,
```
leonardoalt, line 1260: not sure I understand this. Why are Numbers Sends and UnaryOperations Receives?

leonardoalt: ah, positive multiplicity are sends and negative receives? Is it always like this?

gzanitti: It's what it says in the prelude: 
`- An expression for the multiplicity. Negative for bus receives.`

leonardoalt (approved):

PR #2510, opened 2025-02-26
Create bus.asm call to expose native BusInteraction to the backend

qwang98:
Currently just the bus.asm call is created. Next step will be editing `bus.rs` linker to do this call for backends that needs the native bus interaction. 

A key quesiton remains whether we need to import the backend crate or create another linker mode in the linker crate in order to distinguish which cases to pass down the native bus interaction vs which cases to call `bus_multi_linker`.

A third solution is to make both `bus_multi_linker` and `bus_native_linker` call in `bus.rs`, and this will create both `BusInteraction` and `PhantomBusInteraction` for all backends, and some backends can just ignore the native `BusInteraction`, similar to what we've done for native lookup and permutation.

### DIFF
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -19,8 +19,8 @@ jobs:
     ##### The block below is shared between cache build and PR build workflows #####
     - name: Install EStarkPolygon prover dependencies
       run: sudo apt-get update && sudo apt-get install -y nlohmann-json3-dev libpqxx-dev nasm libgrpc++-dev libprotobuf-dev protobuf-compiler-grpc uuid-dev build-essential cmake pkg-config git
-    - name: Install Rust toolchain nightly-2024-09-21 (with clippy and rustfmt)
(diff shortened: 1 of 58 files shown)


qwang98:
Ready for review.

Schaeff:
@qwang98 A lot of the changes look like #2508, for example in halo2. Why is that?

qwang98:
> @qwang98 A lot of the changes look like #2508, for example in halo2. Why is that?

Oh it simply needs a rebase. :)

ShuangWu121 (approved):
The code looks good to me. 

The next step is "editing bus.rs linker to do this call for backends that needs the native bus interaction"  and also remove the logup related constraint and stage 1 witness from pil?

qwang98:
> The code looks good to me.
> 
> The next step is "editing bus.rs linker to do this call for backends that needs the native bus interaction" and also remove the logup related constraint and stage 1 witness from pil?

Yes! This will be done in a new "pil tuning" step after pilopt instead of the linker. We've been working on the infrastructure for that for the past weeks.

qwang98:
For now, you'll have all the stage 1 pil constraints as well as the native bus interactions for all backends, so maybe you can start using the native bus interactions for now and also create stage 1 pil constraints there using native logup.

PR #2515, opened 2025-03-04
Fix else if formatting.

chriseth:


executor/src/witgen/jit/compiler.rs
```diff
@@ -412,21 +405,17 @@ fn format_effect<T: FieldElement>(effect: &Effect<T, Variable>, is_top_level: bo
                 "".to_string()
             };
 
-            if matches!(second[..], [Effect::Branch(..)]) {
-                format!(
-                    "{var_decls}if {} {{\n{}\n}} else if {}",
```
chriseth, line 417: The actual fix is to change `else if` to just `else` here (the inner branch will create another `if`). The rest is just some code rearranging - sorry, couldn't help it.

executor/src/witgen/jit/compiler.rs
```diff
@@ -1082,7 +1071,7 @@ extern \"C\" fn witgen(
         let expectation = "    let p_1;
     if 7 <= IntType::from(p_0) && IntType::from(p_0) <= 20 {
         p_1 = (p_0 + FieldElement::from(1));
-    } else if if 7 <= IntType::from(p_2) && IntType::from(p_2) <= 20 {
+    } else if 7 <= IntType::from(p_2) && IntType::from(p_2) <= 20 {
```
chriseth, line 1074: omg, even the test had the double `if`...

lvella (approved):

PR #2516, opened 2025-03-04
Refactor public declaration and public reference

qwang98:
Started as a small fix but expanded to a complete refactoring of `PublicDeclaration` and `PublicReference`.

Now public reference work the same way as any reference and the syntax is without the colon (`:out` -> `out`). `PublicDeclaration` are also put under definitions in the analyzer. Both public reference and public declaration use the absolute path, so it's easy to refer from one to the other now.

This is needed for #2502 to match public reference to public declaration in the backend. Currently, different namespaces can have public references with the same name, and it will confuse the backend on which public reference value to fetch.


pil-analyzer/src/expression_processor.rs
```diff
@@ -76,7 +76,10 @@ impl<'a, D: AnalysisDriver> ExpressionProcessor<'a, D> {
                     .map_err(|e| src.with_error(e))?;
                 Expression::Reference(src, reference)
             }
-            PExpression::PublicReference(src, name) => Expression::PublicReference(src, name),
+            PExpression::PublicReference(src, name) => {
+                let name = self.driver.resolve_decl(&name);
```
qwang98, line 80: `resolve_decl` is also used by `PublicDeclaration` in the pil analyzer to resolve from the local name to the absolute name, so this can ensure that both paths match.

chriseth: Hm, this is a reference, so it should use `resolve_ref` instead. `resolve_decl` is for declarations: `let x = ...` - turn `x` into a global path. So in the public declaration you should use `resolve_decl`, but for references, there is `resolve_ref`. And I think you should use `resolve_value_ref`.

A similar change is needed in the assembly processing, I think.

qwang98: Thanks for the clarification!

qwang98:
I encountered a few more issues when trying to integrate this PR to #2502. Would really appreciate your expertise on the language @chriseth!

The main issue is from the CI fail: https://github.com/powdr-labs/powdr/actions/runs/13653910465/job/38169720150. Because this PR changes the `parsed::Expression::PublicReference` struct non-trivially, when PIL is reparsed in later stages (such as the optimizer in the CI and then again in the backend (https://github.com/powdr-labs/powdr/blob/main/backend-utils/src/lib.rs#L313-L324), the parser won't allow `::` in the name of a public reference. To solve this issue, I thought of two solutions:
> Edit `powdr.lalrpop` to allow `::` characters, but I'm confused on where to edit: https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L592
> Add the same id used by `PublicDeclaration` to `PublicReference`. The id is defined here: https://github.com/powdr-labs/powdr/blob/main/pil-analyzer/src/statement_processor.rs#L511-L552. Because we can't edit `parsed::Expression::PublicReference` due to the error above, we will need to add this id field to `AlgebraicExpression::PublicReference`, which is converted from `parsed::Expression::PublicReference` by the `pil_analyzer::Evaluator`: https://github.com/powdr-labs/powdr/blob/main/pil-analyzer/src/evaluator.rs#L876-L878. However, this `Evaluator` doesn't have context to public declarations, so there's no way to lookup from the absolute name of public reference to the absolute name of public declaration and then to its ID. I might need to edit the `Evaluator` struct to include a `Driver` similar to `ExpressionProcessor` or `StatementProcessor`.

To recap, I need to match public declaration (with absolute name) to public reference (with local name). This PR tries to convert local name of public reference to absolute name, but encounter more issues down the pipeline, so I propose another solution to add public ID to public reference, but I'm not sure about its feasibility.

qwang98:
> Ok, I haven't thought this through all the way, but there is no reason publics should be treated differently than any other symbol. So I think we should get rid of `parsed::Expression::PublicReference` and just use `Reference` for both (also in the grammar!).
> 
> Note that we can use `PolynomialReference` also for publics - it is already used for anything that is not a local variable and the naming is just for historical reasons.
> 
> This probably also means that you have to move `Analyzed::public_declarations` inside `Analyzed::definitions` (something we wanted to do anyway). You can extend `SymbolKind` with `Public` if that helps.
> 
> So I think there are quite some steps to get there but in the end it will be much cleaner.

When you say "use reference for both (also in the grammar)", I think I would just remove the following two things:
- https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L887-L889
- https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L592
- Then in PIL, I will no longer use the colon in `:out` to refer to a public reference, so that they will be parsed as `GenericReference` in the future: https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L619-L622
- Because it looks like `GenericReference` contains the full `SymbolPath`, after `PublicDeclaration` has a resolved absolute path, I should be able to match it to its reference via the absolute path, and therefore no unique public ID is needed?

Another related question is that I can't find a 1 to 1 mapping between the following two chunks:
- https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L587-L604
- https://github.com/powdr-labs/powdr/blob/main/ast/src/parsed/mod.rs#L575-L593
- For example, `GenericReference` in the first link became `Reference` in the second. Am I missing some intermediate steps?

I also have a few basic questions that would really help clarify the context:
- Would you mind providing a short definition for the differences among `Symbol`, `Polynomial`, and `Expression`? 
- Also what different IDs (e.g. `PolyID` but there might be more) we use across the pipeline?

Thanks!

qwang98:
I was able to gradually learn/work my way through the pipeline. Currently I'm half way (hopefully) through a first pass and nothing compiles yet. I made two primary changes:
1. Remove `parsed::Expression::PublicReference` to use generic `parsed::Expression::Reference`.
2. Move `Analyzed::public_delcaration` inside `Analyzed::definitions` with the help of a new `SymbolKind::Public`.

I encountered four more questions:
1. I created a `parsed::FunctionDefinition` because it's required by `handle_symbol_definition`, because now `Analyzed::public_declaration` is moved inside `Analyzed::definitions`: https://github.com/powdr-labs/powdr/pull/2516/files#diff-15ed3db4f4d607867e3034476061efe4db196479f9f50b18b4eebcb530c4ba1bR141-R146. Given this change, I'm not sure if I should edit `powdr.lalrpop`, which defines `FunctionDefinition`: https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L151-L165. I think I might need to change the lalrpop definition for `PublicDeclaration` to `public <identifier> = <FunctionDefinition>`: https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L144-L149, similar to how `PolynomialCommitDeclaration` is defined: https://github.com/powdr-labs/powdr/blob/main/parser/src/powdr.lalrpop#L181-L195.
2. A basic type is required for `handle_symbol_definition` but I'm not sure what to put for public declaration: https://github.com/powdr-labs/powdr/pull/2516/files#diff-15ed3db4f4d607867e3034476061efe4db196479f9f50b18b4eebcb530c4ba1bR134-R140. I don't really see a specific fit for public declaration: https://github.com/powdr-labs/powdr/blob/main/ast/src/parsed/types.rs#L16-L42.
3. Should we dispense any public id any more for public declaration now that it's a symbol definition? I think it might be good to just use the symbol ID for looking up its value down the pipeline as it seems to be a unique ID.
4. Should `FunctionValueDefinition::PublicDeclaration` have any children? I think the referenced polynomial column should be the only one, but am not sure if it's appropriate as its "children" and what is this function used for: https://github.com/powdr-labs/powdr/pull/2516/files#diff-75fd3abe93ba950534e60dc6078b10557719e35da881b0126dd0f2bb564c11dbR809-R811.

I have a few more "general questions" to help understand the code base:
1. What is `Type` used for in the pipeline and why do we need to assign one to each `Symbol`?
2. I'll probably work further down `AlgebraicExpression` tomorrow, so would you mind explaining how is it different from `analyzed::Expression` and why is it needed?

pipeline/tests/executor.rs
```diff
@@ -1,35 +0,0 @@
...
-
-    namespace Machine2(8);
-        col witness x, y;
-        y = x + 1;
-
-        // x will be 42 on all rows
-        x - :public = 0;
-    "#;
-    run_witgen_pil::<GoldilocksField>(src);
-}
```
qwang98, line 35: This entire test is removed, because public references across different machines can no longer be confused. They now all have absolute names instead of the local reference name. In this example, they become `Machine1::public` and `Machine2::public`. Now that public reference is just a reference to public declaration, publics must be declared first (so a dangling `:public` is no longer possible). This test is therefore obsolete.

ast/src/analyzed/mod.rs
```diff
@@ -799,6 +806,9 @@ impl Children<Expression> for FunctionValueDefinition {
             FunctionValueDefinition::TypeConstructor(_, variant) => variant.children(),
             FunctionValueDefinition::TraitDeclaration(trait_decl) => trait_decl.children(),
             FunctionValueDefinition::TraitFunction(_, trait_func) => trait_func.children(),
+            // not sure if public declaration should have no children or have the polynomial it refers to as the only child
+            // how is the children function used in our pipeline?
+            FunctionValueDefinition::PublicDeclaration(pub_decl) => pub_decl.children(),
```
qwang98, line 811: Here's a question

ast/src/analyzed/display.rs
```diff
@@ -95,6 +95,15 @@ impl<T: Display> Display for Analyzed<T> {
                                 format_witness_column(&name, symbol, definition),
                             )?,
                             SymbolKind::Poly(PolynomialType::Intermediate) => unreachable!(),
+                            SymbolKind::Public() => {
+                                if let FunctionValueDefinition::PublicDeclaration(decl) =
+                                    definition.as_ref().unwrap()
```
chriseth, line 100: Does this work?
```suggestion
                                if let Some(FunctionValueDefinition::PublicDeclaration(decl) = definition
```

ast/src/analyzed/mod.rs
```diff
@@ -122,7 +120,13 @@ impl<T> Analyzed<T> {
     }
     /// @returns the number of public inputs
     pub fn publics_count(&self) -> usize {
-        self.public_declarations.len()
+        self.definitions
+            .iter()
+            .filter_map(move |(_name, (symbol, _))| match symbol.kind {
```
chriseth, line 125: What about using `filter()` and then `count()`?

ast/src/analyzed/mod.rs
```diff
@@ -559,6 +560,10 @@ pub fn type_from_definition(
                     ty: trait_func.ty.clone(),
                 })
             }
+            FunctionValueDefinition::PublicDeclaration(_) => {
+                // I'm not sure what to do with the type of publicdeclaration and how is this function used
```
chriseth, line 564: I think this is good.

ast/src/analyzed/mod.rs
```diff
@@ -856,6 +867,15 @@ impl PublicDeclaration {
     }
 }
 
+impl Children<Expression> for PublicDeclaration {
+    fn children(&self) -> Box<dyn Iterator<Item = &Expression> + '_> {
+        Box::new(empty())
```
chriseth, line 872: I think to make this proper, we should return an expression that contains the reference. It could also help to remove some special cases when we count references from publics. But to do that, we nee to change PublicDeclaration to contain an `Expression`, and we should somehow assert that it has a certain structure.

ast/src/parsed/display.rs
```diff
@@ -555,7 +555,10 @@ impl Display for FunctionDefinition {
                 )
             }
             FunctionDefinition::Expression(e) => write!(f, " = {e}"),
-            FunctionDefinition::TypeDeclaration(_) | FunctionDefinition::TraitDeclaration(_) => {
+            // not sure if we need to print public declaration definition anywhere
```
chriseth, line 558: I think it would be better to put the formatting function here instead.

pil-analyzer/src/statement_processor.rs
```diff
@@ -508,47 +512,61 @@ where
...
+            Some(FunctionValueDefinition::PublicDeclaration(
+                PublicDeclaration {
+                    id,
+                    source: source.clone(),
+                    name: absolute_name,
+                    polynomial: self
+                        .expression_processor(&Default::default())
+                        .process_namespaced_polynomial_reference(polynomial)
+                        .map_err(|err| source.with_error(err))?,
+                    array_index: array_index
```
chriseth, line 540: Can you extract some of those into local variables? It is a bit hard to read this way, in my opinion.

pipeline/tests/executor.rs
chriseth: Wait but you can still reference a "global" public from all machines, right?

qwang98: I think so, as long as you can reference a "global" column from all machines, because public declarations are now added to definitions and public references are the same as any other references.

ast/src/analyzed/display.rs
qwang98: Done.

ast/src/analyzed/display.rs
```diff
@@ -95,6 +95,15 @@ impl<T: Display> Display for Analyzed<T> {
...
+                                {
+                                    writeln_indented(f, format_public_declaration(&name, decl))?;
+                                } else {
+                                    unreachable!() // public symbol should always have a public declaration
```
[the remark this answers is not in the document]
qwang98: Alright noted. Currently public references are defined and solved via constraints. However, I think we might still need to "bind" public to a specific column, e.g. `public x = y;`, where `y` is a column? The plan was to remove the trace row number for sure, e.g. `public x = y(N)`.

ast/src/analyzed/mod.rs
qwang98: Done.

ast/src/analyzed/mod.rs
qwang98: Done.

ast/src/analyzed/mod.rs
```diff
@@ -771,6 +776,7 @@ impl Symbol {
 pub enum SymbolKind {
     /// Fixed, witness or intermediate polynomial
     Poly(PolynomialType),
+    Public(),
```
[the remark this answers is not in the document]
qwang98: Done.

pil-analyzer/src/statement_processor.rs
qwang98: Done.

ast/src/analyzed/mod.rs
qwang98: Done. I changed PublicDeclaration to contain Expression instead of PolynomialReference. I also asserted that it must be Expression::Reference.

ast/src/parsed/display.rs
qwang98: Done

chriseth:
Can you re-add the test file you removed? I think it was not about confusing different publics for the same, it was about assigning different values to the same public, and that is still possible.

ast/src/analyzed/mod.rs
```diff
@@ -852,27 +845,43 @@ pub struct PublicDeclaration {
     pub id: u64,
     pub source: SourceRef,
     pub name: String,
-    pub polynomial: PolynomialReference,
```
chriseth, line 844: What I meant was (sorry for not being concrete enough) to replace all of `polynomial`, `array_index` and `index` by a single `Expression`, and then assume that `Expression` is always a `FunctionCall` with a single argument that is a concrete number. I think this complication about `array_index` is also not really needed, as long as the inner expression resolves to a single column.

ast/src/analyzed/display.rs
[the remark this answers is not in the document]
chriseth: I thought the idea was (but not sure if I understood it correctly) to make publics their own thing, i.e. not tied to a specific cell inside the trace, not even tied to a specific column. If the backend supports it and the declaration shows which cell to use, the public can be tied to a specific cell inside an existing witness column. On the other hand, if the backend supports publics as their own entities, we can use `public p;` without a definition.

ast/src/analyzed/mod.rs
qwang98: Ok, so basically inside the `FuncionCall`, `function` is the polynomial, and the only `argument` is the `index` in `Expression::Number` form?

Regarding `array_index`, I think it's still needed in some cases like this test: https://github.com/powdr-labs/powdr/blob/main/test_data/pil/fib_arrays.pil#L16. In this case, `x[1]` should resolve to a single column, but I'm not sure what data types it's converted to throughout the parsed and analyzed pipeline.

```
pub struct FunctionCall<E = Expression<NamespacedPolynomialReference>> {
    pub function: Box<E>,
    pub arguments: Vec<E>,
}
```

ast/src/analyzed/display.rs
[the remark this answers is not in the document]
qwang98: I see your point. Yes. I think the eventual plan is to declare `public p;` and then use constraint (public reference) to calculate their values, e.g. `ISLAST * (witness_y - p) = 0`.

Currently backends only support the trace cell style public declaration, but witgen calculates public reference values alongside the witness (but are not used in the backend), so we would need to change the backend, which is what I've been working on but found many roadblocks for (including this PR).

ast/src/analyzed/mod.rs
chriseth: Yes exactly, we still need the ability to access elements of arrays, but for that we just have a generic `Expression`, which in turn can be an `Expression::IndexAccess`.

qwang98: Done

qwang98:
> Can you re-add the test file you removed? I think it was not about confusing different publics for the same, it was about assigning different values to the same public, and that is still possible.



> Can you re-add the test file you removed? I think it was not about confusing different publics for the same, it was about assigning different values to the same public, and that is still possible.

Done

pipeline/tests/executor.rs
```diff
@@ -12,24 +12,22 @@ fn run_witgen_pil<T: FieldElement>(pil: &str) -> Arc<Columns<T>> {
...
     // public, but assign different values to it.
     let src = r#"
     namespace Machine1(4);
         col witness y;
         [ 42, y ] in [ Machine2.x, Machine2.y ];
-
         // y will be 43 on all rows
-        y - :public = 0;
-
+        public pub = y(3);
```
qwang98, line 24: Note that this is required because we need to declare public to refer to it.

ast/src/analyzed/display.rs
```diff
@@ -245,11 +249,11 @@ fn format_witness_column(
 fn format_public_declaration(name: &str, decl: &PublicDeclaration) -> String {
     format!(
         "public {name} = {}{}({});",
-        decl.polynomial,
```
chriseth, line 248: Can't you just directly format `decl.polynomial` now?

ast/src/analyzed/mod.rs
```diff
@@ -836,22 +844,85 @@ impl Children<Expression> for NamedType {
...
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Hash, PartialEq, Eq)]
 pub struct PublicDeclaration {
     pub id: u64,
     pub source: SourceRef,
     pub name: String,
-    pub polynomial: PolynomialReference,
-    pub array_index: Option<usize>,
-    /// The evaluation point of the polynomial, not the array index.
-    pub index: DegreeType,
+    pub polynomial: Expression,
```
chriseth, line 852: Could also be called `value` or something else.

ast/src/analyzed/mod.rs
```diff
@@ -836,22 +844,85 @@ impl Children<Expression> for NamedType {
...
+        let name = self.referenced_poly().name.clone();
+        let index = self.referenced_poly_array_index();
+        if let Some(index) = index {
+            format!("{name}[{index}]")
+        } else {
+            name
+        }
+    }
+
+    pub fn index(&self) -> DegreeType {
```
chriseth, line 897: Maybe rename to `row`?

ast/src/analyzed/mod.rs
```diff
@@ -836,22 +844,85 @@ impl Children<Expression> for NamedType {
...
+                }
+            }
+            _ => panic!("Expected FunctionCall."),
+        }
+    }
+}
+
+impl Children<Expression> for PublicDeclaration {
+    fn children(&self) -> Box<dyn Iterator<Item = &Expression> + '_> {
+        match &self.polynomial {
```
chriseth, line 913: Can you just use `self.polynomial.children()`? Or just return `self.polynomial`?

ast/src/parsed/mod.rs
```diff
@@ -738,6 +736,20 @@ impl<Ref> From<u32> for Expression<Ref> {
         BigUint::from(value).into()
     }
 }
+
+impl<Ref> From<u64> for Expression<Ref> {
```
chriseth, line 740: Are these really needed?

ast/src/parsed/mod.rs
```diff
@@ -1388,6 +1402,7 @@ impl Children<Expression> for FunctionDefinition {
             FunctionDefinition::Expression(e) => Box::new(once(e)),
             FunctionDefinition::TypeDeclaration(_enum_declaration) => todo!(),
             FunctionDefinition::TraitDeclaration(trait_declaration) => trait_declaration.children(),
+            FunctionDefinition::PublicDeclaration(_, _, _) => todo!(),
```
chriseth, line 1405: Is it fine to leave it like that?

chriseth: I'm not sure this code is actually used.

pil-analyzer/src/statement_processor.rs
```diff
@@ -141,7 +130,19 @@ where
...
             }
             PilStatement::PublicDeclaration(source, name, polynomial, array_index, index) => {
-                self.handle_public_declaration(source, name, polynomial, array_index, index)
+                let (name, ty) = self.name_and_type_from_polynomial_name(
+                    PolynomialName {
+                        name,
+                        array_size: None,
+                    },
+                    Type::Expr,
+                )?;
```
chriseth, line 139:
```suggestion
```
This doesn't really do anything here because we know it is not an array (array of publics that would be)

pil-analyzer/src/statement_processor.rs
```diff
@@ -508,47 +513,92 @@ where
...
-    fn handle_public_declaration(
+    fn process_public_declaration(
         &mut self,
-        source: SourceRef,
-        name: String,
-        poly: parsed::NamespacedPolynomialReference,
+        symbol: Symbol,
+        polynomial: NamespacedPolynomialReference,
         array_index: Option<parsed::Expression>,
         index: parsed::Expression,
```
chriseth, line 521: Can we call this `row` to reduce confusion?

pipeline/tests/executor.rs
```diff
@@ -12,24 +12,22 @@ fn run_witgen_pil<T: FieldElement>(pil: &str) -> Arc<Columns<T>> {
 }
 
 #[test]
-#[should_panic = "Publics are referenced by more than one machine: public"]
+#[should_panic = "Publics are referenced by more than one machine: Machine1::pub"]
```
chriseth, line 15: Nice!

chriseth:
I went through the whole PR now. I added quite some comments, but I think they are all easy to address.

ast/src/analyzed/mod.rs
```diff
@@ -836,22 +844,85 @@ impl Children<Expression> for NamedType {
...
 pub struct PublicDeclaration {
     pub id: u64,
     pub source: SourceRef,
     pub name: String,
-    pub polynomial: PolynomialReference,
-    pub array_index: Option<usize>,
-    /// The evaluation point of the polynomial, not the array index.
-    pub index: DegreeType,
+    pub polynomial: Expression,
+    // pub array_index: Option<usize>,
```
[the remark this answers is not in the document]
qwang98: Done.

ast/src/analyzed/display.rs
qwang98: Done. Quote clever because IndexedAccess and FunctionCall should automatically have the square bracket and circle bracket syntax. :)

ast/src/analyzed/mod.rs
qwang98: Done

ast/src/analyzed/mod.rs
qwang98: Done. Also renamed `index` local variables in `process_public_declarations` and other places when I see them, to avoid confusion.

ast/src/analyzed/mod.rs
```diff
@@ -836,22 +844,85 @@ impl Children<Expression> for NamedType {
...
+                    Expression::Number(_, index) => Some(index.value.clone().try_into().unwrap()),
+                    _ => panic!("Expected Number."),
+                },
+                _ => panic!("Expected Reference or IndexAccess."),
+            },
+            _ => panic!("Expected FunctionCall."),
+        }
+    }
+
     pub fn referenced_poly_name(&self) -> String {
```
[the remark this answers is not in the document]
qwang98: Done

ast/src/analyzed/mod.rs
qwang98: Done.

ast/src/parsed/mod.rs
qwang98: Not necessarily after I changed some number types else where in the code, so these are removed.

ast/src/parsed/mod.rs
qwang98: I don't think it's actually used, because todo! will panic. However, I'd propose we do this in a follow up PR, to put the three `_, _, _` to a separate type and create proper children functions for it.

backend-utils/src/lib.rs
```diff
@@ -188,11 +188,10 @@ fn split_by_namespace<F: FieldElement>(
...
     // Publics end up in the empty namespace, but we can find the correct namespace by looking
     // at the referenced polynomial.
-    let public_to_namespace = pil
-        .public_declarations
-        .iter()
+    let _public_to_namespace = pil
+        .public_declarations_in_source_order()
         .map(|(public_name, public)| {
-            let ns = extract_namespace(&public.polynomial.name);
+            let ns = extract_namespace(&public.referenced_poly().name);
```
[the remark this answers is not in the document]
qwang98: I removed this entire chunk because `public_to_namespace` is no longer required. It's previously used to lookup `PublicDeclaration` name to namespace to split them to different machines. Now, it's simply a part of `Definitions`.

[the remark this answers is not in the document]
qwang98: I also see your point now, because `PublicDeclaration` now contain the absolute name, thanks to namespace resolution via `Symbol`, so even if we need this mapping in the future, we can directly extract it from the public declaration itself.

pil-analyzer/src/statement_processor.rs
```diff
@@ -141,7 +130,19 @@ where
...
+                        array_size: None,
+                    },
+                    Type::Expr,
+                )?;
+                let value = Some(FunctionDefinition::PublicDeclaration(
+                    polynomial,
+                    array_index,
+                    index,
+                ));
+                self.handle_symbol_definition(source, name, SymbolKind::Public(), None, ty, value)
```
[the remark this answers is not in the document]
qwang98: Done

pil-analyzer/src/statement_processor.rs
qwang98: Done and tried my best to catch them else where.

qwang98:
Thanks for the review and they should be all fixed for a review_final_final. :) @chriseth

chriseth (approved):

PR #2523, opened 2025-03-05
build bus using counter

ShuangWu121:
changing the bus_id build from hash to constant

### DIFF
--- a/linker/src/bus.rs
+++ b/linker/src/bus.rs
@@ -75,15 +75,26 @@ impl LinkerBackend for BusLinker {
                     (indices, sizes)
                 },
             );
-
+        // generate a map of interaction id for each link.to
         let interaction_id_map: BTreeMap<LinkTo, u32> = graph
             .objects
-            .values()
-            .flat_map(|object| object.links.iter().map(|link| link.to.clone()))
-            .collect::<HashSet<_>>()
             .iter()
+            .flat_map(|(location, object)| {
+                object
+                    .operations
+                    .keys()
+                    .map(move |operation_name| (location, operation_name.clone()))
+            })
             .enumerate()
-            .map(|(id, link_to)| (link_to.clone(), id as u32))
+            .map(|(id, (location, operation_name))| {
+                (
+                    LinkTo {
+                        machine: location.clone(),
+                        operation: operation_name.clone(),
+                    },
+                    id as u32,
+                )
+            })
             .collect();
 
         Ok(Self {


linker/src/bus.rs
```diff
@@ -86,6 +75,21 @@ impl LinkerBackend for BusLinker {
...
+        // generate a map of interaction id for each link.to
+        let mut id_counter = 0;
+
+        let interaction_id_map: HashMap<LinkTo, u32> = graph
+            .objects
+            .values()
+            .flat_map(|object| object.links.iter().map(|link| link.to.clone()))
+            .collect::<std::collections::HashSet<_>>()
+            .into_iter()
+            .map(|link_to| {
```
Schaeff, line 87: If we use `flat_map().enumerate()` then we don't need the global `id_counter`

linker/src/bus.rs
```diff
@@ -86,6 +75,21 @@ impl LinkerBackend for BusLinker {
...
             );
+        // generate a map of interaction id for each link.to
+        let mut id_counter = 0;
+
+        let interaction_id_map: HashMap<LinkTo, u32> = graph
+            .objects
+            .values()
+            .flat_map(|object| object.links.iter().map(|link| link.to.clone()))
+            .collect::<std::collections::HashSet<_>>()
+            .into_iter()
```
Schaeff, line 86: Iterating on a `HashSet` does not maintain the order, is it fine to do this here? Could we instead iterate on the operations of each machine? `graph.objects.values().flat_map(|o| o.operations).enumerate(|(global_operation_id, operation)| ...)`

Schaeff (commented):
Looks good! The tests probably need to be adjusted

linker/src/bus.rs
```diff
@@ -27,7 +27,7 @@ pub struct BusLinker {
     /// arguments for `bus_multi_linker`
     bus_multi_linker_args: ArrayLiteral,
     /// interaction id map for each link.to
-    interaction_id_map: HashMap<LinkTo, u32>,
+    interaction_id_map: BTreeMap<(Operation, Location), u32>,
```
Schaeff, line 30: iirc `LinkTo` has the location and operation, so it seems clearer to use that as you started with!

linker/src/bus.rs
```diff
@@ -76,18 +76,20 @@ impl LinkerBackend for BusLinker {
...
+            .flat_map(|(location, object)| {
+                object
+                    .operations
+                    .iter()
+                    .map(move |(operation_name, operation)| {
+                        (location, operation_name.clone(), operation.clone())
+                    })
+            })
+            .enumerate()
+            .map(|(id, (location, _operation_name, operation))| {
```
Schaeff, line 91: `_operation_name` isn't used so could be removed here and above

Schaeff: And we can create a `LinkTo`

linker/src/bus.rs
ShuangWu121: The order doesn't matter here since later a BTreeMap is created, and link.to is used as the index.

linker/src/bus.rs
ShuangWu121: Done

linker/src/bus.rs
```diff
@@ -75,22 +75,15 @@ impl LinkerBackend for BusLinker {
...
-                    .map(move |(operation_name, operation)| {
-                        (location, operation_name.clone(), operation.clone())
-                    })
-            })
             .enumerate()
-            .map(|(id, (location, _operation_name, operation))| {
-                ((operation.clone(), location.clone()), id as u32)
-            })
+            .map(|(id, link_to)| (link_to.clone(), id as u32))
             .collect();
```
Schaeff, line 87: Why did we revert this? We need one id per operation, so the clearest imo it to go through all operations as before this commit, and construct the `LinkTo` from that: a LinkTo uniquely identifies an operation

ShuangWu121: sorry, got a misunderstanding before. Now it uses LinkTo to index, and the list is created from going through operations.

Schaeff (approved):

PR #2532, opened 2025-03-10
Better optimization of intermediate polynomials

Schaeff:
Before this PR, we different optimizers for the cases:
- `col i = constant` -> inline `i`
- `w_0 = w_1` for two witnesses, only keep `w_0`
- `w = lin`, replace witness `w` with intermediate, only if lin is NOT a single witness, to avoid overlapping with the previous optimizer

After this PR, we have:
- `w = lin`, replace witness `w` (without the edge case above)
- `col i = e` if `e` is a number or a polynomial reference -> inline `i`

This reduces complexity in the optimizer, and also produces more optimized code, as seen in the tests.

One "breaking" change is that now when many witness are constrained to be equal, the kept on is not the first, but the last, as shown below:

```
col witness x;
col witness y;
x = y; 
// `y` is linear, therefore we replace the identity by
col x = y;
// `y` is a number of a polynomial reference, therefore we replace usages of x by y
```

`x` does disappear.

This works for larger sets of equal-constrained witnesses, as shown in the tests.


pilopt/src/lib.rs
```diff
@@ -769,23 +768,24 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
     substitute_polynomial_references(pil_file, constant_polys);
 }
 
-/// Identifies intermediate columns that are constrained to a single value, replaces every
-/// reference to this column by the value and deletes the column.
-fn remove_constant_intermediate_columns<T: FieldElement>(pil_file: &mut Analyzed<T>) {
+/// Inlines `col i = e` into the references to `x` where `e` is a non-shifted expression with no operations.
```
chriseth, line 771:
```suggestion
/// Inlines `col i = e` into the references to `i` where `e` is a non-shifted expression with no operations.
```
?

pilopt/src/lib.rs
```diff
@@ -769,23 +768,24 @@ fn remove_constant_witness_columns<T: FieldElement>(pil_file: &mut Analyzed<T>)
...
             let mut symbols_and_definitions = symbol.array_elements().zip_eq(definitions);
             match symbol.is_array() {
                 true => None,
                 false => {
-                    let ((name, poly_id), definition) = symbols_and_definitions.next().unwrap();
-                    match definition {
-                        AlgebraicExpression::Number(value) => {
+                    let ((name, poly_id), value) = symbols_and_definitions.next().unwrap();
+                    match value {
+                        AlgebraicExpression::Number(_) | AlgebraicExpression::Reference(_) => {
```
chriseth, line 784: Where do you check for "no shift"?

gzanitti (approved):
less complexity and better results, nice! :+1:

PR #2534, opened 2025-03-10
Autoprecompiles: BusInteractionKind

gzanitti:
Temporary hack for autoprecompiles that allows us to track the kind of bus interaction after powdr optimization.


autoprecompiles/src/lib.rs
```diff
@@ -95,12 +93,6 @@ impl<T: Clone + Ord + std::fmt::Display> SymbolicBusInteraction<T> {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Hash)]
-pub enum BusInteractionKind {
-    Send,
-    Receive,
-}
-
```
leonardoalt, line 103: I think we should still have this here otherwise users of this crate will need to depend on `ast` too

autoprecompiles/src/lib.rs
```diff
@@ -916,20 +892,25 @@ fn symbolic_machine_to_pilfile<P: FieldElement>(symbolic_machine: SymbolicMachin
...
-                acc.join(sym)
-            });
+        let path = SymbolPath::from_parts(
+            ["std", "prelude", "Constr", "BusInteraction"]
+                .into_iter()
+                .map(|p| Part::Named(p.to_string())),
+        );
+
+        let kind_expr = match interaction.kind {
+            BusInteractionKind::Send => BigUint::from(0u32).into(),
```
leonardoalt, line 902: I think we should keep this as 1 here and -1 below for consistency

pil-analyzer/src/condenser.rs
```diff
@@ -796,12 +796,24 @@ fn to_constraint<T: FieldElement>(
...
             multiplicity: to_expr(&fields[0]),
-            bus_id: to_expr(&fields[1]),
-            payload: ExpressionList(match fields[2].as_ref() {
+            kind: match to_expr(&fields[1]) {
+                AlgebraicExpression::Number(n) => {
+                    if n == T::zero() {
+                        BusInteractionKind::Send
+                    } else if n == T::one() {
+                        BusInteractionKind::Receive
+                    } else {
```
leonardoalt, line 805: 1 and -1

autoprecompiles/src/lib.rs
gzanitti: I assume you meant `powdr-openvm`. I leave both and do the transformation here in `autoprecompiles`, this way `powdr-openvm` doesn't need the one in `ast` (and I avoid having to import `BusInteractionKind` from `autoprecompiles` to `ast` which is also weird).

PR #2535, opened 2025-03-11
PoC for two passes of pilopt

qwang98:
After discussing with @Schaeff, we formulated a plan:
1. Currently, our pipeline is `compile pilopt tune pilopt witgen0 prove0 witgen1 prove1 finalize`, where `prove0 witgen1 prove1 finalize` is backend aware.
2. In this PR, `tune pilopt witgen0 prove0 witgen1 prove1 finalize` are all backend aware, meaning that a backend argument is required for these steps to run.

I made some specific implementation decisions:
3. In this PR, the `tune` step is a `BackendFactory` API, while all other APIs remain the same places as before. For example, `witgen0` is backend aware because it requires a `backend_tuned_pil` to run. However, `witgen0` only runs differently because different backend tuned pil's are provided by different backend, not *directly* because of the different backend (it has the same algorithm regardless of the backend). Therefore, while `witgen0` requires a backend to run, I didn't put it under `BackendFactory`. Design choices like this should ensure maximum backward compatibility.

Summary of all other changes on the high level, mostly fixing pipeline test and CI incompatibilities:

CLI changes:
4. `powdr-rs` requires a backend for witgen.
5. `powdr pil` only runs witgen if a backend is provided.

Test crates changes:
6. For tests that don't require a backend but require witgen, use the mock backend to run witgen.
7. For tests that require a backend and also require witgen but runs witgen before setting the backend, modify the test to set the backend before running witgen.

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -179,7 +179,7 @@ pub trait BackendFactory<F: FieldElement> {
         Err(Error::NoSetupAvailable)
     }
 
-    fn tune_analyzed_pil(&self, pil: Analyzed<F>) -> Analyzed<F> {
+    fn specialize_pil(&self, pil: Analyzed<F>) -> Analyzed<F> {
         // TODO: currently defaults to the identity function
         // Move `bus_multi_linker` calls here in the future
         pil
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -178,6 +178,12 @@ pub trait BackendFactory<F: FieldElement> {
     fn generate_setup(&self, _size: DegreeType, _output: &mut dyn io::Write) -> Result<(), Error> {
         Err(Error::NoSetupAvailable)
     }
+
+    fn tune_analyzed_pil(&self, pil: Analyzed<F>) -> Analyzed<F> {
+        // TODO: currently defaults to the identity function
+        // Move `bus_multi_linker` calls here in the future
+        pil
+    }
 }
 
 /// Dynamic interface for a backend.
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -179,7 +179,7 @@ pub trait BackendFactory<F: FieldElement> {
         Err(Error::NoSetupAvailable)
     }
 
-    fn tune_analyzed_pil(&self, pil: Analyzed<F>) -> Analyzed<F> {
+    fn tune_analyzed_pil(&self, mut pil: Analyzed<F>) -> Analyzed<F> {
         // TODO: currently defaults to the identity function
         // Move `bus_multi_linker` calls here in the future
         pil
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -179,7 +179,7 @@ pub trait BackendFactory<F: FieldElement> {
         Err(Error::NoSetupAvailable)
     }
 
-    fn tune_analyzed_pil(&self, mut pil: Analyzed<F>) -> Analyzed<F> {
+    fn tune_analyzed_pil(&self, pil: Analyzed<F>) -> Analyzed<F> {
         // TODO: currently defaults to the identity function
         // Move `bus_multi_linker` calls here in the future
         pil
--- a/backend/src/plonky3/stark.rs
+++ b/backend/src/plonky3/stark.rs
@@ -20,6 +20,8 @@ use powdr_plonky3::{
     ProverData, StarkProvingKey, StarkVerifyingKey, TableProvingKey, TableProvingKeyCollection,
 };
 
+use crate::BackendType;
+
 use p3_uni_stark::StarkGenericConfig;
 
 pub struct Plonky3Prover<T: FieldElementMap>
@@ -349,8 +351,10 @@ mod tests {
         ProverData<F>: Send + serde::Serialize + for<'a> serde::Deserialize<'a>,
         Commitment<F>: Send,
     {
-        let mut pipeline = Pipeline::<F>::default().from_pil_string(pil.to_string());
-        let pil = pipeline.compute_optimized_pil().unwrap();
+        let mut pipeline = Pipeline::<F>::default()
+            .with_backend_type(BackendType::Plonky3)
+            .from_pil_string(pil.to_string());
+        let pil = pipeline.compute_backend_tune_pil().unwrap();
         let witness_callback = pipeline.witgen_callback().unwrap();
         let witness = &mut pipeline.compute_witness().unwrap();
         let fixed = pipeline.compute_fixed_cols().unwrap();
--- a/backend/src/plonky3/stark.rs
+++ b/backend/src/plonky3/stark.rs
@@ -20,8 +20,6 @@ use powdr_plonky3::{
     ProverData, StarkProvingKey, StarkVerifyingKey, TableProvingKey, TableProvingKeyCollection,
 };
 
-use crate::BackendType;
-
 use p3_uni_stark::StarkGenericConfig;
 
 pub struct Plonky3Prover<T: FieldElementMap>
@@ -330,7 +328,7 @@ mod tests {
 
     use super::Plonky3Prover;
     use powdr_number::{BabyBearField, GoldilocksField, Mersenne31Field};
-    use powdr_pipeline::Pipeline;
+    use powdr_pipeline::{BackendType, Pipeline};
     use test_log::test;
 
     use powdr_plonky3::{Commitment, FieldElementMap, ProverData};
@@ -352,9 +350,9 @@ mod tests {
         Commitment<F>: Send,
     {
         let mut pipeline = Pipeline::<F>::default()
-            .with_backend_type(BackendType::Plonky3)
+            .with_backend(BackendType::Plonky3, None)
             .from_pil_string(pil.to_string());
-        let pil = pipeline.compute_backend_tune_pil().unwrap();
+        let pil = pipeline.compute_backend_tuned_pil().unwrap();
         let witness_callback = pipeline.witgen_callback().unwrap();
         let witness = &mut pipeline.compute_witness().unwrap();
         let fixed = pipeline.compute_fixed_cols().unwrap();
--- a/backend/src/plonky3/stark.rs
+++ b/backend/src/plonky3/stark.rs
@@ -326,7 +326,6 @@ where
 #[cfg(test)]
 mod tests {
 
-    use super::Plonky3Prover;
     use powdr_number::{BabyBearField, GoldilocksField, Mersenne31Field};
     use powdr_pipeline::{BackendType, Pipeline};
     use test_log::test;
@@ -352,25 +351,17 @@ mod tests {
         let mut pipeline = Pipeline::<F>::default()
             .with_backend(BackendType::Plonky3, None)
             .from_pil_string(pil.to_string());
-        let pil = pipeline.compute_backend_tuned_pil().unwrap();
-        let witness_callback = pipeline.witgen_callback().unwrap();
-        let witness = &mut pipeline.compute_witness().unwrap();
-        let fixed = pipeline.compute_fixed_cols().unwrap();
 
-        let mut prover = Plonky3Prover::new(pil, fixed);
-        prover.setup();
-        let proof = prover.prove(witness, witness_callback);
-
-        assert!(proof.is_ok());
+        let proof = pipeline.compute_proof().unwrap().clone();
 
         if let Some(publics) = malicious_publics {
-            prover
+            pipeline
                 .verify(
-                    &proof.unwrap(),
-                    &publics
+                    &proof,
+                    &[publics
                         .iter()
                         .map(|i| F::from(*i as u64))
-                        .collect::<Vec<_>>(),
+                        .collect::<Vec<_>>()],
                 )
                 .unwrap()
         }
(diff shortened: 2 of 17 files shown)


backend/src/lib.rs
```diff
@@ -178,6 +178,12 @@ pub trait BackendFactory<F: FieldElement> {
     fn generate_setup(&self, _size: DegreeType, _output: &mut dyn io::Write) -> Result<(), Error> {
         Err(Error::NoSetupAvailable)
     }
+
+    fn tune_analyzed_pil(&self, pil: Analyzed<F>) -> Analyzed<F> {
+        // TODO: currently defaults to the identity function
+        // Move `bus_multi_linker` calls here in the future
+        pil
+    }
```
qwang98, line 186: This will be an API for BackendFactory instead of Backend, because constructing Backend requires input witness etc., which aren't yet known at this point. BackendFactory is tied to the command line backend argument, which is appropriate for backend aware tuning which this API will include in the future.

pipeline/src/pipeline.rs
```diff
@@ -966,12 +966,19 @@ impl<T: FieldElement> Pipeline<T> {
...
-        self.maybe_write_pil_object(&optimized, "_opt")?;
+        self.log("Optimizing pil (pre backend tuning)...");
+        let optimized_pre_tuning_pil = powdr_pilopt::optimize(analyzed_pil);
+        let optimized_post_tuning_pil = if let Some(backend_type) = self.arguments.backend {
+            let factory = backend_type.factory::<T>();
+            let backend_tuned_pil = factory.tune_analyzed_pil(optimized_pre_tuning_pil);
+            powdr_pilopt::optimize(backend_tuned_pil)
+        } else {
+            optimized_pre_tuning_pil
+        };
```
qwang98, line 977: If no backend argument is provided in the pipeline, this will simply return the same PIL (one pilopt pass instead of two).

backend/src/lib.rs
Schaeff: That makes sense I think!

qwang98:
> I'm wondering if it would make sense to make this a separate pipeline step / artefact

I also discussed this with @Schaeff and sounds good to us. :)

qwang98:
This new version is also a PoC and will fail tests. The work flow is: pilopt -> backend specific tuning -> pilopt -> witgen and etc. In this design, `compute_witness` (stage 0 witgen) will require knowing the backend type (there are stage 0 bus columns to generate for some backend types but not other types). This requirement is enforced by calling `compute_backend_tuned_pil` in `compute_witness`. However, this creates a new problem: many benchmarking/testing functions runs all the way till right before the backend. Previously, backend type is not required as stage 0 witgen is the same for all backends. Now, we would have to specify a backend type for these benchmarking/testing functions. I propose two solutions to address this:

1. Default these benchmarking/testing functions to a backend type that requires stage 0 bus column witgen. This will be a "conservative" estimate of our benchmarks, because it has longer witgen than backends that use native bus interaction.
2. Move bus witgen to the backend and run it as the first step of the backend. This removes the need for two passes of pilopt, because all "backend specific tuning" are now in the backend itself.

P.S. I made comments throughout the PR to mark these scenarios (which a lot more complicated than described above), so you can get a sense of the various cases we encounter here.

backend/src/plonky3/stark.rs
```diff
@@ -349,8 +349,10 @@ mod tests {
         ProverData<F>: Send + serde::Serialize + for<'a> serde::Deserialize<'a>,
         Commitment<F>: Send,
     {
-        let mut pipeline = Pipeline::<F>::default().from_pil_string(pil.to_string());
-        let pil = pipeline.compute_optimized_pil().unwrap();
+        let mut pipeline = Pipeline::<F>::default()
+            .with_backend(BackendType::Plonky3, None)
+            .from_pil_string(pil.to_string());
+        let pil = pipeline.compute_backend_tuned_pil().unwrap();
```
Schaeff, line 355: Looking at the whole function, could we implement it only using `pipeline`?
```
let proof = pipeline.prove();
```
etc instead of computing these things by hand

cli-rs/src/main.rs
```diff
@@ -417,6 +426,7 @@ fn execute<F: FieldElement>(
 ) -> Result<(), Vec<String>> {
     let mut pipeline = Pipeline::<F>::default()
         .from_asm_file(file_name.to_path_buf())
+        .with_backend(backend, None) // backend option is not needed because execute only runs till stage 0 witgen
```
Schaeff, line 429: Can you explain this?

Schaeff: Oh is it because we need to know which backend we're targeting, but since we're not creating a proof we don't need proof specific options?

Could we then split `with_backend` into two methods? The first one would choose the factory (that's the method `tune` is defined on!) and the second would pass the options which are passed when calling `create` to get the actual backend

pipeline/src/pipeline.rs
```diff
@@ -367,6 +370,14 @@ impl<T: FieldElement> Pipeline<T> {
         self
     }
 
+    pub fn with_backend_if_none(&mut self, backend: BackendType, options: Option<BackendOptions>) {
```
Schaeff, line 373: Could we instead make `Backend` non optional on `Pipeline` and make it default to `Mock`?

Schaeff: That would remove the comments about stuff panicking because the backend is not set

pipeline/src/test_util.rs
```diff
@@ -509,14 +510,25 @@ pub fn assert_proofs_fail_for_invalid_witnesses_pilcom(
...
-    pipeline.compute_witness().unwrap();
+        .from_file(resolve_test_file(file_name));
 
-    assert!(run_pilcom_with_backend_variant(pipeline.clone(), BackendVariant::Monolithic).is_err());
-    assert!(run_pilcom_with_backend_variant(pipeline, BackendVariant::Composite).is_err());
+    assert!(run_pilcom_with_backend_variant(
+        pipeline
+            .clone()
+            .with_backend(powdr_backend::BackendType::EStarkDump, None)
+            .set_witness(convert_witness(witness)),
```
Schaeff, line 521: Here we assume that the passed witness works for both the monolithic and composite variants?

pipeline/tests/asm.rs
```diff
@@ -298,23 +307,33 @@ fn vm_to_block_array() {
...
-        .collect::<BTreeMap<_, _>>();
-
-    // Spot-check some witness columns to have the expected length.
-    assert_eq!(witness_by_name["main::X"].len(), 128);
-    assert_eq!(witness_by_name["main_arith::y"].len(), 32);
-    assert_eq!(witness_by_name["main_memory::m_addr"].len(), 32);
-
-    test_mock_backend(pipeline_gl.clone());
-    test_plonky3_pipeline(pipeline_gl);
+        .for_each(|backend| {
```
Schaeff, line 317: Let's use a for loop instead of for_each

pipeline/tests/asm.rs
```diff
@@ -450,7 +469,7 @@ fn read_poly_files() {
             })
             .with_backend(BackendType::EStarkDump, None);
         pipeline.compute_witness().unwrap();
```
Schaeff, line 471: Not a change in this PR, but why do we need to explicitly compute the witness here? Wouldn't `compute_proof` call this internally?

pipeline/tests/mock_backend.rs
```diff
@@ -16,7 +16,8 @@ fn fibonacci_wrong_initialization() {
     // Initializes y with 2 instead of 1
     // -> fails `ISLAST * (y' - 1) = 0;` in the last row
     let f = "pil/fibonacci.pil";
-    let pipeline = make_simple_prepared_pipeline::<GoldilocksField>(f, LinkerMode::Bus);
+    let pipeline = make_simple_prepared_pipeline::<GoldilocksField>(f, LinkerMode::Bus)
+        .with_backend(powdr_backend::BackendType::Mock, None);
```
Schaeff, line 20: These would also go away if `Mock` was the default

pipeline/tests/pil.rs
```diff
@@ -399,10 +402,10 @@ fn serialize_deserialize_optimized_pil() {
...
     let path = powdr_pipeline::test_util::resolve_test_file(f);
 
-    let optimized = powdr_pipeline::Pipeline::<powdr_number::Bn254Field>::default()
-        .from_file(path)
-        .compute_optimized_pil()
-        .unwrap();
+    let mut pipeline =
+        powdr_pipeline::Pipeline::<powdr_number::Bn254Field>::default().from_file(path);
+
+    let optimized = pipeline.compute_optimized_pil().unwrap();
```
Schaeff, line 408: This change doesn't seem to bring anything?

cli-rs/src/main.rs
```diff
@@ -161,6 +162,11 @@ enum Commands {
         #[arg(value_parser = clap_enum_variants!(FieldArgument))]
         field: FieldArgument,
 
+        /// Generate a proof with a given backend.
```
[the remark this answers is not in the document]
qwang98: This comment incorrect and fixed. :)

pipeline/src/pipeline.rs
```diff
@@ -971,21 +982,49 @@ impl<T: FieldElement> Pipeline<T> {
...
+            return Ok(backend_tuned_pil.clone());
+        }
+
+        self.compute_optimized_pil()?;
+
+        let backend_type = self.arguments.backend.expect("no backend selected!");
+
+        // If backend option is set, take the optimized pil from memory.
+        // Then, compute and add the backend-tuned pil to artifacts and return backend-tuned pil.
+        let optimized_pil = self.artifact.optimized_pil.take().unwrap();
```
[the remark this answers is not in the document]
qwang98: Yes. `optimized_pil` is not cached anymore because `take()` is used here. I can also make a separate cached copy instead.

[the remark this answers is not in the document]
qwang98: I think caching makes more sense, because I foresee future test cases for multiple backends, so that we can cache the optimized pil before the backend tuning diverge to multiple backend tuned pils.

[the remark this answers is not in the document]
qwang98: Done.

cli-rs/src/main.rs
qwang98: Exactly :).

backend/src/plonky3/stark.rs
qwang98: Done.

cli-rs/src/main.rs
qwang98: The two methods is a good idea. However, there are actually 30+ `with_backend` changes in this PR and they can require case by case judgements, so to avoid piling up potential CI breaks, I'm doing these changes in another PR that you'll see.

pipeline/src/test_util.rs
qwang98: Yes. I also checked that this function is mostly used in simple tests where you can specify the witness "by hand" instead of having our witgen pipeline compute it. For example: https://github.com/powdr-labs/powdr/blob/main/pipeline/tests/pil.rs#L26-L48

I think in theory it's good to break it up to two functions, especially if more complex cases arise and we shall need two different witnesses. I'm happy to do it in another PR (because it's a different topic) and it should be quite simple to do. :)

pipeline/tests/asm.rs
qwang98: Done.

pipeline/tests/asm.rs
qwang98: Exactly and that's the same I noticed in many places as well. One speculation is that we've edited the pipeline to call the prior step that the current step depends on. This is a backward compatible change but we didn't edit all tests which previously call each step one by one.

qwang98: So I agree that this is redundant and should be removed for clarity in another PR.

pipeline/tests/pil.rs
qwang98: `pipeline` needs to be materialized so that `optimized` can persist. This change is needed after I modified `compute_optimized_pil` to return a reference instead of a clone.

backend/src/lib.rs
```diff
@@ -178,6 +178,12 @@ pub trait BackendFactory<F: FieldElement> {
     fn generate_setup(&self, _size: DegreeType, _output: &mut dyn io::Write) -> Result<(), Error> {
         Err(Error::NoSetupAvailable)
     }
+
+    fn tune_analyzed_pil(&self, pil: Analyzed<F>) -> Analyzed<F> {
```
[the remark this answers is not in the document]
qwang98: I think `specialize_pil`, so that we know it's a backend factory function for pil instead of for backend, to avoid confusion with `create`.

pipeline/src/pipeline.rs
qwang98: `with_backend` and `Backend` related additional changes in another PR.

pipeline/tests/mock_backend.rs
qwang98: Done in another PR.

pipeline/src/test_util.rs
Schaeff: Cool not high priority

cli-rs/src/main.rs
Schaeff: excellent

Schaeff (approved):

Schaeff (approved):

PR #2551, opened 2025-03-14
Rename machines to main to avoid dispatch.

chriseth:


georgwiese (dismissed):

test_data/asm/single_function_vm.asm
```diff
@@ -1,7 +1,10 @@
-machine VM with degree: 8 {
+machine Main with degree: 8 {
     reg pc[@pc];
+    reg X[<=];
+    reg A;
 
     function main {
+        A <=X= A + 1;
```
chriseth, line 7: Since the dispatcher is removed, without this, the machine would be empty, i.e. the row count is zero. This breaks some assumptions somewhere.

georgwiese (approved):

PR #2555, opened 2025-03-17
Function definition public declaration

qwang98:
Small one, just the Children implementation.

### DIFF
--- a/ast/src/parsed/mod.rs
+++ b/ast/src/parsed/mod.rs
@@ -524,35 +524,6 @@ impl<R> Children<Expression<R>> for TraitDeclaration<Expression<R>> {
 #[derive(
     Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize, JsonSchema, Hash,
 )]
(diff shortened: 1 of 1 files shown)


ast/src/parsed/mod.rs
```diff
@@ -1389,7 +1389,9 @@ impl Children<Expression> for FunctionDefinition {
             FunctionDefinition::Expression(e) => Box::new(once(e)),
             FunctionDefinition::TypeDeclaration(_enum_declaration) => todo!(),
             FunctionDefinition::TraitDeclaration(trait_declaration) => trait_declaration.children(),
-            FunctionDefinition::PublicDeclaration(_, _, _) => todo!(),
+            FunctionDefinition::PublicDeclaration(_, array_index, row) => {
+                Box::new(array_index.iter().chain(once(row)))
+            }
```
qwang98, line 1394: Should be good as it matches what we did here: https://github.com/powdr-labs/powdr/blob/main/ast/src/parsed/mod.rs#L220

chriseth (approved):

PR #2569, opened 2025-03-20
Split compile function and use interpreter for large code.

chriseth:
This first derives the code and then makes the decision of whether or not to use the interpreter based on the derived code.

Depends on #2582

### DIFF
--- a/executor/src/witgen/jit/function_cache.rs
+++ b/executor/src/witgen/jit/function_cache.rs
@@ -18,10 +18,14 @@ use crate::witgen::{
 use super::{
     block_machine_processor::BlockMachineProcessor,
     compiler::{compile_effects, CompiledFunction},
+    effect::Effect,
     interpreter::EffectsInterpreter,
+    prover_function_heuristics::ProverFunction,
     variable::Variable,
     witgen_inference::CanProcessCall,
 };
+
+const MAX_COMPILED_CODE_SIZE: usize = 500;
 
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 struct CacheKey<T: FieldElement> {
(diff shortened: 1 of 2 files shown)


executor/src/witgen/jit/function_cache.rs
```diff
@@ -18,11 +18,15 @@ use crate::witgen::{
...
     block_machine_processor::BlockMachineProcessor,
     compiler::{compile_effects, CompiledFunction},
+    effect::Effect,
     interpreter::EffectsInterpreter,
+    prover_function_heuristics::ProverFunction,
     variable::Variable,
     witgen_inference::CanProcessCall,
 };
 
+const MAX_COMPILED_CODE_SIZE: usize = 500;
```
pacheco, line 28: might be good to leave a small comment here for usage/context

executor/src/witgen/jit/function_cache.rs
```diff
@@ -212,25 +209,45 @@ impl<'a, T: FieldElement> FunctionCache<'a, T> {
...
             .iter()
             .enumerate()
             .filter_map(|(i, b)| if b { Some(Variable::Param(i)) } else { None })
             .collect::<Vec<_>>();
 
+        let has_prover_function_call = has_prover_function_call(&result.code);
+        // Use the compiler for goldilocks with at most MAX_COMPILED_CODE_SIZE statements and
+        // the interpreter otherwise.
+        let interpreted = matches!(T::known_field(), Some(KnownField::GoldilocksField))
+            || (code_size(&result.code) > MAX_COMPILED_CODE_SIZE && !has_prover_function_call);
```
pacheco, line 233: `has_prover_function_call` should not be tested here due to the following `if`

chriseth: I changed it again. The idea is that this test is where we want to get to.

pacheco (dismissed):

executor/src/witgen/jit/block_machine_processor.rs
```diff
@@ -570,7 +570,9 @@ params[3] = main_binary::C[3];"
     #[test]
     fn poseidon() {
         let input = read_to_string("../test_data/pil/poseidon_gl.pil").unwrap();
-        generate_for_block_machine(&input, "main_poseidon", 12, 4).unwrap();
+        generate_for_block_machine(&input, "main_poseidon", 12, 4)
+            .map_err(|e| eprintln!("{e}"))
```
pacheco, line 574: leftover `eprintln` or intended?

chriseth: I thought about removing it, but why not keep it?

pacheco (approved):

PR #2570, opened 2025-03-20
Fix clippy

lfgtwo:


leonardoalt:
Thanks for the PR, but we fixed clippy in a different PR together with some other adjustments we had to do because of Rust's new version.

PR #2574, opened 2025-03-21
Fix pilcom proof fail test functions (reapprove)

qwang98:
Incorrectly merged to a non-main branch previously, so needs a reapproval.

Depends on https://github.com/powdr-labs/powdr/pull/2535 and https://github.com/powdr-labs/powdr/pull/2557. Split assert_proofs_fail_for_invalid_witnesses_pilcom into assert_proofs_fail_for_invalid_witnesses_pilcom_composite and assert_proofs_fail_for_invalid_witnesses_pilcom_monolithic. These two APIs are needed because we might generate different witnesses for pilcom composite vs pilcom monolithic under two scenarios:

1. The two backends can have different witnesses to start with. This hasn't been an issue because most of our test cases are very simple machines.
2. After updates in https://github.com/powdr-labs/powdr/pull/2535, if we eventually implement different optimizations for the two different backends, they might compute two different witnesses.


Schaeff:
@qwang98 this might not be needed after #2575 ?

PR #2575, opened 2025-03-21
Remove pilcom

Schaeff:
- remove the `connect_no_witgen` test
- use the mock prover instead, when applicable


riscv/tests/riscv.rs
```diff
@@ -94,7 +94,7 @@ fn bn254_sanity_check() {
         &[],
         Default::default(),
     );
-    run_pilcom_with_backend_variant(pipeline, BackendVariant::Composite).unwrap();
+    test_mock_backend(pipeline, BackendVariant::Composite);
```
Schaeff, line 97:
```suggestion
    test_mock_backend(pipeline);
```

pipeline/src/verify.rs
```diff
@@ -1,42 +0,0 @@
...
-    if result.is_err() {
-        log::error!(
-            "Pil verifier run was unsuccessful.\nStdout: {}\nStderr: {}\n",
-            output,
-            String::from_utf8_lossy(&verifier_output.stderr)
-        );
-    }
-
-    result
-}
```
qwang98, line 42: I wonder why we had this file to start with. Quite odd to me that it's called `verify` but appears in `pipeline` instead of `backend`.

pipeline/tests/asm.rs
```diff
@@ -817,14 +816,6 @@ fn keccak() {
...
 }
 
-#[test]
-fn connect_no_witgen() {
-    let f = "asm/connect_no_witgen.asm";
-    let pipeline = make_simple_prepared_pipeline::<GoldilocksField>(f, LinkerMode::Native);
-    // TODO Mock prover doesn't support this test yet.
-    test_pilcom(pipeline);
-}
-
```
qwang98, line 827: Seems that this isn't used anywhere after this deletion (except auto compiled reparse_tests). Any concerns?

qwang98: Obviously we can't do much if we can't support it with mock prover.

qwang98:
High level question: How is Pilcom similar to our Mock prover? Why is it needed initially but no longer?

qwang98 (dismissed):
LGTM except some potential concern about `connect_no_witgen.asm`.

pipeline/src/verify.rs
leonardoalt: This was used to verify that the witness satisfies the constraints

pipeline/tests/asm.rs
leonardoalt: `TODO yet`, so maybe leave the test but comment it out?

leonardoalt:
> High level question: How is Pilcom similar to our Mock prover? Why is it needed initially but no longer?

It's very similar. In essence it does the same thing, only checks that a given witness satisfies the constraints. We used it at first to keep our PIL a superset of Polygon's, and Pilcom comes from their tooling. After we diverged we needed our own constraint checker, so we built the Mock prover.

pipeline/tests/asm.rs
Schaeff: Done

.github/workflows/nightly-tests.yml
```diff
@@ -77,12 +71,10 @@ jobs:
...
     - name: Install EStarkPolygon prover dependencies
       run: sudo apt-get update && sudo apt-get install -y nlohmann-json3-dev libpqxx-dev nasm libgrpc++-dev libprotobuf-dev protobuf-compiler-grpc uuid-dev build-essential cmake pkg-config git
-    - name: Install pilcom
-      run: git clone https://github.com/0xPolygonHermez/pilcom.git  && cd pilcom && npm install
     - name: Check without Halo2
       run: cargo check --all --no-default-features
     - name: Build
       run: cargo build --all --release --all-features
     - name: Run tests
       # Number threads is set to 1 because the runner does not have enough memory for more.
```
chriseth, line 79: Is this still the case without pilcom?

.github/workflows/pr-tests.yml
```diff
@@ -231,8 +207,6 @@ jobs:
         run: rustup component add rust-src --toolchain nightly-2024-08-01-x86_64-unknown-linux-gnu
       - name: Install riscv target
         run: rustup target add riscv32imac-unknown-none-elf --toolchain nightly-2024-08-01-x86_64-unknown-linux-gnu
-      - name: Install pilcom
-        run: git clone https://github.com/0xPolygonHermez/pilcom.git  && cd pilcom && npm install
       - uses: taiki-e/install-action@nextest
       - name: Run slow tests
         # Number threads is set to 2 because the runner does not have enough memory for more.
```
chriseth, line 212: Same here, is this still the case without pilcom?

.github/workflows/nightly-tests.yml
Schaeff: removed

.github/workflows/pr-tests.yml
Schaeff: removed

qwang98 (approved):

Issue #2543, opened 2025-03-13
Add Logup-GKR support


ShuangWu121:
# Logup-GKR Integration in STARK Provers  

## Context  
I have been researching how different implementations integrate Logup-GKR with STARK proofs. My focus is on two implementations:  

- **Miden VM**, which uses **[Winterfell](https://github.com/facebook/winterfell/tree/logup-gkr)** (Facebook’s STARK prover).  
- **Stwo Logup-GKR** implementation.  

## Current Progress  

The main challenge is understanding how these implementations use **a univariate polynomial IOP (STARK) to integrate GKR**, which requires a **multivariate polynomial IOP**.  

- **Miden VM** provides a clear approach to this integration (detailed in [[this doc](https://github.com/0xPolygonMiden/miden-vm/issues/1182)] and [[this doc](https://hackmd.io/@plafer/ByuGXywwC)]). However, its GKR and related APIs are from Winterfell ([logup-gkr branch](https://github.com/facebook/winterfell/tree/logup-gkr)), making direct application to the Stwo backend difficult.   - This may not be a blocker, as Stwo already implements GKR and sumcheck, meaning integration could be possible with modifications.  

- **Stwo's GKR implementation** is clear for proving a **logup lookup** alone, but its integration with STARK is less clear.  
  - One key aspect is understanding the **"univariate IOP for multilinear evaluation at a point"** in this [example](https://github.com/starkware-libs/stwo/blob/305cabc2b06dd86aa1af678af1064dbc038323a8/crates/prover/src/examples/xor/gkr_lookups/mle_eval.rs#L47), which seems central to the integration.  
  - My current research is focused on understanding this component.  

## Next Steps  
I will continue investigating this and update the issue as I gain a clearer picture. Any insights or related references are welcome!

PR #2579, opened 2025-03-22
Stwo logup-gkr, proving a single lookup

ShuangWu121:
This PR adds logup-gkr support for stwo backend. 

Only supports a single lookup operation for now.

more descriptions will be added later.

### DIFF
--- a/pipeline/tests/pil.rs
+++ b/pipeline/tests/pil.rs
@@ -283,12 +283,6 @@ fn stwo_gkr_lookup_invalid() {
 }
 
 #[test]
-fn stwo_bus_lookup() {
-    let f = "pil/bus_lookup.pil";
-    test_stwo(f, Default::default());
-}
-
-#[test]
 fn stwo_stage1_publics() {
     let f = "pil/stage1_publics.pil";
     test_stwo_stage1_public(


leonardoalt:
@georgwiese @Schaeff @ShuangWu121 do we still want to merge this?

Schaeff:
@ShuangWu121 could you please explain the state of this PR? What is supported and what isn't? Maybe updating the description.

ShuangWu121:
> @ShuangWu121 could you please explain the state of this PR? What is supported and what isn't? Maybe updating the description.

here #2543 I wrote more descriptions. 
The GKR method is able to prove single lookup and multi-lookup **without** dynamic machine size now. But I did some initial tests of GKR performance, it is much slower than none-GKR when stwo uses Blake2 as hash function, it is better when stwo use Poseidon, the same as mentioned by Miden VM Logup-gkr implementation benchmark result. but I didn't do a vigorous performance test, as now the logup-GKR use BusInteration, but pil still generate bus constraints, stage 1 witness, accumulators etc. and in addition, it proves BusInteration using logup-GKR.  To complete this PR, it would need generating BusInteraction Identity without pil generating stage 1 accumulators, thus needs PR #2629 #2631 #2636 and maybe witgen (I remembered @qwang98 mentioned witgen might not sound without accumulators ...). 

Getting it to work with dynamic machine sizes will require more effort, and the final performance is still uncertain. So this PR might be paused for now, since we're currently focusing on pre-compiles?

leonardoalt:

PR #2580, opened 2025-03-24
Optimize PR test workflow

qwang98:
This PR makes 16 GitHub Workflow jobs each with cargo nextest to run on 4 threads (max allowed for ubuntu for GitHub hosted servers), so that each thread can run maximum 1 of the 55 tests.

We are using two workflow runners:
1. One is from provider "warp build", for jobs: build, bench, and run_examples. I assume that this is because these jobs requrie compiling to x86 architecture, which might not be supported by the GitHub Workflow runners? These runners seem to have associated cost with them, so I think we should be careful when optimizing the `run_examples` job, which is a current bottleneck of 38 minutes. It can be optimized by creating more runners to run the 5 examples in parallel, depending on the timing of each. (Need to calculate associated cost, not sure if it's charged by runner or by running minutes).
2. The other is from GitHub Workflow runner, which seem to have 4 cores for the ubuntu instances that we are using. They are also free for public repos. I'm not sure how many concurrent runners are supported as it's unlimited and free in theory, so I'm allocating 16 concurrent runners for the slow tests. The test_quick job uses 1 runner.

We have 55 slow tests, with the following timing distribution:

PASS [  21.329s] powdr-pipeline::powdr_std poseidon2_gl_test
PASS [  25.665s] powdr-pipeline::powdr_std rotate_large_test
PASS [   4.357s] powdr-pipeline::powdr_std split_bb_test
PASS [  11.462s] powdr-pipeline::powdr_std split_gl_test
PASS [   7.930s] powdr-pipeline::powdr_std rotate_small_test
PASS [ 510.098s] powdr-riscv::riscv halt
PASS [  52.306s] powdr-riscv::riscv many_chunks_dry
PASS [ 637.928s] powdr-riscv::riscv keccak
PASS [ 332.218s] powdr-riscv::riscv vec_median
PASS [   6.723s] powdr-pipeline::pil block_lookup_or_permutation
PASS [  19.041s] powdr-pipeline::powdr_std arith256_small_test
PASS [ 471.428s] powdr-pipeline::powdr_std arith256_memory_large_test
PASS [   3.878s] powdr-pipeline::powdr_std arith_small_test
PASS [  88.260s] powdr-riscv::riscv output_syscall
PASS [ 564.755s] powdr-pipeline::powdr_std arith_large_test
PASS [ 552.897s] powdr-riscv::riscv std_hello_world
PASS [ 531.372s] powdr-riscv::riscv sum_serde
[...]

PR #2581, opened 2025-03-24
Optimize PR test workflow (Simple starter version)

qwang98:
This PR increases the number of `test_slow` PR test bins from 8 to 17, thereby strictly reducing the time spent on `test_slow`.

We have three types of runners: 
- `warp-ubuntu-2404-x64-8x` for: build, run_examples, bench. I think this is a paid server: https://www.warpbuild.com/, which @leonardoalt last changed in #2187, so I have a question: why don't we use the GitHub Workflow free ones? Besides, I think we can parallelize `run_examples`, which is a bottleneck of 38 minutes currently run single threaded, so there can be some immediate time improvement if we run it on multiple runner instances. Is this server charged by the number of runners or the number of minutes?
- `ubuntu-24.04` for: test_quick, test_estark_polygon, test_slow.
- `ubuntu-22.04` for: udeps.

**BEFORE Optimization**
Screen Shot 2025-03-24 at 17 25 11

**AFTER Optimization**
See run time of this PR.

**Future Optimization**
In #2580, I'm working on creating 14 bins with 4 threads each because we have 55 tests in total and each can run on a separate thread. This would reduce the `test_slow` run time to that of the longest test, which is ~10 minutes.

This requires computing the bin-thread assignment because `nextest` only supports the hash method for partitions, which isn't ideal in this case.


.github/workflows/pr-tests.yml
```diff
@@ -235,10 +243,10 @@ jobs:
         run: git clone https://github.com/0xPolygonHermez/pilcom.git  && cd pilcom && npm install
       - uses: taiki-e/install-action@nextest
       - name: Run slow tests
-        # Number threads is set to 2 because the runner does not have enough memory for more.
+        # Number threads is set to 4 because the runner does not have enough memory for more.
         run: |
           NIGHTLY_TESTS=$(cat .github/workflows/nightly_tests_list.txt)
-          cargo nextest run --archive-file tests.tar.zst --workspace-remap . --verbose --run-ignored=ignored-only -E "!($NIGHTLY_TESTS)" --test-threads 2 --partition hash:"${{ matrix.test }}"/8
+          cargo nextest run --archive-file tests.tar.zst --workspace-remap . --verbose --run-ignored=ignored-only -E "!($NIGHTLY_TESTS)" --test-threads 4 --partition hash:"${{ matrix.test }}"/16
```
qwang98, line 249: According to this page, we can now run 4 threads instead of 2 on GitHub Workflow's standard server: 
https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

This was last updated in our repo 8 months ago.

.github/workflows/pr-tests.yml
```diff
@@ -204,6 +204,14 @@ jobs:
...
           - "7"
           - "8"
+          - "9"
+          - "10"
+          - "11"
+          - "12"
+          - "13"
+          - "14"
+          - "15"
+          - "16"
```
qwang98, line 214: These 16 runners are all run concurrently. I honestly have no idea what's the maximum allowed, because there's no official documentation. ChatGPT says GitHub allows a maximum of 20 concurrency runners for private repos and likely more for public ones, but I'm not sure if we should rely on that.

leonardoalt:
> so I have a question: why don't we use the GitHub Workflow free ones?

@qwang98 they're too slow and have too little RAM

leonardoalt:
> Is this server charged by the number of runners or the number of minutes?

Number of computation minutes, so it shouldn't matter how many runners we're using if we're faster/same.

leonardoalt:
>  I think we can parallelize run_examples, which is a bottleneck of 38 minutes currently run single threaded

Yea those are bad in general, because they don't re-use the build binaries. I think I/we need to re-think those tests anyway.

qwang98:
> > I think we can parallelize run_examples, which is a bottleneck of 38 minutes currently run single threaded
> 
> Yea those are bad in general, because they don't re-use the build binaries. I think I/we need to re-think those tests anyway.

I mean we can just create multiple bins for them to run, because we currently run 5 tests single threaded. (Unless our subscription doesn't support multiple concurrent runners).

qwang98:
> > I think we can parallelize run_examples, which is a bottleneck of 38 minutes currently run single threaded
> 
> Yea those are bad in general, because they don't re-use the build binaries. I think I/we need to re-think those tests anyway.

It's also not a bottleneck yet, because even with this PR, our bottleneck is `build` + `test_slow` (the latter depends on the former)...

qwang98:
> > so I have a question: why don't we use the GitHub Workflow free ones?
> 
> @qwang98 they're too slow and have too little RAM

Got it

qwang98:
I tested a few parameters and think 17 bins for test_slow is the best so far.

I think this can be merged for some immediate speedups.

Schaeff (approved):

PR #2584, opened 2025-03-24
Fix cache

lvella:
It seems WarpBuild removed the ability to overwrite a cache with the same key.

But since they allow for removing a cache, this fix simply removes the old one before saving the new one.

Please make sure the build cache test run was successful before merging.


.github/workflows/build-cache.yml
```diff
@@ -39,6 +41,22 @@ jobs:
 
     - name: Save date of cache build
       run: date -R -u > target/cache-build-date.txt
+
+    # It seems that in an attempt to be more compatible with GitHub, WarpBuild removed
+    # the ability to overwrite an old cache and they longer charge for cache storage.
```
chriseth, line 46: "no longer"?

lvella: No. I misread the billing statement. They are still charging it.

They just removed the ability to automatically replace a cache if I try to save with an identical key.

lvella: removed comment

Schaeff (approved):

PR #2594, opened 2025-03-27
CI status fix

qwang98:


qwang98:
Doesn't seem to work (trying to remove the forever pending status from test_quick and test_slow). They don't seem to affect merging though.

PR #2595, opened 2025-03-27
Remove name from public

Schaeff:


chriseth (approved):

qwang98 (approved):
FYI #2596 passes so I think it's good :)

Schaeff:
Running CI again since somehow the result was not being reported...

PR #2596, opened 2025-03-27
Test simplify pub

qwang98:
WILL NOT BE MERGED.

Just to run CI to test that #2595 is compatible with the public infrastructure.


qwang98:
It's just a test so now closed.

PR #2601, opened 2025-03-28
Powdr PIL runs witgen with mock backend by default

qwang98:
Now that witgen depends on backend type, `powdr pil` requires a backend type to run through witgen. It only runs till optimized pil if `--prove-with` is not provided.

To match the description, this PR defaults backend to Mock if `--prove-with` is not provided, so `powdr pil` can still run till witgen. To prove with any backend, `--prove-with` is still required.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -725,6 +725,12 @@ fn run<F: FieldElement>(
             .with_backend(backend, backend_options.clone())
             .compute_proof()
             .unwrap();
+    } else {
+        // If no backend is specified, we run witgen via Mock backend
+        pipeline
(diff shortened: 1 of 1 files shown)


Schaeff (approved):

cli/src/main.rs
```diff
@@ -105,6 +105,7 @@ struct Cli {
 #[derive(Subcommand)]
 enum Commands {
     /// Runs compilation and witness generation for .pil and .asm files.
+    /// Also runs backend if `--prove-with` is set.
```
chriseth, line 108: What does `runs backend` mean? Generates a proof? Also maybe put this at the end (such that it is in sequence of operations)

PR #2602, opened 2025-03-28
Fix column by name.

chriseth:


executor/src/witgen/mod.rs
```diff
@@ -533,17 +529,12 @@ impl<'a, T: FieldElement> FixedData<'a, T> {
...
 
+    /// Returns a PolyID given the name of it. Supports array elements
+    /// in that for `"X[7]"` it returns the PolyID of element 7 of the array `X`.
     pub fn try_column_by_name(&self, name: &str) -> Option<PolyID> {
         self.column_by_name.get(name).cloned()
     }
 
-    pub fn try_symbol_by_name(&self, name: &str) -> Option<&Symbol> {
-        self.analyzed
-            .definitions
```
chriseth, line 542: The bug was that it didn't look at the intermediate columns.

lvella (approved):

PR #2608, opened 2025-03-28
Prover functions for small arith.

chriseth:
Tests working for "small":
- memory
- add_sub
- rotate
- shift
- memory
- arith256

Not working:
- keccak
- poseidon
- poseidon2


std/machines/small_field/arith.asm
```diff
@@ -96,13 +48,32 @@ machine Arith(byte: Byte, byte2: Byte2) with
...
+                // Quotient is unconstrained, use zero for y1
+                // and set remainder x2 = y3.
+                [0, 0, 0, 0] + int_to_limbs(y3_value)
+            } else {
+                let quotient = y3_value / x1_value;
+                let remainder = y3_value % x1_value;
+                int_to_limbs(quotient) + int_to_limbs(remainder)
+            }
+        }
+    );
```
gzanitti, line 74: I have reviewed the mathematics behind all the computation and it makes sense (it is even clearer than before). What I don't quite understand is the use of this query function. I understand that the results are assigned in y3 + x1, but when is it “called” for that to happen?

chriseth: In the runtime solver, (all) query functions are executed when there is no more progress, I think. It works because `eval` silently fails if the value is not there yet. If all `eval` calls succeed, then the computed value is assigned to the target columns.

leonardoalt:
@chriseth I'm guessing we do want this soon?

PR #2610, opened 2025-03-31
Reparse fail sample

qwang98:
Sample that `analyze(parse(analyzed.to_string()))` fails.


chriseth:
Don't we have the reparse tests for that? Do you have an input (maybe forgot to commit)?

gzanitti:
Maybe it's related to [this issue](https://github.com/powdr-labs/powdr/issues/2052)?

leonardoalt:
Do we want to merge or fix this?

PR #2611, opened 2025-03-31
Native bus tuning

qwang98:
NOT READY FOR REVIEW. Now only the `block_to_block.asm` example works. I foresee lots of debugging.

### DIFF
--- a/pil-analyzer/src/type_inference.rs
+++ b/pil-analyzer/src/type_inference.rs
@@ -926,6 +926,11 @@ impl TypeChecker {
                     format_type_scheme_around_name(&name, &Some(inferred)),
                     format_type_scheme_around_name(&name, &Some(declared_type),
                 ))));
+                // println!(
+                //     "Inferred type scheme for symbol {name} does not match the declared type.\nInferred: let{}\nDeclared: let{}",
+                //     format_type_scheme_around_name(&name, &Some(inferred.clone())),
+                //     format_type_scheme_around_name(&name, &Some(declared_type.clone())),
+                // );
             }
             let declared_type_vars = declared_type.ty.contained_type_vars();
             let inferred_type_vars = inferred_type.contained_type_vars();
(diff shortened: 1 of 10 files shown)


pil-analyzer/src/type_inference.rs
```diff
@@ -915,28 +915,41 @@ impl TypeChecker {
...
+                //     "Inferred type scheme for symbol {name} does not match the declared type.\nInferred: let{}\nDeclared: let{}",
+                //     format_type_scheme_around_name(&name, &Some(inferred)),
+                //     format_type_scheme_around_name(&name, &Some(declared_type),
+                // ))));
+                // println!(
+                //     "Inferred type scheme for symbol {name} does not match the declared type.\nInferred: let{}\nDeclared: let{}",
+                //     format_type_scheme_around_name(&name, &Some(inferred.clone())),
+                //     format_type_scheme_around_name(&name, &Some(declared_type.clone())),
+                // );
+                // }
```
qwang98, line 936: Note that these are currently commented out, but to do it properly I'll eventually need to make two `analyze_ast` versions, one with this check for first pass analyze and the other without for second pass analyze.

Schaeff:
> Remove remove_unreferenced_definitions() step from optimize (first pass), because otherwise the optimizer will remove many required PIL definitions needed later which are no longer referenced by the linker.

Alternatively, we could inject the stdlib again in the tuning. This would have the advantage to keep the first and second pilopt passes identical and limit edge cases.

pil-analyzer/src/type_inference.rs
Schaeff: Maybe there's a way around this by going to string again? After injecting, we should have valid pil, so if the compiler cannot deal with it then we should fix that and not go around it?

Schaeff: Also if the issue is in the btree library, we could try removing that and see if it fixes it.

qwang98: > Maybe there's a way around this by going to string again? After injecting, we should have valid pil, so if the compiler cannot deal with it then we should fix that and not go around it?

I just tried going to string again via `let analyzed = powdr_pil_analyzer::analyze_string(&PILFile(all_statements).to_string()).unwrap();`. Unfortunately the same error happened.

leonardoalt:
@georgwiese @Schaeff do we still want to go ahead with this?

PR #2612, opened 2025-03-31
Perform bit decomposition as assignment if offset is known.

chriseth:
If we detect a bit decomposition pattern like `X + Y * 256 + Z * 65536 = T`, the witgen inference returns a bit decomposition effect. But if the value of `T` is a known number, we can actually extract the concrete values for `X`, `Y` and `Z` at compile-time and continue solving with this additional information.


executor/src/witgen/jit/block_machine_processor.rs
```diff
@@ -520,19 +520,19 @@ main_binary::operation_id_next[0] = main_binary::operation_id[1];
 call_var(9, 0, 0) = main_binary::operation_id_next[0];
 main_binary::operation_id_next[1] = main_binary::operation_id[2];
 call_var(9, 1, 0) = main_binary::operation_id_next[1];
-2**24 * main_binary::A_byte[2] + 2**0 * main_binary::A[2] := main_binary::A[3];
+2**0 * main_binary::A[2] + 2**24 * main_binary::A_byte[2] := main_binary::A[3];
```
Schaeff, line 523: These are just reordering. Is this ordering better? Could we change the code to maintain the original ordering?

executor/src/witgen/jit/block_machine_processor.rs
```diff
@@ -816,4 +816,68 @@ S::Z[0] = params[1];
...
+        ";
+        let code = format_code(&generate_for_block_machine(input, "S", 2, 1).unwrap().code);
+        assert_eq!(
+            code,
+            "\
+S::Y[0] = params[0];
+S::Z[0] = params[1];
+S::X[0] = 253;
+S::carry[0] = 1;
+params[2] = 1;"
```
Schaeff, line 848: noob: what is params here?

Schaeff (commented):
LGTM from an outsider perspective

executor/src/witgen/jit/block_machine_processor.rs
chriseth: I added code to sort by exponent. Before, it it was sorted by the order it appears in the constraint. Note that when we execute such a bit decomposition, we will sort it by exponent as well, so I think it's better to sort it like this already when creating the effect.

executor/src/witgen/jit/block_machine_processor.rs
chriseth: params is the array of bus receive parameters / lookup parameters. Special variables are needed because you could have complex expressions: `[a, b] in [x + 2, 2 * y]`. Assuming the first one is an input and the second one is an output, then we have:
`param[0]` is known -> `x = param[0] - 2`. At some point we compute `y` and then we can solve `param[1] = y / 2`, and that is what is returned from the submachine call.

Schaeff: Great thanks, that makes sense

Schaeff (approved):

PR #2620, opened 2025-04-01
session builder: create pipeline from ELF files

gballet:
Add a `pipeline_from_elf` method  to the session builder, in order to be capable of creating sessions from a non-rust ELF riscv32imc file.

The proposed change is to still use `guest_path` as an input function, and change the behavior based on what item is pointed at by the path:

 * if the path is a directory, it will try to compile it as a rust project
 * if the path is a file, it will try to interpret it as a riscv ELF file


gballet:
Tested and it worked (well, it failed on #2622, which means it loaded the file)

PR #2622, opened 2025-04-01
riscv: support lifting for S[BHL]U instructions

gballet:
While running my program, I got an error saying that sbu wasn't supported after auipc.

There is some code tasked to "lift" whatever instruction comes after auipc in order to merge all the immediate computations into a single immediate. SBU is currently missing from it.

### DIFF
--- a/riscv/src/elf/mod.rs
+++ b/riscv/src/elf/mod.rs
@@ -859,7 +859,7 @@ impl TwoOrOneMapper<MaybeInstruction, HighLevelInsn> for InstructionLifter<'_> {
 
                         HighLevelInsn { op, args, loc }
                     }
-                    // l{b|h|w} rd, symbol
+                    // l{b|h|w}[u] rd, symbol
                     Ins {
                         opc: l_op,
                         rd: Some(rd_l),
@@ -867,7 +867,7 @@ impl TwoOrOneMapper<MaybeInstruction, HighLevelInsn> for InstructionLifter<'_> {
                         rs2: None,
                         imm: Some(lo),
                         ..
-                    } if matches!(l_op, Op::LB | Op::LH | Op::LW)
+                    } if matches!(l_op, Op::LB | Op::LH | Op::LW | Op::LBU | Op::LHU | Op::LWU)
                         && rd_auipc == rd_l
                         && rd_l == rs1_l =>
                     {


riscv/src/elf/mod.rs
```diff
@@ -859,15 +859,15 @@ impl TwoOrOneMapper<MaybeInstruction, HighLevelInsn> for InstructionLifter<'_> {
...
+                    // l{b|h|w}[u] rd, symbol
                     Ins {
                         opc: l_op,
                         rd: Some(rd_l),
                         rs1: Some(rs1_l),
                         rs2: None,
                         imm: Some(lo),
                         ..
-                    } if matches!(l_op, Op::LB | Op::LH | Op::LW)
+                    } if matches!(l_op, Op::LB | Op::LH | Op::LW | Op::LBU | Op::LHU | Op::LWU)
```
lvella, line 870: I am not sure `LWU` should be included, as it is a RV64-only instruction and all the rest of the code assumes RV32.

lvella (changes requested):
Thanks. Left a comment.

riscv/src/elf/mod.rs
gballet: it's definitely rv32, my compiler generates it.

lvella: It is probably either `LBU` or `LHU`, because our codegen doesn't implement `LWU`, either, so this fix wouldn't be enough.

`LWU` is defined in section [4.3. Load and Store Instructions](https://riscv.github.io/riscv-isa-manual/snapshot/unprivileged/#_load_and_store_instructions) of the spec, under [4. RV64I Base Integer Instruction Set](https://riscv.github.io/riscv-isa-manual/snapshot/unprivileged/#rv64).

lvella:
Please amend the title and description to L[BH]U instead of S[BHL]U.

riscv/src/elf/mod.rs
gballet: totally misread your comment, yeah that's right, just B and H 👍

gballet:
```suggestion
                    } if matches!(l_op, Op::LB | Op::LH | Op::LW | Op::LBU | Op::LHU)
```

lvella (approved):

PR #2628, opened 2025-04-03
Plonky3 parameters check

ShuangWu121:
print plonky3 parameters info,

when the circuit constraint degree is less than the degree bound, print potential parameter optimization


leonardoalt:

PR #2629, opened 2025-04-03
Native bus tuning (condenser without reparsing route)

qwang98:
WIP. Currently still fails all tests but I'd just like to use this as a reference.


leonardoalt:
@georgwiese @Schaeff do we still want this?

leonardoalt:

PR #2637, opened 2025-04-08
Solving for inverse.

chriseth:
The main obstacle here is solving the following (`low_addr` is known, `inverse` is not):

`(inverse * (low_addr - 65532) - 1) * (low_addr - 65532) = 0;`

This is similar to https://github.com/powdr-labs/powdr/pull/2619 but not exactly (maybe we can still combine both):

We try to solve both branches. The RHS just gives us a range constraint `low_addr = 65532` - but `low_addr` was already known (not that this only  works for the "only concrete known is known").

On the LHS we can only do the division if we know that `low_addr - 65532` is not zero.

This is a slight problem with the new data structures, since we need to solve one branch in the "only concrete is known" version and the other branch in the "also symbolically known is known" version.

### DIFF
--- a/executor/src/witgen/jit/block_machine_processor.rs
+++ b/executor/src/witgen/jit/block_machine_processor.rs
@@ -936,4 +936,33 @@ S::carry[0] = 0;
 params[2] = 0;"
         );
     }
+
+    #[test]
+    fn addr_increment() {
+        let input = "
+namespace Main(256);
+    col witness a, b;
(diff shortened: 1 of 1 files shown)


chriseth:
We can probably solve in regular mode and then convert the symbolic expressions to QSEs by translating all references to unknown variables (known concrete variables will not appear anyway!)

leonardoalt:

PR #2655, opened 2025-04-18
Check out cache, set mtime and re-check out actual commit.

chriseth:
Instead of setting `mtime` to the committed date, which @lvella noted is error-prone, this PR does the following so that we can utilize the build cache:

- check out the commit the build cache was built for
- set mtime to the time the cache was built
- check out the commit we are running the tests on

This hopefully has the effect that all modified files will have the current time as mtime and will trigger re-builds correctly

### DIFF
--- a/.github/workflows/build-cache.yml
+++ b/.github/workflows/build-cache.yml
@@ -39,6 +39,8 @@ jobs:
 
     - name: Save date of cache build
       run: date -R -u > target/cache-build-date.txt
+    - name: Save commit hash of cache build
+      run: git rev-parse HEAD > target/cache-commit-hash.txt
 
     - name: Delete the old cache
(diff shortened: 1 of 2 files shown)


.github/workflows/build-cache.yml
```diff
@@ -16,6 +16,11 @@ jobs:
       with:
         submodules: recursive
 
+    - name: Save date of cache build
```
chriseth, line 19: We need to save it before we do the actual build so that if we later on set the mtime to that timestamp, the build artefacts will be slightly newer.

chriseth:
I guess this still needs some tuning for the examples (it still re-compiles a lot there), but I would say this is good to go already.

leonardoalt (approved):

PR #2656, opened 2025-04-18
Autoprecompiles: support kind implied from multiplicity

leonardoalt:
change autoprecompile optimizations for the case where bus interaction kind is not available but implied from the multiplicity

### DIFF
--- a/autoprecompiles/src/lib.rs
+++ b/autoprecompiles/src/lib.rs
@@ -20,6 +20,7 @@ use powdr_pil_analyzer::analyze_ast;
 use powdr_pilopt::optimize;
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, BTreeSet};
+use std::fmt::Display;
 
 use powdr_number::{BigUint, FieldElement, LargeInt};
 use powdr_pilopt::simplify_expression;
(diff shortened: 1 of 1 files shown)


autoprecompiles/src/lib.rs
```diff
@@ -768,7 +794,7 @@ pub fn generate_precompile<T: FieldElement>(
 
                 let is_valid: AlgebraicExpression<T> = exec_receive(&machine).mult.clone();
                 let one = AlgebraicExpression::Number(1u64.into());
-                local_constraints.push((is_valid.clone() - one).into());
+                local_constraints.push((is_valid.clone() + one).into());
```
leonardoalt, line 797: exec receives have `-expr` as the multiplicity now, so we make it `mult = -1`

autoprecompiles/src/lib.rs
```diff
@@ -560,14 +546,43 @@ pub fn optimize_precompile<T: FieldElement>(mut machine: SymbolicMachine<T>) ->
                 Some(addr) => addr,
             };
 
-            match mem_int.op {
-                MemoryOp::Read => {}
-                MemoryOp::Write => {
-                    last_store.insert(addr, i);
+            if receive {
+                match local_reg_mem.get(&addr) {
+                    Some(data) => {
```
leonardoalt, line 551: This block was unmodified, a loop below was removed but github diff is confused, as usual

PR #2657, opened 2025-04-20
Use column type instead of names in apc

Schaeff:
- use existing visitor traits (Children, AllChildren)
- introduce Column type which keeps name and id together
- introduce UniqueColumns trait to get all Columns in an AST node
- change API to return `Column -> Column` subs instead of `name -> name`

### DIFF
--- a/autoprecompiles/src/powdr.rs
+++ b/autoprecompiles/src/powdr.rs
@@ -131,7 +131,6 @@ pub fn substitute_algebraic_algebraic<T: Clone + std::cmp::Ord>(
     );
 }
 
-// @leo: This is where I'm not sure. I think after this change this function should return `Column -> Column`, which it does, but I'm not sure my implementation is correct
 pub fn append_suffix_algebraic<T: Clone>(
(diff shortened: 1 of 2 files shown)


autoprecompiles/src/powdr.rs
```diff
@@ -128,18 +131,24 @@ pub fn substitute_algebraic_algebraic<T: Clone + std::cmp::Ord>(
     );
 }
 
+// @leo: This is where I'm not sure. I think after this change this function should return `Column -> Column`, which it does, but I'm not sure my implementation is correct
```
Schaeff, line 134: @leonardoalt

leonardoalt: looks fine I think?

leonardoalt: the `expr` keeps its own poly id which is important, and the name is changed in place. The return is only for the substitution map returned later on, so looks fine

leonardoalt (approved):

PR #2659, opened 2025-04-22
Solve some quadratic constraints

chriseth:


constraint-solver/src/quadratic_symbolic_expression.rs
```diff
@@ -403,6 +425,100 @@ impl<T: FieldElement, V: Ord + Clone + Hash + Eq + Display> QuadraticSymbolicExp
...
+            }
+            (Ok(left), Ok(right)) => (left, right),
+        };
+
+        if let Some(result) =
+            combine_to_conditional_assignment(left_solution, right_solution, range_constraints)
+        {
+            Ok(result)
+        } else {
+            // TODO Now at least try to combine joint range constraints
```
chriseth, line 453: Done in https://github.com/powdr-labs/powdr/pull/2653

constraint-solver/src/quadratic_symbolic_expression.rs
```diff
@@ -897,6 +1013,41 @@ c = (((10 + Z) & 0xff000000) >> 24) [negative];
...
+        );
+    }
+
+    #[test]
+    fn solve_quadratic() {
+        let rc = RangeConstraint::from_mask(0xffu32);
+        let a = Qse::from_unknown_variable("a");
+        let b = Qse::from_known_symbol("b", rc.clone());
+        let range_constraints =
+            HashMap::from([("a", rc.clone()), ("b", rc.clone()), ("c", rc.clone())]);
```
georgwiese, line 1026:
```suggestion
            HashMap::from([("a", rc.clone()), ("b", rc.clone())]);
```
Doesn't seem to be needed?

constraint-solver/src/quadratic_symbolic_expression.rs
```diff
@@ -403,6 +425,100 @@ impl<T: FieldElement, V: Ord + Clone + Hash + Eq + Display> QuadraticSymbolicExp
...
+        .combine_sum(&RangeConstraint::from_value(diff))
+        .is_disjoint(&rc)
+    {
+        return None;
+    }
+
+    Some(ProcessResult {
+        effects: vec![Effect::ConditionalAssignment {
+            variable: first_var,
+            condition: BranchCondition {
```
georgwiese, line 505:
```suggestion
            condition: Condition {
```
Guess we should rename it?

constraint-solver/src/range_constraint.rs
```diff
@@ -244,6 +244,12 @@ impl<T: FieldElement> RangeConstraint<T> {
             None
         }
     }
+
+    /// Returns true if no value can satisfy both range constraints at the same time.
+    pub fn is_disjoint(&self, other: &RangeConstraint<T>) -> bool {
+        // We cannot use masks because they always allow zero.
+        interval_intersection((self.min, self.max), (other.min, other.max)).is_none()
```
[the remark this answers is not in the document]
chriseth: No, masks are never disjoint, because they always allow zero, see line 250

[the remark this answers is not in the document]
georgwiese: OK, but I mean `[0, 2**24] & 0xff00ff` and `[2**8, 2**16] & 0x00ff00`. They would be disjoint, yet `is_disjoint()` would return `false`.

georgwiese:
LGTM! Last comment is that `PartialEq` for `QuadraticSymbolicExpression` could be derived.

georgwiese (approved):

PR #2666, opened 2025-04-24
More apc clean up

Schaeff:


autoprecompiles/src/lib.rs
```diff
@@ -763,14 +763,14 @@ pub fn generate_precompile<T: FieldElement>(
     let mut bus_interactions: Vec<SymbolicBusInteraction<T>> = Vec::new();
     let mut col_subs: Vec<BTreeMap<Column, Column>> = Vec::new();
     let mut global_idx: u64 = 3;
-    let mut global_idx_subs: BTreeMap<Column, u64> = BTreeMap::new();
-    let mut global_idx_subs_rev: BTreeMap<u64, Column> = BTreeMap::new();
```
Schaeff, line 767: I don't think this reverse map is needed but I may be missing something

autoprecompiles/src/lib.rs
```diff
@@ -763,14 +763,14 @@ pub fn generate_precompile<T: FieldElement>(
     let mut bus_interactions: Vec<SymbolicBusInteraction<T>> = Vec::new();
     let mut col_subs: Vec<BTreeMap<Column, Column>> = Vec::new();
     let mut global_idx: u64 = 3;
-    let mut global_idx_subs: BTreeMap<Column, u64> = BTreeMap::new();
```
Schaeff, line 766: This map is not needed globally, but only by local instruction. Moved there.

autoprecompiles/src/powdr.rs
```diff
@@ -219,37 +164,37 @@ impl<'a, T: Clone + Ord + std::fmt::Display + 'a, E: AllChildren<AlgebraicExpres
     }
 }
 
-pub fn reassign_ids_algebraic<T: Clone + Ord>(
+/// Given an expression `expr`
+/// - Reassign the IDs of all references in the expression to be unique and larger or equal than `curr_id`
+/// - Add a suffix to the name of each reference, so that `foo` becomes `foo_suffix`
```
leonardoalt, line 169: I kinda feel like these should be kept separate, but fine to try this first and separate as needed in the future.

autoprecompiles/src/lib.rs
```diff
@@ -759,19 +760,34 @@ pub fn generate_precompile<T: FieldElement>(
...
+                name: name.to_string(),
+                id: PolyID {
+                    ptype: PolynomialType::Constant,
+                    id: global_idx,
+                },
+            };
+            global_idx += 1;
+            (col.clone(), col)
+        })
+        .collect();
```
leonardoalt, line 783: I think we actually don't need this rn

autoprecompiles/src/lib.rs
```diff
@@ -816,57 +832,36 @@ pub fn generate_precompile<T: FieldElement>(
...
+                            powdr::substitute_algebraic(&mut expr, &sub_map);
+                            powdr::substitute_algebraic(&mut expr, &sub_map_loadstore);
+                            let mut expr = simplify_expression(expr);
+                            global_idx =
+                                powdr::local_to_global(&mut expr, global_idx, &mut local_subs, i);
+                            expr
+                        };
+                        ControlFlow::Continue::<()>(())
+                    },
+                    VisitOrder::Pre,
```
leonardoalt, line 854: Nice

autoprecompiles/src/lib.rs
```diff
@@ -816,57 +832,36 @@ pub fn generate_precompile<T: FieldElement>(
...
+                );
+
+                result.constraints.extend(machine.constraints);
+                result.bus_interactions.extend(machine.bus_interactions);
                 col_subs.push(local_subs);
 
                 // after the first round of simplifying,
                 // we need to look for register memory bus interactions
                 // and replace the addr by the first argument of the instruction
```
leonardoalt, line 863: this sounds weird

autoprecompiles/src/lib.rs
```diff
@@ -0,0 +1,1213 @@
...
+                        AlgebraicExpression::Number(n) => n.to_integer().try_into_u32().unwrap(),
+                        _ => panic!(
+                            "Address space must be a constant but got {}",
+                            bus_int.args[0]
+                        ),
+                    };
+
+                    if addr_space != 1 {
+                        continue;
+                    }
```
Schaeff, line 879: is it correct that we run this on the entire apc so far, as opposed to on the added interactions, or a single time on the whole thing at the end?

autoprecompiles/src/lib.rs
```diff
@@ -756,142 +757,142 @@ pub fn optimize_exec_bus<T: FieldElement>(mut machine: SymbolicMachine<T>) -> Sy
...
+                    );
+
+                    apc.constraints.extend(machine.constraints);
+                    apc.bus_interactions.extend(machine.bus_interactions);
+                    col_subs.push(subs);
+
+                    // after the first round of simplifying,
+                    // we need to look for register memory bus interactions
+                    // and replace the addr by the first argument of the instruction
+                    for bus_int in &mut apc.bus_interactions {
```
Schaeff, line 854: here

autoprecompiles/src/lib.rs
```diff
@@ -756,142 +757,142 @@ pub fn optimize_exec_bus<T: FieldElement>(mut machine: SymbolicMachine<T>) -> Sy
...
-                    sub_map_loadstore.extend(loadstore_chip_info(&machine, instr.opcode));
-                }
+    let CompilationState { apc, col_subs, .. } = statements.iter().enumerate().fold(
+        Default::default(),
+        |CompilationState {
+             mut apc,
+             mut col_subs,
+             mut global_idx,
+         },
+         (i, instr)| {
```
leonardoalt, line 779: I'm personally not a fan of huge blocks inside closures because they're so far apart and find it rather hard to read.

leonardoalt: imo extracting the inner block into a function would look better

Schaeff: sure

leonardoalt (approved):

PR #2685, opened 2025-05-06
Some small changes to "substitute_by_unknown".

chriseth:


constraint-solver/src/quadratic_symbolic_expression.rs
```diff
@@ -208,22 +208,23 @@ impl<T: FieldElement, V: Ord + Clone + Hash + Eq> QuadraticSymbolicExpression<T,
...
+    ///
+    /// Note this does NOT work properly if the variable is used inside a
+    /// known SymbolicExpression.
     pub fn substitute_by_unknown(
         &mut self,
         variable: &V,
         substitution: &QuadraticSymbolicExpression<T, V>,
     ) {
-        if !self.referenced_variables().any(|v| v == variable) {
+        if !self.referenced_unknown_variables().any(|v| v == variable) {
```
chriseth, line 220: This should be much faster

constraint-solver/src/quadratic_symbolic_expression.rs
```diff
@@ -252,13 +253,7 @@ impl<T: FieldElement, V: Ord + Clone + Hash + Eq> QuadraticSymbolicExpression<T,
             })
             .collect();
 
-        if !to_add
```
chriseth, line 255: If it is zero, the addition function will more or less do the same checks here and adding zero also has no effect on a SymbolicExpression.

gzanitti (approved):

PR #2690, opened 2025-05-06
Restrict access to the constraint system.

chriseth:
This is the first step in allowing some optimizations, for example to store a lookup table for the occurrences of variables and only substituting in those constraints or keeping a queue of modified constraints.


chriseth:
This might need some changes in the autoprecompiles branch...

gzanitti (dismissed):

PR #2703, opened 2025-05-12
some bugfixes in autoprecompiles

leonardoalt:


chriseth (approved):
Not sure what native means but looks good otherwise!

leonardoalt:
> Not sure what native means but looks good otherwise!

It's used to represent FE words in memory natively vs byte limbs

PR #2715, opened 2025-05-15
Serde for constraint solver.

chriseth:


number/src/traits.rs
```diff
@@ -32,14 +32,15 @@ pub trait LargeInt:
     + Sub<Output = Self>
     + fmt::Display
     + fmt::Debug
-    + Copy
```
chriseth, line 35: duplicate

chriseth:
This should contain some mechanism to check that the finite field is the same at export and re-import time.

chriseth:
I'll add the mechanism to check the field once powdr-openvm is merged - then the SymbolicMachine should get `serialize` and `deserialize` functions where it also writes and reads the field.

pacheco (dismissed):

constraint-solver/src/range_constraint.rs
```diff
@@ -12,7 +13,8 @@ use powdr_number::{log2_exact, FieldElement, LargeInt};
 /// and bit masks. The actual constraint is the conjunction of the two.
 ///
 /// Note that the same constraint can have multiple representations.
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash, Serialize, Deserialize)]
+#[serde(bound(deserialize = ""))]
```
pacheco, line 17: what does this line do?

chriseth: The derive macro has some logic to find out what the correct trait bounds on generic parameters (like `T` here) are. This heuristic fails and `FieldElement` already requires serialize/deserialize, so everything is fine.

Schaeff:
@chriseth can you add a description here? The cbor file is not generated by this PR, so I assume this is part of a larger thing?

chriseth:
The cbor file is generated from a three line snippet in the powdr-openvm repo, that's why I didn't add it here. I want to create commandline tools in a later PR (that also adds the name of the field).

chriseth:
Ah ok, also let me rename the cbor file, it's very non-descriptive.

chriseth:
Ok, I just realized that I added serialize/deserialize for stuff that is not actually used here. Will fix this later.

chriseth:
Yeah, this PR is not really consistent but I think all parts are good in isolation. Could merge as is but also will not oppose if someone objects.

PR #2717, opened 2025-05-16
fix apc register memory optimizer bug

leonardoalt:
The bug that happened was that the last send of a register memory bus interaction was removed due to having multiplicity zero. In the end the bus was left unbalanced. Therefore, we need to run the solver optimizer before the reg mem opt, so that these multiplicities are removed in pairs before the apc optimizer runs and keeps only the first receive and last send, because those multiplicities need to match. This is the main solution implemented here in `build`. I also added some checks that the reg mem interactions need to be even per address and in total.

A side effect of this fix is that we can no longer easily remove the timestamp columns of removed mem int receives, the constraint that decomposes it into bytes, the decomposing columns, and the range checks on them. The pattern matching becomes too brittle and would be even hackier if we wanted to keep supporting it that way. That's what causes the regression in the test, and all the code that is removed in this PR, since we won't do this anymore. The ultimate fix is to rely on the optimizer to remove these columns, since:
- The reg mem int receive is removed
- Then the timestamp column becomes a free variable in its constraint
- The constraint can be removed
- The limb decomposition columns become unused in polynomial constraints and only appear in range checks
- Therefore the range checks and columns can also be removed

At this point the result should be the same as the pattern matching alternative removed by this PR.
I anticipate that we'll also need a fixpoint loop because it may be the case that we need a few iterations to propagate everything possible when this is implemented.

### DIFF
--- a/autoprecompiles/src/lib.rs
+++ b/autoprecompiles/src/lib.rs
@@ -524,10 +524,10 @@ pub fn exec_receive<T: FieldElement>(machine: &SymbolicMachine<T>) -> SymbolicBu
     r
 }
 
-// Check that the number of register memory bus interactions in the precompile is even.
-// The check is performed on the total number and per address.
+// Check that the number of register memory bus interactions for each concrete address in the precompile is even.
+// Assumption: all register memory bus interactions feature a concrete address.
 pub fn check_precompile<T: FieldElement>(machine: &SymbolicMachine<T>) -> bool {
-    let (total_count, count_per_addr) = machine
+    let count_per_addr = machine
         .bus_interactions
         .iter()
         .filter_map(|bus_int| bus_int.clone().try_into().ok())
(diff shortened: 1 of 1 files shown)


autoprecompiles/src/lib.rs
```diff
@@ -510,11 +524,34 @@ pub fn exec_receive<T: FieldElement>(machine: &SymbolicMachine<T>) -> SymbolicBu
...
+                    mem_int.addr
+                )
+            })
+        })
+        .fold((0, BTreeMap::new()), |(count, mut map), addr| {
+            *map.entry(addr).or_insert(0) += 1;
+            (count + 1, map)
+        });
+
+    total_count % 2 == 0 && count_per_addr.values().all(|&v| v % 2 == 0)
```
Schaeff, line 548: doesn't the latter imply the former? also, the docs seem incorrect here, as we do more than check the number of register bus interactions is even, we check evenness for each address.

leonardoalt: it does  assuming that the assertion before it catches the non-constant addresses, just added for extra safety

georgwiese (commented):
LGTM, but what Thibaut said.

autoprecompiles/src/lib.rs
```diff
@@ -510,11 +524,34 @@ pub fn exec_receive<T: FieldElement>(machine: &SymbolicMachine<T>) -> SymbolicBu
     r
 }
 
+// Check that the number of register memory bus interactions in the precompile is even.
+// The check is performed on the total number and per address.
```
leonardoalt, line 528: @Schaeff here

Schaeff:
```rust
// Check that the number of register memory bus interactions for each concrete address in the precompile is even.
// Assumption: all register memory bus interactions feature a concrete address
```

then we don't have to count the total?

But maybe I'm misunderstanding something

leonardoalt: I think it's safer, but sure can remove

Schaeff (approved):

Schaeff (approved):

PR #2727, opened 2025-05-19
apc: optimizer should target (degree_bound - 1)

pacheco:
because we need to add is_valid guards to constraints later, which increases the degree by 1


pacheco:
its not only for chip stacking: the bound is reached even in keccak main... the command is:
`cargo run -r prove guest-keccak --mock --autoprecompiles 1 --skip 268 --pgo --input 10`
let me see if I can add a test

pacheco:
actually its `--skip 267`

pacheco:
solved by PR #2799

PR #2735, opened 2025-05-20
remove starky as dependency of cli tests

leonardoalt:
Without this compiling things in the server with AVX fails because starky uses it as default


leonardoalt:
Closing in favor of https://github.com/powdr-labs/powdr/pull/2739

PR #2743, opened 2025-05-20
reduce deps in powdr-openvm

leonardoalt:
The goal here is to make more explicit which powdr crates are needed for powdr-openvm.
After this PR, next concrete steps are:

- Extract the elf stuff from powdr-riscv into another crate so that this new crate can be used by powdr-openvm instead of powdr-riscv, since that one has tons of other dependencies
- constraint-solver only depends on powdr-number which is great!
- powdr-autoprecompiles depends on a bunch of other powdr crates. Ideally we'd eventually be able to remove from the dep list: ast, parser, parser-util, pil-analyzer, pilopt

### DIFF
--- a/openvm/Cargo.toml
+++ b/openvm/Cargo.toml
@@ -43,7 +43,9 @@ openvm-stark-backend = { git = "https://github.com/powdr-labs/stark-backend.git"
   "jemalloc",
 ] }
 
-powdr.workspace = true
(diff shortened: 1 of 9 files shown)


Schaeff (approved):

Schaeff:
Nice one, for pilopt at this point we only rely on `simplify_expression` which could be moved elsewhere

PR #2745, opened 2025-05-21
Remove parser dependency.

chriseth:


Schaeff (approved):
Interesting udeps doesn't catch that

PR #2747, opened 2025-05-21
split riscv modules in crates so other crates can have minimal deps

leonardoalt:


openvm/Cargo.toml
```diff
@@ -47,7 +47,7 @@ openvm-stark-backend = { git = "https://github.com/powdr-labs/stark-backend.git"
 
 powdr-ast.workspace = true
 powdr-number.workspace = true
-powdr-riscv.workspace = true
+powdr-riscv-elf.workspace = true
```
leonardoalt, line 50: This was the main reason for this PR

riscv-types/src/lib.rs
```diff
@@ -0,0 +1,138 @@
+use powdr_isa_utils::SingleDataValue;
+use std::fmt;
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct Register {
```
leonardoalt, line 5: These are all needed by both `riscv` and `riscv-elf`, so needed to extract in another crate too.

pacheco (approved):
afaiu mostly rearranging code around, lgtm

pacheco (approved):

PR #2754, opened 2025-05-22
Witgen: improve tracegen

Schaeff:
- Following the insight by @georgwiese that `eval` is actually not called many times, this PR changes `PowdrAir` to store the AST nodes based on `AlgebraicExpression` (as opposed to openvm SymbolicExpression).
- During trace generation, we do work on each row of the table, which currently includes some table lookups in the table of airs. Move as much as possible out of the part which runs for each row, saving map lookups.
- Avoid collecting into a vector when processing bus interactions

### DIFF
--- a/openvm/src/powdr_extension/chip.rs
+++ b/openvm/src/powdr_extension/chip.rs
@@ -39,6 +39,7 @@ use openvm_stark_backend::{
     rap::{AnyRap, BaseAirWithPublicValues, PartitionedBaseAir},
     Chip, ChipUsageGetter,
 };
+use powdr_ast::analyzed::AlgebraicExpression;
 use powdr_autoprecompiles::powdr::{Column, UniqueColumns};
 use serde::{Deserialize, Serialize};
 
@@ -135,14 +136,9 @@ impl<F: PrimeField32> PowdrChip<F> {
         periphery: SharedChips,
     ) -> Self {
         let air: PowdrAir<F> = PowdrAir::new(precompile.machine);
-        let original_airs = precompile
-            .original_airs
-            .into_iter()
-            .map(|(k, v)| (k, v.into()))
-            .collect();
         let executor = PowdrExecutor::new(
             precompile.original_instructions,
-            original_airs,
+            precompile.original_airs,
             precompile.is_valid_column,
             memory,
             base_config,
@@ -221,7 +217,7 @@ pub struct PowdrAir<F> {
     /// The mapping from poly_id id to the index in the list of columns.
     /// The values are always unique and contiguous
     column_index_by_poly_id: BTreeMap<u64, usize>,
-    machine: SymbolicMachine<F>,
+    machine: powdr_autoprecompiles::SymbolicMachine<F>,
 }
 
 impl<F: PrimeField32> ColumnsAir<F> for PowdrAir<F> {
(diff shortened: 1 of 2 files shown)


openvm/src/powdr_extension/chip.rs
```diff
@@ -393,26 +389,25 @@ where
 
         let witness_evaluator = WitnessEvaluator::<AB>::new(&witness_values);
 
+        let eval_expr = |expr: &AlgebraicExpression<_>| {
```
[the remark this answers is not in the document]
qwang98: So that this is only called at proving stage and we get to keep it Powdr type before?

openvm/src/powdr_extension/chip.rs
```diff
@@ -393,26 +422,25 @@ where
...
-            } = interaction;
-
-            let mult = witness_evaluator.eval_expr(mult);
-            let args = args
-                .iter()
-                .map(|arg| witness_evaluator.eval_expr(arg))
-                .collect_vec();
-
-            builder.push_interaction(*id, args, mult, *count_weight);
+            let powdr_autoprecompiles::SymbolicBusInteraction { id, mult, args, .. } = interaction;
```
qwang98, line 436: Nice! :)

openvm/src/powdr_extension/executor.rs
```diff
@@ -231,46 +228,63 @@ impl<F: PrimeField32> PowdrExecutor<F> {
...
                             self.periphery
                                 .range_checker
                                 .remove_count(value, max_bits as usize);
                         }
                     }
 
-                    for (dummy_trace_index, apc_index) in
-                        &dummy_trace_index_to_apc_index_by_instruction[instruction_id]
-                    {
+                    for (dummy_trace_index, apc_index) in dummy_trace_index_to_apc_index {
```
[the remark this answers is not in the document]
qwang98: Yeah...I figured we don't need the enumerate this way.

openvm/src/powdr_extension/executor.rs
```diff
@@ -281,17 +295,17 @@ impl<F: PrimeField32> PowdrExecutor<F> {
                 let evaluator = RowEvaluator::new(row_slice, Some(column_index_by_poly_id));
 
                 // replay the side effects of this row on the main periphery
-                for bus_interaction in bus_interactions.iter() {
+                // TODO: this could be done in parallel since `self.periphery` is thread safe, but is it worth it? cc @qwang98
```
qwang98, line 298: Doesn't make a much difference per my prior benchmarking.

openvm/src/powdr_extension/chip.rs
```diff
@@ -135,14 +142,9 @@ impl<F: PrimeField32> PowdrChip<F> {
...
         let air: PowdrAir<F> = PowdrAir::new(precompile.machine);
-        let original_airs = precompile
-            .original_airs
-            .into_iter()
-            .map(|(k, v)| (k, v.into()))
-            .collect();
         let executor = PowdrExecutor::new(
             precompile.original_instructions,
-            original_airs,
+            precompile.original_airs,
```
qwang98, line 147: So this seems to be where all the original conversions happen?

qwang98 (approved):
lgtm :)

openvm/src/powdr_extension/executor.rs
Schaeff: Ok removing

openvm/src/powdr_extension/chip.rs
Schaeff: Yes, before this PR, we converted to povm data structures here

PR #2764, opened 2025-05-23
Witgen: avoid timestamp LT checks on irrelevant memory interactions

Schaeff:
- [x] Collect sets of `RecordId` for each APC row during record generation
- [ ] During trace generation, avoid generating lt check column values for memory access beyond the first one

### DIFF
--- a/openvm/src/powdr_extension/executor.rs
+++ b/openvm/src/powdr_extension/executor.rs
@@ -54,7 +54,6 @@ pub struct PowdrExecutor<F: PrimeField32> {
     inventory: SdkVmInventory<F>,
     number_of_calls: usize,
     periphery: SharedChips,
-    offline_memory: Arc<Mutex<OfflineMemory<F>>>,
 }
 
 impl<F: PrimeField32> PowdrExecutor<F> {
@@ -62,7 +61,6 @@ impl<F: PrimeField32> PowdrExecutor<F> {
         instructions: Vec<OriginalInstruction<F>>,
         air_by_opcode_id: BTreeMap<usize, SymbolicMachine<F>>,
         is_valid_column: Column,
-        offline_memory: Arc<Mutex<OfflineMemory<F>>>,
         base_config: SdkVmConfig,
         periphery: SharedChips,
     ) -> Self {
@@ -71,7 +69,7 @@ impl<F: PrimeField32> PowdrExecutor<F> {
             air_by_opcode_id,
             is_valid_poly_id: is_valid_column.id.id,
             inventory: create_chip_complex_with_memory(
-                offline_memory.clone(),
+                offline_memory,
                 periphery.range_checker.clone(),
                 base_config.clone(),
             )
@@ -79,7 +77,6 @@ impl<F: PrimeField32> PowdrExecutor<F> {
             .inventory,
             number_of_calls: 0,
             periphery,
-            offline_memory,
         }
     }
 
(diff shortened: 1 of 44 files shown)


openvm/src/powdr_extension/executor.rs
```diff
@@ -116,9 +127,10 @@ impl<F: PrimeField32> PowdrExecutor<F> {
         SC: StarkGenericConfig,
         <SC::Pcs as Pcs<SC::Challenge, SC::Challenger>>::Domain: PolynomialSpace<Val = F>,
     {
+        let number_of_calls = self.number_of_calls();
```
qwang98, line 130: Cool that we manage to use number of memory ranges as `number_of_calls` as they should be the same.

openvm/src/powdr_extension/executor.rs
```diff
@@ -207,93 +219,153 @@ impl<F: PrimeField32> PowdrExecutor<F> {
...
-                            .args
+            .zip(self.record_ranges.par_iter())
+            .for_each(
+                |((row_slice, dummy_values), (from_record_id, to_record_id))| {
+                    // For each register, find the first read and last write, if they exist
+                    let table: BTreeMap<u8, (Option<MemoryRecord<F>>, Option<MemoryRecord<F>>)> =
+                        self.offline_memory
+                            .lock()
+                            .unwrap()
+                            .records_for_range(*from_record_id, *to_record_id)
```
qwang98, line 256: So offline memory log should match memory log.

qwang98, line 256: So offline memory log should match memory log.

qwang98:
Another thing I notice after printing all dummy traces for the APC row is that not all timestamps of the register memory accesses can be found. I mean we skipped creating aux columns but I think the timestamps should still be there.

qwang98:
I investigated a bit about mapping dummy aux columns to apc aux columns:
1. After filtering to the memory records we want, map them to the dummy rows via timestamp, which all dummy rows have. (However, need to double check that the memory record timestamps we get are correct, as not all of them map to dummy rows. It's also possible that there isn't a one to one mapping, in which case we need to investigate how memory record timestamp differ from dummy row ones.)
2. Get the indices of read aux and write aux columns in a given dummy row. This can be calculated at compile time as long as we have all the data structures for adapter and core columns of a given chip. However, it's not the most straight forward as it requires knowing the indices of a field in a struct. Worst case, we might need to calculate and hardcode these indices manually, which we should try to avoid as much as possible, as it's not super maintainable.
3. We already have mapping from dummy to apc column indices (for the columns we want) from `global_index` function.
4. From the prior steps, we can get a mapping from dummy aux column indices to apc column indices, which is all we need to generate the aux columns.

qwang98:
> Another thing I notice after printing all dummy traces for the APC row is that not all timestamps of the register memory accesses can be found. I mean we skipped creating aux columns but I think the timestamps should still be there.

I think this is related to the timestamp in dummy trace being the main timestamp for that instruction while the memory record timestamps can be the same or incremented by 1, 2.

openvm/src/powdr_extension/executor.rs
```diff
@@ -242,7 +254,6 @@ impl<F: PrimeField32> PowdrExecutor<F> {
                     self.instructions.iter().zip_eq(dummy_values).enumerate()
                 {
                     let evaluator = RowEvaluator::new(dummy_row, None);
-
```
Schaeff, line 245: Let's revert these

openvm/src/powdr_extension/executor.rs
```diff
@@ -207,93 +219,153 @@ impl<F: PrimeField32> PowdrExecutor<F> {
...
+                    &table.values[start..end]
+                })
+                .collect_vec()
+        });
 
         // go through the final table and fill in the values
         values
             // a record is `width` values
             .par_chunks_mut(width)
+            .take(1)
```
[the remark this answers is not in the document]
qwang98: No longer relevant now it's simplified :)

qwang98:
Should be all cleaned up. :)

qwang98 (approved):
LGTM

qwang98 (approved):
Should be good once CI passes.

PR #2768, opened 2025-05-23
Split constraints into independent ones.

chriseth:


georgwiese (commented):
Could you provide some context how that fits into the bigger picture? Seems like a generalization of the limb decomposition solver we have. Does that mean that we can remove it? What examples appear in practice that can be solved after splitting but not by the existing solver?

I think I found an example where the constraint is split into unsatisfiable constraints, even though there is a unique solution, which I think should not be expected?

constraint-solver/src/quadratic_symbolic_expression.rs
```diff
@@ -433,6 +434,125 @@ impl<T: FieldElement, V: Ord + Clone + Hash + Eq + Display> QuadraticSymbolicExp
...
+            }
+
+            // If `rest` change by one, the result will "step over" the range constraint of
+            // the candidate. Now what remains is to check that the others do not wrap.
+
+            let rest: QuadraticSymbolicExpression<_, _> = rest
+                .into_iter()
+                .map(|(coeff, expr)| expr * &SymbolicExpression::from(coeff / smallest_coeff))
+                .sum();
+            // The original constraint is equivalent to `candidate + smallest_coeff * rest = 0`.
```
chriseth, line 506:
```suggestion
            // The original constraint is equivalent to `candidate + smallest_coeff * rest + constant / candidate_coeff = 0`.
```

constraint-solver/src/quadratic_symbolic_expression.rs
```diff
@@ -433,6 +434,125 @@ impl<T: FieldElement, V: Ord + Clone + Hash + Eq + Display> QuadraticSymbolicExp
...
+
+            let rest_rc = rest
+                .range_constraint(range_constraints)
+                .multiple(smallest_coeff)
+                .combine_sum(&RangeConstraint::from_value(constant / candidate_coeff));
+
+            if !rest_rc.is_unconstrained() {
```
chriseth, line 513: Check that this is correct

chriseth:
Ok, here is the proof for a special case:

Assume `0 <= x < t`, `t <= k`, `0 <= y <= v` and `v * k` (integer multiplication) is less than the field modulus.
Then `x = k * y` is equivalent to `x = 0` and y = 0`.

The direction `<=` is obvious, so let's focus on `=>` and assume `y != 0`. This means `k <= k * y <= k * v` and these
inequalities also hold in the integers. But if `x = k * y`, then this contradicts `x < t <= k`. This means `y = 0` and then of course `x = 0` easily follows.

chriseth:
Now I think the only step to the generic case is relaxing the lower bound for `y` to `u <= y < v` and `k * (v - u)` is less than the field modulus.

To see this, assume the range constraint for `x` is `s <= x < t`, where `(t - s) < k` and we have an additional constant `c` such that the constraint reads
`x = k * y + c`. Transform this into `(x - s) = k * y + c - s = k * (y + (c - s) * k^(-1))`

The range constraint for `(x - s)` is now `0 .. (t - s)` as required. Furthermore, the range constraint for the term after `k *` has only shifted by a constant, it's width has not changed.

leonardoalt:
Still relevant?

chriseth:
This has all been moved to https://github.com/powdr-labs/powdr/pull/2751

PR #2771, opened 2025-05-27
Make plonk expression evaluated to be zero

ShuangWu121:
Current plonk compiler transfer an air expression to plonk gate, but it doesn't evaluate the expression. 
for example: 

`a * b * c = 0`

it will compile it to 

```
a * b = temp0
temp0 * c = temp1
```
This PR set the last gate output, which is the output of the expression (temp1 in the above example) to be zero

### DIFF
--- a/openvm/src/plonk/air_to_plonkish.rs
+++ b/openvm/src/plonk/air_to_plonkish.rs
@@ -14,6 +14,8 @@ where
 {
     let mut plonkish_expr = PlonkCircuit::new();
     air_to_plonkish(algebraic_expr, &mut plonkish_expr, temp_id_offset);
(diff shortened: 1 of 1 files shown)


ShuangWu121:
> I think this case would break:
> 
> ```
> col witness x;
> x = 0
> ```
> 
> => `x = 0` would not actually add the gate (just return the variable), and then the outer code would set the last output gate to 0.
nice catch. In this case the gates will be empty though, since the function `build_plonk_expr` only transfers one expression to plonk gates. `x = 0` would actually create a plonkCircuit with empty gates. I add a dummy gate to record this assignment, otherwise it kind of ignores this information, or it is not necessary?

PR #2777, opened 2025-05-27
Remove equal constraints.

chriseth:


leonardoalt (approved):

chriseth:
Ah ok, but it does when it's run in a loop, so all good.

PR #2778, opened 2025-05-27
Boolean extractor only.

chriseth:
Simplified version of #2749 that can be used by the memory optimizer but is not used by the solver yet.


georgwiese (approved):
LGTM!

PR #2781, opened 2025-05-28
Pgo trace cell

qwang98:
Depend on: https://github.com/powdr-labs/powdr/pull/2808

Mostly the same as: https://github.com/powdr-labs/powdr-openvm/pull/121

Differences:
1. Given that we generate apc for all eligible basic blocks, some apc fail due to some new optimization applied (currently narrowed down to `exhaustive_search`). Note that previously in powdr-openvm, apc can be generated for all eligible basic blocks.
2. Remove basic block that fail to generate apc for from the list of eligible basic blocks, which requires propogating error information all the way up to `customize`.


openvm/src/lib.rs
```diff
@@ -673,7 +673,7 @@ mod tests {
     const GUEST: &str = "guest";
     const GUEST_ITER: u32 = 1 << 10;
     const GUEST_APC: usize = 1;
-    const GUEST_SKIP: usize = 56;
+    const GUEST_SKIP: usize = 64;
```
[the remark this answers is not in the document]
qwang98: I thought the same. I just wasn't sure if we were keen on keeping this test (for which we accelerate a very specific block, whose index keeps changing depending on how we update the block sorting algorithm).

openvm/src/customize_exe.rs
```diff
@@ -100,42 +113,130 @@ pub fn customize<F: PrimeField32>(
...
+    // print start index of all retained blocks
+    blocks.iter().enumerate().for_each(|(i, block)| {
+        tracing::info!("Block {}: start_idx {}", i, block.start_idx);
+    });
+
+    // store air width by opcode, so that we don't repetitively calculate them later
+    // filter out opcodes that contain next references in their air, because they are not supported yet in apc
+    let air_width_by_opcode = airs
+        .iter()
+        .filter(|&(i, _)| (!opcode_next_reference.contains(i)))
```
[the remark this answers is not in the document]
leonardoalt: hmm I don't think that happens. The no apc opcodes should be from hand-written precompiles and variants of. Do you have an example of that?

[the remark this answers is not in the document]
qwang98: > I'm fairly certain some of these opcodes also use next references and will crash when a program that uses them is run, which doesn't happen in our tests.

Ahh this is actually a good point, so I'll change to exclude all "bad" opcodes and remove `opcode_next_reference`.

qwang98:
> Very nice!
> 
> Two things:
> 
> * I think generating apcs for every single block will be too slow for large use cases. So I think it should be put behind a non-default option where the default is the current approach.
> * Can you open another PR to `main`, to be merged before this one, with a test similar to `test_keccak_machine` but with 10+ (maybe 50 or 100) apcs and PGO (no ovm execution or proofs)? Then we can add another variant of that test here using this new approach and hopefully we'd see some gains.

I updated as per your first bullet. Will work on the second bullet point another day :)

I ran all openvm tests locally and currently just the `guest_machine` test fails, which will be removed in another PR as per your second bullet point which will be merged before this PR, so I'm just leaving it there.

openvm/src/customize_exe.rs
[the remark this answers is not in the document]
qwang98: Fixed :)

openvm/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition.workspace = true
 license.workspace = true
 homepage.workspace = true
 repository.workspace = true
+cli-openvm.workspace = true
```
leonardoalt, line 8: I don't think `openvm` should use `cli-openvm` as a dependency

qwang98: Yeah...it's something I did originally but forgot to remove. It creates circular dependency so isn't good.

qwang98: Done.

openvm/src/lib.rs
```diff
@@ -80,6 +80,41 @@ mod instruction_formatter;
...
+    Instruction,
+    /// disable PGO
+    None,
+}
+
+pub struct PgoConfig {
+    /// The type of PGO to use.
+    pub pgo_type: PgoType,
+    /// Number of time a pc is executed, Some for PgoType::Cell or PgoType::Instruction, None for PgoType::None.
+    pub pc_idx_count: Option<HashMap<u32, u32>>,
```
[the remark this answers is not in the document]
qwang98: Clever :)

openvm/src/customize_exe.rs
```diff
@@ -273,6 +311,22 @@ pub fn customize<F: PrimeField32>(
             acc_block.pretty_print(openvm_instruction_formatter)
         );
 
+        // all apcs would have been generated earlier if we had Pgo::Cell
+        let (apc_opcode, autoprecompile, subs) = if let Some(apcs) = &mut apcs {
+            apcs.remove(&acc_block.start_idx).unwrap()
```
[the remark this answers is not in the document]
qwang98: Remove returns the APC stored in the map. I used remove because I don't want to clone the whole struct and the same apc won't be used twice. :)

openvm/src/lib.rs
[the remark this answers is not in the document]
qwang98: Done.

openvm/src/customize_exe.rs
```diff
@@ -141,103 +144,141 @@ pub fn customize<F: PrimeField32>(
...
-                (acc_block.start_idx, (apc_opcode, autoprecompile, subs)),
-            ))
-        })
-        .unzip();
+                .unzip();
+
+            // filter out the basic blocks where the apc generation errored out
+            blocks.retain(|b| cells_saved_per_row_by_bb.contains_key(&b.start_idx));
+
+            let pgo_program_idx_count = pc_idx_count.unwrap(); // guaranteed to be Some
```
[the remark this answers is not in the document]

cli-openvm/src/main.rs
```diff
@@ -178,3 +169,41 @@ fn stdin_from(input: Option<u32>) -> StdIn {
...
+
+#[derive(Copy, Clone, Debug, ValueEnum)]
+pub enum CliPgoType {
+    /// cost = cells saved per apc * times executed
+    Cell,
+    /// cost = instruction per apc * times executed
+    Instruction,
+    /// disable PGO
+    None,
+}
```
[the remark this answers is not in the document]
qwang98: Done :)

openvm/src/customize_exe.rs
```diff
@@ -141,103 +144,141 @@ pub fn customize<F: PrimeField32>(
...
+        }
 
-    // sort basic blocks by:
-    // 1. if pc_idx_count (frequency) is provided, cost = frequency * cells_saved_per_row
-    // 2. if pc_idx_count (frequency) is not provided, cost = cells_saved_per_row
-    if let Some(pgo_program_idx_count) = pc_idx_count.clone() {
-        blocks.sort_by(|a, b| {
-            let a_cells_saved = cells_saved_per_row_by_bb[&a.start_idx];
-            let b_cells_saved = cells_saved_per_row_by_bb[&b.start_idx];
+        PgoType::Instruction => {
```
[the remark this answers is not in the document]

leonardoalt (commented):
It'd be good to add a test with the new option

qwang98:
> It'd be good to add a test with the new option

We now have one compilation test and one prove test:
1. The prior test (now expanded to the new PGO mode) to make sure that PGO and no-PGO accelerate the same Keccak block: https://github.com/powdr-labs/powdr/pull/2781/files#diff-1b181b34fae13d7e3ce8cc2b5354deb45a2c8a06bc8b66a2c242a8cc44b956e1R904-R910
2. This newly added test, now with both PGO modes and 10 APCs: https://github.com/powdr-labs/powdr/pull/2781/files#diff-1b181b34fae13d7e3ce8cc2b5354deb45a2c8a06bc8b66a2c242a8cc44b956e1R807-R837

I understand that you commented: "Can you open another PR to main, to be merged before this one, with a test similar to test_keccak_machine but with 10+ (maybe 50 or 100) apcs and PGO (no ovm execution or proofs)? Then we can add another variant of that test here using this new approach and hopefully we'd see some gains."

However, I think any gain should come with ovm execution and/or proofs, because compilation alone for the latest PGO::Cell mode will only be strictly worse, as we make APC for everything and that's why I did #2.

qwang98:
> > It'd be good to add a test with the new option
> 
> We now have one compilation test and one prove test:
> 
> 1. The prior test (now expanded to the new PGO mode) to make sure that PGO and no-PGO accelerate the same Keccak block: https://github.com/powdr-labs/powdr/pull/2781/files#diff-1b181b34fae13d7e3ce8cc2b5354deb45a2c8a06bc8b66a2c242a8cc44b956e1R904-R910
> 2. This newly added test, now with both PGO modes and 10 APCs (): https://github.com/powdr-labs/powdr/pull/2781/files#diff-1b181b34fae13d7e3ce8cc2b5354deb45a2c8a06bc8b66a2c242a8cc44b956e1R807-R837. I also run it with 100 APCs and worked but am keeping it low here so we don't overwhelm CI.
> 
> I understand that you commented: "Can you open another PR to main, to be merged before this one, with a test similar to test_keccak_machine but with 10+ (maybe 50 or 100) apcs and PGO (no ovm execution or proofs)?"
> 
> However, I think any gain should come with ovm execution and/or proofs, because compilation alone for the latest PGO::Cell mode will only be strictly worse, as we make APC for everything and that's why I did number 2.

qwang98:
In terms of gains, when proving 1000 keccak iterations with 25 APC, PGO::Instruction takes 64 secs whereas PGO::Cell takes 61 seconds. It's a very specific example so probably can't extrapolate too much beyond that.

For the version pushed (because CI can't run 1000 keccak), we prove 10 keccak iterations with 25 APC, in which case PGO::Cell actually takes longer, which is expected because the fixed compilation overhead of creating APC for all BB in PGO::Cell gets spread over the more we prove.

leonardoalt:
> In terms of gains, when proving 1000 keccak iterations with 25 APC, PGO::Instruction takes 64 secs whereas PGO::Cell takes 61 seconds. It's a very specific example so probably can't extrapolate too much beyond that.
> 
> For the version pushed (because CI can't run 1000 keccak), we prove 10 keccak iterations with 25 APC, in which case PGO::Cell actually takes longer, which is expected because the fixed compilation overhead of creating APC for all BB in PGO::Cell gets spread over the more we prove.

Nice, but note that we don't really care about compilation time. When measuring performance only proof generation time matters (and execution somewhat).

openvm/src/lib.rs
```diff
@@ -854,9 +866,14 @@ mod tests {
...
         let pgo_data = get_pc_idx_count(GUEST, GuestOptions::default(), stdin);
-        let machines = compile_guest(GUEST, GuestOptions::default(), config, Some(pgo_data))
-            .unwrap()
-            .powdr_airs_metrics();
+        let machines = compile_guest(
+            GUEST,
+            GuestOptions::default(),
+            config,
+            PgoConfig::Cell(pgo_data),
+        )
```
leonardoalt: It would be good to still test both modes with the small example

openvm/src/customize_exe.rs
```diff
@@ -104,6 +119,8 @@ pub fn customize<F: PrimeField32>(
         0x51c, // not sure yet what this is
         0x523, // not sure yet what this is
         0x526, // not sure yet what this is
+        Rv32HintStoreOpcode::HINT_STOREW.global_opcode().as_usize(), // contain next references that don't work with apc
+        Rv32HintStoreOpcode::HINT_BUFFER.global_opcode().as_usize(), // contain next references that don't work with apc
```
leonardoalt: Maybe bringing this back to the beginning of the list makes sense?

openvm/src/customize_exe.rs
```diff
@@ -584,3 +612,164 @@ fn transpose_symbolic_machine<F: PrimeField32, P: FieldElement>(
...
+            let apc_opcode = POWDR_OPCODE + i;
+            let (autoprecompile, subs) =
+                match generate_autoprecompile::<F, powdr_number::BabyBearField>(
+                    acc_block,
+                    airs,
+                    apc_opcode,
+                    config.bus_map.clone(),
+                    config.degree_bound,
+                    branch_opcodes_set,
+                ) {
```
leonardoalt: I'm slightly weary of this block being duplicated (here and in the old location while generating the apcs), because if we change one thing eg the opcode (POWDR_OPCODE + 1) we might forget to update the other one. Maybe we can extract this into a function and centralize the logic there? Basically the same idea of the "cached apcs" but the generation is only computed there and can be called from different places.

openvm/src/customize_exe.rs
```diff
@@ -584,3 +612,164 @@ fn transpose_symbolic_machine<F: PrimeField32, P: FieldElement>(
...
+        .filter(|&(i, _)| (!opcodes_no_apc.contains(i)))
+        .map(|(i, air)| (*i, air.unique_columns().count()))
+        .collect::<HashMap<_, _>>();
+
+    // generate apc and cache it for all basic blocks
+    // calculate number of trace cells saved per row for each basic block to sort them by descending cost
+    let (cells_saved_per_row_by_bb, apcs): (HashMap<_, _>, HashMap<_, (_, _, _)>) = blocks
+        .iter()
+        .enumerate()
+        .filter_map(|(i, acc_block)| {
```
leonardoalt, line 608: This block is quite large now and is doing many different things, I think it'd be good to split it up

openvm/src/customize_exe.rs
```diff
@@ -584,3 +612,164 @@ fn transpose_symbolic_machine<F: PrimeField32, P: FieldElement>(
...
+    );
+
+    // cost = cells_saved_per_row
+    blocks.sort_by(|a, b| {
+        let a_cnt = pgo_program_idx_count[&(a.start_idx as u32)];
+        let b_cnt = pgo_program_idx_count[&(b.start_idx as u32)];
+        (b_cnt * (b.statements.len() as u32)).cmp(&(a_cnt * (a.statements.len() as u32)))
+    });
+
+    // Debug print blocks by decsending cost
```
leonardoalt:
```suggestion
    // Debug print blocks by descending cost
```

openvm/src/lib.rs
qwang98: Done.

openvm/src/customize_exe.rs
qwang98: Done. I think they get messed up during merges.

openvm/src/customize_exe.rs
qwang98: > Done.

openvm/src/customize_exe.rs
qwang98: Broken up into a function to take the opcode as a parameter, but in practice still calculates the opcode from the loop index `i`, so still feels similar. Maybe we are out of options.

openvm/src/customize_exe.rs
qwang98: I think it now looks much leaner now that we have the `generate_apc_cache` helper function and no longer return a tuple. Let me know if you think it's not good enough.

qwang98:
Got some errors caused by the latest main, but now merged again so hopefully CI all good.

qwang98:
@leonardoalt Merged but new test error appears so I had to change `GUEST_KECCAK_APC_PGO` back to 2 from 25.

This error is likely related to one of the commits between when CI was last run yesterday and now and I don't think it's PGO related because all CI passed yesterday.

I recommend that we merge this now and figure out the error (which is not related to this PR) later. Otherwise, it's been quite inefficient now that main has been merged 6 times.

autoprecompiles/src/constraint_optimizer.rs
```diff
@@ -34,13 +34,11 @@ pub fn optimize_constraints<P: FieldElement>(
     constraint_system: ConstraintSystem<P, Variable>,
     bus_interaction_handler: impl BusInteractionHandler<P> + IsBusStateful<P> + Clone,
     degree_bound: usize,
-) -> Result<SymbolicMachine<P>, Error> {
-    let constraint_system = symbolic_machine_to_constraint_system(symbolic_machine);
-
-    let mut stats_logger = StatsLogger::start(&constraint_system);
+    stats_logger: &mut StatsLogger,
```
leonardoalt, line 37: hm why do we need to pass this around now?

openvm/src/lib.rs
```diff
@@ -789,7 +789,7 @@ mod tests {
     const GUEST_KECCAK_ITER: u32 = 1000;
     const GUEST_KECCAK_ITER_SMALL: u32 = 10;
     const GUEST_KECCAK_APC: u64 = 1;
-    const GUEST_KECCAK_APC_PGO: u64 = 25;
+    const GUEST_KECCAK_APC_PGO: u64 = 2;
```
leonardoalt, line 792: why change back to 2?

PR #2789, opened 2025-05-28
make degree bound part of config

leonardoalt:


openvm/src/lib.rs
```diff
@@ -216,6 +219,13 @@ impl PowdrConfig {
     pub fn with_bus_map(self, bus_map: BusMap) -> Self {
         Self { bus_map, ..self }
     }
+
+    pub fn with_degree_bound(self, degree_bound: usize) -> Self {
+        Self {
+            degree_bound,
```
chriseth, line 225: Does it make sense that it's zero by default?

chriseth: Ah no it's not, it's some magic number. Then it's fine!

PR #2790, opened 2025-05-28
Remove equal bus interactions

chriseth:


georgwiese (dismissed):
Cool! Did you check which get removed? Because I checked an old PIL file I had lying around and it didn't contain any duplicate interactions. But could be that with the newer solver steps there are some.

leonardoalt (dismissed):

leonardoalt (approved):

PR #2797, opened 2025-05-29
inliner indexing

pacheco:
keep track of which constraints/expressions contain a given variable to avoid going through the whole constraint system

### DIFF
--- a/pilopt/src/inliner.rs
+++ b/pilopt/src/inliner.rs
@@ -104,11 +104,11 @@ fn is_valid_substitution<T: FieldElement, V: Ord + Clone + Hash + Eq>(
         .constraints_referencing_variables(std::iter::once(var.clone()))
         .all(|cref| match cref {
             ConstraintRef::AlgebraicConstraint(identity) => {
-                let degree = qse_degree_with_virtual_substitution(identity, &var, replacement_deg);
+                let degree = qse_degree_with_virtual_substitution(identity, var, replacement_deg);
                 degree <= max_degree
             }
             ConstraintRef::BusInteraction(interaction) => interaction.fields().all(|expr| {
-                let degree = qse_degree_with_virtual_substitution(expr, &var, replacement_deg);
+                let degree = qse_degree_with_virtual_substitution(expr, var, replacement_deg);
                 degree <= max_degree
             }),
         })
--- a/pilopt/src/inliner.rs
+++ b/pilopt/src/inliner.rs
@@ -1,5 +1,5 @@
 use powdr_constraint_solver::constraint_system::ConstraintRef;
-use powdr_constraint_solver::indexed_constraint_system::{IndexedConstraintSystem};
+use powdr_constraint_solver::indexed_constraint_system::IndexedConstraintSystem;
 use powdr_constraint_solver::{
     constraint_system::ConstraintSystem, quadratic_symbolic_expression::QuadraticSymbolicExpression,
 };
@@ -25,12 +25,7 @@ pub fn replace_constrained_witness_columns<
         let constraint = &constraint_system.algebraic_constraints()[curr_idx];
 
         for (var, expr) in find_inlinable_variables(constraint) {
-            if is_valid_substitution(
-                &var,
-                &expr,
-                &constraint_system,
-                max_degree,
-            ) {
+            if is_valid_substitution(&var, &expr, &constraint_system, max_degree) {
                 log::trace!("Substituting {var} = {expr}");
                 log::trace!("  (from identity {constraint})");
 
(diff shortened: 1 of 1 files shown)


pilopt/src/inliner.rs
```diff
@@ -1,10 +1,46 @@
...
     constraint_system::ConstraintSystem, quadratic_symbolic_expression::QuadraticSymbolicExpression,
 };
 use powdr_number::FieldElement;
+use std::collections::{HashMap, HashSet};
 use std::fmt::Display;
 use std::hash::Hash;
 
+/// Keeps track of which constraints and bus interactions reference a given variable.
+/// Maps variables to constraint/bus interaction indices.
+struct VarIndex<V> {
```
chriseth, line 12: Can we use `IndexedConstraintSystem` from the constraint-solver crate here?

pilopt/src/inliner.rs
```diff
@@ -14,43 +50,86 @@ pub fn replace_constrained_witness_columns<
...
+                inlined_vars.push(var);
+
+                break;
             }
         }
     }
 
-    false
+    // remove inlined constraints from system
+    for idx in to_remove_idx {
```
chriseth, line 122: Maybe better to use `into_iter.filter(..).collect()`

pilopt/src/inliner.rs
```diff
@@ -89,17 +168,45 @@ fn find_inlinable_variables<T: FieldElement, V: Ord + Clone + Hash + Eq>(
...
-    !constraint_system
-        .expressions()
-        .map(|constraint| qse_degree_with_virtual_substitution(constraint, var, replacement_deg))
-        .any(|deg| deg > max_degree)
+    let is_valid_constraints = constraint_system
+        .algebraic_constraints
+        .iter()
+        .enumerate()
+        .filter(|(idx, _)| {
+            var_index.constraints[var].contains(idx) && !ignore_constraints.contains(idx)
```
chriseth, line 186: Why not iter over `var_index.constaints[var]` directly? Because it does not store the index?

pilopt/src/inliner.rs
pacheco: done, much cleaner now!

pilopt/src/inliner.rs
```diff
@@ -11,46 +13,48 @@ pub fn replace_constrained_witness_columns<
...
-    false
+    let mut constraint_system = ConstraintSystem::from(constraint_system);
+
+    // remove inlined constraints from system
+    constraint_system.algebraic_constraints = constraint_system
+        .algebraic_constraints
+        .into_iter()
+        .enumerate()
+        .filter_map(|(idx, qse)| (!to_remove_idx.contains(&idx)).then_some(qse))
+        .collect();
```
Schaeff, line 49: Could this be a method on `ConstraintSystem`? Also is linear search (`Vec::contains`) ok here?

pilopt/src/inliner.rs
```diff
@@ -11,46 +14,48 @@ pub fn replace_constrained_witness_columns<
...
+
+    // sanity check
+    assert!(constraint_system.expressions_mut().all(|expr| {
```
Schaeff, line 53:
```suggestion
    assert!(constraint_system.expressions_mut().all(|expr| {
```

the mut is a bit surprising here, is there an immutable equivalent?

Schaeff (approved):
LGTM, just some non-blocking comments

pilopt/src/inliner.rs
```diff
@@ -11,46 +14,48 @@ pub fn replace_constrained_witness_columns<
...
-    constraint_system: &mut ConstraintSystem<T, V>,
-    max_degree: usize,
-) -> bool {
-    let indices: Vec<usize> = (0..constraint_system.algebraic_constraints.len()).collect();
+    let mut constraint_system = IndexedConstraintSystem::from(constraint_system);
 
-    for idx in indices.into_iter().rev() {
-        let constraint = &constraint_system.algebraic_constraints[idx];
+    for curr_idx in (0..constraint_system.algebraic_constraints().len()).rev() {
+        let constraint = &constraint_system.algebraic_constraints()[curr_idx];
```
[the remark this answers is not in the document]
pacheco: doesn't work because then the constraint system is borrowed for the whole loop :/

pilopt/src/inliner.rs
```diff
@@ -11,46 +14,48 @@ pub fn replace_constrained_witness_columns<
     T: FieldElement,
     V: Ord + Clone + Hash + Eq + Display,
 >(
-    mut constraint_system: ConstraintSystem<T, V>,
+    constraint_system: ConstraintSystem<T, V>,
     max_degree: usize,
 ) -> ConstraintSystem<T, V> {
-    loop {
-        if !try_apply_substitution(&mut constraint_system, max_degree) {
```
chriseth, line 18: Why can we do this without this loop now? Before, it would re-try searching for a variable from the beginning. Is it obvious that we would not find another variable until the first substitution point? Should we at least re-consider the constraint we just substituted in?

chriseth (approved):

PR #2798, opened 2025-05-29
Fix xor bus

chriseth:


openvm/src/bus_interaction_handler/bitwise_lookup.rs
```diff
@@ -40,7 +40,8 @@ pub fn handle_bitwise_lookup<T: FieldElement>(
                 ]
             } else {
                 // The result of an XOR can only be a byte and have bits set that are set in either x or y
-                let z_constraint = x.disjunction(y).conjunction(&byte_constraint());
```
chriseth, line 43: The bug here was that min/max range constraints were also transferred to z, i.e. if we have x=1..4, y= 3..8,then we would get z = 3..4

openvm/tests/optimizer.rs
```diff
@@ -55,3 +55,33 @@ fn test_optimize() {
...
+    let machine: SymbolicMachine<BabyBearField> = serde_cbor::from_reader(reader).unwrap();
+
+    let machine = optimize(
+        machine,
+        OpenVmBusInteractionHandler::new(BusMap::openvm_base()),
+        None, // opcode
+        5,
+    );
+
+    println!(
```
leonardoalt, line 73: leftover print
I think we have test tracing also if you do want to print stuff

openvm/tests/optimizer.rs
```diff
@@ -55,3 +55,33 @@ fn test_optimize() {
         [3194, 2861, 160]
     );
 }
+
+#[test]
+fn test_conflicting_constraints_in_bus_interaction() {
+    let file =
+        std::fs::File::open("tests/conflicting_constraints_in_bus_interaction.cbor").unwrap();
```
leonardoalt, line 62: hmm do we wanna keep artifacts like these in the repo? Feels a bit odd since we can't read its contents, so the test doesn't mean much

chriseth: A better way would be to add unit tests, but I won't add unit tests for units I have not written...

openvm/src/bus_interaction_handler/bitwise_lookup.rs
```diff
@@ -40,7 +40,8 @@ pub fn handle_bitwise_lookup<T: FieldElement>(
...
+                let z_constraint = RangeConstraint::from_mask(*x.mask() | *y.mask())
```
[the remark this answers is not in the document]
georgwiese: Sorry, that's what I meant. I would expect `disjunction` to do something similar (looks like it does, but also handles min and max somehow). So was this is a bug, and if yes, why was the previous implementation wrong? Asking because I wrote the code.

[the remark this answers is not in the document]
chriseth: Ah, just realized that the example in the first comment above is wrong. So disjunction on range constraints means that the range constraint of z is rc1 or rc2. So if `rc1=5..8` and `rc2 = 7..9` the disjunction would be `5..9` (ignoring bit constraints here). What we want here is that we approximate the range constraint effects of xor by the range constraint effects of the bit-wise or. But a bit-wise or is different from a "logical or" on range constraints.

In other words, `RangeConstraint::disjunction` models the constraint `(z - x) * (z - y) = 0`.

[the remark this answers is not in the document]
georgwiese: Yeah, that makes sense, but did that cause an issue, and how? Did it lead to a less precise range constraint, or did it not allow the correct value?

georgwiese (approved):
I don't fully get what the problem was, but the new code looks good to me ^^

openvm/src/bus_interaction_handler/bitwise_lookup.rs
georgwiese: Wouldn't you get `z = 1..8`? And also the correct mask?

openvm/src/bus_interaction_handler/bitwise_lookup.rs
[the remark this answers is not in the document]
chriseth: Yes, it led to a wrong (too tight) range constraint.

PR #2799, opened 2025-05-29
Inliner bus degree

pacheco:
depends on #2797


openvm/tests/optimizer.rs
```diff
@@ -35,7 +35,10 @@ fn test_optimize() {
         machine,
         OpenVmBusInteractionHandler::new(BusMap::openvm_base()),
         None, // opcode
-        5,
+        DegreeBound {
+            identities: 5,
+            bus_interactions: 5,
```
Schaeff, line 40: Are they supposed to be the same?

pilopt/src/inliner.rs
```diff
@@ -169,7 +182,8 @@ mod test {
             bus_interactions: vec![],
         };
 
-        let constraint_system = replace_constrained_witness_columns(constraint_system, 3);
+        let constraint_system =
+            replace_constrained_witness_columns(constraint_system, bounds(3, 3));
```
Schaeff: Same in these tests?

openvm/tests/optimizer.rs
pacheco: This keeps the tests with the same behaviour as before when there was a shared degree for both

openvm/src/lib.rs
```diff
@@ -220,7 +220,10 @@ impl PowdrConfig {
             autoprecompiles,
             skip_autoprecompiles,
             bus_map: BusMap::openvm_base(),
-            degree_bound: customize_exe::OPENVM_DEGREE_BOUND,
+            degree_bound: DegreeBound {
+                identities: customize_exe::OPENVM_DEGREE_BOUND,
+                bus_interactions: customize_exe::OPENVM_DEGREE_BOUND - 1,
```
pacheco, line 307: @leonardoalt do we still need this after your `is_valid` PR that avoids increasing degrees?

leonardoalt: yea because now I just pass a very low degree bound to make sure everything works

pacheco: i meant having the default for bus interactions be the ovm bound `-1`

leonardoalt: this makes sense to me now

PR #2801, opened 2025-05-30
Remove pure neg gate from plonk circuit

ShuangWu121:

-1 * var = temp
this PR remove this gate and plug -var into the gate that has temp.


ShuangWu121:
handle the case in another PR

PR #2807, opened 2025-06-02
Remove neg gate

ShuangWu121:

-1 * var = temp
this PR remove this gate and plug -var into the gate that has temp, if the last gate is unary operation, it still create a new gate and assert it to zero.


leonardoalt:
Is it ready for review after resolving conflits?

PR #2808, opened 2025-06-02
Refactor tests for PGO

qwang98:
All changes explained in comments. Prerequisite for https://github.com/powdr-labs/powdr/pull/2781/


openvm/src/lib.rs
```diff
@@ -798,29 +824,20 @@ mod tests {
...
-        let config = PowdrConfig::new(GUEST_KECCAK_APC, GUEST_KECCAK_SKIP);
-        let machines = compile_guest(GUEST_KECCAK, GuestOptions::default(), config, pc_idx_count)
-            .unwrap()
-            .powdr_airs_metrics();
-        assert_eq!(machines.len(), 1);
-        let m = &machines[0];
-        assert_eq!(m.width, 3195);
-        assert_eq!(m.constraints, 160);
-        assert_eq!(m.bus_interactions, 2861);
-    }
```
qwang98: Removed this helper function because now we only have one Keccak compilation test with PGO. (We don't need the non-PGO version, because they have the same result and PGO is becoming quite common in most runs).

openvm/src/lib.rs
```diff
@@ -838,39 +855,15 @@ mod tests {
...
-        // Input via StdIn
-        let mut stdin = StdIn::default();
-        stdin.write(&GUEST_ITER);
-
-        let guest_opts = GuestOptions::default();
-        // Guest machine should have more optimized results with pgo
-        // because we didn't accelerate the "costliest block" in the non-pgo version.
-        let pc_idx_count = get_pc_idx_count(GUEST, guest_opts.clone(), stdin);
-        // We don't skip any sorted basic block here to accelerate the "costliest" block.
-        let config = PowdrConfig::new(GUEST_APC, GUEST_SKIP_PGO);
```
qwang98, line 851: Now only have one guest test left which uses PGO, so removing this one.

openvm/src/lib.rs
```diff
@@ -838,39 +855,15 @@ mod tests {
...
         let m = &machines[0];
-        assert_eq!(m.width, 53);
-        assert_eq!(m.constraints, 21);
-        assert_eq!(m.bus_interactions, 39);
-    }
-
-    #[test]
-    fn keccak_machine() {
-        test_keccak_machine(None);
-    }
```
qwang98: Removed this redundant one now that both Keccak tests use PGO.

openvm/src/lib.rs
```diff
@@ -784,7 +790,27 @@ mod tests {
...
+    }
+
+    // Create 2 APC for 10 Keccak iterations to test different PGO modes
+    #[test]
+    fn keccak_prove_multiple_pgo_modes() {
+        use std::time::Instant;
+        // Config
+        let mut stdin = StdIn::default();
+        stdin.write(&GUEST_KECCAK_ITER_SMALL); // 10
+        let config = PowdrConfig::new(GUEST_KECCAK_APC_PGO, GUEST_KECCAK_SKIP); // 2, 0
```
qwang98: Now use 2 APCs to test PGO but will bump this number up (so that we don't get APC compilation error here as this PR doesn't handle those errors yet).

openvm/src/lib.rs
```diff
@@ -672,23 +672,28 @@ mod tests {
...
+        pgo: bool,
     ) -> Result<(), Box<dyn std::error::Error>> {
-        let program = compile_guest(guest, GuestOptions::default(), config, None).unwrap();
+        let pgo_data = pgo.then(|| {
+            let guest_opts = GuestOptions::default();
+            get_pc_idx_count(guest, guest_opts, stdin.clone())
+        });
+        let program = compile_guest(guest, GuestOptions::default(), config, pgo_data).unwrap();
         prove(&program, mock, recursion, stdin)
     }
```
qwang98: Add PGO option here and in other functions so that it's more explicit in all tests, now that PGO becomes a very common option.

openvm/src/lib.rs
leonardoalt: I'd say for keccak it's actually useful to have a single run without PGO and just 1 apc, because it's a good reference for the PGO test with 1 apc since the results should be the same.

openvm/src/lib.rs
leonardoalt: so more than 2 already crash?

qwang98: I think 10 does. Didn't test the lower ones. It's a random factor depending on which APC happens to crash in compilation, not an indicator of things like out of memory error.

With the new PGO PR, these compilation errors should be filtered out before we create APC for everything.

openvm/src/lib.rs
qwang98: Brought it back.

openvm/src/lib.rs
```diff
@@ -784,7 +801,27 @@ mod tests {
...
+    #[test]
+    fn keccak_prove_multiple_pgo_modes() {
+        use std::time::Instant;
+        // Config
+        let mut stdin = StdIn::default();
+        stdin.write(&GUEST_KECCAK_ITER_SMALL); // 10
+        let config = PowdrConfig::new(GUEST_KECCAK_APC_PGO, GUEST_KECCAK_SKIP); // 2, 0
+
+        let start = Instant::now();
+        prove_simple(GUEST_KECCAK, config.clone(), stdin.clone(), None);
```
leonardoalt: This isn't using PGO though?

openvm/src/lib.rs
```diff
@@ -867,10 +885,8 @@ mod tests {
     #[test]
     fn keccak_machine_pgo() {
         let mut stdin = StdIn::default();
-        stdin.write(&GUEST_KECCAK_ITER);
-        // Keccak machine should have the same results with pgo
-        // because we already accelerate the "costliest" block with the non-pgo version.
-        let pc_idx_count = get_pc_idx_count(GUEST_KECCAK, GuestOptions::default(), stdin);
-        test_keccak_machine(Some(pc_idx_count));
+        stdin.write(&GUEST_ITER);
```
leonardoalt: shouldn't this be GUEST_KECCAK_ITER?

openvm/src/lib.rs
qwang98: Yeah. Just placeholders. They are filled in the next PR which I'm running CI for right now.

openvm/src/lib.rs
qwang98: Agree. Passing CI as it's just compilation.

openvm/src/lib.rs
qwang98: Fixed.

openvm/src/lib.rs
qwang98: Fixed.

leonardoalt (approved):

PR #2809, opened 2025-06-02
change is_valid guard to not increase the degree

leonardoalt:
The `is_valid` guard is needed to enforce that every constraint works with zeroes as the witness. Before this PR, we simply took each constraint `C` and multiplied by `is_valid` which increases the max degree of the constraints. However, we found a better way to do that:

- We only need to care about constants, since any expression over columns only will always solve to 0.
- We assume the expressions are already simplified trivially, eg: `3 + 4` and `col * 1` will not show up at this point.
- We only need to add `* is_valid` to constants, but not all:
- For `(x + 3) * (x + 4)`, it suffices to enforce that one side is 0, so we only continue on the LHS: `(x + 3 * is_valid) * (x + 4)` is enough and keeps the same max degree.
- For `(x + 3) + (x + 4)` we need to make sure both sides converge, so we apply the transformation to both sides: `(x + 3 * is_valid) + (x + 4 * is_valid)`

### DIFF
--- a/autoprecompiles/src/lib.rs
+++ b/autoprecompiles/src/lib.rs
@@ -326,6 +326,8 @@ fn satisfies_zero_witness<T: FieldElement>(expr: &AlgebraicExpression<T>) -> boo
 }
 
 /// Adds `is_valid` guards to constraints without increasing its degree.
+/// This implementation always guards the LHS of multiplications.
+/// In the future this could be changed to minimize the number of guards added.
 /// Assumption:
 /// - `expr` is already simplified, i.e., expressions like (3 + 4) and (x * 1) do not appear.
 fn add_guards_constraint<T: FieldElement>(


autoprecompiles/src/lib.rs
```diff
@@ -120,6 +123,17 @@ impl<T: Display> Display for SymbolicMachine<T> {
     }
 }
 
+impl<T: Clone + Ord + std::fmt::Display> SymbolicMachine<T> {
+    pub fn degree(&self) -> usize {
```
Schaeff, line 127: Just curious: we were talking about having two degrees, one for constraints and one for bus interactions. How does  this method relate to that? Is the bus interaction one this one minus one?

autoprecompiles/src/lib.rs
```diff
@@ -293,11 +307,52 @@ pub fn build<T: FieldElement>(
...
+    }
+
+    match expr {
+        AlgebraicExpression::BinaryOperation(AlgebraicBinaryOperation { left, op, right }) => {
+            let left = add_guards_constraint(*left, is_valid);
+            let right = match op {
+                AlgebraicBinaryOperator::Add | AlgebraicBinaryOperator::Sub => {
+                    Box::new(add_guards_constraint(*right, is_valid))
+                }
+                _ => right,
```
Schaeff: It would be clearer to explicitly match `Mul`

autoprecompiles/src/lib.rs
leonardoalt: Could be that yea, but I guess we can also make this 2 functions at that point

autoprecompiles/src/lib.rs
leonardoalt: done

autoprecompiles/src/lib.rs
```diff
@@ -293,11 +307,56 @@ pub fn build<T: FieldElement>(
...
+fn satisfies_zero_witness<T: FieldElement>(expr: &AlgebraicExpression<T>) -> bool {
+    let mut zeroed_expr = expr.clone();
+    powdr::make_refs_zero(&mut zeroed_expr);
+    let zeroed_expr = simplify_expression(zeroed_expr);
+    powdr::is_zero(&zeroed_expr)
+}
+
+/// Adds `is_valid` guards to constraints without increasing its degree.
+/// Assumption:
+/// - `expr` is already simplified, i.e., expressions like (3 + 4) and (x * 1) do not appear.
```
Schaeff, line 332: Maybe let's add a comment that the implementation has a bias towards guarding the lhs of multiplications? In theory there could be an implementation that minimizes the number of `is_valid` occurrences?

leonardoalt: done

Schaeff (approved):

PR #2810, opened 2025-06-02
add bigint branching opcodes as well

leonardoalt:
This came up in the Reth benchmarks. Without this our basic blocks collection fails to see these extension branch instructions. These don't come up in the ELF file because they're extensions.

### DIFF
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -418,16 +418,19 @@ fn add_extra_targets<F: PrimeField32>(
     mut labels: BTreeSet<u32>,
 ) -> BTreeSet<u32> {
     let branch_opcodes_bigint = branch_opcodes_bigint_set();
-    for (i, instr) in program.instructions_and_debug_infos.iter().enumerate() {
-        let instr = instr.as_ref().unwrap().0.clone();
-        // Program address are given as multiples of 4 bytes.
-        let adjusted_pc = OPENVM_INIT_PC + (i as u32) * 4;
-        let op = instr.opcode.as_usize();
-        if branch_opcodes_bigint.contains(&op) {
-            let target: u32 = adjusted_pc + instr.c.as_canonical_u32();
-            labels.insert(target);
-        }
-    }
+    let new_labels = program
+        .instructions_and_debug_infos
+        .iter()
+        .enumerate()
+        .filter_map(|(i, instr)| {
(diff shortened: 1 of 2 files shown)


openvm/src/customize_exe.rs
```diff
@@ -93,15 +97,29 @@ pub fn customize<F: PrimeField32>(
...
         0x526, // not sure yet what this is
+        1024,
+        1025,
+        1028,
+        1033,
+        1104,
+        1030,
+        1033,
+        1027,
+        1029,
```
Schaeff, line 115: This list is getting quite long, shall we add comments?
Also, do these opcode ids always refer to the same instruction, or does it depend on the VM configuration/extensions?

leonardoalt: let's do that in a different PR, there were already some before that were cryptic, it'd be good to do it in a single run

leonardoalt: > Also, do these opcode ids always refer to the same instruction, or does it depend on the VM configuration/extensions?

The former, but it's tricky because they're not all named. For example some curve ones have "base" names for the opcodes, but then they're indexed over different curves causing a lot more global opcodes than the named ones

leonardoalt: Re comments, I think using the actual type names is the best outcome, but again, I'd do it in a separate PR just doing that cus it'll be more work to figure everything out

openvm/src/customize_exe.rs
```diff
@@ -334,6 +407,30 @@ pub fn collect_basic_blocks<F: PrimeField32>(
...
+    for (i, instr) in program.instructions_and_debug_infos.iter().enumerate() {
+        let instr = instr.as_ref().unwrap().0.clone();
+        // Program address are given as multiples of 4 bytes.
+        let adjusted_pc = OPENVM_INIT_PC + (i as u32) * 4;
+        let op = instr.opcode.as_usize();
+        if branch_opcodes_bigint.contains(&op) {
+            let target: u32 = adjusted_pc + instr.c.as_canonical_u32();
+            labels.insert(target);
+        }
+    }
```
Schaeff: labels.extend(program.instructions_and_debug_infos.iter().enumerate().filter_map(|(i, instr)| {
   branch_opcodes_bigint.contains(&op).then_some(adjusted_pc + instr.c.as_canonical_u32())
}));

PR #2818, opened 2025-06-02
Extract stats logger.

chriseth:


autoprecompiles/src/stats_logger.rs
```diff
@@ -0,0 +1,67 @@
...
+use powdr_number::FieldElement;
+use powdr_pilopt::qse_opt::Variable;
+
+pub struct StatsLogger {
+    start_time: Instant,
+}
+
+impl StatsLogger {
+    pub fn start<P: FieldElement>(constraint_system: &ConstraintSystem<P, Variable>) -> Self {
+        log_constraint_system_stats("Starting optimization", constraint_system);
```
pacheco, line 15: it seems this is still closely tied to the optimizer, shouldn't this be a module of it instead of autoprecompiles?
otherwise i think the type of argument (`ConstraintSystem`) and logging function should be parametrized, no?

chriseth: -> https://github.com/powdr-labs/powdr/pull/2820 :)

pacheco (dismissed):

leonardoalt:
conflcits

Schaeff (approved):

PR #2826, opened 2025-06-03
Rust bench for OVM execution

qwang98:
WIP

### DIFF
--- a/cli-openvm/Cargo.toml
+++ b/cli-openvm/Cargo.toml
@@ -22,6 +22,7 @@ eyre = "0.6.12"
 tracing = "0.1.40"
 
 clap = { version = "^4.3", features = ["derive"] }
+criterion = { version = "0.4" }
 
 serde_cbor = "0.11.2"
 
--- a/cli-openvm/benches/execution_benchmark.rs
+++ b/cli-openvm/benches/execution_benchmark.rs
@@ -1,0 +1,32 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+
+const GUEST_KECCAK: &str = "guest_keccak";
+const GUEST_KECCAK_ITER: u32 = 1000;
+const GUEST_KECCAK_APC: u64 = 1;
+const GUEST_KECCAK_SKIP: u64 = 0;
+
+fn keccak_benchmark(c: &mut Criterion) {
+    // To see the component run times (e.g. powdr chip vs non-powdr chip trace gen time, powdr chip dummy trace gen time for each dummy chip)
+    // run with RUST_LOG=debug
+    let mut group = c.benchmark_group("keccak-benchmark");
+
+    // Compile the program to execute
+    let guest_opts = powdr_openvm::GuestOptions::default();
+    let powdr_config = powdr_openvm::PowdrConfig::new(GUEST_KECCAK_APC, GUEST_KECCAK_SKIP);
+    let program =
+        powdr_openvm::compile_guest(GUEST_KECCAK, guest_opts, powdr_config, None).unwrap(); // we don't need PGO because it's guaranteed to run the biggest Keccak basic block without PGO
+    let mut stdin = openvm_sdk::StdIn::default();
+    stdin.write(&GUEST_KECCAK_ITER);
+
+	// Run benchmark
+    group.bench_function("Execute 1000 Keccaks with 1 APC", |b| {
+        b.iter(|| {
+            powdr_openvm::execute(program.clone(), stdin.clone()).unwrap();
+        })
+    });
+
+    group.finish();
+}
+
+criterion_group!(execution_benchmarks, keccak_benchmark);
+criterion_main!(execution_benchmarks);


cli-openvm/benches/execution_benchmark.rs
```diff
@@ -0,0 +1,251 @@
...
+    let mut stdin = StdIn::default();
+    stdin.write(&GUEST_KECCAK_ITER);
+
+    // Default uses 100 samples, which is too many, but 10 samples is the required minimum
+    group.sample_size(10);
+
+    // Run benchmark
+    group.bench_function("Execute 1000 Keccaks with 1 APC", |b| {
+        b.iter(|| {
+            execute_and_generate(program.clone(), stdin.clone()).unwrap();
```
qwang98: This is the function benchmarked.

cli-openvm/benches/execution_benchmark.rs
```diff
@@ -0,0 +1,251 @@
...
+        // visit the fields and extract id
+        let mut visitor = Visitor {
+            id: None,
+            air_name: None,
+        };
+        attrs.record(&mut visitor);
+
+        // turn fields into strings or None
+        let id_str = visitor.id.map(|i| format!("id:{i}"));
+        let air_str = visitor.air_name.map(|a| format!("air:{a}"));
```
pacheco: maybe this logic could be inside a Display implementation of `Visitor`?

cli-openvm/benches/execution_benchmark.rs
```diff
@@ -0,0 +1,251 @@
...
+        let count = durs.len() as u32;
+        let total: Duration = durs.iter().copied().sum();
+        let avg = if count > 0 {
+            total / count
+        } else {
+            Duration::ZERO
+        };
+
+        tracing::info!(
+            "{:indent$}{} → ran {} times, avg = {:?}",
```
pacheco: maybe we're more interested in total time spent rather than avg here?

cli-openvm/benches/execution_benchmark.rs
```diff
@@ -0,0 +1,251 @@
...
+fn print_tree() {
+    let parents = SPAN_PARENTS.lock().unwrap();
+    let times = SPAN_TIMES.lock().unwrap();
+
+    // Build name to set of child names
+    let mut tree: BTreeMap<String, HashSet<String>> = BTreeMap::new();
+    let mut has_parent = HashSet::new();
+
+    for (name, parent_opt) in parents.iter() {
+        if let Some(parent) = parent_opt {
```
pacheco: i wonder if you could insert into `roots` here already and avoid this `has_parent`?

cli-openvm/benches/execution_benchmark.rs
```diff
@@ -0,0 +1,251 @@
...
+            "{:indent$}{} → ran {} times, avg = {:?}",
+            "",
+            name,
+            count,
+            avg,
+            indent = level * 2
+        );
+
+        if let Some(children) = tree.get(name) {
+            let mut kids: Vec<_> = children.iter().cloned().collect();
```
pacheco: is it possible to do `for child in children` directly here?

cli-openvm/benches/execution_benchmark.rs
```diff
@@ -0,0 +1,251 @@
...
+            kids.sort();
```
pacheco: maybe this could be sorted by total time spent, so its easy to see the most expensive spans

cli-openvm/benches/execution_benchmark.rs
qwang98: Sounds good. Not sure how that will impact creating the `key` for the span but will attempt.

cli-openvm/benches/execution_benchmark.rs
qwang98: I think it's actually the `avg` we interested here, because total can depend on the number of sampling, which is a configurable parameter.

cli-openvm/benches/execution_benchmark.rs
qwang98: Actually would prefer to keep the original order in some cases, at least in the out-most layer, as it shows the sequential run of functions, for example, in order of compilation --> execution --> prove.

For substeps within execution, it can be good to either not sort the chips so they keep the order they are executed, or as you point out, sort them by total time spent, so we see the most expensive ones.

So the answer is that we might need EITHER no sorting OR no sorting for some plus time sorted for others. Regardless, my current solution can be improved, so good point made. :)

cli-openvm/benches/execution_benchmark.rs
qwang98: Yes. It's because I attempted to use sort here so had to collect the iter into a concrete type first.

cli-openvm/benches/execution_benchmark.rs
qwang98: Ahh, I agree that I can just collect those nodes that have `None` as `parent_opt`, so you right.

qwang98: Done.

cli-openvm/benches/execution_benchmark.rs
qwang98: Done.

cli-openvm/benches/execution_benchmark.rs
qwang98: I ended up using no sorting at all but just keeping the insertion order via an IndexMap.

cli-openvm/benches/execution_benchmark.rs
qwang98: Done.

cli-openvm/benches/execution_benchmark.rs
```diff
@@ -0,0 +1,47 @@
...
+const GUEST_KECCAK: &str = "guest-keccak";
+const GUEST_KECCAK_ITER: u32 = 1000;
+const GUEST_KECCAK_APC: u64 = 1;
+const GUEST_KECCAK_SKIP: u64 = 0;
+
+fn keccak_benchmark(c: &mut Criterion) {
+    // create a fresh collector for this benchmark
+    let mut collector = BenchmarkCollector::new();
+    collector.install();
+    collector.clear();
```
pacheco: is the clear needed?

cli-openvm/Cargo.toml
```diff
@@ -20,10 +20,20 @@ powdr-openvm.workspace = true
 
 eyre = "0.6.12"
 tracing = "0.1.40"
+once_cell = "1.17"
```
pacheco: not being used anymore?

pacheco (commented):
minor comments, otherwise lgtm

cli-openvm/benches/execution_benchmark.rs
qwang98: Not for this specific example. This API is more for running multiple benchmarks with the same collector and clearing the data in between, so I removed this line. :)

cli-openvm/Cargo.toml
qwang98: Good catch! I switched away from using LazyStatic and this is no longer needed, so I removed it.

cli-openvm/benches/execution_benchmark.rs
qwang98: I moved it to the end with a comment, because still need to use the function so clippy doesn't ask to remove it. Still want to keep this function as a needed feature in the future.

openvm/src/lib.rs
```diff
@@ -458,6 +458,24 @@ pub fn execute(
     Ok(())
 }
 
+pub fn execute_and_generate(
+    program: CompiledProgram<PowdrBB>,
+    inputs: StdIn,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let CompiledProgram { exe, vm_config } = program;
+
+    let app_log_blowup = 2;
```
Schaeff, line 461: iirc this should depend on the max constraint degree. How do we set it in other places?

cli-openvm/benches/execution_benchmark.rs
```diff
@@ -0,0 +1,47 @@
...
+
+const GUEST_KECCAK: &str = "guest-keccak";
+const GUEST_KECCAK_ITER: u32 = 10;
+const GUEST_KECCAK_APC: u64 = 1;
+const GUEST_KECCAK_SKIP: u64 = 0;
+
+fn keccak_benchmark(c: &mut Criterion) {
+    // create a fresh collector for this benchmark
+    let mut collector = BenchmarkCollector::new();
+    collector.install();
```
Schaeff: Could `install` be part of `new` ? Is there a use case for creating a collector and not installing it?

cli-openvm/benches/execution_benchmark.rs
```diff
@@ -0,0 +1,47 @@
...
+        b.iter(|| {
+            execute_and_generate(program.clone(), stdin.clone()).unwrap();
+        })
+    });
+
+    group.finish();
+
+    // Print the tree of spans and their durations
+    collector.print_tree();
+    collector.clear(); // not necessarily needed in this example, but need to clear the collector in between benchmarks when we run multiple with the same collector
```
Schaeff: Could these two be combined and added to the `Drop` implementation of `Collector`? Then when the collector goes out of scope it would print. Is there a use case for running many benchmarks with the same collector?

cli-openvm/Cargo.toml
```diff
@@ -22,8 +22,17 @@ eyre = "0.6.12"
...
 
+[dev-dependencies]
+indexmap = "2.9.0"
+
 [lints]
 workspace = true
+
+[[bench]]
+name = "execution_benchmark"
+harness = false
```
Schaeff: Any reason why the benches should be in the cli crate as opposed to the openvm crate?

cli-openvm/benches/execution_benchmark.rs
Schaeff: This PR already has a lot of code so I reckon we should stick to what we are using now and only extend in the future if needed

cli-openvm/Cargo.toml
qwang98: Original thought was that I might use some CLI functions, but ends up not being true, so I'm moving them.

cli-openvm/benches/execution_benchmark.rs
qwang98: Yes, for example this test group benchmarks multiple functions, and it'll be great to clear the collector in between instead of initializing a new collector each time: https://github.com/powdr-labs/powdr/blob/main/pipeline/benches/evaluator_benchmark.rs#L32-L87

I think I can combine print_tree and clear to print_tree_and_clear, so that when we print some results, which should only happen after a benchmark function, we also clear it for the next benchmark function, if any.

openvm/src/lib.rs
qwang98: I think it's always `app_log_blowup = 2` if you do a full search of the whole code base. It's taken directly from `openvm/src/lib.rs`, the only place this param is set.

cli-openvm/benches/execution_benchmark.rs
qwang98: Cool. Removed it.

cli-openvm/benches/execution_benchmark.rs
qwang98: Done.

cli-openvm/benches/execution_benchmark.rs
qwang98: Done.

cli-openvm/Cargo.toml
qwang98: Done.

qwang98:
Ready for a final.

openvm/Cargo.toml
```diff
@@ -49,6 +49,7 @@ itertools = "0.14.0"
 
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.17", features = ["std", "env-filter"] }
+criterion = { version = "0.4" }
```
Schaeff: Can this be a dev dep?

openvm/benches/utils/duration_recorder.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+impl DurationRecorderLayer {
+    /// Create a new collector with empty maps.
+    pub fn new() -> Self {
+        Self {
+            span_times: Arc::new(Mutex::new(IndexMap::new())),
+            span_parents: Arc::new(Mutex::new(IndexMap::new())),
+        }
+    }
```
Schaeff: derive(Default)?

openvm/benches/utils/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+        tracing::subscriber::set_global_default(subscriber)
+            .expect("failed to set global tracing subscriber");
+
+        Self { layer }
+    }
+
+    /// Print the timing tree for all collected spans and clear any previously collected data.
+    pub fn print_tree_and_clear(&self) {
+        self.layer.print_tree_and_clear();
+    }
```
Schaeff: It's still unclear to me why we would want the same collector for many benchmarks, so since it's not used in this PR, in the interest of simplifying, I think this function should take `self` by value and be called something like `finish` to match what we are doing with criterion groups. Also, it doesn't seem expensive to create a new one?

openvm/benches/utils/duration_recorder.rs
qwang98: Done.

openvm/benches/utils/mod.rs
qwang98: Done. :)

qwang98:
Ready for a final.

openvm/Cargo.toml
Schaeff: @qwang98 ?

qwang98: Sorry I missed this!

qwang98: Fixed :)

PR #2836, opened 2025-06-04
Fix OVM Witgen

qwang98:
Pair here: https://github.com/powdr-labs/openvm/pull/24

OVM tests for creating multiple APCs (introduced in PGO branch) fails.

This patches the error that we cannot skip witgen for the last read/write access to memory.

Previously we thought this was the last access, but some APC has timestamp increment as the last access, which is not a "real" access.

### DIFF
--- a/openvm/src/lib.rs
+++ b/openvm/src/lib.rs
@@ -789,7 +789,7 @@ mod tests {
     const GUEST_KECCAK_ITER: u32 = 1000;
     const GUEST_KECCAK_ITER_SMALL: u32 = 10;
     const GUEST_KECCAK_APC: u64 = 1;
-    const GUEST_KECCAK_APC_PGO: u64 = 2;
+    const GUEST_KECCAK_APC_PGO: u64 = 100;
     const GUEST_KECCAK_SKIP: u64 = 0;
 
     #[test]
--- a/openvm/src/powdr_extension/executor.rs
+++ b/openvm/src/powdr_extension/executor.rs
@@ -15,7 +15,7 @@ use openvm_circuit::{
         ExecutionState, InstructionExecutor, Result as ExecutionResult, VmChipComplex,
         VmInventoryError,
     },
-    system::memory::OfflineMemory,
+    system::memory::{online::MemoryLogEntry, OfflineMemory},
 };
 use openvm_circuit::{
     arch::{VmConfig, VmInventory},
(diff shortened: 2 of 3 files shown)


openvm/src/lib.rs
```diff
@@ -789,7 +789,7 @@ mod tests {
     const GUEST_KECCAK_ITER: u32 = 1000;
     const GUEST_KECCAK_ITER_SMALL: u32 = 10;
     const GUEST_KECCAK_APC: u64 = 1;
-    const GUEST_KECCAK_APC_PGO: u64 = 2;
+    const GUEST_KECCAK_APC_PGO: u64 = 100;
```
qwang98: Now we can run 100 APCs and maybe even more :)

openvm/src/powdr_extension/executor.rs
```diff
@@ -106,12 +106,32 @@ impl<P: IntoOpenVm> PowdrExecutor<P> {
             });
 
         self.number_of_calls += 1;
-        let to_record_id = memory.get_memory_logs().len() - 1;
+        let to_record_id = memory.get_memory_logs().len(); // exclusive range
+
+        let memory_logs = memory.get_memory_logs();
```
Schaeff: Calling `get_memory_logs` twice

openvm/src/powdr_extension/executor.rs
```diff
@@ -106,12 +106,32 @@ impl<P: IntoOpenVm> PowdrExecutor<P> {
...
+
+        let last_read_write = memory_logs[from_record_id..to_record_id]
+            .iter()
+            .rposition(|entry| {
+                matches!(
+                    entry,
+                    MemoryLogEntry::Read { .. } | MemoryLogEntry::Write { .. }
+                )
+            })
+            .unwrap() // assume that there is at least one read/write within the apc range
```
Schaeff: We don't need this assumption if we store an `Option`

qwang98: Thought about that but was hard to imagine that we have a Powdr Chip with no real memory access. Can make this edit to make this more robust :)

openvm/src/powdr_extension/executor.rs
qwang98: Fixed.

openvm/src/powdr_extension/executor.rs
qwang98: Done.

openvm/src/powdr_extension/executor.rs
```diff
@@ -118,20 +121,19 @@ impl<P: IntoOpenVm> PowdrExecutor<P> {
...
-            .unwrap() // assume that there is at least one read/write within the apc range
-            + from_record_id;
+            .map(|idx| idx + from_record_id);
 
         tracing::debug!(
             "APC range (exclusive): {}..{} (last read/write at {})",
             from_record_id,
             to_record_id,
-            last_read_write
+            last_read_write.unwrap_or(to_record_id)
```
Schaeff, line 130: This is incorrect in the None case, let's print nothing in that case

Schaeff (approved):

openvm/src/powdr_extension/executor.rs
```diff
@@ -106,12 +109,31 @@ impl<P: IntoOpenVm> PowdrExecutor<P> {
...
+            .iter()
+            .rposition(|entry| {
+                matches!(
+                    entry,
+                    MemoryLogEntry::Read { .. } | MemoryLogEntry::Write { .. }
+                )
+            })
+            .map(|idx| idx + from_record_id);
+
+        tracing::debug!(
```
leonardoalt: maybe debug is too granular and this should be trace?

qwang98: Should be good for smaller inputs, but basically for the Keccak example will be printed Sum over `i` of (# of times executed * # of APCs * # of Keccaks) so I agree.

PR #2839, opened 2025-06-04
Abstract constraint optimizer.

chriseth:
Turns the variable in the constraint optimizer into a generic parameter. The goal of this is that we can later add auxiliary variables which are enum alternatives in an enum that also contains the original variable - without having to touch the optimizer code.


georgwiese (approved):
Cool! I actually changed the type in #2831, but this should be compatible.

PR #2843, opened 2025-06-05
fix default degree bound

leonardoalt:
Read comment in the code


georgwiese (approved):
Interesting that this doesn't seem to cause any changes to the snapshot tests?

PR #2845, opened 2025-06-05
Option for `prove` to use a smaller segment

qwang98:
Just doing this for `prove` so far because that's needed for our test, but technically can be added for `execute` as well because execution already uses `SegmentationStrategy`.

### DIFF
--- a/openvm/src/lib.rs
+++ b/openvm/src/lib.rs
@@ -3,8 +3,8 @@ use eyre::Result;
 use itertools::{multiunzip, Itertools};
 use openvm_build::{build_guest_package, find_unique_executable, get_package, TargetFilter};
 use openvm_circuit::arch::{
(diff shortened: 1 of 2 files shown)


openvm/src/lib.rs
```diff
@@ -509,8 +509,18 @@ pub fn prove(
     mock: bool,
     recursion: bool,
     inputs: StdIn,
+    segment_height: Option<usize>,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let CompiledProgram { exe, vm_config } = program;
+    let exe = &program.exe;
+    let mut vm_config = program.vm_config.clone();
+
+    // DefaultSegmentationStrategy { max_segment_len: 4194204, max_cells_per_chip_in_segment: 503304480 }
```
qwang98, line 526: These are private constants I'm just putting here for reference.

qwang98: From OVM.

leonardoalt (dismissed):

PR #2850, opened 2025-06-06
Ovm heap witgen

qwang98:
- [x] Track index of heap memory bus interactions before any optimization (which are supposed to match execution memory access 1 to 1)
- [ ] Pass the heap memory bus interaction indices to execution
- [ ] Compute memory records to skip (if both send and receive are optmized away)

### DIFF
--- a/autoprecompiles/src/optimizer.rs
+++ b/autoprecompiles/src/optimizer.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 
 use itertools::Itertools;
 use powdr_ast::analyzed::AlgebraicExpression;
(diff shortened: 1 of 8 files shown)


autoprecompiles/src/optimizer.rs
```diff
@@ -277,3 +327,45 @@ fn bus_interaction_to_symbolic_bus_interaction<P: FieldElement>(
...
+    let mut removed_indices = BTreeSet::new();
+
+    bus_int_optimization_tracker
+        .iter()
+        .for_each(|optimization| {
+            tracing::debug!("bus interaction optimization: {:?}", optimization);
+            tracing::debug!("applying optimization to {:?}", orig_mem_bus_int_idx);
+            match optimization {
+                BusInteractionOptimization::Remove { index, keep } => {
+                    for idx in index.iter().rev() {
```
qwang98, line 393: Remove in reverse order so we don't need to shift `orig_mem_bus_int_idx`

autoprecompiles/src/optimizer.rs
```diff
@@ -277,3 +327,45 @@ fn bus_interaction_to_symbolic_bus_interaction<P: FieldElement>(
...
+                            ));
+                        } else {
+                            // bus interactions we don't keep are expected to be non heap memory bus interactions
+                            orig_mem_bus_int_idx.remove(*idx);
+                        }
+                    }
+                }
+                BusInteractionOptimization::Append { number_appended } => {
+                    // appended bus interactions are not memory bus interactions
+                    orig_mem_bus_int_idx.extend(std::iter::repeat(None).take(*number_appended));
```
qwang98: Appended bus interactions are not memory related, so they don't have a `Some(mem_bus_int_index)`.

autoprecompiles/src/optimizer.rs
```diff
@@ -46,41 +52,124 @@ pub fn optimize<T: FieldElement>(
...
+                        None
+                    }
+                }
+            };
+            // return what we found (or None)
+            Some(out)
+        })
+        // filter_map to drop the Nones
+        .filter_map(|x| x)
+        .collect();
```
qwang98: Applies the logic that a memory record is removed if both of its bus interactions are removed.

For example if bus interactions removed are [0, 1, 2, 4, 7, 8, 9, 10], memory records removed are [0, 4], where 0 corresponds to [0, 1] and 4 corresponds to [8, 9].

PR #2852, opened 2025-06-06
Journal

chriseth:


constraint-solver/src/indexed_constraint_system.rs
```diff
@@ -74,6 +79,26 @@ impl<T: FieldElement, V> IndexedConstraintSystem<T, V> {
     }
 }
 
+impl<T: FieldElement, V> IndexedConstraintSystem<T, V> {
+    /// Removes all constraints that do not fulfill the predicate.
+    pub fn retain_constraints(
+        &mut self,
+        f: impl FnMut(&QuadraticSymbolicExpression<T, V>) -> bool,
+    ) {
+        // TODO updated index
```
chriseth: These are causing the test failure currently.

autoprecompiles/src/constraint_optimizer.rs
```diff
@@ -34,48 +31,45 @@ pub fn optimize_constraints<
...
     bus_interaction_handler: impl BusInteractionHandler<P> + IsBusStateful<P> + Clone,
     degree_bound: usize,
     stats_logger: &mut StatsLogger,
-) -> Result<ConstraintSystem<P, V>, Error> {
-    let constraint_system =
-        solver_based_optimization(constraint_system, bus_interaction_handler.clone())?;
-    stats_logger.log("solver-based optimization", &constraint_system);
+) -> Result<(), Error> {
+    solver_based_optimization(constraint_system, bus_interaction_handler.clone())?;
+    stats_logger.log("solver-based optimization", &*constraint_system);
```
qwang98, line 41: Why dereference and reference?

constraint-solver/src/journalled_constraint_system.rs
```diff
@@ -0,0 +1,93 @@
...
+        // We do not track removal of constraints yet, but we could.
+        self.system.retain_algebraic_constraints(f);
+    }
+
+    /// Removes all bus interactions that do not fulfill the predicate.
+    pub fn retain_bus_interactions(
+        &mut self,
+        f: impl FnMut(&BusInteraction<QuadraticSymbolicExpression<T, V>>) -> bool,
+    ) {
+        // TODO track
```
qwang98: Sounds like we can simply return the `replacement_map` from `retain_bus_interactions` and store it as a field of the journaled system?

autoprecompiles/src/constraint_optimizer.rs
chriseth: Because it's a double reference

chriseth: And the single `*` would dereference both I think. I think this is what the compiler usually does automatically ("re-borrowing") but it doesn't work in this case.

constraint-solver/src/journalled_constraint_system.rs
chriseth: yes exactly

autoprecompiles/src/constraint_optimizer.rs
```diff
@@ -35,48 +32,45 @@ pub fn optimize_constraints<
     P: FieldElement,
     V: Ord + Clone + Eq + Hash + Display + IsWitnessColumn,
 >(
-    constraint_system: ConstraintSystem<P, V>,
+    constraint_system: &mut JournalledConstraintSystem<P, V>,
```
Schaeff: What is the rationale behind switching to taking mutable references instead of values?

autoprecompiles/src/optimizer.rs
```diff
@@ -61,14 +62,17 @@ fn optimization_loop_iteration<T: FieldElement>(
...
+    let mut constraint_system = JournalledConstraintSystem::from(constraint_system);
+    optimize_constraints(
+        &mut constraint_system,
         bus_interaction_handler.clone(),
         degree_bound,
         stats_logger,
     )?;
     // TODO: avoid these conversions
-    let machine = constraint_system_to_symbolic_machine(constraint_system);
+    // TODO continue here with the journalled system once the memory machen is changed to
```
Schaeff:
```suggestion
    // TODO continue here with the journalled system once the memory machine is changed to
```

constraint-solver/src/constraint_system.rs
```diff
@@ -17,6 +17,33 @@ pub struct ConstraintSystem<T: FieldElement, V> {
...
 }
 
+impl<T: FieldElement, V> Default for ConstraintSystem<T, V> {
+    fn default() -> Self {
+        Self {
+            algebraic_constraints: Vec::new(),
+            bus_interactions: Vec::new(),
+        }
+    }
+}
```
Schaeff: Can this be derived?

constraint-solver/src/indexed_constraint_system.rs
```diff
@@ -74,6 +79,74 @@ impl<T: FieldElement, V> IndexedConstraintSystem<T, V> {
...
+                            replacement_map[*i].map(ConstraintSystemItem::AlgebraicConstraint)
+                        }
+                        ConstraintSystemItem::BusInteraction(_) => Some(*item),
+                    })
+                    .collect();
+            });
+    }
+
+    /// Removes all bus interactions that do not fulfill the predicate.
+    pub fn retain_bus_interactions(
```
Schaeff, line 95: It would be good to share code between these two methods

constraint-solver/src/inliner.rs
```diff
@@ -23,19 +22,25 @@ pub fn replace_constrained_witness_columns<
...
-    let mut constraint_system = IndexedConstraintSystem::from(constraint_system);
-
-    for curr_idx in (0..constraint_system.algebraic_constraints().len()).rev() {
-        let constraint = &constraint_system.algebraic_constraints()[curr_idx];
+    let constraint_count = constraint_system
+        .indexed_system()
+        .algebraic_constraints()
+        .len();
+    for curr_idx in (0..constraint_count).rev() {
+        let constraint = &constraint_system.indexed_system().algebraic_constraints()[curr_idx];
```
Schaeff, line 35: Could this be implemented as an iterator on `constraint_system.indexed_system()`?

constraint-solver/src/journalled_constraint_system.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl<T: FieldElement, V: Clone + Ord + Hash> From<ConstraintSystem<T, V>>
+    for JournalledConstraintSystem<T, V>
+{
+    fn from(system: ConstraintSystem<T, V>) -> Self {
+        Self {
+            system: system.into(),
+        }
+    }
+}
```
Schaeff: Could be generalized to `impl<..., C: Into<IndexedConstraintSystem<T, V>> From<C>`

constraint-solver/src/constraint_system.rs
chriseth: No, because `V` needs to be `Default` for that, it's this weird limitation of Vector...

chriseth: Or at least of the macro

autoprecompiles/src/constraint_optimizer.rs
chriseth: It looked more ergonomic to me. Also what we want to avoid is to return a new journaled constraint system. I mean in the end, what we actually need would be a monad here, but ....

constraint-solver/src/constraint_system.rs
chriseth: I was wrong :)

constraint-solver/src/inliner.rs
chriseth: Not sure what you mean. Iterate over all expressions or not using index access? I translated the existing implementation and the indexed access is used because we mutate the system in each iteration (if we do inline). The assumption of course is that the indices do not change when we substitute.

Schaeff: I see yeah thanks, we mutate the whole collection while iterating over it. Yes the index access was bothering me. This is fine but I got curious why we don't do something like

```rust
    let mut to_remove_idx = HashSet::new();
    let mut inlined_vars = HashSet::new();
    let substitutions = constraint_system
        .indexed_system()
        .algebraic_constraints()
        .iter()
        .rev()
        .enumerate()
        .flat_map(|(curr_idx, constraint)| {
            find_inlinable_variables(constraint)
                .iter()
                .flat_map(|(var, expr)| {
                    if is_valid_substitution(
                        var,
                        expr,
                        constraint_system.indexed_system(),
                        degree_bound,
                    ) {
                        log::trace!("Substituting {var} = {expr}");
                        log::trace!("  (from identity {constraint})");

                        to_remove_idx.insert(curr_idx);
                        inlined_vars.insert(var.clone());

                        Some((var.clone(), expr.clone()))
                    } else {
                        None
                    }
                }).collect_vec()
        })
        .collect_vec();

    for (var, expr) in substitutions {
        constraint_system.substitute_by_unknown(&var, &expr);
    }
```
    
Is it correct? It requires some cloning and more collection but applying the substitution at the end could be made to use a single pass, instead of now doing a full pass for every inlinable variable?

Non blocking :)

constraint-solver/src/journalled_constraint_system.rs
Schaeff: This was not addressed, non blocking though.

autoprecompiles/src/constraint_optimizer.rs
Schaeff: But passing `JournalingConstraintSystem<P, V>` by value we could still mutate it and return it from the function without it being a new one? Got your point on ergonomic, although I think for transformations like that a pure approach seems better?

constraint-solver/src/journalled_constraint_system.rs
chriseth: Hm, I think I did. I also renamed the file, maybe because of that?

constraint-solver/src/inliner.rs
chriseth: The problem is that if we delay substitution, the degree values will be wrong:
```
x - y * z = 0;
u * x  + t = 8;
```
If we substitute `x = y * z` into `u * x + t = 8`, then the degree will increase and we might not be able to choose `t` for substitution any more.

PR #2855, opened 2025-06-06
more nightly tests

leonardoalt:
This should be merged only if https://github.com/powdr-labs/powdr/actions/runs/15495026836/job/43629576763 passes.


georgwiese (approved):
LGTM

PR #2859, opened 2025-06-06
Collect used instructions into hashset

Schaeff:
Without this PR, we get the executor a linear amount of times in the size of the program. We need them only once per used opcode.


qwang98 (approved):
LGTM.

Technically if multiple opcodes share the same AIR, might be able to further reduce this by just calling this on a set of unique executors/AIRs instead of unique opcodes.

Do we know the rough run time impact of this? Might be worth doing this further optimization or not depending on that.

PR #2870, opened 2025-06-10
Share dummy periphery across all APCs

Schaeff:
All APCs need access to periphery chips such as the range checker or the xor chip.
These chips are expensive to create in memory, but they are not needed in the end and just get thrown away.
Before this PR, each APC creates one instance of them.
After this PR, a single instance is created and then shared across all APCs.

### DIFF
--- a/openvm/src/powdr_extension/executor/inventory.rs
+++ b/openvm/src/powdr_extension/executor/inventory.rs
@@ -17,7 +17,7 @@ pub type DummyInventory<F> = VmInventory<DummyExecutor<F>, DummyPeriphery<F>>;
 pub type DummyChipComplex<F> = VmChipComplex<F, DummyExecutor<F>, DummyPeriphery<F>>;
 
 #[allow(clippy::large_enum_variant)]
-#[derive(ChipUsageGetter, Chip, InstructionExecutor, AnyEnum)]
+#[derive(ChipUsageGetter, Chip, InstructionExecutor, AnyEnum, From)]
 pub enum DummyExecutor<F: PrimeField32> {
     #[any_enum]
     Sdk(SdkVmConfigExecutor<F>),
@@ -29,26 +29,7 @@ pub enum DummyExecutor<F: PrimeField32> {
     System(SystemExecutor<F>),
 }
 
-impl<F: PrimeField32> From<SharedExecutor<F>> for DummyExecutor<F>
-where
-    F: PrimeField32,
-{
-    fn from(executor: SharedExecutor<F>) -> Self {
-        DummyExecutor::Shared(executor)
-    }
-}
-
-// Here we keep the `SystemExecutor` variant to allow for system-level operations.
-impl<F> From<SystemExecutor<F>> for DummyExecutor<F>
-where
-    F: PrimeField32,
-{
-    fn from(system_executor: SystemExecutor<F>) -> Self {
-        DummyExecutor::System(system_executor)
-    }
-}
-
-#[derive(ChipUsageGetter, Chip, AnyEnum)]
+#[derive(ChipUsageGetter, Chip, AnyEnum, From)]
 pub enum DummyPeriphery<F: PrimeField32> {
     #[any_enum]
     Sdk(SdkVmConfigPeriphery<F>),
@@ -56,24 +37,6 @@ pub enum DummyPeriphery<F: PrimeField32> {
     Shared(SharedPeriphery<F>),
     #[any_enum]
     System(SystemPeriphery<F>),
-}
-
-impl<F: PrimeField32> From<SharedPeriphery<F>> for DummyPeriphery<F>
-where
-    F: PrimeField32,
-{
-    fn from(periphery: SharedPeriphery<F>) -> Self {
-        DummyPeriphery::Shared(periphery)
-    }
-}
-
-impl<F> From<SystemPeriphery<F>> for DummyPeriphery<F>
-where
-    F: PrimeField32,
-{
-    fn from(system_periphery: SystemPeriphery<F>) -> Self {
-        DummyPeriphery::System(system_periphery)
-    }
 }
 
 #[derive(ChipUsageGetter, Chip, InstructionExecutor, From, AnyEnum)]
--- a/openvm/src/powdr_extension/executor/periphery.rs
+++ b/openvm/src/powdr_extension/executor/periphery.rs
@@ -35,6 +35,9 @@ impl PowdrPeripheryInstances {
                 range_checker: range_checker.clone(),
                 tuple_range_checker: tuple_range_checker.cloned(),
             },
+            // Bitwise lookup and tuple range checker do not need to be shared with the main execution:
+            // If we did share, we'd have to roll back the side effects of execution and apply the side effects from the apc air onto the main periphery.
+            // By not sharing them, we can throw away the dummy ones after execution and only apply the side effects from the apc air onto the main periphery.
             dummy: SharedPeripheryChips {
                 bitwise_lookup_8: SharedBitwiseOperationLookupChip::new(bitwise_8.bus()),
                 range_checker: range_checker.clone(),


openvm/src/customize_exe.rs
```diff
@@ -152,11 +154,6 @@ pub fn customize<P: IntoOpenVm>(
         blocks.len()
     );
 
-    // print start index of all retained blocks
-    blocks.iter().enumerate().for_each(|(i, block)| {
-        tracing::info!("Block {}: start_idx {}", i, block.start_idx);
-    });
-
```
Schaeff: Too much data on reth

openvm/src/customize_exe.rs
```diff
@@ -265,14 +278,10 @@ pub fn customize<P: IntoOpenVm>(
...
-        let opcodes_in_acc = acc
-            .iter()
-            .map(|x| x.opcode.as_usize())
-            .unique()
-            .collect_vec();
+        tracing::info!("Create extension");
 
         extensions.push(PowdrPrecompile::new(
-            format!("PowdrAutoprecompile_{i}"),
+            format!("PowdrAutoprecompile_{apc_opcode}"),
```
Schaeff: This is a breaking change, we use the opcode not the index in the list

openvm/src/customize_exe.rs
```diff
@@ -291,10 +290,7 @@ pub fn customize<P: IntoOpenVm>(
                 .zip_eq(subs)
                 .map(|(instruction, subs)| OriginalInstruction::new(instruction, subs))
                 .collect(),
-            airs.iter()
-                .filter(|(i, _)| opcodes_in_acc.contains(*i))
-                .map(|(i, air)| (*i, air.clone()))
-                .collect(),
+            airs.clone(),
```
Schaeff: Change here: we pass a reference to the original air table instead of copying all airs

openvm/src/powdr_extension/chip.rs
```diff
@@ -53,95 +52,12 @@ pub struct PowdrChip<P: IntoOpenVm> {
     pub air: Arc<PowdrAir<P>>,
 }
 
-/// The shared chips which can be used by the PowdrChip.
```
Schaeff, line 56: Moved to a different file as is

openvm/src/customize_exe.rs
Schaeff: reverted

Cargo.toml
```diff
@@ -83,33 +83,38 @@ powdr-autoprecompiles = { path = "./autoprecompiles", version = "0.1.4" }
...
-openvm-rv32im-guest = { git = "https://github.com/powdr-labs/openvm.git", rev = "f12a2fc", default-features = false }
-openvm-transpiler = { git = "https://github.com/powdr-labs/openvm.git", rev = "f12a2fc" }
-openvm-circuit = { git = "https://github.com/powdr-labs/openvm.git", rev = "f12a2fc" }
-openvm-circuit-derive = { git = "https://github.com/powdr-labs/openvm.git", rev = "f12a2fc" }
-openvm-circuit-primitives = { git = "https://github.com/powdr-labs/openvm.git", rev = "f12a2fc" }
-openvm-circuit-primitives-derive = { git = "https://github.com/powdr-labs/openvm.git", rev = "f12a2fc" }
-openvm-instructions = { git = "https://github.com/powdr-labs/openvm.git", rev = "f12a2fc" }
-openvm-instructions-derive = { git = "https://github.com/powdr-labs/openvm.git", rev = "f12a2fc" }
-openvm-sdk = { git = "https://github.com/powdr-labs/openvm.git", rev = "f12a2fc", default-features = false, features = [
+openvm = { git = "https://github.com/powdr-labs/openvm.git", rev = "76be56c" }
```
leonardoalt: hash can be updated now that the openvm side PR has been merged

openvm/src/powdr_extension/executor/mod.rs
```diff
@@ -378,26 +383,27 @@ fn global_index<F>(
...
 
     let this = base_config;
-    let mut complex = this.system.config.create_chip_complex()?.transmute();
+    let mut complex: DummyChipComplex<F> = this.system.config.create_chip_complex()?.transmute();
 
     // CHANGE: inject the correct memory here to be passed to the chips, to be accessible in their get_proof_input
     complex.base.memory_controller.offline_memory = memory.clone();
-    complex.base.range_checker_chip = range_checker;
+    complex.base.range_checker_chip = shared_chips.range_checker.clone();
+    // END CHANGE
```
leonardoalt, line 403: wdym end change?

Schaeff: See above, this is taken from an internal openvm function and extended to inject memory and shared chips.

leonardoalt:
maybe merge main again so we can test nightly separately on this?

openvm/src/lib.rs
```diff
@@ -502,13 +502,6 @@ pub fn pgo(
     // print the total and by pc counts at the warn level (default level in powdr-openvm)
     tracing::warn!("Pgo captured {} pc's", pc.len());
 
-    // print pc_index map in descending order of pc_index count
-    let mut pc_index_count_sorted: Vec<_> = pc_index_count.iter().collect();
-    pc_index_count_sorted.sort_by(|a, b| b.1.cmp(a.1));
-    pc_index_count_sorted.iter().for_each(|(pc, count)| {
-        tracing::warn!("pc_index {}: {}", pc, count);
-    });
```
qwang98: I'm proposing in #2894 to keep this but de-prioritizing tracing level to debug as it's useful in debugging, which means it shouldn't show up in `cli-openvm` uses, whose default level is `WARN`.

openvm/src/powdr_extension/executor/inventory.rs
```diff
@@ -0,0 +1,163 @@
...
+
+// Here we keep the `SystemExecutor` variant to allow for system-level operations.
+impl<F> From<SystemExecutor<F>> for DummyExecutor<F>
+where
+    F: PrimeField32,
+{
+    fn from(system_executor: SystemExecutor<F>) -> Self {
+        DummyExecutor::System(system_executor)
+    }
+}
```
qwang98: Can derive these via `From`?

openvm/src/powdr_extension/executor/inventory.rs
```diff
@@ -0,0 +1,163 @@
...
+
+impl<F> From<SystemPeriphery<F>> for DummyPeriphery<F>
+where
+    F: PrimeField32,
+{
+    fn from(system_periphery: SystemPeriphery<F>) -> Self {
+        DummyPeriphery::System(system_periphery)
+    }
+}
```
qwang98: Can derive these via `From`?

qwang98: Or is it because you need to derive the `SdkVm` variants using the macro below and you have to manually create the other variants here?

openvm/src/powdr_extension/executor/mod.rs
```diff
@@ -378,26 +383,27 @@ fn global_index<F>(
...
+
+    // CHANGE: inject the periphery chips so that they are not created by the extensions. This is done for memory footprint: the dummy periphery chips are thrown away anyway, so we reuse a single one for all APCs.
```
qwang98, line 405: Dummy question (pun intended): Why are dummy chips thrown away? I mean in the trace gen it looks like we remove the side effects by looking at bus interactions in the dummy airs so doesn't seem that dummy range checker is used. Are there other reasons?

openvm/src/powdr_extension/executor/mod.rs
```diff
@@ -378,26 +383,27 @@ fn global_index<F>(
...
     use openvm_native_circuit::Native;
     use openvm_rv32im_circuit::{Rv32I, Rv32Io};
     use openvm_sha256_circuit::Sha256;
 
     let this = base_config;
-    let mut complex = this.system.config.create_chip_complex()?.transmute();
+    let mut complex: DummyChipComplex<F> = this.system.config.create_chip_complex()?.transmute();
 
     // CHANGE: inject the correct memory here to be passed to the chips, to be accessible in their get_proof_input
     complex.base.memory_controller.offline_memory = memory.clone();
```
qwang98, line 401: Memory is shared with main (non-Powdr) execution?

openvm/src/powdr_extension/executor/periphery.rs
```diff
@@ -0,0 +1,150 @@
...
+    var_range::SharedVariableRangeCheckerChip,
+};
+use openvm_stark_backend::p3_field::PrimeField32;
+
+/// The shared chips which can be used by the PowdrChip.
+#[derive(Clone)]
+pub struct PowdrPeripheryInstances {
+    /// The real chips used for the main execution.
+    pub real: SharedPeripheryChips,
+    /// The dummy chips used for all APCs. They share the range checker but create new instances of the bitwise lookup chip and the tuple range checker.
```
qwang98, line 15: Why not make them also share bitwise lookup chip and tuple range checker?

openvm/src/powdr_extension/executor/mod.rs
qwang98: Wondering if there are execution implications and maybe we can skip applying side effects to the dummy periphery? Can be another execution optimization?

openvm/src/powdr_extension/executor/periphery.rs
```diff
@@ -0,0 +1,150 @@
...
+                bitwise_lookup_8: bitwise_8.clone(),
+                range_checker: range_checker.clone(),
+                tuple_range_checker: tuple_range_checker.cloned(),
+            },
+            dummy: SharedPeripheryChips {
+                bitwise_lookup_8: SharedBitwiseOperationLookupChip::new(bitwise_8.bus()),
+                range_checker: range_checker.clone(),
+                tuple_range_checker: tuple_range_checker.map(|tuple_range_checker| {
+                    SharedRangeTupleCheckerChip::new(*tuple_range_checker.bus())
+                }),
```
qwang98, line 46: Why make new ones here.

openvm/src/powdr_extension/vm.rs
```diff
@@ -196,28 +196,24 @@ impl<P: IntoOpenVm> VmExtension<OpenVmField<P>> for PowdrExtension<P> {
...
                     precompile.clone(),
                     offline_memory.clone(),
                     self.base_config.clone(),
-                    SharedChips::new(
-                        bitwise_lookup.clone(),
-                        range_checker.clone(),
-                        tuple_range_checker.cloned(),
-                    ),
+                    shared_chips_pair.clone(),
                     self.bus_map.clone(),
```
qwang98, line 217: To summarize:
1. Range checker is shared across all APC but not shared with main execution.
2. Bitwise lookup and tuple range checker is spawned for each APC and not shared with main execution.
3. Memory is shared across all APC and across the main execution.

Is this correct understanding?

openvm/src/lib.rs
Schaeff: Fine with that

openvm/src/powdr_extension/vm.rs
Schaeff: No:

- Range checker and memory are shared across main execution and all apcs
- All other periphery chips are shared across all apcs, but not with main execution

Schaeff: This is already the case before  this PR, with the difference that the second bullet point has one instance per apc

openvm/src/powdr_extension/executor/periphery.rs
Schaeff: Added a comment here

openvm/src/powdr_extension/executor/periphery.rs
Schaeff: Added comment, see previous

openvm/src/powdr_extension/executor/mod.rs
Schaeff: Explained in the other comment. In other words, if we have a single periphery, we need to do:
```
periphery = side_effects_of_original_execution - side_effects_of_dummy_rows + side_effects_of_apc_rows
```
so we execute, then rollback the side effects of dummy rows, and then aply the side effects of the apc rows. This is what happens for the range checker, because we have to share it with main execution, because it's used by memory*.

Now, by using a dummy periphery when executing the apc blocks, it means that `side_effects_of_original_execution` above does *not* include side effects of the apc blocks. Therefore, `side_effects_of_dummy_rows` is empty. So we do not need to roll back anything. We only need to apply `side_effects_of_apc_rows`.

Does this make sense?

*I'm not completely sure about this, maybe we can also have a dummy range checker, in which case we save the time spent rolling back in trace generation.

openvm/src/powdr_extension/executor/mod.rs
Schaeff: Yes

openvm/src/powdr_extension/executor/inventory.rs
Schaeff: Good catch, it works!

openvm/src/powdr_extension/executor/mod.rs
Schaeff: Just tried using a dummy range checker as well, but that doesn't work because it is used by memory for timestamp range checks.

PR #2872, opened 2025-06-10
Move bus map

qwang98:
Solves #2856.

### DIFF
--- a/autoprecompiles/src/lib.rs
+++ b/autoprecompiles/src/lib.rs
@@ -180,7 +180,7 @@ impl<T: FieldElement> TryFrom<SymbolicBusInteraction<T>> for PcLookupBusInteract
     type Error = ();
 
     fn try_from(bus_interaction: SymbolicBusInteraction<T>) -> Result<Self, ()> {
-        (bus_interaction.id == PC_LOOKUP_BUS_ID)
+        (bus_interaction.id == DEFAULT_PC_LOOKUP)
             .then(|| {
                 let from_pc = bus_interaction.args[0].clone();
                 let op = bus_interaction.args[1].clone();
@@ -196,12 +196,92 @@ impl<T: FieldElement> TryFrom<SymbolicBusInteraction<T>> for PcLookupBusInteract
     }
 }
 
-pub const EXECUTION_BUS_ID: u64 = 0;
-pub const MEMORY_BUS_ID: u64 = 1;
-pub const PC_LOOKUP_BUS_ID: u64 = 2;
-pub const VARIABLE_RANGE_CHECKER_BUS_ID: u64 = 3;
-pub const BITWISE_LOOKUP_BUS_ID: u64 = 6;
-pub const TUPLE_RANGE_CHECKER_BUS_ID: u64 = 7;
+pub const DEFAULT_EXECUTION_BRIDGE: u64 = 0;
+pub const DEFAULT_MEMORY: u64 = 1;
+pub const DEFAULT_PC_LOOKUP: u64 = 2;
+pub const DEFAULT_VARIABLE_RANGE_CHECKER: u64 = 3;
+pub const DEFAULT_BITWISE_LOOKUP: u64 = 6;
+pub const DEFAULT_TUPLE_RANGE_CHECKER: u64 = 7;
+
+#[derive(Debug, Copy, Clone, Deserialize, Serialize, PartialEq)]
+pub enum BusType {
+    ExecutionBridge,
+    Memory,
+    PcLookup,
+    VariableRangeChecker,
+    BitwiseLookup,
+    TupleRangeChecker,
+    Sha,
+}
+
+impl std::fmt::Display for BusType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let name = match self {
+            BusType::ExecutionBridge => "EXECUTION_BRIDGE",
+            BusType::Memory => "MEMORY",
+            BusType::PcLookup => "PC_LOOKUP",
+            BusType::VariableRangeChecker => "VARIABLE_RANGE_CHECKER",
+            BusType::BitwiseLookup => "BITWISE_LOOKUP",
+            BusType::TupleRangeChecker => "TUPLE_RANGE_CHECKER",
+            BusType::Sha => "SHA",
+        };
+        write!(f, "{name}")
+    }
+}
+
+#[derive(Clone, Deserialize, Serialize)]
+pub struct BusMap {
+    bus_ids: BTreeMap<u64, BusType>,
+}
+
+impl BusMap {
+    pub fn openvm_base() -> Self {
+        let bus_ids = [
+            (DEFAULT_EXECUTION_BRIDGE, BusType::ExecutionBridge),
+            (DEFAULT_MEMORY, BusType::Memory),
+            (DEFAULT_PC_LOOKUP, BusType::PcLookup),
+            (
+                DEFAULT_VARIABLE_RANGE_CHECKER,
+                BusType::VariableRangeChecker,
+            ),
+            (DEFAULT_BITWISE_LOOKUP, BusType::BitwiseLookup),
+            (DEFAULT_TUPLE_RANGE_CHECKER, BusType::TupleRangeChecker),
+        ]
+        .into_iter()
+        .collect();
+
+        Self { bus_ids }
+    }
+
+    pub fn bus_type(&self, bus_id: u64) -> BusType {
+        self.bus_ids[&bus_id]
+    }
+
+    pub fn with_sha(mut self, id: u64) -> Self {
+        self.bus_ids.insert(id, BusType::Sha);
+        self
+    }
+
+    pub fn with_bus_type(mut self, id: u64, bus_type: BusType) -> Self {
+        self.bus_ids.insert(id, bus_type);
+        self
+    }
+
+    pub fn with_bus_map(mut self, bus_map: BusMap) -> Self {
+        self.bus_ids.extend(bus_map.bus_ids);
+        self
+    }
+
+    pub fn inner(&self) -> &BTreeMap<u64, BusType> {
+        &self.bus_ids
+    }
+
+    pub fn get_bus_id(&self, bus_type: &BusType) -> Option<u64> {
+        self.bus_ids
+            .iter()
+            .find_map(|(id, bus)| if bus == bus_type { Some(*id) } else { None })
+    }
+}
 
 /// A configuration of a VM in which execution is happening.
 pub struct VmConfig<'a, T: FieldElement, B> {
@@ -209,7 +289,8 @@ pub struct VmConfig<'a, T: FieldElement, B> {
     pub instruction_machines: &'a BTreeMap<usize, SymbolicMachine<T>>,
     /// The bus interaction handler, used by the constraint solver to reason about bus interactions.
     pub bus_interaction_handler: B,
-    // TODO: Add bus map
+    /// The bus map that maps bus id to bus type
+    pub bus_map: BusMap,
 }
 
 pub fn build<T: FieldElement, B: BusInteractionHandler<T> + IsBusStateful<T> + Clone>(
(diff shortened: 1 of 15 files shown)


autoprecompiles/src/lib.rs
```diff
@@ -189,27 +190,108 @@ impl<T: FieldElement> TryFrom<SymbolicBusInteraction<T>> for PcLookupBusInteract
...
     }
 }
 
-pub const EXECUTION_BUS_ID: u64 = 0;
-pub const MEMORY_BUS_ID: u64 = 1;
-pub const PC_LOOKUP_BUS_ID: u64 = 2;
-pub const VARIABLE_RANGE_CHECKER_BUS_ID: u64 = 3;
-pub const BITWISE_LOOKUP_BUS_ID: u64 = 6;
-pub const TUPLE_RANGE_CHECKER_BUS_ID: u64 = 7;
+pub const DEFAULT_EXECUTION_BRIDGE: u64 = 0;
```
chriseth: Maybe move all these into a file `bus_map.rs`? I think `lib.rs` files should be as minimal as possible, because everything they define is `pub` inside the module.

autoprecompiles/src/lib.rs
```diff
@@ -189,27 +190,108 @@ impl<T: FieldElement> TryFrom<SymbolicBusInteraction<T>> for PcLookupBusInteract
...
+        .collect();
+
+        Self { bus_ids }
+    }
+
+    pub fn bus_type(&self, bus_id: u64) -> BusType {
+        self.bus_ids[&bus_id]
+    }
+
+    pub fn with_sha(mut self, id: u64) -> Self {
```
chriseth: please document these functions. Maybe it would also make sense to define a `BusMap` trait and separate that from specific implementations like the default openvm bus configuration?

autoprecompiles/src/lib.rs
```diff
@@ -189,27 +190,108 @@ impl<T: FieldElement> TryFrom<SymbolicBusInteraction<T>> for PcLookupBusInteract
...
+        self.bus_ids.insert(id, bus_type);
+        self
+    }
+
+    pub fn with_bus_map(mut self, bus_map: BusMap) -> Self {
+        self.bus_ids.extend(bus_map.bus_ids);
+        self
+    }
+
+    pub fn inner(&self) -> &BTreeMap<u64, BusType> {
```
chriseth: Maybe `all_types_by_id`?

autoprecompiles/src/lib.rs
```diff
@@ -189,27 +190,108 @@ impl<T: FieldElement> TryFrom<SymbolicBusInteraction<T>> for PcLookupBusInteract
...
+        &self.bus_ids
+    }
+
+    pub fn get_bus_id(&self, bus_type: &BusType) -> Option<u64> {
```
chriseth: Does it make sense to only return the first of the type? Should we add that as a restriction (only one bus id per type)?

autoprecompiles/src/lib.rs
chriseth: Ok maybe a trait is overkill, but what about defining free-standing functions inside an openvm sub-module that returns the default openvm bus map?

autoprecompiles/src/memory_optimizer.rs
```diff
@@ -99,8 +105,9 @@ impl<T: FieldElement> MemoryBusInteraction<T> {
     /// Otherwise returns `Ok(Some(memory_bus_interaction))`
     fn try_from_symbolic_bus_interaction(
         bus_interaction: &SymbolicBusInteraction<T>,
+        bus_map: &BusMap,
     ) -> Result<Option<Self>, ()> {
-        if bus_interaction.id != MEMORY_BUS_ID {
+        if bus_interaction.id != bus_map.get_bus_id(&BusType::Memory).unwrap() {
```
chriseth: I think it would be better to supply the bus id as argument instead of the bus map

autoprecompiles/src/memory_optimizer.rs
```diff
@@ -15,16 +15,19 @@ use powdr_number::{FieldElement, LargeInt};
...
 /// Optimizes bus sends that correspond to general-purpose memory read and write operations.
 /// It works best if all read-write-operation addresses are fixed offsets relative to some
 /// symbolic base address. If stack and heap access operations are mixed, this is usually violated.
-pub fn optimize_memory<T: FieldElement>(mut machine: SymbolicMachine<T>) -> SymbolicMachine<T> {
-    let (to_remove, new_constraints) = redundant_memory_interactions_indices(&machine);
+pub fn optimize_memory<T: FieldElement>(
+    mut machine: SymbolicMachine<T>,
+    bus_map: &BusMap,
+) -> SymbolicMachine<T> {
+    let (to_remove, new_constraints) = redundant_memory_interactions_indices(&machine, bus_map);
```
chriseth: We could do a single lookup of the memory bus id at this point and if there is no memory, we could just return the unmodified machine.

autoprecompiles/src/optimizer.rs
```diff
@@ -97,10 +99,11 @@ fn system_size<T: FieldElement>(
 pub fn optimize_pc_lookup<T: FieldElement>(
     mut machine: SymbolicMachine<T>,
     opcode: u32,
+    bus_map: &BusMap,
 ) -> SymbolicMachine<T> {
     let mut first_pc = None;
```
chriseth, line 117: Although unlikely, I would also retrieve the pc lookup bus id here and return the unmodified machine if there is no pc lookup bus.

autoprecompiles/src/optimizer.rs
```diff
@@ -124,14 +127,17 @@ pub fn optimize_pc_lookup<T: FieldElement>(
     machine
 }
 
-pub fn optimize_exec_bus<T: FieldElement>(mut machine: SymbolicMachine<T>) -> SymbolicMachine<T> {
+pub fn optimize_exec_bus<T: FieldElement>(
+    mut machine: SymbolicMachine<T>,
+    bus_map: &BusMap,
+) -> SymbolicMachine<T> {
     let mut first_seen = false;
```
chriseth, line 147: Same here.

autoprecompiles/src/lib.rs
qwang98: sg will do

autoprecompiles/src/memory_optimizer.rs
qwang98: sg will do

autoprecompiles/src/optimizer.rs
qwang98: sg

autoprecompiles/src/lib.rs
qwang98: How would it be different from just calling `BusMap::openvm_base()`? I mean it's not a free standing function but if I were to create one probably just calls `BusMap::openvm_base()` inside.

qwang98: My interpretation is that your proposal "sort of" separates the implementation from `BusMap`? Basically we will remove the `openvm_base()` and the free-standing function will create the same `BusMap` using other APIs that aren't specific to any "implementation" of `BusMap`?

autoprecompiles/src/lib.rs
qwang98: Yeah. I agree. This is a reverse lookup but we haven't really enforced that the values are unique.

autoprecompiles/src/lib.rs
qwang98: Done.

autoprecompiles/src/lib.rs
qwang98: This has been done.

Also I originally proposed defining a trait in powdr-apc and implement it in powdr-ovm but after talking to Georg we decided not to go that route till in the future when we have the apc fully abstracted away from ovm.

autoprecompiles/src/lib.rs
qwang98: Done.

autoprecompiles/src/memory_optimizer.rs
qwang98: Done.

autoprecompiles/src/memory_optimizer.rs
qwang98: Done.

autoprecompiles/src/optimizer.rs
qwang98: Done.

autoprecompiles/src/optimizer.rs
qwang98: Done.

autoprecompiles/src/lib.rs
qwang98: This has been done by enforcing uniqueness on all methods that builds or mutates `BusMap`, so uniqueness is ensured for all view and lookup methods like this.

autoprecompiles/src/bus_map.rs
```diff
@@ -0,0 +1,92 @@
...
+        }
+        BusMap { bus_ids }
+    }
+
+    /// Lookup the `BusType` for a given ID.
+    pub fn bus_type(&self, bus_id: u64) -> BusType {
+        self.bus_ids[&bus_id]
+    }
+
+    /// Insert a new SHA bus with given ID, ensuring uniqueness.
```
chriseth: How does it ensure it?

autoprecompiles/src/bus_map.rs
```diff
@@ -0,0 +1,92 @@
...
+        self
+    }
+
+    /// Insert a new bus type under the given ID, ensuring no other ID
+    /// already maps to the same `BusType` value.
+    pub fn with_bus_type(mut self, id: u64, bus_type: BusType) -> Self {
+        if let Some(existing) = self.get_bus_id(&bus_type) {
+            panic!("BusType `{bus_type:?}` already exists under ID `{existing}`");
+        }
+        self.bus_ids.insert(id, bus_type);
```
chriseth, line 60:
```suggestion
        if let Some(existing) = self.bus_ids.insert(id, bus_type) {
            panic!("BusType `{bus_type:?}` already exists under ID `{existing}`");
        }
```

autoprecompiles/src/bus_map.rs
```diff
@@ -0,0 +1,92 @@
...
+        BusMap { bus_ids }
+    }
+
+    /// Lookup the `BusType` for a given ID.
+    pub fn bus_type(&self, bus_id: u64) -> BusType {
+        self.bus_ids[&bus_id]
+    }
+
+    /// Insert a new SHA bus with given ID, ensuring uniqueness.
+    pub fn with_sha(mut self, id: u64) -> Self {
```
chriseth: Why is sha special? Callers could just use `with_bus_type(id, BusType::Sha)`.

autoprecompiles/src/bus_map.rs
```diff
@@ -0,0 +1,92 @@
...
+        if let Some(existing) = self.get_bus_id(&bus_type) {
+            panic!("BusType `{bus_type:?}` already exists under ID `{existing}`");
+        }
+        self.bus_ids.insert(id, bus_type);
+        self
+    }
+
+    /// Extend with another `BusMap`, ensuring no duplicate `BusType`s between them.
+    pub fn with_bus_map(mut self, other: BusMap) -> Self {
+        for bus_type in other.bus_ids.values() {
```
chriseth: Could just iteratively call `with_bus_type`, then we would have less code duplication

autoprecompiles/src/bus_map.rs
```diff
@@ -0,0 +1,92 @@
...
+        self.bus_ids.extend(other.bus_ids);
+        self
+    }
+
+    /// View the entire map.
+    pub fn all_types_by_id(&self) -> &BTreeMap<u64, BusType> {
+        &self.bus_ids
+    }
+
+    /// Find the ID for a given `BusType` (if any), ensuring uniqueness.
```
chriseth:
```suggestion
    /// Find the ID for a given `BusType` (if any).
```

autoprecompiles/src/openvm.rs
```diff
@@ -0,0 +1,26 @@
+// To support an abstracted autoprecompile layer, this module stores type implementations specific to OpenVM
```
chriseth:
```suggestion
//! To support an abstracted autoprecompile layer, this module stores type implementations specific to OpenVM
```

autoprecompiles/src/bus_map.rs
qwang98: I'm good with removing as it's not the most accurate description, but FYI this is technically insured in all mut methods so not necessarily in this function.

autoprecompiles/src/bus_map.rs
qwang98: Agree on this. It's just some legacy stuff moved over from the OVM crate, so I'm good with removing.

autoprecompiles/src/bus_map.rs
qwang98: I was planning to use the `with_bus_type` method which ensures uniqueness but some how it didn't get in the commit. I'm good with removing this method entirely tho.

autoprecompiles/src/bus_map.rs
qwang98: Done.

autoprecompiles/src/bus_map.rs
qwang98: Done.

autoprecompiles/src/bus_map.rs
qwang98: Done.

autoprecompiles/src/openvm.rs
qwang98: Done.

chriseth (approved):

PR #2874, opened 2025-06-10
Use journalled constraint system for memory and bitwise

chriseth:


pacheco (commented):
what's the purpose of the journalling again?

autoprecompiles/src/memory_optimizer.rs
```diff
@@ -19,21 +20,20 @@ const REGISTER_ADDRESS_SPACE: u32 = 1;
...
-        .filter_map(|(i, bus)| (!to_remove.contains(&i)).then_some(bus))
-        .collect();
-    // TODO perform substitutions instead
-    system.algebraic_constraints.extend(new_constraints);
+    let mut counter = 0;
+    system.retain_bus_interactions(|_| {
+        let retain = !to_remove.contains(&counter);
+        counter += 1;
+        retain
+    });
```
Schaeff, line 35: Maybe `retain_bus_interactions` could take a closure that takes `(usize, bus_interaction)` instead?

autoprecompiles/src/optimizer.rs
```diff
@@ -44,43 +44,44 @@ pub fn optimize<T: FieldElement>(
...
         constraint_system = optimization_loop_iteration(
             constraint_system,
             bus_interaction_handler.clone(),
             degree_bound,
             &mut stats_logger,
             bus_map,
         )?;
-        if stats == stats_logger::Stats::from(&constraint_system) {
-            return Ok(constraint_system_to_symbolic_machine(constraint_system));
+        if stats == stats_logger::Stats::from(constraint_system.system()) {
```
Schaeff, line 59: Maybe the `AsRef` trait could be useful here to be able to pass `constraint_system` directly

chriseth:
@pacheco the purpose of the journal is to be able to match items of the final constraint system back to the original constraint system, triggered by witgen, where it was important to know which bus interactions got removed and which are newly added.

In the meantime (realized today), I think this could also be useful to output a trace of optimization steps which can then be formally verified more easily.

autoprecompiles/src/memory_optimizer.rs
chriseth: It's all horrible. Because there situations where we would want to use the index after removal and there are situations where the index before removal would be better. I'm thinking about passing a vector of indices instead...

autoprecompiles/src/optimizer.rs
chriseth: Hm, I tried, but it's complicated and somehow we would still need the `.system()` function anyway since we call `c.system().clone()` and stuff.

constraint-solver/src/journaling_constraint_system.rs
```diff
@@ -92,6 +92,26 @@ impl<T: FieldElement, V> JournalingConstraintSystem<T, V> {
     }
 }
 
+impl<T: FieldElement, V: Clone + Ord + Hash> JournalingConstraintSystem<T, V> {
+    /// Adds a new algebraic constraints to the system.
```
leonardoalt, line 96:
```suggestion
    /// Adds new algebraic constraints to the system.
```

constraint-solver/src/journaling_constraint_system.rs
```diff
@@ -92,6 +92,26 @@ impl<T: FieldElement, V> JournalingConstraintSystem<T, V> {
...
+    pub fn add_algebraic_constraints(
+        &mut self,
+        constraints: impl IntoIterator<Item = QuadraticSymbolicExpression<T, V>>,
+    ) {
+        // TODO track
+        self.system.add_algebraic_constraints(constraints);
+    }
+
+    /// Adds a new bus interactions to the system.
```
leonardoalt, line 105:
```suggestion
    /// Adds new bus interactions to the system.
```

constraint-solver/src/journaling_constraint_system.rs
```diff
@@ -92,6 +92,26 @@ impl<T: FieldElement, V> JournalingConstraintSystem<T, V> {
...
+    pub fn add_bus_interactions(
+        &mut self,
+        bus_interactions: impl IntoIterator<Item = BusInteraction<QuadraticSymbolicExpression<T, V>>>,
+    ) {
+        // TODO track
```
leonardoalt, line 110: In this PR or later?

constraint-solver/src/journaling_constraint_system.rs
```diff
@@ -92,6 +92,26 @@ impl<T: FieldElement, V> JournalingConstraintSystem<T, V> {
     }
 }
 
+impl<T: FieldElement, V: Clone + Ord + Hash> JournalingConstraintSystem<T, V> {
+    /// Adds a new algebraic constraints to the system.
+    pub fn add_algebraic_constraints(
+        &mut self,
+        constraints: impl IntoIterator<Item = QuadraticSymbolicExpression<T, V>>,
+    ) {
+        // TODO track
```
leonardoalt, line 101: In this PR or later?

constraint-solver/src/journaling_constraint_system.rs
Schaeff: bitte was

autoprecompiles/src/memory_optimizer.rs
qwang98, line 35: Maybe for the next PR but can `redundant_memory_interactions_indice` be adjusted to become an input of `retain_bus_interactions` instead of via `to_remove`?

qwang98: This way it follows how `retain_bus_interactions` has been used else where.

constraint-solver/src/journaling_constraint_system.rs
qwang98: Is the "tracking" here gonna include something like an "addition map" similar to the "replacement map" in the case of removal?

qwang98:
Probably a few PRs down the road but where will the "tracked optimization" objects be stored eventually I wonder? :)

PR #2877, opened 2025-06-11
Add constraints to indexed constraint system.

chriseth:


constraint-solver/src/constraint_system.rs
```diff
@@ -68,6 +68,14 @@ impl<T: FieldElement, V> ConstraintSystem<T, V> {
             ),
         )
     }
+
+    /// Extends the constraint system by the constraints of another system.
+    /// No de-duplication is performed.
+    pub fn extend(&mut self, system: ConstraintSystem<T, V>) {
```
Schaeff, line 74: I almost want to suggest not having this function and rather inlining it into `IndexedConstraintSystem::extend`, since it's not a lot of code and it is easy to misuse if exposed like this

chriseth: Wait how would you misuse it? Also note that the fields of this struct are `pub`, so it's not safe either way.

Schaeff: Oh, I thought by "no-deduplication is performed" you meant that one could output an invalid CS, but it would just have duplicated constraints? Then that's fine, no problem!

Schaeff (approved):

PR #2878, opened 2025-06-11
Use StatsLogger stats for fixpoint detection.

chriseth:


autoprecompiles/src/optimizer.rs
```diff
@@ -48,15 +47,15 @@ pub fn optimize<T: FieldElement>(
     let mut constraint_system = symbolic_machine_to_constraint_system(machine);
 
     loop {
-        let size = system_size(&constraint_system);
+        let size = stats_logger::Stats::from(&constraint_system);
```
Schaeff:
```suggestion
        let stats = stats_logger::Stats::from(&constraint_system);
```

?

Schaeff (approved):

PR #2885, opened 2025-06-12
Refactor bus map retrieval

Schaeff:
We currently include the bus map in the powdr config. However, it is not a parameter we can chose, but rather depends on the original vm. Move it there, still returning the default one. Next step would be ideally generating it from the vm config.

Also use `OriginalProgram` when possible, to keep the program and config together.


openvm/src/lib.rs
```diff
@@ -634,37 +630,42 @@ where
     // Note that we could use chip_complex.inventory.available_opcodes() instead of used_instructions,
     // which depends on the program being executed. But this turns out to be heavy on memory, because
     // it includes large precompiles like Keccak.
-    used_instructions
```
Schaeff, line 637: only indentation changes due to returning a tuple

PR #2889, opened 2025-06-12
Move openvm default bus map out of apc crate

Schaeff:


openvm/src/lib.rs
```diff
@@ -62,8 +62,11 @@ use crate::traits::OpenVmField;
 use crate::utils::symbolic_to_algebraic;
 
 mod air_builder;
+pub mod bus_map;
 mod utils;
 
+pub use bus_map::default_openvm_bus_map;
```
pacheco: nitpick: is the `pub` needed? imports seem to refer to the `bus_map` module

Schaeff: Good catch, let's always go through the module!

qwang98 (approved):
LGTM. Also merging this now because #2876 will depend on it.

PR #2895, opened 2025-06-13
Dont skip non apc instruction

chriseth:
If we don't put the non-apc instruction into at least a basic block, then the address list is not complete and we cannot do jump analysis.


openvm/src/customize_exe.rs
```diff
@@ -144,18 +145,28 @@ pub fn customize<P: IntoOpenVm>(
...
 
-    let mut blocks =
-        collect_basic_blocks(&exe.program, &labels, &opcodes_no_apc, &branch_opcodes_set);
+    let blocks = collect_basic_blocks(&exe.program, &labels, &opcodes_no_apc, &branch_opcodes_set);
     tracing::info!(
         "Got {} basic blocks from `collect_basic_blocks`",
         blocks.len()
     );
 
+    let mut blocks = blocks
```
chriseth, line 161: We don't filter right away because in between here I plan to do the analysis.

PR #2896, opened 2025-06-13
Actually store journal.

chriseth:


constraint-solver/src/journaling_constraint_system.rs
```diff
@@ -76,19 +103,37 @@ impl<T: FieldElement, V> JournalingConstraintSystem<T, V> {
...
+        self.system.retain_algebraic_constraints(|expr| {
+            let retain = f(expr);
+            if !retain {
+                self.operations
+                    .push(Operation::RemoveAlgebraicConstraint { index });
+            } else {
+                index += 1;
+            }
+            retain
+        });
```
qwang98, line 118: Nice that this is as granular as a single removal instead of a whole optimization :)

constraint-solver/src/journaling_constraint_system.rs
```diff
@@ -9,6 +9,26 @@ use std::{fmt::Display, hash::Hash};
...
 pub struct JournalingConstraintSystem<T: FieldElement, V> {
     system: IndexedConstraintSystem<T, V>,
+    operations: Vec<Operation<T, V>>,
+}
+
+pub enum Operation<T: FieldElement, V> {
+    Substitution {
+        variable: V,
+        substitution: QuadraticSymbolicExpression<T, V>,
+    },
```
qwang98, line 19: Dummy question: on the high level which optimization(s) typically use substitution?

chriseth:
This is too outdated now.

PR #2897, opened 2025-06-13
Transform simplified.

chriseth:


chriseth:
The goal is to get rid of some code duplication as well. But I think it would be better to just transform the QSE to an `Expression<String>` and then format it.

PR #2909, opened 2025-06-17
Make pil export work with unknown bus ids

ShuangWu121:
This PR changes `BusMap: bus_ids: BTreeMap<u64, BusType>` to `BusMap: bus_ids: BTreeMap<u64, Option<BusType>>`, by doing this we can avoid adding CopyConstraintLookup into BusType enum when adding this new bus type by plonk chip, get_pil can behave correctly, see discussion [here](https://github.com/powdr-labs/powdr/pull/2871#discussion_r2149829356)

### DIFF
--- a/autoprecompiles/src/bus_map.rs
+++ b/autoprecompiles/src/bus_map.rs
@@ -29,7 +29,7 @@ impl std::fmt::Display for BusType {
 
 #[derive(Clone, Deserialize, Serialize)]
 pub struct BusMap {
-    bus_ids: BTreeMap<u64, BusType>,
+    pub bus_ids: BTreeMap<u64, BusType>,
 }
 
 impl BusMap {
--- a/autoprecompiles/src/bus_map.rs
+++ b/autoprecompiles/src/bus_map.rs
@@ -29,7 +29,7 @@ impl std::fmt::Display for BusType {
 
 #[derive(Clone, Deserialize, Serialize)]
 pub struct BusMap {
-    pub bus_ids: BTreeMap<u64, BusType>,
+    bus_ids: BTreeMap<u64, BusType>,
 }
 
 impl BusMap {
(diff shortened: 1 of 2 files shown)


Schaeff:
I think this is too invasive. If the issue is in `export_pil`, it should be fixed there. 
- From `get_constraints` in `export_pil` we have all bus interactions.
- From the `bus_map` we have the bus interactions that apc understands

By crossing these two sets, we should be able to output all bus interactions, marking the ones that apc does not understand as `unknown` or something.

autoprecompiles/src/bus_map.rs
```diff
@@ -29,7 +29,7 @@ impl std::fmt::Display for BusType {
 
 #[derive(Clone, Deserialize, Serialize)]
 pub struct BusMap {
-    bus_ids: BTreeMap<u64, BusType>,
+    pub bus_ids: BTreeMap<u64, BusType>,
```
Schaeff: There is a method which returns a reference to the internal map, so this would not be required.

openvm/src/utils.rs
```diff
@@ -203,7 +203,7 @@ namespace {name};
         pil.push_str(&format!("    col witness {column};\n"));
     }
 
-    let bus_interactions_by_bus = constraints
+    let mut bus_interactions_by_bus = constraints
```
Schaeff: If I understand correctly, to build `A` and `B`, this is:
- building `A + B`
- then building `B` from that
- then removing `B` from `A + B` to end up with `A`

Would it be possible to build `A` and `B` directly, for example using the `Iterator::partition` method?

autoprecompiles/src/bus_map.rs
ShuangWu121: Done.

openvm/src/utils.rs
ShuangWu121: Done. thanks for the tip!

PR #2914, opened 2025-06-17
Clean up API

Schaeff:
Some refactoring:
- avoid collecting a large string in `export_pil`, add basic test
- introduce `OriginalVmConfig` which wraps SdkVmConfig and serves airs, bus map, etc, with internal caching
- refactor to make the external API higher level. `compile_exe_with_elf` can be used directly by `openvm-reth-benchmark`, see https://github.com/powdr-labs/openvm-reth-benchmark/pull/7

### DIFF
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -2,7 +2,7 @@ use std::collections::HashSet;
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::sync::Arc;
 
-use crate::extraction_utils::SdkVmConfigWrapper;
+use crate::extraction_utils::OriginalVmConfig;
 use crate::utils::UnsupportedOpenVmReferenceError;
 use crate::IntoOpenVm;
 use crate::OpenVmField;
@@ -70,9 +70,9 @@ pub fn customize(
     config: PowdrConfig,
     pgo_config: PgoConfig,
 ) -> CompiledProgram<BabyBearField> {
-    let config_wrapper = SdkVmConfigWrapper::new(sdk_vm_config.clone());
-    let (airs, bus_map) =
-            (config_wrapper.airs().expect("Failed to convert the AIR of an OpenVM instruction, even after filtering by the blacklist!"), config_wrapper.bus_map());
+    let config_wrapper = OriginalVmConfig::new(sdk_vm_config.clone());
+    let airs = config_wrapper.airs().expect("Failed to convert the AIR of an OpenVM instruction, even after filtering by the blacklist!");
+    let bus_map = config_wrapper.bus_map();
 
     // If we use PgoConfig::Cell, which creates APC for all eligible basic blocks,
     // `apc_cache` will be populated to be used later when we select which basic blocks to accelerate.
@@ -238,8 +238,8 @@ pub fn customize(
     }
 
     CompiledProgram {
-        vm_config: SpecializedConfig::new(config_wrapper, extensions, config.implementation),
         exe,
+        vm_config: SpecializedConfig::new(config_wrapper, extensions, config.implementation),
     }
 }
 
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -65,11 +65,11 @@ pub fn customize(
     OriginalCompiledProgram {
         mut exe,
         sdk_vm_config,
-    }: OriginalCompiledProgram<BabyBearField>,
+    }: OriginalCompiledProgram,
     labels: &BTreeSet<u32>,
     config: PowdrConfig,
     pgo_config: PgoConfig,
-) -> CompiledProgram<BabyBearField> {
+) -> CompiledProgram {
     let original_config = OriginalVmConfig::new(sdk_vm_config.clone());
     let airs = original_config.airs().expect("Failed to convert the AIR of an OpenVM instruction, even after filtering by the blacklist!");
     let bus_map = original_config.bus_map();
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -70,9 +70,9 @@ pub fn customize(
     config: PowdrConfig,
     pgo_config: PgoConfig,
 ) -> CompiledProgram<BabyBearField> {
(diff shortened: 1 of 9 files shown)


openvm/src/powdr_extension/executor/periphery.rs
```diff
@@ -18,28 +18,29 @@ pub struct PowdrPeripheryInstances {
 
 #[derive(Clone)]
 pub struct SharedPeripheryChips {
-    pub bitwise_lookup_8: SharedBitwiseOperationLookupChip<8>,
+    pub bitwise_lookup_8: Option<SharedBitwiseOperationLookupChip<8>>,
```
Schaeff, line 21: This change was introduced because technically we can have an SdkVm without a lookup chip, see test for export_pil

openvm/src/powdr_extension/plonk/chip.rs
```diff
@@ -48,14 +48,14 @@ impl<P: IntoOpenVm> PlonkChip<P> {
     #[allow(dead_code)]
     pub(crate) fn new(
         precompile: PowdrPrecompile<P>,
+        original_airs: BTreeMap<usize, SymbolicMachine<P>>,
```
Schaeff, line 52: We do not store the original airs in the precompile anymore, since they are common to all precompiles. We pass them here instead.

openvm/src/powdr_extension/chip.rs
```diff
@@ -55,14 +55,14 @@ pub struct PowdrChip<P: IntoOpenVm> {
 impl<P: IntoOpenVm> PowdrChip<P> {
     pub(crate) fn new(
         precompile: PowdrPrecompile<P>,
+        original_airs: BTreeMap<usize, powdr_autoprecompiles::SymbolicMachine<P>>,
```
Schaeff, line 58: Same here

openvm/src/powdr_extension/vm.rs
```diff
@@ -42,6 +42,7 @@ pub struct PowdrExtension<P: IntoOpenVm> {
     pub base_config: SdkVmConfig,
     pub implementation: PrecompileImplementation,
     pub bus_map: BusMap,
+    pub airs: BTreeMap<usize, SymbolicMachine<P>>,
```
Schaeff, line 45: The airs are common to all precompiles, so we store them a single time here

openvm/src/customize_exe.rs
```diff
@@ -61,17 +61,19 @@ impl From<powdr_autoprecompiles::constraint_optimizer::Error> for Error {
     }
 }
 
-pub fn customize<P: IntoOpenVm>(
+pub fn customize(
```
Schaeff, line 64: We lose field genericity here because air extraction is not generic. Field genericity is not a priority though.

openvm/src/extraction_utils.rs
```diff
@@ -117,46 +144,71 @@ pub fn get_airs_and_bus_map<P: IntoOpenVm>(
...
-    bus_map: &BusMap,
-) where
-    VC::Executor: Chip<BabyBearSC>,
-    VC::Periphery: Chip<BabyBearSC>,
-{
+pub fn export_pil(
+    writer: &mut impl std::io::Write,
+    vm_config: SpecializedConfig<powdr_number::BabyBearField>,
+) {
+    let blacklist: [&'static str; 1] = ["KeccakVmAir"];
```
Schaeff: I don't see a reason to expose this blacklist outside this function.

openvm/src/lib.rs
```diff
@@ -355,34 +383,15 @@ pub fn compile_exe(
...
-        .filter(|opcode| !blacklist.contains(&opcode.as_usize()))
-        .collect();
-    let (airs, bus_map) =
-        get_airs_and_bus_map(original_program.sdk_vm_config.clone(), &used_instructions)
-            .expect("Failed to convert the AIR of an OpenVM instruction, even after filtering by the blacklist!");
-
-    let sdk_vm_config = original_program.sdk_vm_config.clone();
-
-    let (exe, extension) = customize_exe::customize(
+    let compiled_program = customize_exe::customize(
```
Schaeff: This is where a lot of the simplification happens.

openvm/src/customize_exe.rs
```diff
@@ -61,17 +61,19 @@ impl From<powdr_autoprecompiles::constraint_optimizer::Error> for Error {
...
     OriginalCompiledProgram {
         mut exe,
         sdk_vm_config,
-    }: OriginalCompiledProgram<P>,
+    }: OriginalCompiledProgram,
     labels: &BTreeSet<u32>,
```
Schaeff, line 69: airs and bus_map are not passed anymore, they are generated inside this function based on the config

openvm/src/extraction_utils.rs
```diff
@@ -117,46 +149,81 @@ pub fn get_airs_and_bus_map<P: IntoOpenVm>(
...
+            )
+            .chain(
+                shared_range_tuple_checker
+                    .into_iter()
+                    .map(|chip| (chip.bus().inner.index, BusType::TupleRangeChecker)),
+            )
+            .map(|(id, bus_type)| (id as u64, bus_type)),
+        )
+    }
+
```
Schaeff, line 191: These free functions are moved into the implementation of `OriginalVmConfig`

openvm/src/customize_exe.rs
leonardoalt: sick!

openvm/src/extraction_utils.rs
```diff
@@ -117,46 +149,81 @@ pub fn get_airs_and_bus_map<P: IntoOpenVm>(
...
+        BusMap::from_id_type_pairs(
+            {
+                let base = &chip_complex.base;
+                [
+                    (base.execution_bus().inner.index, BusType::ExecutionBridge),
+                    (base.memory_bus().inner.index, BusType::Memory),
+                    (base.program_bus().inner.index, BusType::PcLookup),
+                    (
+                        base.range_checker_bus().inner.index,
```
leonardoalt, line 172: ah so this `inner.index` thing is the bus id?

Schaeff: oui

PR #2915, opened 2025-06-18
Use allow list for opcode

qwang98:
Currently we use a black list for opcode that shouldn't appear in a BB for APC. This method has the disadvantage of encountering many bugs due to new opcodes appearing that we can't create APC for (whether due to unsupported expression type or compile time memory overload).

With the new reth-benchmark, we identified a few custom opcodes (likely manual precompiles) that get injected in an APC, which causes an abnormal amount of memory usage up to the 260 gigs limit of our server. This calls for a systematic way to use an allow list for opcodes so that we are robust against future manual opcodes.

A few key changes:
1. Use a macro to derive the vector of allowed opcodes from public constants.
2. Allow list = basic opcodes - hint opcodes + bigint branch opcodes


openvm/src/opcode.rs
```diff
@@ -1,54 +1,62 @@
...
+      /// A slice containing *all* opcode constants.
+      pub const ALL_OPCODES: &[u32] = &[
+          $( $name ),*
+      ];
+  }
+}
+
+define_opcodes!(
+    // Rv32BaseAluChip
+    OPCODE_ADD = 512,
```
leonardoalt, line 20: I know these were there before, but why do they use constants instead of the OpenVM types? Feels less robust this way compared to the removed `instruction_blacklist` way

qwang98: Sounds good. I think maybe I'll inject a PR before this that takes all const from OVM instead.

leonardoalt: let's do it after so this PR has priority to unblock expeirments

leonardoalt (approved):

PR #2918, opened 2025-06-19
Make opcode constant derivation more robust

qwang98:
Replace hard coded opcode numbers with OVM imports.

### DIFF
--- a/openvm/src/opcode.rs
+++ b/openvm/src/opcode.rs
@@ -1,62 +1,80 @@
+use openvm_instructions::LocalOpcode;
+use openvm_rv32im_transpiler::*;
+
 /// Defines each opcode as a `pub const u32` and also generates
 /// a `pub const ALL_OPCODES: &[u32]` containing all of them.
 macro_rules! define_opcodes {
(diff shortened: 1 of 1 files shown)


openvm/src/opcode.rs
```diff
@@ -1,62 +1,80 @@
...
-          $( $name ),*
-      ];
-  }
+    (
+        $( $name:ident = $ty:ident :: $variant:ident, )*
+    ) => {
+        $(
+            // e.g. OPCODE_ADD = BaseAluOpcode::ADD as usize + BaseAluOpcode::CLASS_OFFSET
+            pub const $name: u32 =
+                ($ty::$variant as usize + <$ty as LocalOpcode>::CLASS_OFFSET) as u32;
```
qwang98, line 13: Somehow can't use `global_opcode()` method, so had to do it from scratch here.

leonardoalt: I guess because it isn't `const`

leonardoalt (approved):

PR #2919, opened 2025-06-19
Cache fewer APC for Pgo::Cell

qwang98:
Currently Pgo::Cell creates APCs for all BBs, but immediately after the sorting the first `n_acc` BBs are created again regardless of PgoType.

This PR isolates the apc_caching for the first `n_acc` BBs as a separate API that's only invoked by Pgo::Instruction and Pgo::None.

As per my reth bench mark test `MODE="execute"  APC=1000 APC_SKIP=0 PGO_TYPE=cell /usr/bin/time -v ./run.sh` PLUS `panic!` immediately after compilation, this reduces maximum resident size (memory consumption) from 92 gigs to 72 gigs.

### DIFF
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -111,7 +111,7 @@ pub fn customize(
     // 3. if PgoConfig::None, cost = number_of_instructions
     match pgo_config {
         PgoConfig::Cell(pgo_program_idx_count) => {
-            sort_blocks_by_pgo_cell_cost(
+            sort_blocks_by_pgo_cell_cost_and_cache_apc(
                 &mut blocks,
                 &mut apc_cache,
                 pgo_program_idx_count,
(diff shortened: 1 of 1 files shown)


openvm/src/customize_exe.rs
```diff
@@ -111,7 +111,7 @@ pub fn customize(
     // 3. if PgoConfig::None, cost = number_of_instructions
     match pgo_config {
         PgoConfig::Cell(pgo_program_idx_count) => {
-            sort_blocks_by_pgo_cell_cost(
+            sort_blocks_by_pgo_cell_cost_and_cache_apc(
```
qwang98, line 114: Different naming makes it clearer that we already cached APC here.

openvm/src/customize_exe.rs
```diff
@@ -145,42 +147,6 @@ pub fn customize(
...
-                    apc_opcode,
-                    bus_map.clone(),
-                    config.degree_bound,
-                )
-                .unwrap(),
-            )
-        })
-        .collect::<Vec<_>>();
-
-    apc_cache.extend(apcs);
```
qwang98, line 183: This has been done for all Pgo modes but it shouldn't be done for Pgo::Cell.

PR #2920, opened 2025-06-19
Cache even fewer APC for Pgo::Cell

qwang98:
Because we know the number of costliest APCs to compute at compile time, we only need to cache the top `APC + APC_SKIP` costliest APCs for Pgo::Cell mode instead of all of them. This is achieved by:
1. Maintaining a binary heap data structure which tracks the cost of blocks and if we are over the max number of cache, insert the new cache and eject the lowest costed block if the former is not the minimum costed block.
2. Still keeping APC generation parallel by adding mutex guard around the binary heap which is used for cost ordering as well as the map of APC caches.

### DIFF
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -576,8 +576,8 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
         max_cache,
     );
 
-    // each created apc becomes a candidate
-    // apc candidate carries all needed data but is only ordered by cost
+    // each generated apc becomes a candidate for caching
+    // it is only ordered by cost, but carries all needed data for modifying the blocks, extending the cache, and debug print
     struct ApcCandidate<P: IntoOpenVm> {
         block: BasicBlock<OpenVmField<P>>,
         apc_cache_entry: CachedAutoPrecompile<P>,
@@ -608,7 +608,7 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
     }
 
     // map–reduce over blocks into a single BinaryHeap<ApcCandidate<P>> capped at max_cache
-    // created from the min heap, so should be sorted already
+    // returned caches and blocks are ordered by descending cost already because they are unzipped from the min heap
     let (new_apc_cache, retained_blocks_with_stats): (
         HashMap<usize, CachedAutoPrecompile<_>>,
         Vec<_>,


leonardoalt:
Hmm this is starting to get complicated. I feel like we should find a simpler architecture for the whole thing

leonardoalt:
Let's discuss sync in the chat with @Schaeff

Schaeff:
We discussed it and there's a simpler parallel map reduce implementation.

openvm/src/customize_exe.rs
```diff
@@ -571,83 +567,140 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
+            Some(self.cmp(other))
+        }
+    }
+
+    // order nodes by reverse cost, so that our BinaryHeap pops the smallest costed node first
+    impl<P> Ord for Node<P> {
+        fn cmp(&self, other: &Self) -> Ordering {
+            other.cost.cmp(&self.cost)
+        }
+    }
```
qwang98: Seems a lot, but yes, each of these is needed.

Tried implementing with `Reverse(cost, start, apc_cache_entry, cells_saved)`, but this requires implementing `Ord` for each of the tuple elements including `apc_cache_entry`, which is not as simple as the `Ord` implementation here, because we only care about cost.

openvm/src/customize_exe.rs
```diff
@@ -571,83 +567,140 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
+        blocks.len(),
+        max_cache,
     );
 
-    let (cells_saved_per_row_by_bb, new_apc_cache): (
-        HashMap<_, _>,
-        HashMap<usize, CachedAutoPrecompile<P>>,
-    ) = blocks
+    // node carries all needed data but is only ordered by cost
+    struct Node<P> {
```
Schaeff:
```suggestion
    struct ApcCandidate<P> {
```

openvm/src/customize_exe.rs
```diff
@@ -571,83 +567,140 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
+        cost: u32,
+        start: usize,
+        entry: CachedAutoPrecompile<P>,
+        cells_saved: usize,
```
Schaeff: It's not immediately clear what these mean. Better naming and/or documentation would be helpful.

openvm/src/customize_exe.rs
Schaeff: Let's use `derive_more::{PartialEq, Eq, PartialOrd, Ord}` which have macros to derive these when comparing only some members.

openvm/src/customize_exe.rs
```diff
@@ -571,83 +567,140 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
+        .filter_map(|(i, block)| {
+            // attempt apc; if it errors, skip this block entirely
+            let entry = generate_apc_cache(
+                block,
                 airs,
                 POWDR_OPCODE + i,
                 bus_map.clone(),
                 config.degree_bound,
             )
             .ok()?;
```
Schaeff: We silently ignore the error here. If it's fine, let's add a comment.

openvm/src/customize_exe.rs
```diff
@@ -571,83 +567,140 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
+             }| { ((start, cells_saved), (start, entry)) },
+        )
         .unzip();
 
     apc_cache.extend(new_apc_cache);
 
-    // filter out the basic blocks where the apc generation errored out
-    blocks.retain(|b| cells_saved_per_row_by_bb.contains_key(&b.start_idx));
+    // filter to the top `max_cache` blocks by cost
+    blocks.retain(|b| cells_saved_by_bb.contains_key(&b.start_idx));
```
Schaeff: Can't we avoid this by putting the block inside the `Node` as well? Then the Nodes are already ranked by cells saved, and we just have to unzip them, putting the apc in the cache and the block in the output?

openvm/src/customize_exe.rs
```diff
@@ -571,83 +567,140 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
 
         tracing::debug!(
             "Basic block start_idx: {}, cost: {}, frequency: {}, cells_saved_per_row: {}",
             start_idx,
             cost,
-            count,
+            freq,
             cells_saved,
         );
     }
```
Schaeff: Same here, wouldn't that just be going through the nodes in the order they are returned already?

Schaeff (commented):
Much better. Can be simplified still though, the whole thing sounds like a big `blocks.map(create_apc).map(measure_cost).reduce(keep_top_n).unzip()`

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs
qwang98: I tried using this but it requires deriving the same traits for sub types including `CachedAutoPrecompile`, `SymbolicMachine`, and many more.

And even if we derive these traits, they will compare on all fields (which is the default option) rather than just on cost, so I think manual implementation is the easiest way around here.

openvm/src/customize_exe.rs
```diff
@@ -571,86 +567,142 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
+            apc_cost,
+            execution_frequency,
+            cells_saved_per_row,
         );
     }
+
+    let retained_blocks: Vec<BasicBlock<<P as IntoOpenVm>::Field>> = retained_blocks_with_stats
+        .into_iter()
+        .map(|(block, _, _, _)| block)
+        .collect();
```
qwang98, line 702: Was hoping to just return the `Vec<Block>` but unzip can only return two elements, so I had to squish `Block` with the other block stats and unpack them here.

openvm/src/customize_exe.rs
Schaeff: Ah I meant the `derivative` crate, but maybe this is fine as is.

PR #2929, opened 2025-06-20
Refactor opcode utilities in powdr_apc

qwang98:
Depends on #2928.

New features:
1. Moved opcode related utilites from `customize_exe.rs` to `opcode.rs`, so we don't overload the file and group all opcode stuffs in the same file.
2. Bigint related opcodes are derived an updated macro in `opcode.rs`. `ALL_OPCODES` contain bigint related opcodes.
3. Opcode lists can stay as `&[usize]` instead of forced to `BTreeSet`, because both data structure can invoke `contains()` call.


openvm/src/opcode.rs
```diff
@@ -1,21 +1,37 @@
...
+                $ty::$variant as usize
+                + < $ty as LocalOpcode >::CLASS_OFFSET
+            ) as usize;
         )*
 
-        /// A slice containing *all* opcode constants.
-        pub const ALL_OPCODES: &[u32] = &[
-            $( $name ),*
+        $(
+            pub const $big_name: usize = (
```
leonardoalt: so this will be created for every opcode?

qwang98: Nope, because of the pattern difference, `big_name` will only be generated for the bigint opcodes, e.g. input like `BIGINT_OPCODE_BEQ = Rv32BranchEqual256Opcode; BranchEqualOpcode::BEQ,`, whereas the non-bigint opcode have input like `OPCODE_BEQ = BranchEqualOpcode::BEQ,`

The const creation line in the macro will be invoked if all its inputs are available, and because `big_name` is not parsed for the non-bigint opcodes, `pub const $big_name` won't be generated for them. The reverse is also true, which means that for each macro input, only one of the `pub const` pattern will be parsed and created.

To confirm, I'm also adding some tests.

openvm/src/customize_exe.rs
```diff
@@ -317,8 +230,8 @@ impl<F: PrimeField32> BasicBlock<F> {
 pub fn collect_basic_blocks<F: PrimeField32>(
     program: &Program<F>,
     labels: &BTreeSet<u32>,
-    opcode_allowlist: &HashSet<usize>,
-    branch_opcodes: &BTreeSet<usize>,
+    opcode_allowlist: &[usize],
+    branch_opcodes: &[usize],
```
Schaeff: I find a set clearer and has constant time access.

openvm/src/opcode.rs
qwang98: Tests added.

openvm/src/customize_exe.rs
qwang98: I can bring them back as utilities in `opcode.rs`

openvm/src/opcode.rs
```diff
@@ -64,4 +80,78 @@ define_opcodes!(
...
+    fn test_instruction_allowlist() {
+        let allowlist = instruction_allowlist();
+        let expected = &[
+            512, 513, 514, 515, 516, 517, 518, 519, 520, 521, 528, 529, 530, 531, 532, 533, 534,
+            535, 544, 545, 549, 550, 551, 552, 560, 561, 565, 576, 592, 593, 594, 595, 596, 597,
+            598, 599, 1056, 1057, 1061, 1062, 1063, 1064,
+        ];
+        assert_eq!(allowlist.len(), ALL_OPCODES.len() - 2); // Excluding HINT_STOREW and HINT_BUFFER
+        assert_eq!(allowlist, expected);
+    }
```
qwang98: Added these tests so we sure that the macro didn't mess up :)

openvm/src/customize_exe.rs
qwang98: Brought back. :)

Schaeff (approved):

PR #2931, opened 2025-06-20
Clean up customize

Schaeff:
- Remove cache

### DIFF
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -1,6 +1,7 @@
 use std::cmp::Ordering;
 use std::collections::HashSet;
 use std::collections::{BTreeMap, BTreeSet, BinaryHeap, HashMap};
+use std::iter::once;
 use std::sync::Arc;
 
 use crate::extraction_utils::OriginalVmConfig;
@@ -23,11 +24,11 @@ use openvm_stark_backend::{
 };
 use powdr_autoprecompiles::expression::try_convert;
 use powdr_autoprecompiles::powdr::UniqueReferences;
-use powdr_autoprecompiles::DegreeBound;
 use powdr_autoprecompiles::VmConfig;
 use powdr_autoprecompiles::{
     bus_map::BusMap, SymbolicBusInteraction, SymbolicInstructionStatement, SymbolicMachine,
 };
+use powdr_autoprecompiles::{Apc, DegreeBound};
 use powdr_number::{BabyBearField, FieldElement};
 
 use crate::bus_interaction_handler::OpenVmBusInteractionHandler;
@@ -44,13 +45,6 @@ const OPENVM_INIT_PC: u32 = 0x0020_0800;
 
 pub const POWDR_OPCODE: usize = 0x10ff;
 
-#[derive(Clone, Debug)]
-pub struct CachedAutoPrecompile<F> {
-    apc_opcode: usize,
-    autoprecompile: SymbolicMachine<F>,
-    subs: Vec<Vec<u64>>,
-}
-
 use crate::{PgoConfig, PowdrConfig};
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


openvm/src/customize_exe.rs
```diff
@@ -64,6 +58,44 @@ impl From<powdr_autoprecompiles::constraint_optimizer::Error> for Error {
     }
 }
 
+struct BlockWithApc<P: IntoOpenVm> {
```
Schaeff, line 58: Define a struct for the result of creating an APC for a given block.

openvm/src/customize_exe.rs
```diff
@@ -145,71 +156,70 @@ pub fn customize(
...
     };
 
-    let mut extensions = Vec::new();
-    let n_acc = config.autoprecompiles as usize;
-    let n_skip = config.skip_autoprecompiles as usize;
-
     tracing::info!("Adjust the program with the autoprecompiles");
 
-    // now the blocks have been sorted by cost
-    for acc_block in blocks.iter().skip(n_skip).take(n_acc) {
```
Schaeff, line 151: No need to skip and take anymore, `blocks_with_apcs` contains everything we need

openvm/src/customize_exe.rs
```diff
@@ -579,16 +572,20 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
     // each generated apc becomes a candidate for caching
     // it is only ordered by cost, but carries all needed data for modifying the blocks, extending the cache, and debug print
     struct ApcCandidate<P: IntoOpenVm> {
-        block: BasicBlock<OpenVmField<P>>,
-        apc_cache_entry: CachedAutoPrecompile<P>,
-        apc_cost: usize,
```
Schaeff, line 507: No need to store the cost, it can be computed cheaply based on the other members

openvm/src/customize_exe.rs
```diff
@@ -145,71 +156,70 @@ pub fn customize(
...
-
-        let len_before = program.len();
-        program.splice(pc..pc + n_acc, new_instrs);
-        assert_eq!(program.len(), len_before);
-
-        let is_valid_column = autoprecompile
-            .unique_references()
-            .find(|c| &*c.name == "is_valid")
-            .unwrap();
+    let extensions = blocks_with_apcs
```
Schaeff, line 157: Diff looks big but I just turned a loop into a map. There is still a side effect on the program though, it gets modified to insert the new instructions.

openvm/src/customize_exe.rs
qwang98: Noice :)

openvm/src/customize_exe.rs
qwang98: Yeah. It's just cells saved * frequency.

openvm/src/customize_exe.rs
```diff
@@ -637,18 +630,20 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
+                once(ApcCandidate {
+                    block_with_apc: BlockWithApc {
+                        opcode: POWDR_OPCODE + i,
+                        block,
+                        apc,
+                    },
+                    execution_frequency,
+                    cells_saved_per_row,
+                })
+                .collect(),
```
qwang98, line 566: Can we add comment that this is a heap with a single element? I mean can be inferred from the compiler but it's not that obvious. :)

openvm/src/customize_exe.rs
```diff
@@ -665,50 +660,28 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
-            execution_frequency,
-            cells_saved_per_row,
-        );
-    }
-
-    let retained_blocks: Vec<BasicBlock<<P as IntoOpenVm>::Field>> = retained_blocks_with_stats
-        .into_iter()
-        .map(|(block, _, _, _)| block)
-        .collect();
+        .skip(config.skip_autoprecompiles as usize)
```
qwang98, line 584: Nice :) Skipped here.

qwang98 (commented):
Haven't taken a very detailed look as need to go afk but looks pretty good :) Left some small comments and not sure why CI isn't running.

openvm/src/customize_exe.rs
```diff
@@ -569,6 +560,8 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
 
     // generate apc for all basic blocks and only cache the ones we eventually use
     // calculate number of trace cells saved per row for each basic block to sort them by descending cost
+    // Note: we could reduce `max_cache` to `config.autoprecompiles` by discarding the first `config.skip_autoprecompiles` candidates.
+    // This would require remembering their costs, which adds complexity here. We do not do this because the `skip_autoprecompiles` parameter is for debugging purposes only.
```
qwang98: Yeah...I was also thinking about this as it's easy to discard the low cost ones as we are sure that they are gone, but it's harder to do for the high cost ones, as we aren't sure if they'll eventually still be the top high cost ones due to the min heap structure.

I do agree as well that skip is more like a debugging param, as there's no point to explicitly not accelerate the costliest block in a performance scenario.

openvm/src/customize_exe.rs
```diff
@@ -326,37 +336,19 @@ fn add_extra_targets<F: PrimeField32>(
...
-
-    Ok(CachedAutoPrecompile {
-        apc_opcode,
-        autoprecompile,
-        subs,
-    })
-}
-
 // Only used for PgoConfig::Instruction and PgoConfig::None,
 // because PgoConfig::Cell caches all APCs in sorting stage.
```
qwang98, line 340: Can change to "creates all APCs in sorting stage" now.

openvm/src/customize_exe.rs
```diff
@@ -61,6 +55,44 @@ impl From<powdr_autoprecompiles::constraint_optimizer::Error> for Error {
...
+            bus_map,
+            opcodes_allowlist,
+        ),
+        PgoConfig::Instruction(pgo_program_idx_count) => {
+            create_apcs_with_instruction_pgo(blocks, pgo_program_idx_count, airs, config, bus_map)
+        }
+        PgoConfig::None => create_apcs_with_no_pgo(blocks, airs, config, bus_map),
+    };
+
+    assert!(res.len() <= config.autoprecompiles as usize);
```
qwang98, line 91: Just to confirm that it's acceptable to generate fewer autoprecompiles than `config.autoprecompiles` asks for. This  can happen for all PGO modes.

This has been the case before PGO exists, because `take()` early exits without erroring out if there aren't enough elements in the iterator.

openvm/src/customize_exe.rs
```diff
@@ -502,16 +493,20 @@ fn sort_blocks_by_pgo_cell_cost_and_cache_apc<P: IntoOpenVm>(
...
+        block_with_apc: BlockWithApc<P>,
         execution_frequency: usize,
         cells_saved_per_row: usize,
     }
 
+    impl<P: IntoOpenVm> ApcCandidate<P> {
+        fn cost(&self) -> usize {
+            self.execution_frequency * self.cells_saved_per_row
+        }
+    }
```
qwang98, line 505: Nice :)

openvm/src/customize_exe.rs
Schaeff: Correct

PR #2933, opened 2025-06-20
Better instruction air retrieval

Schaeff:
- Introduce `OriginalAirs` struct in `powdr_openvm` to reduce air extraction from once per opcode to once per air
- Expose a method to get air widths per opcode, useful in PGO
- Introduce `InstructionAirHandler` trait in `powdr_autoprecompiles` to avoid this change leaking into that crate. The only thing that `powdr_autoprecompiles` needs from the instruction airs is to fetch them by opcode id.
- Implement this trait for `&OriginalAirs`

Currently, `powdr_autoprecompile` unwraps the result from getting the air for a given opcode. In the future, we could encode "blocklisted" opcodes as `get_instruction_air` returning Err.

### DIFF
--- a/autoprecompiles/src/lib.rs
+++ b/autoprecompiles/src/lib.rs
@@ -13,7 +13,7 @@ use powdr_expression::{
 use serde::{Deserialize, Serialize};
 use std::fmt::Display;
 use std::sync::Arc;
-use std::{collections::BTreeMap, iter::once};
+use std::{iter::once};
 use symbolic_machine_generator::statements_to_symbolic_machine;
 
 use powdr_number::FieldElement;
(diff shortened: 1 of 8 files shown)


openvm/src/extraction_utils.rs
```diff
@@ -27,6 +31,68 @@ use crate::utils::{get_pil, UnsupportedOpenVmReferenceError};
...
+    ) -> Result<(), UnsupportedOpenVmReferenceError> {
+        if self.opcode_to_air.contains_key(&opcode) {
+            panic!("Opcode {opcode} already exists");
+        }
+        // Insert the machine only if `air_name` isn't already present
+        if !self.air_name_to_machine.contains_key(&air_name) {
+            let machine_instance = machine()?;
+            self.air_name_to_machine
+                .insert(air_name.clone(), machine_instance);
+        }
```
qwang98, line 65: Noice :)

openvm/src/extraction_utils.rs
```diff
@@ -27,6 +31,68 @@ use crate::utils::{get_pil, UnsupportedOpenVmReferenceError};
...
+                |width_by_air: &mut HashMap<&String, usize>,
+                 (opcode, air_name): (&VmOpcode, &String)| {
+                    let width = if let Some(width) = width_by_air.get(air_name) {
+                        *width
+                    } else {
+                        let machine = &self.air_name_to_machine[air_name];
+                        let width = machine.unique_references().count();
+                        width_by_air.insert(air_name, width);
+                        width
+                    };
```
qwang98, line 88: Cool. Avoids calculating air width for the same AIR but differen opcode.

qwang98:
> Currently, `powdr_autoprecompile` unwraps the result from getting the air for a given opcode. In the future, we could encode "blocklisted" opcodes as `get_instruction_air` returning Err.

Does this mean that `get_instruction_air` no longer take a list and `OriginalAir` maybe will use the allow list when inserting opcodes?

autoprecompiles/src/lib.rs
```diff
@@ -232,29 +232,38 @@ impl<T: FieldElement> PcLookupBusInteraction<T> {
...
-pub struct VmConfig<'a, T: FieldElement, B> {
+pub struct VmConfig<'a, M, B> {
     /// Maps an opcode to its AIR.
-    pub instruction_machines: &'a BTreeMap<usize, SymbolicMachine<T>>,
+    pub instruction_machine_handler: &'a M,
     /// The bus interaction handler, used by the constraint solver to reason about bus interactions.
     pub bus_interaction_handler: B,
     /// The bus map that maps bus id to bus type
     pub bus_map: BusMap,
 }
```
qwang98, line 242: Why do we have generic `M` here instead of `InstructionMachineHandler`?

Probably a similar question for `B` and `BusInteractionHandler`.

Schaeff: It's best practice to avoid bounds on struct declarations if possible. A rule of thumb is that they are only needed when accessing associated types in the struct fields, which is not the case here.

PR #2938, opened 2025-06-23
Make instruction machine retrieval part of apc API

Schaeff:
Currently, the apc user is responsible for collecting all necessary instruction machines and giving them to the apc builder.

With this PR, the user passes a set of opcodes that it wants to be supported (currently, this is decided in PGO) and the VmConfig takes care of fetching them if it can. Opcodes for which the machine is not available are ignored, but when building the apc, requesting their machine returns an error.


qwang98:
Maybe need a rebase?

PR #2941, opened 2025-06-24
restore pgo API

leonardoalt:
We need the `CompiledProgram` entry point because in the reth benchmark we already have the elf binary

### DIFF
--- a/openvm/src/lib.rs
+++ b/openvm/src/lib.rs
@@ -600,14 +600,19 @@ pub fn prove(
     Ok(())
 }
 
-// Produces execution count by pc_index
-// Used in Pgo::Cell and Pgo::Instruction to help rank basic blocks to create APCs for
-pub fn get_pc_idx_count(guest: &str, guest_opts: GuestOptions, inputs: StdIn) -> HashMap<u32, u32> {
+// Same as execution_profile below but for guest path inputs.
+pub fn execution_profile_from_guest(
+    guest: &str,
+    guest_opts: GuestOptions,
+    inputs: StdIn,
+) -> HashMap<u32, u32> {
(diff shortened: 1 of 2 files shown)


openvm/src/lib.rs
```diff
@@ -603,7 +603,12 @@ pub fn prove(
 // Produces execution count by pc_index
 // Used in Pgo::Cell and Pgo::Instruction to help rank basic blocks to create APCs for
 pub fn get_pc_idx_count(guest: &str, guest_opts: GuestOptions, inputs: StdIn) -> HashMap<u32, u32> {
-    let OriginalCompiledProgram { exe, sdk_vm_config } = compile_openvm(guest, guest_opts).unwrap();
+    let program = compile_openvm(guest, guest_opts).unwrap();
+    pgo(program, inputs)
+}
+
+pub fn pgo(program: OriginalCompiledProgram, inputs: StdIn) -> HashMap<u32, u32> {
+    let OriginalCompiledProgram { exe, sdk_vm_config } = program;
```
qwang98: I just don't like the `pgo` name here as it's technically just building a map of PC execution count while the actual PGO happens in `customize_exe.rs`.

But the out level function is already called `get_pc_idx_count`, so I'm not sure what to call the inner function instead of `pgo`...

qwang98: Need suggestions :)

qwang98: Maybe `get_pc_idx_count_inner`?

leonardoalt: well both are pub so I don't think `inner` works

Schaeff (approved):

PR #2942, opened 2025-06-24
Get allowlist from airs

Schaeff:


openvm/src/customize_exe.rs
```diff
@@ -106,7 +100,7 @@ pub fn customize(
     let airs = original_config.airs().expect("Failed to convert the AIR of an OpenVM instruction, even after filtering by the blacklist!");
```
qwang98, line 100: So we call the `instruction_allowlist()` within `airs()` and filter the entries to the allowed opcodes already, and this airs can return the allow list by calling `airs.allow_list()`? Sounds like we solve the problem at "the source"?

PR #2945, opened 2025-06-25
Assembly builder branch update

qwang98:


openvm/tests/apc_builder.rs
```diff
@@ -220,7 +220,7 @@ mod complex_tests {
             add(8, 8, 16777200, 0),
             storew(4, 8, 12, 2, 1, 0),
             auipc(4, 0, 0, 1, 0),
-            blt(4, 4, 1780, 1, 0),
+            jalr(4, 4, 1780, 1, 0),
```
qwang98, line 223: Detected this error because branch opcodes (including `blt`) has fixed `d=1` and `e=1`, so I checked again that this opcode is actually `jalr`, and therefore the machine print out changes.

openvm/src/symbolic_instruction_builder.rs
```diff
@@ -105,28 +105,40 @@ macro_rules! ls_ops {
...
+                        T::one(),
+                        T::one(),
+                        T::zero(),
+                        T::zero(),
+                    ],
+                }
+            }
+        )+
+    };
+}
```
qwang98, line 138: New macro for building branch opcodes, to make sure we only supply the 3 arguments that aren't fixed.

PR #2955, opened 2025-06-27
[PR for Review] Detect and count non-Powdr columns for cell PGO max column use

qwang98:
Pgo input is now "max total columns" instead of "max apc columns", because there can be many non APC columns. To be exact, roughly ~580 for RV32 chip set and ~15K in the set used in `reth-benchmark`. Tested and works for Keccak. 

This PR calculates the total non APC columns number, logs them by chip in case we interested in what airs they belong to, and then subtract the non APC total from the max total the user supplies to PGO to obtain the max APC columns, which is then used in the KnapSack algorithm to limit APC generation in Pgo::Cell mode.

This PR's method is a bit of simplification, we are not sure if all chips in the original config's chip inventory get eventually proven. There are three scenarios (in order of ascending restrictiveness): 
1. All columns in original inventory stay in. 
2. Those that get executed stay in. 
3. Those that get proven stay in. 

Scenario 3 is more restrictive than scenario 2, because we need all dummy chips executed for witgen but not necessarily all dummy columns for proving.

### DIFF
--- a/openvm/src/extraction_utils.rs
+++ b/openvm/src/extraction_utils.rs
@@ -271,6 +271,20 @@ impl OriginalVmConfig {
                        // Create a new chip complex
         self.sdk_config.create_chip_complex()
     }
+
+    pub fn chip_inventory_air_widths(&self) -> HashMap<String, usize> {
+        self.chip_complex()
+            .inventory
+            .executors()
+            .iter()
+            .map(|executor| {
+                let air: Arc<dyn AnyRap<BabyBearSC>> = executor.air();
+                let name = air.name();
+                let width = air.width();
+                (name, width)
+            })
+            .collect::<HashMap<_, _>>()
+    }
 }
 
 pub fn export_pil(writer: &mut impl std::io::Write, vm_config: &SpecializedConfig) {
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -65,6 +65,7 @@ fn generate_apcs_with_pgo<P: IntoOpenVm>(
     airs: &OriginalAirs<P>,
     bus_map: &BusMap,
     config: &PowdrConfig,
+    original_config: &OriginalVmConfig,
     pgo_config: PgoConfig,
 ) -> Vec<BlockWithApc<P>> {
     // sort basic blocks by:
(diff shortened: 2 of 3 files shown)


openvm/src/extraction_utils.rs
```diff
@@ -271,6 +271,20 @@ impl OriginalVmConfig {
...
+            .executors()
+            .iter()
+            .map(|executor| {
+                let air: Arc<dyn AnyRap<BabyBearSC>> = executor.air();
+                let name = air.name();
+                let width = air.width();
+                (name, width)
+            })
+            .collect::<HashMap<_, _>>()
+    }
```
qwang98, line 287: This function seems duplicative from `air_width_per_opcode`, but it's not, because:
1. `air_width_per_opcode()` operates on `symbolic_machine.unique_references()`, which are after Powdr conversions in `insert_opcode()`. 
2. `insert_opcode()` is called in `airs()`, which already filters opcode by the allowlist, because `airs()` purposefully builds airs for consumption when creating APCs.
3. Here in this function, we are not interested in airs that we can create APCs for, but in all other non-Powdr columns, and therefore we start from `OriginalConfig`.
4. This function, different from `air_width_per_opcode()`, is much cheaper, because instead of looking for number of unique references in a symbolic machine, it simply loops over all executors. The `width()` API used here simply counts the number of field in a chip's column struct, e.g. `BaseAluCoreCols`.

qwang98: That's to say, this function has different purpose from `air_width_per_opcode()`, and therefore both should stay.

Schaeff: > The width() API used here simply counts the number of field in a chip's column struct, e.g. BaseAluCoreCols.

Just checking the adapter columns are also included?

Schaeff (approved):
lgtm except comment

openvm/src/extraction_utils.rs
qwang98: I think should be good. Did some println in the function:
```
inventory air VmAirWrapper<Rv32BaseAluAdapterAir, BaseAluCoreAir<4, 8> has width 36
inventory air VmAirWrapper<Rv32BaseAluAdapterAir, LessThanCoreAir<4, 8> has width 37
inventory air VmAirWrapper<Rv32BaseAluAdapterAir, ShiftCoreAir<4, 8> has width 53
inventory air VmAirWrapper<Rv32LoadStoreAdapterAir, LoadStoreCoreAir<4> has width 41
inventory air VmAirWrapper<Rv32LoadStoreAdapterAir, LoadSignExtendCoreAir<4, 8> has width 36
inventory air VmAirWrapper<Rv32BranchAdapterAir, BranchEqualCoreAir<4> has width 26
inventory air VmAirWrapper<Rv32BranchAdapterAir, BranchLessThanCoreAir<4, 8> has width 32
inventory air VmAirWrapper<Rv32CondRdWriteAdapterAir, Rv32JalLuiCoreAir> has width 18
inventory air VmAirWrapper<Rv32JalrAdapterAir, Rv32JalrCoreAir> has width 28
inventory air VmAirWrapper<Rv32RdWriteAdapterAir, Rv32AuipcCoreAir> has width 20
inventory air Rv32HintStoreAir has width 32
inventory air KeccakVmAir has width 3163
inventory air VmAirWrapper<Rv32MultAdapterAir, MultiplicationCoreAir<4, 8> has width 31
inventory air VmAirWrapper<Rv32MultAdapterAir, MulHCoreAir<4, 8> has width 39
inventory air VmAirWrapper<Rv32MultAdapterAir, DivRemCoreAir<4, 8> has width 59
```

```
impl<F, A, C> BaseAir<F> for VmAirWrapper<A, C>
where
    A: BaseAir<F>,
    C: BaseAir<F>,
{
    fn width(&self) -> usize {
        self.adapter.width() + self.core.width()
    }
}
```

PR #2956, opened 2025-06-27
Add sha256 guest

ShuangWu121:


openvm/guest-sha256/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+use openvm::io::{reveal, read};
+use sha2::{Sha256, Digest};
+
+pub fn main() {
+    let n: u32 = read();
+    let mut output = black_box([0u8; 32]);
+     for _ in 0..n {
+        let mut hasher = Sha256::new();
+        hasher.update(&output);
+        output.copy_from_slice(&hasher.finalize());
```
leonardoalt: this sounds expensive (to copy every iteration). In the Keccak example, we use `hasher.finalize(&mut output);`. Can't we do the same here?

Schaeff:
@ShuangWu121 @leonardoalt 

what about

```rust
pub fn main() {
    let n: u32 = read();
    let mut output = black_box([0u8; 32]);
     for _ in 0..n {
        output = Sha256::digest(&output).into();
    }

    reveal_u32(output[0] as u32, 0);
}
```

ShuangWu121:
> ```rust
> for _ in 0..n {
>         output = Sha256::digest(&output).into();
>     }
> ```

Nice, this works!

openvm/guest-sha256/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use openvm::io::{read,reveal_u32};
+use sha2::{Sha256, Digest};
+use digest::generic_array::GenericArray;
+
+pub fn main() {
+    let n: u32 = read();
+    let mut output = black_box([0u8; 32]);
+     for _ in 0..n {
+        for _ in 0..n {
```
leonardoalt: 2 for loops

ShuangWu121: fixed, sorry!

leonardoalt:
> I think this somehow uses way more cycles, checking locally

Ok this was fake news

openvm/guest-sha256/src/main.rs
```diff
@@ -0,0 +1,20 @@
+#![cfg_attr(target_os = "zkvm", no_main)]
+#![cfg_attr(target_os = "zkvm", no_std)]
+
+openvm::entry!(main);
+
+use core::hint::black_box;
+
+use openvm::io::{read,reveal_u32};
+use sha2::{Sha256, Digest};
+use digest::generic_array::GenericArray;
```
leonardoalt: Can you run clippy on this file? This import for example is not needed

leonardoalt (commented):
Please add a test to `openvm/lib.rs` similar to the keccak ones (can use `[ignored]`) that runs and proves this with 1000 inputs and something like 10 apcs

openvm/guest-sha256/src/main.rs
ShuangWu121: fixed

openvm/src/lib.rs
```diff
@@ -1016,6 +1022,15 @@ mod tests {
         tracing::debug!("Proving with PgoConfig::Cell took {:?}", elapsed);
     }
 
+    #[test]
+    #[ignore = "Too long"]
+    fn sha256_prove_simple() {
+        let mut stdin = StdIn::default();
+        stdin.write(&GUEST_SHA256_ITER);
+        let config = PowdrConfig::new(GUEST_SHA256_APC, GUEST_SHA256_SKIP);
+        prove_simple(GUEST_SHA256, config, stdin, PgoConfig::None, None);
```
leonardoalt: why pgo none?

leonardoalt: ah keccak does that too.. let's at least use `PgoConfig::Instruction` here cus I don't know if the largest is Sha

leonardoalt (commented):
I don't know exactly which Keccak tests run on pr-tests or nightly only (with [ignore]). Can you check which keccak tests always run and make equivalent tests for Sha. I think this will prevent issues in the future

openvm/guest-sha256-manual-precompile/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+use core::hint::black_box;
+
+use openvm::io::{read, reveal_u32};
+use openvm_sha2::sha256;
+
+openvm::entry!(main);
+
+pub fn main() {
+    let n = read();
+    let mut output = [0u8; 32];
```
Schaeff: Not sure if it makes a difference, but herer we call black_box on each iteration whereas in the other guest it's called only once. Can we harmonize this so that only the sha256 call differs?

Schaeff (dismissed):
LGTM except the small comment

openvm/guest-sha256-manual-precompile/src/main.rs
ShuangWu121: Done

PR #2960, opened 2025-06-30
Calculate non-APC LogUp columns for PGO

qwang98:
Original comment by @Schaeff:
```
so here's my suspicion:

to get the number of columns of an air, we use executor.air().width(), in fully qualified syntax openvm_stark_backend::p3_air::BaseAir::width(&executor.air())
this is the number of main trace columns, which does not include interaction columns
in order to get the full number of columns, we need to do what we already do for get_constraints but additionally call InteractionPhaseAirBuilder::finalize_interactions on the symbolic builder to materialize the interactions. Sadly InteractionPhaseAirBuilder is pub(crate) so it seems we need to change stark-backend
```

### DIFF
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -5,9 +5,9 @@ use std::sync::Arc;
 use crate::extraction_utils::{OriginalAirs, OriginalVmConfig};
 use crate::opcode::{branch_opcodes_bigint_set, branch_opcodes_set};
 use crate::utils::{fractional_knapsack, KnapsackItem, UnsupportedOpenVmReferenceError};
-use crate::IntoOpenVm;
 use crate::OpenVmField;
 use crate::OriginalCompiledProgram;
+use crate::{AirMetrics, IntoOpenVm};
 use crate::{CompiledProgram, SpecializedConfig};
 use itertools::Itertools;
 use openvm_instructions::instruction::Instruction;
@@ -519,13 +519,13 @@ fn create_apcs_with_cell_pgo<P: IntoOpenVm>(
         }
     }
 
-    let max_total_apc_columns = max_total_columns.map(|max_total_columns| {
-        let chip_inventory_air_widths = original_config.chip_inventory_air_widths();
-        let total_non_apc_columns = chip_inventory_air_widths
+    let max_total_apc_columns: Option<usize> = max_total_columns.map(|max_total_columns| {
+        let chip_inventory_air_metrics = original_config.chip_inventory_air_metrics();
+        let total_non_apc_columns = chip_inventory_air_metrics
             .iter()
-            .map(|(air_name, air_width)| {
-                tracing::debug!("Chip inventory air {} has {}", air_name, air_width);
-                air_width.base_width + air_width.log_up_width
+            .map(|AirMetrics { name, width, .. }| {
+                tracing::debug!("Chip inventory air {} has {}", name, width);
+                width.base_width + width.log_up_width
             })
             .sum::<usize>();
         max_total_columns - total_non_apc_columns
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -523,9 +523,9 @@ fn create_apcs_with_cell_pgo<P: IntoOpenVm>(
         let chip_inventory_air_widths = original_config.chip_inventory_air_widths();
         let total_non_apc_columns = chip_inventory_air_widths
             .iter()
-            .map(|(air_name, width)| {
-                tracing::debug!("Chip inventory air {} has width {}", air_name, width);
-                width
+            .map(|(air_name, air_width)| {
+                tracing::debug!("Chip inventory air {} has {}", air_name, air_width);
+                air_width.base_width + air_width.log_up_width
             })
             .sum::<usize>();
         max_total_columns - total_non_apc_columns
--- a/openvm/src/extraction_utils.rs
+++ b/openvm/src/extraction_utils.rs
@@ -9,6 +9,11 @@ use openvm_circuit_primitives::bitwise_op_lookup::SharedBitwiseOperationLookupCh
 use openvm_circuit_primitives::range_tuple::SharedRangeTupleCheckerChip;
 use openvm_instructions::VmOpcode;
 use openvm_sdk::config::{SdkVmConfig, SdkVmConfigExecutor, SdkVmConfigPeriphery};
+use openvm_stark_backend::air_builders::symbolic::SymbolicRapBuilder;
+use openvm_stark_backend::interaction::fri_log_up::{
+    find_interaction_chunks, STARK_LU_NUM_CHALLENGES,
+};
+use openvm_stark_backend::interaction::rap::InteractionPhaseAirBuilder;
 use openvm_stark_backend::{
     air_builders::symbolic::SymbolicConstraints, config::StarkGenericConfig, rap::AnyRap, Chip,
 };
@@ -25,6 +30,7 @@ use powdr_number::BabyBearField;
 use serde::{Deserialize, Serialize};
 use std::ops::Deref;
 use std::sync::MutexGuard;
+use tracing::Dispatch;
 
 use crate::utils::{get_pil, UnsupportedOpenVmReferenceError};
 
@@ -272,7 +278,7 @@ impl OriginalVmConfig {
         self.sdk_config.create_chip_complex()
     }
 
-    pub fn chip_inventory_air_widths(&self) -> HashMap<String, usize> {
+    pub fn chip_inventory_air_widths(&self) -> HashMap<String, AirWidth> {
         self.chip_complex()
             .inventory
             .executors()
@@ -280,8 +286,15 @@ impl OriginalVmConfig {
             .map(|executor| {
                 let air: Arc<dyn AnyRap<BabyBearSC>> = executor.air();
                 let name = air.name();
-                let width = air.width();
-                (name, width)
+                let base_width = air.width();
+                let log_up_width = get_log_up_width(air.clone());
+                (
+                    name,
+                    AirWidth {
+                        base_width,
+                        log_up_width,
+                    },
+                )
             })
             .collect::<HashMap<_, _>>()
     }
@@ -329,12 +342,46 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
 pub fn get_constraints(
     air: Arc<dyn AnyRap<BabyBearSC>>,
 ) -> SymbolicConstraints<p3_baby_bear::BabyBear> {
+    let builder = get_symbolic_builder(air);
+    builder.constraints()
+}
+
+pub fn get_log_up_width(air: Arc<dyn AnyRap<BabyBearSC>>) -> usize {
+    let builder = get_symbolic_builder(air);
+    let max_degree = builder.max_constraint_degree();
+    let perm_width = find_interaction_chunks(&builder.constraints().interactions, max_degree)
+        .interaction_partitions()
+        .len()
+        + 1;
+    let challenge_width = STARK_LU_NUM_CHALLENGES;
+    perm_width + challenge_width
+}
+
+pub fn get_symbolic_builder(
+    air: Arc<dyn AnyRap<BabyBearSC>>,
+) -> SymbolicRapBuilder<p3_baby_bear::BabyBear> {
     let perm = default_perm();
     let security_params = SecurityParameters::standard_fast();
     let config = config_from_perm(&perm, security_params);
     let air_keygen_builder = AirKeygenBuilder::new(config.pcs(), air);
-    let builder = air_keygen_builder.get_symbolic_builder(None);
-    builder.constraints()
+    air_keygen_builder.get_symbolic_builder(None)
+}
+
+pub struct AirWidth {
+    pub base_width: usize,
+    pub log_up_width: usize,
+}
+
+impl std::fmt::Display for AirWidth {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Total Width: {} (Base: {}, Log Up: {})",
+            self.base_width + self.log_up_width,
+            self.base_width,
+            self.log_up_width
+        )
+    }
 }
 
 #[cfg(test)]
--- a/openvm/src/extraction_utils.rs
+++ b/openvm/src/extraction_utils.rs
@@ -2,8 +2,8 @@ use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::sync::{Arc, Mutex};
 
 use crate::air_builder::AirKeygenBuilder;
-use crate::IntoOpenVm;
 use crate::{opcode::instruction_allowlist, BabyBearSC, SpecializedConfig};
+use crate::{AirMetrics, IntoOpenVm};
 use openvm_circuit::arch::{VmChipComplex, VmConfig, VmInventoryError};
 use openvm_circuit_primitives::bitwise_op_lookup::SharedBitwiseOperationLookupChip;
 use openvm_circuit_primitives::range_tuple::SharedRangeTupleCheckerChip;
@@ -278,25 +278,16 @@ impl OriginalVmConfig {
         self.sdk_config.create_chip_complex()
     }
 
-    pub fn chip_inventory_air_widths(&self) -> HashMap<String, AirWidth> {
+    pub fn chip_inventory_air_metrics(&self) -> Vec<AirMetrics> {
         self.chip_complex()
             .inventory
             .executors()
             .iter()
             .map(|executor| {
                 let air: Arc<dyn AnyRap<BabyBearSC>> = executor.air();
-                let name = air.name();
-                let base_width = air.width();
-                let log_up_width = get_log_up_width(air.clone());
-                (
-                    name,
-                    AirWidth {
-                        base_width,
-                        log_up_width,
-                    },
-                )
+                get_air_metrics(air)
             })
-            .collect::<HashMap<_, _>>()
+            .collect::<Vec<_>>()
     }
 }
 
@@ -346,15 +337,29 @@ pub fn get_constraints(
     builder.constraints()
 }
 
-pub fn get_log_up_width(air: Arc<dyn AnyRap<BabyBearSC>>) -> usize {
+pub fn get_air_metrics(air: Arc<dyn AnyRap<BabyBearSC>>) -> AirMetrics {
+    let name = air.name();
+    let base_width = air.width();
     let builder = get_symbolic_builder(air);
     let max_degree = builder.max_constraint_degree();
-    let perm_width = find_interaction_chunks(&builder.constraints().interactions, max_degree)
+    let SymbolicConstraints {
+        constraints,
+        interactions,
+    } = builder.constraints();
+    let perm_width = find_interaction_chunks(&interactions, max_degree)
         .interaction_partitions()
         .len()
         + 1;
     let challenge_width = STARK_LU_NUM_CHALLENGES;
-    perm_width + challenge_width
+    AirMetrics {
+        name,
+        width: AirWidth {
+            base_width,
+            log_up_width: perm_width + challenge_width,
+        },
+        constraints: constraints.len(),
+        bus_interactions: interactions.len(),
+    }
 }
 
 pub fn get_symbolic_builder(
@@ -367,6 +372,7 @@ pub fn get_symbolic_builder(
     air_keygen_builder.get_symbolic_builder(None)
 }
 
+#[derive(Debug)]
 pub struct AirWidth {
     pub base_width: usize,
     pub log_up_width: usize,
--- a/openvm/src/lib.rs
+++ b/openvm/src/lib.rs
@@ -500,6 +500,7 @@ pub struct OriginalCompiledProgram {
     pub sdk_vm_config: SdkVmConfig,
 }
 
+#[derive(Debug)]
 pub struct AirMetrics {
     pub name: String,
     pub width: usize,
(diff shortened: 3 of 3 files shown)


openvm/src/customize_exe.rs
```diff
@@ -519,13 +519,13 @@ fn create_apcs_with_cell_pgo<P: IntoOpenVm>(
...
+    let max_total_apc_columns: Option<usize> = max_total_columns.map(|max_total_columns| {
+        let chip_inventory_air_metrics = original_config.chip_inventory_air_metrics();
+        let total_non_apc_columns = chip_inventory_air_metrics
             .iter()
-            .map(|(air_name, width)| {
-                tracing::debug!("Chip inventory air {} has width {}", air_name, width);
-                width
+            .map(|AirMetrics { name, width, .. }| {
+                tracing::debug!("Chip inventory air {} has {}", name, width);
+                width.base_width + width.log_up_width
```
qwang98: Main change.

openvm/src/extraction_utils.rs
```diff
@@ -272,18 +278,16 @@ impl OriginalVmConfig {
         self.sdk_config.create_chip_complex()
     }
 
-    pub fn chip_inventory_air_widths(&self) -> HashMap<String, usize> {
+    pub fn chip_inventory_air_metrics(&self) -> Vec<AirMetrics> {
```
qwang98: Expose method for `air_metrics` instead of `air_widths`, because we need the `SymbolicBuilder` and `SymbolicConstraints` regardless, so getting `air_metrics` only has very small marginal cost (just calling `len()` twice) on top of getting `air_widths`.

`air_metrics` can also be used in `CompiledProgram::powdr_air_metrics`.

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +333,62 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
...
+    let max_degree = builder.max_constraint_degree();
+    let SymbolicConstraints {
+        constraints,
+        interactions,
+    } = builder.constraints();
+    let perm_width = find_interaction_chunks(&interactions, max_degree)
+        .interaction_partitions()
+        .len()
+        + 1;
+    let challenge_width = STARK_LU_NUM_CHALLENGES;
```
qwang98: `find_interaction_chunks` simply batches up as many bus interactions as possible up to the max constrain degrees and the number of permutation columns equals to the number of chunks plus 1. These are all taken from `finalize_bus_interactions` but without creating the addtional columns and is therefore more efficient.

qwang98: There are three types of columns added in `finalize_bus_interactions`, i.e. `Permutation`, `Challenge`, and `Expose`. I assume they are all "real columns", so added them all to the `log_up_width` variable, but I'm not sure if `Expose` are real columns depending on how they expose (maybe a fixed column with 1 on one row and then multiply?)

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +333,62 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
 pub fn get_constraints(
     air: Arc<dyn AnyRap<BabyBearSC>>,
 ) -> SymbolicConstraints<p3_baby_bear::BabyBear> {
+    let builder = symbolic_builder_with_degree(air, None);
```
qwang98, line 342: Symbolic builder is set to 0 degree in the original `get_constraints` API, because I think variable references from constraints and interactions per se (not lower level columns) don't depend on the degree.

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +331,63 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
...
+        .len()
+        + 1;
+    let challenge_width = STARK_LU_NUM_CHALLENGES;
+    let exposed_width = STARK_LU_NUM_EXPOSED_VALUES;
+    AirMetrics {
+        name,
+        width: AirWidth {
+            base_width,
+            log_up_width: perm_width + challenge_width + exposed_width,
+        },
```
qwang98: find_interaction_chunks simply batches up as many bus interactions as possible up to the max constrain degrees and the number of permutation columns equals to the number of chunks plus 1. These are all taken from finalize_bus_interactions but without creating the addtional columns and is therefore more efficient.

There are three types of columns added in finalize_bus_interactions, i.e. Permutation, Challenge, and Expose. I assume they are all "real columns", so added them all to the log_up_width variable, but I'm not sure if Expose are real columns depending on how they expose (maybe a fixed column with 1 on one row and then multiply?)

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +331,62 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
...
     air: Arc<dyn AnyRap<BabyBearSC>>,
 ) -> SymbolicConstraints<p3_baby_bear::BabyBear> {
+    let builder = symbolic_builder_with_degree(air, None);
+    builder.constraints()
+}
+
+pub fn get_air_metrics(air: Arc<dyn AnyRap<BabyBearSC>>) -> AirMetrics {
+    let name = air.name();
+    let base_width = air.width();
+    let app_log_blow_up = 2;
```
qwang98, line 347: This number appears 2 or 3 times across our repo. I'm not sure if it's a fixed value at this point but think we should make it a constant in another PR.

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +331,62 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
...
+    let max_degree = (1 << app_log_blow_up) + 1;
+    let SymbolicConstraints {
+        constraints,
+        interactions,
+    } = get_constraints(air);
+    let perm_width = find_interaction_chunks(&interactions, max_degree)
```
qwang98: `max_degree` shouldn't be 0 here, because the degree actually matters for the # of log up columns, and we should confirm whether we hardcode it (as currently done) or expose as a parameter.

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +331,62 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
...
     let air_keygen_builder = AirKeygenBuilder::new(config.pcs(), air);
-    let builder = air_keygen_builder.get_symbolic_builder(None);
-    builder.constraints()
+    air_keygen_builder.get_symbolic_builder(max_constraint_degree)
+}
+
+pub struct AirWidth {
+    pub base_width: usize,
+    pub log_up_width: usize,
+}
```
Schaeff:
```suggestion
pub struct AirWidths {
    pub base: usize,
    pub log_up: usize,
}
```

openvm/src/lib.rs
```diff
@@ -502,7 +502,7 @@ pub struct OriginalCompiledProgram {
 
 pub struct AirMetrics {
     pub name: String,
-    pub width: usize,
+    pub width: AirWidth,
```
Schaeff:
```suggestion
    pub widths: AirWidths,
```

openvm/src/extraction_utils.rs
Schaeff: I think this is incorrect:
- `find_interaction_chunks` does return the number of "columns" needed for the log up phase, but these "columns" are of type `Challenge` which is typically in the extension field, so ends up taking many columns. We need to somehow access the width of a challenge to know the multiplier. From the verifier code, it seems like it would be `<SC::Challenge as FieldExtensionAlgebra<Val<SC>>>::D`
- `STARK_LU_NUM_CHALLENGES` is a constant (2) relating to how many challenges FRI log up needs. It is not the number of base elements per extension field element.
- the `STARK_LU_NUM_CHALLENGES` challenges are not part of the log up trace. They are drawn before we fill up the log up trace, based on the values of the main trace. So they should not be counted in `log_up_width`.
- I think we should keep the `exposed_width` stuff out. It is a constant and there are other costs that we do not account for when we simplify verifier complexity to only consider column count, so we should ignore that as well.

All in all, my impression is that the number of columns should be:
`main_trace_width + (interaction_partitions.len() + 1) * extension degree`.

qwang98: - Response to bullet point 1: Good point on extension fields! I think `find_interaction_chunks` actually also creates `Entry::Permutation` columns here: https://github.com/powdr-labs/stark-backend/blob/main/crates/stark-backend/src/air_builders/symbolic/mod.rs#L409, while the challenges are created here: https://github.com/powdr-labs/stark-backend/blob/main/crates/stark-backend/src/air_builders/symbolic/mod.rs#L415. The `Entry::Permutation` columns have length `num_chunks` whereas the `Entry::Challenge` columns have length `phase_shape.num_challenges == STARK_LU_NUM_CHALLENGES` according to this snippet: https://github.com/powdr-labs/stark-backend/blob/main/crates/stark-backend/src/interaction/mod.rs#L259-L270.
- Response to bullet point 2: I agree, and according to the links above, does it mean the number of challenges is `STARK_LU_NUM_CHALLENGES * extension_columns_per_challenge`.
- Response to bullet point 3: should we create another field in `AirWidth` as `challenge_width`?
- Response to bullet point 4: I'm good with this. As a separate question, I was under the impression we simplify column count to accommodate for recursion proofs (maybe also verifier complexity)? Also are `exposed_width` fixed columns or maybe not materialized as columns at all?

Response to final conclusion: with these additional info, my impression is that the number of columns (for each AIR) is `main_trace_width + (interaction_partitions.len() + 1) + STARK_LU_NUM_CHALLENGES * extension_degree`, which is `main + Entry::Perm (log up) width + Challenge width`. Let me know if you disagree though :)

qwang98: As another question, I'm not sure if `Entry::Perm` are extension fields as well or just the challenges. I remember it's just the challenges in Powdr legacy asm implementation, but not sure about here.

Schaeff:
```rust
pub enum Entry {
    Preprocessed {
        offset: usize,
    },
    /// Main may be partitioned
    Main {
        part_index: usize,
        offset: usize,
    },
    Permutation {
        offset: usize,
    },
    Public,
    Challenge,
    Exposed,
}
```

But why do we need to look at `Entry::Permutation`, if `find_interaction_chunks` already gives us the number of permutation columns? The formula I shared above should be enough (adding preprocessed columns which I forgot)

```
preprocessed_trace_width + main_trace_width + (interaction_partitions.len() + 1) * <SC::Challenge as FieldExtensionAlgebra<Val<SC>>>::D`
```

Similar to how it's done [here](https://github.com/powdr-labs/stark-backend/blob/main/crates/stark-backend/src/keygen/mod.rs#L123)

qwang98: Yeah. I mean we've reached consensus after the discussion so my comments above are outdated :)

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +331,71 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
...
+
+    let symbolic_rap_builder = symbolic_builder_with_degree(air, Some(max_degree));
+    let preprocess = symbolic_rap_builder.width().preprocessed.unwrap_or(0);
+
+    let SymbolicConstraints {
+        constraints,
+        interactions,
+    } = symbolic_rap_builder.constraints();
+
+    // TODO: replace hardcoded EXT_DEGREE with `<SC::Challenge as FieldExtensionAlgebra<Val<SC>>>::D`` once `BabyBearSC` is generic.
```
qwang98: I tried a few ways of getting the extension field # of columns per base field, but none worked:
1. Cannot access `BabyBearSC::Challenge`, because it's a type alias, so would need to fully spell out the original type, which would include all associated types of `StarkGenericConfig`, which will take a lot of space.
2. I traced up the `Challenge` type to `plonky3::field::extension::BinomialExtensionField<MontyField31<BabyBearParameters>, 4>`, so we would essentially need to do `plonky3::field::extension::BinomialExtensionField<MontyField31<BabyBearParameters>, 4>::D`, for which we might as well just hardcode `EXT_DEGREE = 4` for now and use the proper syntax once the stark config is generic.

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +331,72 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
...
+
+pub fn get_air_metrics(air: Arc<dyn AnyRap<BabyBearSC>>) -> AirMetrics {
+    let app_log_blow_up = 2;
+    let max_degree = (1 << app_log_blow_up) + 1;
+
+    let name = air.name();
+    let base = air.width();
+
+    let symbolic_rap_builder = symbolic_builder_with_degree(air, Some(max_degree));
+    let preprocess = symbolic_rap_builder.width().preprocessed.unwrap_or(0);
```
qwang98: According to tests in #2962, `preprocess` is always 0, but I think it should be the correct API to fetch it from.

Note that `width()` from symbolic_rap_builder takes `TraceWidth.preprocessed` from `AirKeygenBuilder.prep_key_gen_data`, which is first built here:

```

impl<SC: StarkGenericConfig> AirKeygenBuilder<SC> {
    fn new(pcs: &SC::Pcs, rap_phase_seq_kind: RapPhaseSeqKind, air: Arc<dyn AnyRap<SC>>) -> Self {
        let prep_keygen_data = compute_prep_data_for_air(pcs, air.as_ref());
        AirKeygenBuilder {
            air,
            rap_phase_seq_kind,
            prep_keygen_data,
        }
    }
```

qwang98: `prep_keygen_data` is built here, which will be `None` if `vpdata_opt` is `None`, which will be `None` if `preprocessed_trace` is `None`, which will be `None` if the `preprocessed_trace` trait function of `BaseAir` is the default implementation. This means that the `air` that is the start point of our metrics calculation must implement this `preprocessed_trace` function, which basically says that none of our airs (Powdr ones and non-Powdr ones in the allow list) implement this trait. I checked that the `BaseAir` implementation of `PowdrAir` has no `preprocessed_trace`, but do you think that it's normal that none of the non-Powdr allow list airs has it?

```
fn compute_prep_data_for_air<SC: StarkGenericConfig>(
    pcs: &SC::Pcs,
    air: &dyn AnyRap<SC>,
) -> PrepKeygenData<SC> {
    let preprocessed_trace = air.preprocessed_trace();
    let vpdata_opt = preprocessed_trace.map(|trace| {
        let domain = pcs.natural_domain_for_degree(trace.height());
        let (commit, data) = pcs.commit(vec![(domain, trace.clone())]);
        let vdata = VerifierSinglePreprocessedData { commit };
        let pdata = ProverOnlySinglePreprocessedData {
            trace: Arc::new(trace),
            data: Arc::new(data),
        };
        (vdata, pdata)
    });
    if let Some((vdata, pdata)) = vpdata_opt {
        PrepKeygenData {
            prover_data: Some(pdata),
            _verifier_data: Some(vdata),
        }
    } else {
        PrepKeygenData {
            prover_data: None,
            _verifier_data: None,
        }
    }
}
```

```
pub trait BaseAir<F>: Sync {
    /// The number of columns (a.k.a. registers) in this AIR.
    fn width(&self) -> usize;

    fn preprocessed_trace(&self) -> Option<RowMajorMatrix<F>> {
        None
    }
}
```

qwang98: The other option is to call a keygen function in the link you provided in a comment above: https://github.com/powdr-labs/stark-backend/blob/main/crates/stark-backend/src/keygen/mod.rs#L123

However, this is typically done at proving and I'm not sure if we'd want to go this length to get a preprocessed metric.

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +331,72 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
...
     let air_keygen_builder = AirKeygenBuilder::new(config.pcs(), air);
-    let builder = air_keygen_builder.get_symbolic_builder(None);
-    builder.constraints()
+    air_keygen_builder.get_symbolic_builder(max_constraint_degree)
+}
+
+pub struct AirWidths {
+    pub preprocess: usize,
+    pub base: usize,
+    pub log_up: usize,
```
Schaeff:
```suggestion
    pub preprocessed: usize,
    pub main: usize,
    pub log_up: usize,
```

Schaeff (changes requested):

openvm/src/extraction_utils.rs
```diff
@@ -2,13 +2,17 @@ use std::collections::{BTreeMap, BTreeSet, HashMap};
...
+use crate::{AirMetrics, IntoOpenVm};
 use openvm_circuit::arch::{VmChipComplex, VmConfig, VmInventoryError};
 use openvm_circuit_primitives::bitwise_op_lookup::SharedBitwiseOperationLookupChip;
 use openvm_circuit_primitives::range_tuple::SharedRangeTupleCheckerChip;
 use openvm_instructions::VmOpcode;
 use openvm_sdk::config::{SdkVmConfig, SdkVmConfigExecutor, SdkVmConfigPeriphery};
+use openvm_stark_backend::air_builders::symbolic::SymbolicRapBuilder;
+use openvm_stark_backend::config::{PackedChallenge, Val};
+use openvm_stark_backend::interaction::fri_log_up::find_interaction_chunks;
+use openvm_stark_backend::p3_field::FieldExtensionAlgebra;
```
qwang98, line 15: Should be correct imports re packed challenge.

openvm/src/extraction_utils.rs
```diff
@@ -329,12 +331,72 @@ pub fn get_name(air: Arc<dyn AnyRap<BabyBearSC>>) -> String {
...
+        constraints,
+        interactions,
+    } = symbolic_rap_builder.constraints();
+
+    // TODO: replace hardcoded EXT_DEGREE with `<SC::Challenge as FieldExtensionAlgebra<Val<SC>>>::D` once `BabyBearSC` is generic.
+    let log_up = (find_interaction_chunks(&interactions, max_degree)
+        .interaction_partitions()
+        .len()
+        + 1)
+        * <PackedChallenge<BabyBearSC> as FieldExtensionAlgebra<Val<BabyBearSC>>>::D;
```
qwang98, line 365: Compilation error on my local after `cargo clean`:

```
error[E0277]: the trait bound `BinomialExtensionField<..., 4>: FieldExtensionAlgebra<...>` is not satisfied
   --> openvm/src/extraction_utils.rs:358:12
    |
358 |         * <PackedChallenge<BabyBearSC> as FieldExtensionAlgebra<Val<BabyBearSC>>>::D;
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsatisfied trait bound
    |
    = help: the trait `FieldExtensionAlgebra<p3_monty_31::monty_31::MontyField31<BabyBearParameters>>` is not implemented for `BinomialExtensionField<PackedMontyField31Neon<BabyBearParameters>, 4>`
note: there are multiple different versions of crate `p3_field` in the dependency graph
   --> /Users/steve/.cargo/git/checkouts/plonky3-7d8a3b21a665a86f/539bbc8/field/src/field.rs:409:1
    |
409 | / pub trait FieldExtensionAlgebra<Base: FieldAlgebra>:
410 | |     FieldAlgebra
411 | |     + From<Base>
412 | |     + Add<Base, Output = Self>
...   |
416 | |     + Mul<Base, Output = Self>
417 | |     + MulAssign<Base>
    | |_____________________^ this is the required trait
    |
   ::: /Users/steve/.cargo/git/checkouts/plonky3-7d8a3b21a665a86f/539bbc8/field/src/extension/binomial_extension.rs:25:1
    |
25  |   pub struct BinomialExtensionField<FA, const D: usize> {
    |   ----------------------------------------------------- this type doesn't implement the required trait
    |
   ::: /Users/steve/.cargo/git/checkouts/plonky3-7d8a3b21a665a86f/2192432/field/src/extension/complex.rs:8:1
    |
8   |   pub trait ComplexExtendable: Field {
    |   ---------------------------------- this is the found trait
    |
   ::: openvm/src/lib.rs:5:5
    |
5   |   use openvm_circuit::arch::InitFileGenerator;
    |       -------------- one version of crate `p3_field` used here, as a dependency of crate `p3_air`
...
26  |   use powdr_number::{BabyBearField, FieldElement, LargeInt};
    |       ------------ one version of crate `p3_field` used here, as a dependency of crate `p3_baby_bear`
    = help: you can use `cargo tree` to explore your dependency tree
    = note: the full name for the type has been written to '/Users/steve/Documents/repo/powdr/target/debug/deps/powdr_openvm-a0b2338362f2e68c.long-type-12321201571794860768.txt'
    = note: consider using `--verbose` to print the full type name to the console
```

Schaeff: Builds fine on CI so something to do with your local config

qwang98:
Good to go again :)

Schaeff (approved):

Schaeff (changes requested):
Withdrawing approval until we figure out what's wrong with reth

Cargo.toml
```diff
@@ -83,47 +83,47 @@ powdr-autoprecompiles = { path = "./autoprecompiles", version = "0.1.4" }
...
+openvm-bigint-circuit = { git = "https://github.com/powdr-labs/openvm.git", rev = "d730a36" }
+openvm-bigint-transpiler = { git = "https://github.com/powdr-labs/openvm.git", rev = "d730a36" }
+openvm-pairing-circuit = { git = "https://github.com/powdr-labs/openvm.git", rev = "d730a36" }
+openvm-pairing-transpiler = { git = "https://github.com/powdr-labs/openvm.git", rev = "d730a36" }
+openvm-native-circuit = { git = "https://github.com/powdr-labs/openvm.git", rev = "d730a36", default-features = false }
+openvm-native-recursion = { git = "https://github.com/powdr-labs/openvm.git", rev = "d730a36", default-features = false }
 
 # stark-backend
-openvm-stark-sdk = { git = "https://github.com/powdr-labs/stark-backend.git", rev = "c33fda6", default-features = false, features = [
+openvm-stark-sdk = { git = "https://github.com/powdr-labs/stark-backend.git", rev = "ee4e22b", default-features = false, features = [
```
leonardoalt, line 120: This and next seem correct

leonardoalt (approved):

PR #2962, opened 2025-06-30
Comprehensive Cell Pgo Test

qwang98:
Depends on #2960. I think we've long needed a more comprehensive test for PGO, especially the cell mode.


openvm/src/lib.rs
```diff
@@ -1079,16 +1102,80 @@ mod tests {
...
+        let total_non_apc_metrics = sum_up_air_metrics(non_powdr_metrics);
+        assert_eq!(
+            [
+                total_non_apc_metrics.width.base_width,
+                total_non_apc_metrics.width.log_up_width,
+                total_non_apc_metrics.constraints,
+                total_non_apc_metrics.bus_interactions
+            ],
+            [3657, 272, 4569, 569]
+        );
```
qwang98: In the kecak example, we very originally only count for the 6069 main constraint columns when limiting the max number of columns.

We need to additionally subtract 1341 + 3657 + 272 = 5270 columns from the max number of column limit. This split of 6069 vs 5270 might help explain why we can only create max 25K columns in reth-benchmark vs the alluded 60K in the first place.

Should also add a similar test to reth-benchmark when it's ready in Powdr :)

qwang98:
> I just hope it runs on our server ;)

Ok, seems that the CI is fine :)

openvm/src/extraction_utils.rs
```diff
@@ -380,6 +380,7 @@ pub fn symbolic_builder_with_degree(
     air_keygen_builder.get_symbolic_builder(max_constraint_degree)
 }
 
+#[derive(Debug, Default, Eq, PartialEq)]
 pub struct AirWidths {
     pub preprocess: usize,
     pub base: usize,
```
qwang98: Derived `Eq` so that we can compare the entire type rather than each of its `usize` fields.

openvm/src/lib.rs
```diff
@@ -1079,16 +1101,87 @@ mod tests {
...
+        let expected = AirMetrics {
+            widths: AirWidths {
+                preprocess: 0,
+                base: 3657,
+                log_up: 896,
+            },
+            constraints: 4569,
+            bus_interactions: 569,
+            ..Default::default()
+        };
```
qwang98: As of this PR, in cell PGO, we subtract the total non APC columns from the max column limit, and then use the base APC columns as cost, so here we have `non APC columns = 0 + 3657 + 896 = 4553` and `APC base columns = 5443` where `4553 + 5443 = 9996 < 10000`. This will be fixed in the next down stream PR to also calculate total columns for APC.

openvm/src/lib.rs
```diff
@@ -1085,16 +1107,87 @@ mod tests {
...
+
+        let expected = AirMetrics {
+            widths: AirWidths {
+                preprocessed: 0,
+                main: 494,
+                log_up: 364,
+            },
+            constraints: 322,
+            bus_interactions: 248,
+            ..Default::default()
```
qwang98: Currently we get all Powdr APCs that there are eligible basic blocks for (63), because we haven't counted the log_up columns in Powdr yet, so 7536 + 494 + 364 < 10000. This will be fixed in #2973  that's already pending review.

openvm/src/lib.rs
```diff
@@ -1085,16 +1107,87 @@ mod tests {
...
+        stdin.write(&GUEST_KECCAK_ITER_SMALL);
+        let pgo_data = execution_profile_from_guest(GUEST_KECCAK, GuestOptions::default(), stdin);
         test_keccak_machine(PgoConfig::None);
+        test_keccak_machine(PgoConfig::Instruction(pgo_data.clone()));
+        test_keccak_machine(PgoConfig::Cell(pgo_data, None));
     }
 
     #[test]
-    fn keccak_machine_pgo() {
+    fn keccak_machine_cell_pgo() {
```
Schaeff: I think in this test, rather than exactly which blocks are selected and their detailed stats, we should probably rather only look at the metrics that we care about:
- number of cells saved
- number of columns created

Then as we tweak PGO, we can see how these numbers change here: if we end up with more columns and more trace cells, then it's a regression.

qwang98: Agree, but I'd like to defer this to another PR, as this involves exporting cells saved and APC frequency executed data somehow to `AirMetrics`. I've made an attempt when exporting metrics to a file in #3000, but exporting them to `AirMetrics` will still be different.

This test is more intended to check that our total columns don't exceed the max column input parameter, which I've just added an assertion to be clear about.

qwang98: I also removed the individual block checks.

qwang98: I also renamed this test to indicate it's for the max columns only.

Schaeff (approved):

PR #2967, opened 2025-07-01
Calculate APC log up columns during PGO

qwang98:


openvm/src/customize_exe.rs
```diff
@@ -544,6 +546,9 @@ fn create_apcs_with_cell_pgo<P: IntoOpenVm>(
             )
             .ok()?; // if apc creation fails, filter out this candidate block
 
+            let powdr_air = PowdrAir::new(apc.machine().clone());
+            let metrics = get_air_metrics(Arc::new(powdr_air));
```
qwang98: Currently these two lines don't work, because `get_air_metrics(air: Arc<dyn AnyRap<BabyBearSC>>)` requires input air to be concrete on `BabyBear` while the `powdr_air` created here is generic over `P: IntoOpenVm`. A few things I tried but didn't work:
1. Make `get_air_metrics` generic over `Arc<dyn AnyRap<dyn StarkConfig>>`, but failed.
2. Cast `PowdrAir<P: IntoOpenVm>` to `PowdrAir<OpenVmField<P>>`, but this requires recursively changing all field element usage in its sub fields, and kind of defeats our purpose to remove these casting methods for types that use Powdr field vs OVM field to start with.

qwang98: Found a solution and currently implementing :)

PR #2968, opened 2025-07-01
Rename quadratic symbolic expression.

chriseth:
Renames QuadraticSymbolicExpression to GroupedExpression. The advantage of doing it his way is also that we can slowly change what is called `QuadraticSymbolicExpression` afer this PR, namely the one with SymbolicExpression as RuntimeConstant to GroupedExpression and easily see which on is generic already and which is not.

### DIFF
--- a/constraint-solver/src/grouped_expression.rs
+++ b/constraint-solver/src/grouped_expression.rs
@@ -50,7 +50,7 @@ pub enum Error {
 /// A symbolic expression in unknown variables of type `V` and (symbolically)
 /// known terms, representing a sum of (super-)quadratic, linear and constant parts.
 /// The quadratic terms are of the form `X * Y`, where `X` and `Y` are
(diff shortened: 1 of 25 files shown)


constraint-solver/src/grouped_expression.rs
```diff
@@ -73,10 +75,9 @@ pub struct QuadraticSymbolicExpressionImpl<T, V> {
 
 // TODO: This type is equivalent to a pre-refactoring version of `QuadraticSymbolicExpressionImpl`.
```
pacheco, line 76: is this TODO still relevant/correct?

chriseth: Yes, I will do this in a step by step way one PR at a time and at the end the type will be removed.

pacheco (approved):

PR #2971, opened 2025-07-01
Require field inverse for runtime constant

chriseth:


constraint-solver/src/symbolic_expression.rs
```diff
@@ -508,6 +484,28 @@ impl<T: FieldElement, V: Clone + Eq> RuntimeConstant for SymbolicExpression<T, V
...
+        } else {
+            SymbolicExpression::BinaryOperation(
+                Arc::new(Self::from(T::from(1))),
+                BinaryOperator::Div,
+                Arc::new(self.clone()),
+                Default::default(),
+            )
+        }
+    }
+
```
Schaeff, line 508: It seems like this is just moved from an impl to a trait impl, but not using the changes in `runtime_constant`, correct?

chriseth: Yes exactly, I used specialized impls for SymbolicExpression because I was not sure if it would simplify correctly if this is reduced to multiplying with the inverse.

PR #2973, opened 2025-07-02
Calculate APC non base columns during PGO

qwang98:
Depends on #2960 and #2962


openvm/src/lib.rs
```diff
@@ -1128,60 +1145,76 @@ mod tests {
...
-                log_up: 896,
+                base: 456,
+                log_up: 332,
             },
-            constraints: 4569,
-            bus_interactions: 569,
-            ..Default::default()
+            constraints: 307,
+            bus_interactions: 227,
         };
```
qwang98: This is reduced because now we use the allow list airs, which filters out the KeccakAir and a few more.

openvm/src/customize_exe.rs
```diff
@@ -60,14 +62,13 @@ struct BlockWithApc<P: IntoOpenVm> {
     apc: Apc<P>,
 }
 
-fn generate_apcs_with_pgo<P: IntoOpenVm>(
-    blocks: Vec<BasicBlock<OpenVmField<P>>>,
-    airs: &OriginalAirs<P>,
+fn generate_apcs_with_pgo(
+    blocks: Vec<BasicBlock<BabyBear>>,
```
chriseth, line 66: I have no idea about most of this PR, but what was the reason to make this specific for a single field?

qwang98: It's because in cell_pgo we have to get the number of log up columns, which depends on the specific stark backend used, and it hasn't been made generic yet.

openvm/src/lib.rs
```diff
@@ -1132,56 +1148,72 @@ mod tests {
...
+        assert_eq!(original_air_metrics.len(), 19); // Number of non-APC chips
 
         let expected = AirMetrics {
             widths: AirWidths {
-                preprocessed: 0,
-                main: 3657,
-                log_up: 896,
+                preprocessed: 5,
+                main: 3960,
+                log_up: 920,
```
qwang98: More columns are from the periphery.

openvm/src/extraction_utils.rs
```diff
@@ -347,7 +348,6 @@ pub fn get_air_metrics(air: Arc<dyn AnyRap<BabyBearSC>>) -> AirMetrics {
     let app_log_blow_up = 2;
```
pacheco, line 348: shouldn't this be a parameter? e.g., i think reth uses `app_log_blow_up` 1

qwang98: Yeah...TBH I also had some questions on this as I was hoping this to be a const or some thing generic to our Stark Config. @Schaeff Any thoughts?

PR #2974, opened 2025-07-02
PGO max column CLI

qwang98:
Adding CLI option as a utility and also an example for reth-benchmark, which currently can't parse `Cell(Option<usize>)` and only defaults to `Cell(None)`.

### DIFF
--- a/cli-openvm/src/main.rs
+++ b/cli-openvm/src/main.rs
@@ -197,10 +197,10 @@ fn pgo_config(
                 guest_opts.clone(),
                 stdin_from(input),
             );
-            let max_columns = match cli_max_columns {
-                None => max_columns,
-                _ => panic!("cli --pgo can't parse Cell(Option<usize>), input must be wrong"),
-            };
+            assert!(
+                cli_max_columns.is_none(),
+                "cli --pgo can't parse Cell(Option<usize>), input must be wrong"
+            );
             PgoConfig::Cell(pc_idx_count, max_columns)
         }
         PgoType::Instruction => {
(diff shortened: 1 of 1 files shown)


cli-openvm/src/main.rs
```diff
@@ -32,6 +32,10 @@ enum Commands {
         #[arg(long, default_value_t = PgoType::default())]
         pgo: PgoType,
 
+        /// When `--pgo-mode cell`, the optional max columns
+        #[clap(long)]
+        max_columns: Option<usize>,
+
```
qwang98, line 38: To keep it simplistic, we don't define complex string parsing traits for the argument, but just add an additional `max_columns` which defaults to `None`. In cases where it's `Some` but `PgoType` is not `Cell`, the argument will be ignored automatically.

cli-openvm/src/main.rs
```diff
@@ -172,16 +187,21 @@ fn pgo_config(
...
                 &guest,
                 guest_opts.clone(),
                 stdin_from(input),
             );
-            PgoConfig::Cell(pc_idx_count, max_total_apc_columns)
+            assert!(
+                cli_max_columns.is_none(),
+                "cli --pgo can't parse Cell(Option<usize>), input must be wrong"
+            );
+            PgoConfig::Cell(pc_idx_count, max_columns)
```
qwang98, line 204: The default CLI parsing can just parse enum name, not elements wrapped in the enum, e.g. it can only parse `Cell` but not `Cell(var)`. Therefore, `cli_max_columns` is guaranteed to be `None` here, which we assert, and pass in the other CLI argument `max_columns` instead.

PR #2975, opened 2025-07-02
Var transformable

chriseth:
Yet another trait we need to make GroupedExpression generic.


constraint-solver/src/runtime_constant.rs
```diff
@@ -85,6 +85,11 @@ pub trait Substitutable<V> {
     fn substitute(&mut self, variable: &V, substitution: &Self);
 }
 
+pub trait VarTransformable<V1, V2> {
```
Schaeff, line 91: Can you add docs for what this trait does

Schaeff (approved):

PR #2982, opened 2025-07-02
nightly benchmarks

pacheco:
This PR runs a few benchmarks nightly, saving the results to a separate repo: `bench-results`.
Idea is to get something going on and improve reporting later.
Currently, it runs `keccak` and `reth`.

### DIFF
--- a/openvm/scripts/run_guest_benches.sh
+++ b/openvm/scripts/run_guest_benches.sh
@@ -36,12 +36,12 @@ run_bench() {
     pushd "$dir"
     # prove with manual precompile if given
     if [ -n "$guest_manual" ]; then
-        cargo run --bin powdr_openvm -r prove $guest_manual --input "$input" --metrics manual.json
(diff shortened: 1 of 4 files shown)


openvm/scripts/run_guest_benches.sh
```diff
@@ -0,0 +1,55 @@
...
+run_bench() {
+    guest="$1"
+    guest_manual="$2"
+    apcs="$3"
+    input="$4"
+    dir="results/$guest"
+    mkdir -p "$dir"
+    pushd "$dir"
+    # prove with manual precompile if given
+    if [ -n "$guest_manual" ]; then
```
leonardoalt, line 38: but if it's not given does that change the indices of the arguments?

pacheco: it needs to be given, but should be the empty string `""`

leonardoalt (approved):

PR #2984, opened 2025-07-02
get `SdkVmConfig` from the openvm.toml file

pacheco:
instead of hardcoding extensions, load the SdkVmConfig from the openvm.toml file in the guest

### DIFF
--- a/openvm/src/lib.rs
+++ b/openvm/src/lib.rs
@@ -290,15 +290,6 @@ pub fn compile_openvm(
     guest: &str,
     guest_opts: GuestOptions,
 ) -> Result<OriginalCompiledProgram, Box<dyn std::error::Error>> {
-    // wrap the sdk config (with the standard extensions) in our custom config (with our custom extension)
-    let sdk_vm_config = SdkVmConfig::builder()
-        .system(Default::default())
-        .rv32i(Default::default())
-        .rv32m(Default::default())
-        .io(Default::default())
-        .keccak(Default::default())
-        .build();
-
     let sdk = Sdk::default();
 
     // Build the ELF with guest options and a target filter.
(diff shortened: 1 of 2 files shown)


openvm/src/lib.rs
```diff
@@ -290,15 +290,6 @@ pub fn compile_openvm(
     guest: &str,
     guest_opts: GuestOptions,
 ) -> Result<OriginalCompiledProgram, Box<dyn std::error::Error>> {
-    // wrap the sdk config (with the standard extensions) in our custom config (with our custom extension)
-    let sdk_vm_config = SdkVmConfig::builder()
-        .system(Default::default())
-        .rv32i(Default::default())
-        .rv32m(Default::default())
-        .io(Default::default())
-        .keccak(Default::default())
```
leonardoalt, line 299: Won't the tests fail if we remove keccak from the default list?

leonardoalt: Now I'm kinda confused because the tests do pass but the keccak guest doesn't have the openvm.toml file

Schaeff: Lower down it uses a default one if the file is not provided. Could we make it a hard error instead?

pacheco: I kept the default because thats also the behaviour of openvm, that way programs that don't use the specialized precompiles (e.g., keccak, curves...) don't need the openvm.toml

pacheco: note that the `openvm.toml` can also include fri params (see `AppConfig` members)  but I'm ignoring those in favour what was hardcoded in our code

leonardoalt: > Lower down it uses a default one if the file is not provided. Could we make it a hard error instead?

yes but the default one doesn't have Keccak

pacheco: but the `guest-keccak-manual-precompile` does have the toml

leonardoalt: ah right, and our tests don't actually need it

leonardoalt (approved):

PR #2985, opened 2025-07-03
Keep a single PGO strategy

Schaeff:
Simplification of the PGO code:
- only keep the Cell strategy, remove the enum
- Introduce `ExecutionProfile` struct which covers both pgo runs and no pgo data. Rename `pgo_config` to `execution_profile`
- Move `max_column_count` to `PowdrConfig`: like the degree bound and the number of autoprecompiles, it is a parameter the user sets about the output of the apc process

### DIFF
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -67,6 +67,7 @@ fn generate_apcs_with_pgo<P: IntoOpenVm>(
     original_config: &OriginalVmConfig,
     execution_profile: ExecutionProfile,
 ) -> Vec<BlockWithApc<P>> {
+    // TODO: inline this function call
     let res = create_apcs_with_cell_pgo(
         blocks,
         execution_profile,
--- a/openvm/src/lib.rs
+++ b/openvm/src/lib.rs
@@ -104,60 +104,8 @@ pub struct ExecutionProfile {
 
 impl ExecutionProfile {
     pub fn create(program: OriginalCompiledProgram, inputs: StdIn) -> ExecutionProfile {
-        let OriginalCompiledProgram { exe, sdk_vm_config } = program;
-
-        // in memory collector storage
-        let collected = Arc::new(Mutex::new(Vec::new()));
-        let collector_layer = PgoCollector {
-            pc: collected.clone(),
-        };
-
-        // build subscriber
-        let subscriber = Registry::default().with(collector_layer);
-
-        // prepare for execute
-        let sdk = Sdk::default();
-
-        // dispatch constructs a local subscriber at trace level that is invoked during data collection but doesn't override the global one at info level
-        let dispatch = Dispatch::new(subscriber);
-        tracing::dispatcher::with_default(&dispatch, || {
-            sdk.execute(exe.clone(), sdk_vm_config.clone(), inputs)
-                .unwrap();
-        });
-
-        // collect the pc's during execution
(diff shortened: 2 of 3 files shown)


openvm/src/customize_exe.rs
```diff
@@ -453,197 +488,3 @@ pub fn openvm_bus_interaction_to_powdr<F: PrimeField32, P: FieldElement>(
...
-    max_total_columns: Option<usize>,
-    airs: &OriginalAirs<P>,
-    config: &PowdrConfig,
-    original_config: &OriginalVmConfig,
-    bus_map: &BusMap,
-) -> Vec<BlockWithApc<P>> {
-    // drop any block whose start index cannot be found in pc_idx_count,
-    // because a basic block might not be executed at all.
-    // Also only keep basic blocks with more than one original instruction.
-    blocks.retain(|b| {
```
Schaeff, line 470: We don't need to do this here. We can do it during the single iteration on the blocks lower down.

openvm/src/customize_exe.rs
```diff
@@ -66,27 +65,125 @@ fn generate_apcs_with_pgo<P: IntoOpenVm>(
...
+    // filter out opcodes that contain next references in their air, because they are not supported yet in apc
+    let air_width_by_opcode = airs.air_width_per_opcode();
+
+    // we are only interested in the top `config.autoprecompiles + config.skip_autoprecompiles` blocks
+    let max_cache = (config.autoprecompiles + config.skip_autoprecompiles) as usize;
+
+    tracing::info!(
+        "Generating autoprecompiles for all ({}) basic blocks in parallel and caching costliest {}",
+        blocks.len(),
+        max_cache,
```
Schaeff: I kept this function to make reviewing easier, but it could be inlined here

openvm/src/lib.rs
```diff
@@ -658,66 +706,9 @@ pub fn execution_profile_from_guest(
...
-        let mut pc_index_count_sorted: Vec<_> = pc_index_count.iter().collect();
-        pc_index_count_sorted.sort_by(|a, b| b.1.cmp(a.1));
-        pc_index_count_sorted.iter().for_each(|(pc, count)| {
-            tracing::debug!("pc_index {}: {}", pc, count);
-        });
-    }
-
-    pc_index_count
+    ExecutionProfile::create(program, inputs)
 }
```
Schaeff, line 661: avoiding a clone here, which is all pcs ever visited. A further improvement here would be to already collect into the hashmap instead of into this large vector.

openvm/src/lib.rs
```diff
@@ -984,45 +944,7 @@ mod tests {
         let mut stdin = StdIn::default();
         stdin.write(&GUEST_KECCAK_ITER);
         let config = PowdrConfig::new(GUEST_KECCAK_APC, GUEST_KECCAK_SKIP);
-        prove_mock(GUEST_KECCAK, config, stdin, PgoConfig::None, None);
-    }
-
-    // Create multiple APC for 10 Keccak iterations to test different PGO modes
-    #[test]
-    fn keccak_prove_multiple_pgo_modes() {
```
Schaeff, line 992: Removed this test because Instruction mode does not exist anymore

leonardoalt:
@Schaeff is this still wanted?

Schaeff:
Closing in favour of #3188

PR #2989, opened 2025-07-03
Avoid collecting into pc vector

Schaeff:
When we generate the PGO data, we first collect all visited pcs in execution order which is linear in the execution length.
Then we reduce this to a map of unique PC to number of occurrences, which is linear in the program size.

This PR removes the first collection, generating the map directly.

### DIFF
--- a/openvm/src/lib.rs
+++ b/openvm/src/lib.rs
@@ -669,9 +669,11 @@ pub fn execution_profile(program: OriginalCompiledProgram, inputs: StdIn) -> Has
     let OriginalCompiledProgram { exe, sdk_vm_config } = program;
 
     // in memory collector storage
-    let collected = Arc::new(Mutex::new(Vec::new()));
+    let collected = Arc::new(Mutex::new(HashMap::new()));
     let collector_layer = PgoCollector {
-        pc: collected.clone(),
+        base: exe.program.pc_base,
+        step: exe.program.step,
+        map: collected.clone(),
     };
 
     // build subscriber
@@ -688,25 +690,17 @@ pub fn execution_profile(program: OriginalCompiledProgram, inputs: StdIn) -> Has
     });
 
     // collect the pc's during execution
-    let pc = collected.lock().unwrap().clone();
+    let pc_index_count = collected.lock().unwrap().clone();
 
-    // create pc_index map to times executed, where pc_index = (pc - pc_base) / step
+    // // create pc_index map to times executed, where pc_index = (pc - pc_base) / step
     let pc_base = exe.program.pc_base;
-    let step = exe.program.step;
-    let pc_index_count = pc
-        .iter()
-        .fold(std::collections::HashMap::new(), |mut acc, pc| {
-            let pc_index = (*pc as u32 - pc_base) / step;
-            *acc.entry(pc_index).or_insert(0u32) += 1;
-            acc
-        });
 
     // the smallest pc is the same as the base_pc if there's no stdin
-    let pc_min = pc.iter().min().unwrap();
+    let pc_min = pc_index_count.keys().min().unwrap();
     tracing::debug!("pc_min: {}; pc_base: {}", pc_min, pc_base);
 
     // print the total and by pc counts
-    tracing::debug!("Pgo captured {} pc's", pc.len());
+    tracing::debug!("Pgo captured {} pc's", pc_index_count.len());
 
     if tracing::enabled!(Level::DEBUG) {
         // print pc_index map in descending order of pc_index count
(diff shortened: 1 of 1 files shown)


openvm/src/lib.rs
```diff
@@ -742,7 +729,52 @@ impl tracing::field::Visit for PgoData {
...
 struct PgoCollector {
-    pc: Arc<Mutex<Vec<usize>>>,
+    step: usize,
+    pc_base: usize,
+    map: Arc<Mutex<Vec<u32>>>,
+}
+
+impl PgoCollector {
+    fn new<F>(program: &Program<F>) -> Self {
+        let max_pc = program.instructions_and_debug_infos.len() * program.step as usize;
```
Schaeff: The pc values are multiples of 4 so we make more space

openvm/src/lib.rs
```diff
@@ -742,7 +729,52 @@ impl tracing::field::Visit for PgoData {
...
+    }
+
+    fn take(self) -> HashMap<u32, u32> {
+        // take the map from the mutex
+        let map = std::mem::take(&mut *self.map.lock().unwrap());
+
+        // Turn the map into a HashMap of (pc_index, count)
+        map.chunks(self.step)
+            .enumerate()
+            .filter_map(|(pc_index, chunk)| {
```
Schaeff: Now `pc_index` is the index, not the multiple of 4

openvm/src/lib.rs
```diff
@@ -742,7 +729,52 @@ impl tracing::field::Visit for PgoData {
...
+                // Sanity check that all counts are zero except the first one
```
leonardoalt: Not sure I get this here. All counts of what are 0 except first what?

Schaeff: We have an assumption that during execution, `pc` is always a multiple of `step`.
Therefore, if we chunk by `step`, only the first value of each chunk can be non-zero.
An alternative way would be to divide the `pc` by `step` when we increment its count, as we did before. Then we don't have to chunk here. I'm fine with either.

Schaeff: Will do this, it's cleaner.

openvm/src/lib.rs
```diff
@@ -742,7 +730,44 @@ impl tracing::field::Visit for PgoData {
...
 struct PgoCollector {
-    pc: Arc<Mutex<Vec<usize>>>,
+    step: usize,
+    pc_base: usize,
+    map: Arc<Vec<AtomicU32>>,
+}
+
+impl PgoCollector {
+    fn new<F>(program: &Program<F>) -> Self {
+        let max_pc = program.instructions_and_debug_infos.len();
```
qwang98: Can we call this `max_pc_index`?

I forgot about this detail, but are program PCs contiguous, so that this length can represent `max_pc_index`?

openvm/src/lib.rs
```diff
@@ -742,7 +730,44 @@ impl tracing::field::Visit for PgoData {
 // A Layer that collects data we are interested in using for the pgo from the trace fields.
 #[derive(Clone)]
 struct PgoCollector {
-    pc: Arc<Mutex<Vec<usize>>>,
+    step: usize,
+    pc_base: usize,
+    map: Arc<Vec<AtomicU32>>,
```
qwang98: Can we call this `pc_index_map`?

openvm/src/lib.rs
```diff
@@ -669,13 +671,10 @@ pub fn execution_profile(program: OriginalCompiledProgram, inputs: StdIn) -> Has
...
     // in memory collector storage
-    let collected = Arc::new(Mutex::new(Vec::new()));
-    let collector_layer = PgoCollector {
-        pc: collected.clone(),
-    };
+    let collector = PgoCollector::new(&exe.program);
 
     // build subscriber
-    let subscriber = Registry::default().with(collector_layer);
+    let subscriber = Registry::default().with(collector.clone());
```
qwang98, line 677: Nice :) cleaner

openvm/src/lib.rs
qwang98: In practice, if tests involving PGO passes, it should be contiguous, or otherwise the `increment()` will report an index error.

Schaeff: Yes it's continuous

Schaeff: Done

openvm/src/lib.rs
Schaeff: Done

qwang98: Somehow the changes doesn't seem to be in

openvm/src/lib.rs
qwang98: Some how the changes doesn't seem to be in

qwang98 (changes requested):
Did you forget to push maybe?

openvm/src/lib.rs
Schaeff: Yes, pushed now

qwang98 (approved):
lgtm :)

PR #3002, opened 2025-07-07
Introduce constant for log blowup

Schaeff:
1 is recommended, and having a constant is an improvement over the current way the blowup factor can be set everywhere.

update: 1 does not work, so set to 2 for now, as it was before.


qwang98:
Umm conflicts

qwang98 (changes requested):
Removing approval pending resolution on constraint degree too high on nightly.

leonardoalt:
> Removing approval pending resolution on constraint degree too high on nightly.

I think this was changed back to 2 and we can figure out the degree=1 in another PR

leonardoalt (approved):

PR #3005, opened 2025-07-08
Use hint for pairing guest

ShuangWu121:
Use hint to do pairing check, but the compilation then become very slow and the proving takes too long to finish. 

One problem I can think of is that the pairing hint ([these two values](https://github.com/powdr-labs/powdr/blob/a7712b72019e6476bf4881fdce19a39b92fc5fe7/openvm/guest-pairing/src/pairing_check_with_hint.rs#L23)) now is computed inside the zkvm, but it should be provided to the zkvm directly.  The computation of the hint is expensive.

The program can be tested on the host by cargo build and cargo run from the guest-pairing folder.


ShuangWu121:
Closed due to not implementing pairing guest.

PR #3000, opened 2025-07-04
Utils for future APC optimization

qwang98:
Required for efficient analysis of future APC optimizations:
1. Cell pgo now outputs `cell_pgo.metrics` (sample below)
2. `debug.pil` now writes out powdr opcode, e.g. `namespace powdr_air_for_opcode_4394;`

The user can therefore start from `cell_pgo.metrics` and then cross reference to `debug.pil` via the opcode, potentially using the assembly builder after (thanks to pretty printing of basic block instructions).

There should be no additional computation cost as it just exports what's already calculated during cell pgo.

```
Apc(opcode: 4391, execution_frequency: 30, cells_saved_per_row: 211, cells_per_row_after_saving: 81, percent_saved: 72.26%)
BasicBlock(start_idx: 1386, statements: [
   instr   0:   AND rd_ptr = 52, rs1_ptr = 44, rs2 = 3, rs2_as = 0
   instr   1:   SLTU rd_ptr = 52, rs1_ptr = 52, rs2 = 1, rs2_as = 0
   instr   2:   SLTU rd_ptr = 56, rs1_ptr = 48, rs2 = 1, rs2_as = 0
   instr   3:   OR rd_ptr = 52, rs1_ptr = 52, rs2 = 56, rs2_as = 1
   instr   4:   BNE 52 0 248 1 1
])

Apc(opcode: 4392, execution_frequency: 30, cells_saved_per_row: 159, cells_per_row_after_saving: 75, percent_saved: 67.95%)
BasicBlock(start_idx: 1452, statements: [
   instr   0:   ADD rd_ptr = 52, rs1_ptr = 40, rs2 = 0, rs2_as = 0
   instr   1:   ADD rd_ptr = 56, rs1_ptr = 44, rs2 = 0, rs2_as = 0
   instr   2:   AND rd_ptr = 44, rs1_ptr = 52, rs2 = 3, rs2_as = 0
   instr   3:   BNE 44 0 2013265733 1 1
])
```

PR #3008, opened 2025-07-08
Write APC candidates to disk

Schaeff:
@georgwiese built some graphs from the APC candidates by exporting the main stats to CSV. This PR start extending that by persisting the candidates to disk for further inspection.

This first PR creates one file per APC, and #3015 creates a single JSON which has the main stats and points to this APC file.


openvm/src/customize_exe.rs
```diff
@@ -54,8 +56,9 @@ impl From<powdr_autoprecompiles::constraint_optimizer::Error> for Error {
     }
 }
 
-struct BlockWithApc<P: IntoOpenVm> {
-    block: BasicBlock<OpenVmField<P>>,
+#[derive(Serialize, Deserialize)]
+struct BlockWithApc<P, T> {
```
Schaeff, line 60: Changed because serde doesn't like trait bounds in struct definitions

openvm/src/customize_exe.rs
```diff
@@ -454,6 +457,93 @@ pub fn openvm_bus_interaction_to_powdr<F: PrimeField32, P: FieldElement>(
...
+#[derive(Serialize, Deserialize)]
+struct ApcCandidate<P, T> {
+    block_with_apc: BlockWithApc<P, T>,
+    execution_frequency: usize,
+    cells_saved_per_row: usize,
+    width: usize, // only tag this field in Pgo::Cell, the only place it's needed
+}
+
+impl ApcCandidate<BabyBearField, OpenVmField<BabyBearField>> {
+    pub fn try_create(
```
Schaeff, line 489: Moved into this function, and added the candidates dir path

pacheco: minor: maybe `save_to_disk` could be called separately from creating the candidate (instead of extra arg here)?

openvm/src/lib.rs
```diff
@@ -393,6 +396,11 @@ impl PowdrConfig {
             ..self
         }
     }
+
+    pub fn with_apc_candidates_folder<P: AsRef<Path>>(mut self, path: P) -> Self {
```
pacheco: minor: `_dir`?

pacheco (commented):
just a couple of minor comments, but what is the use case for this?

qwang98:
I did something similar in #3000, so maybe this can be a joint effort to explore potential use cases?

The use case for #3000 was more for discovering additional optimizations. It writes a single `cell_pgo.metrics` file containing multiple Apc candidates of the following format, so the user can look for the most important `ApcCandidate` and cross lookup via the opcode to `Debug.pil` to look at specific constraints/interactions.
```
Apc(opcode: 4391, execution_frequency: 30, cells_saved_per_row: 211, cells_per_row_after_saving: 81, percent_saved: 72.26%)
BasicBlock(start_idx: 1386, statements: [
   instr   0:   AND rd_ptr = 52, rs1_ptr = 44, rs2 = 3, rs2_as = 0
   instr   1:   SLTU rd_ptr = 52, rs1_ptr = 52, rs2 = 1, rs2_as = 0
   instr   2:   SLTU rd_ptr = 56, rs1_ptr = 48, rs2 = 1, rs2_as = 0
   instr   3:   OR rd_ptr = 52, rs1_ptr = 52, rs2 = 56, rs2_as = 1
   instr   4:   BNE 52 0 248 1 1
])
```

I noticed that here we write a `cbor` file for each `ApcCandidate`, so also curious to hear the use case we targeting here!

openvm/src/lib.rs
```diff
@@ -1072,8 +1080,12 @@ mod tests {
 
     // The following are compilation tests only
     fn test_guest_machine(pgo_config: PgoConfig) {
-        let config = PowdrConfig::new(GUEST_APC, GUEST_SKIP_PGO);
-        let machines = compile_guest(GUEST, GuestOptions::default(), config, pgo_config)
+        let apc_candidates_dir = tempfile::tempdir().unwrap();
+        let apc_candidates_dir_path = apc_candidates_dir.path();
+        let config = PowdrConfig::new(GUEST_APC, GUEST_SKIP_PGO)
+            .with_apc_candidates_folder(apc_candidates_dir_path);
+        let should_have_exported_apc_candidates = matches!(pgo_config, PgoConfig::Cell(_, _));
```
qwang98: I wonder if the apc candidates folder should be a part of the `PgoConfig` instead of `PowdrConfig`, because only cell Pgo can have Apc candidates?

Schaeff: Good point. My view is that exporting the APC candidates is not PGO specific. This could even potentially be moved to the autoprecompiles crate. So I think PowdrConfig makes more sense.

openvm/src/customize_exe.rs
Schaeff: Done!

openvm/src/lib.rs
Schaeff: done

openvm/src/lib.rs
```diff
@@ -1084,8 +1092,12 @@ mod tests {
 
     // The following are compilation tests only
     fn test_guest_machine(pgo_config: PgoConfig) {
-        let config = PowdrConfig::new(GUEST_APC, GUEST_SKIP_PGO);
-        let machines = compile_guest(GUEST, GuestOptions::default(), config, pgo_config)
+        let apc_candidates_dir = tempfile::tempdir().unwrap();
+        let apc_candidates_dir_path = apc_candidates_dir.path();
+        let config = PowdrConfig::new(GUEST_APC, GUEST_SKIP_PGO)
+            .with_apc_candidates_dir(apc_candidates_dir_path);
+        let should_have_exported_apc_candidates = matches!(pgo_config, PgoConfig::Cell(_, _));
```
qwang98, line 1099:
```suggestion
        let config = PowdrConfig::new(GUEST_APC, GUEST_SKIP_PGO);
        matches!(pgo_config, PgoConfig::Cell(_, _)).then(|| {
            let apc_candidates_dir = tempfile::tempdir().unwrap();
            let apc_candidates_dir_path = apc_candidates_dir.path();
            config = config.with_apc_candidates_dir(apc_candidates_dir_path);
        });
```

openvm/src/lib.rs
```diff
@@ -1094,6 +1106,14 @@ mod tests {
...
             [49, 22, 31]
         );
+        // In Cell PGO, check that the apc candidates were persisted to disk
+        let files_count = std::fs::read_dir(apc_candidates_dir_path)
+            .expect("Failed to read APC candidates directory")
+            .count();
+        assert!(
+            (files_count > 0) == should_have_exported_apc_candidates,
+            "Unexpected number of APC candidate files"
+        );
```
qwang98, line 1116:
```suggestion
        matches!(pgo_config, PgoConfig::Cell(_, _)).then(|| {
            // In Cell PGO, check that the apc candidates were persisted to disk
            let files_count = std::fs::read_dir(apc_candidates_dir_path)
                .expect("Failed to read APC candidates directory")
                .count();
            assert!(files_count > 0, "Unexpected number of APC candidate files");
        });
```

qwang98:
Did a full review and lgtm.

Had small suggested change that maybe makes the test a bit cleaner.

openvm/src/lib.rs
Schaeff: This tests less than the original, which tests for all PGO modes.

qwang98: Oh I see. We are trying to explicitly give a directory for other modes and make sure it doesn't generate any files.

qwang98 (approved):
lgtm now

PR #3012, opened 2025-07-08
Take other bus interactions into account when computing xor range constraints

chriseth:


autoprecompiles/src/bitwise_lookup_optimizer.rs
```diff
@@ -104,32 +113,46 @@ fn is_simple_multiplicity_bitwise_bus_interaction<T: FieldElement, V: Clone + Ha
...
+    fn get(&self, var: &V) -> RangeConstraint<T> {
+        self.range_constraints.get(var).cloned().unwrap_or_default()
+    }
+}
+
+impl<T: FieldElement, V: Clone + Hash + Ord + Display> RangeConstraints<T, V> {
+    /// Computes range constraints for variables in the machine.
+    /// This is a very basic way to do it, it does not do it iteratively
+    /// and only considers bus interactions.
+    /// We could get better results using the solver.
```
chriseth: Wouldn't be too hard to do it actually....

chriseth: I'll do that tomorrow, but I think in this context, it might be super handy to run the solver, but on the version where bus interaction variables are substituted. Because then, we don't only get range constraints on variables but also on the actual expressions used in the bus interactions, because that is what we want to have.

chriseth:
We can actually do the same for the other bit-range constraint bus send, and here we could maybe do it in a generic way by extending the bus interaction trait.

chriseth:
Oh yeah, after the bugfix, all the improvements in tests that are not sigle-instruction are gone...

chriseth:
Actually no, even the single instruction improvements are gone...

openvm/tests/apc_builder_outputs/memcpy_block.txt
```diff
@@ -29,8 +29,7 @@ mult=is_valid * 1, args=[b__0_0, 3, b__0_0 + 3 - 2 * b__0_1, 1]
 mult=diff_marker__0_1 + diff_marker__1_1 + diff_marker__2_1 + diff_marker__3_1, args=[diff_val_1 - 1, 0, 0, 0]
 mult=diff_marker__0_2 + diff_marker__1_2 + diff_marker__2_2 + diff_marker__3_2, args=[diff_val_2 - 1, 0, 0, 0]
 mult=is_valid * 1, args=[b__0_3, c__0_3, 2 * a__0_4 - (b__0_3 + c__0_3), 1]
-mult=is_valid * 1, args=[b__1_0, b__2_0, 0, 0]
```
chriseth, line 32: It's working!

PR #3015, opened 2025-07-08
Generate APC candidate json

Schaeff:
Follow up to #3008

### DIFF
--- a/openvm/src/customize_exe.rs
+++ b/openvm/src/customize_exe.rs
@@ -1,7 +1,8 @@
 use std::collections::{BTreeSet, HashMap};
 
+use std::io::BufWriter;
 use std::path::Path;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use crate::extraction_utils::{get_air_metrics, OriginalAirs, OriginalVmConfig};
 use crate::opcode::{branch_opcodes_bigint_set, branch_opcodes_set};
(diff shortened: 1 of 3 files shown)


openvm/src/customize_exe.rs
```diff
@@ -587,8 +629,10 @@ fn create_apcs_with_cell_pgo(
         max_total_columns - total_non_apc_columns
     });
 
+    let apc_candidates = Arc::new(Mutex::new(vec![]));
+
     // map–reduce over blocks into a single BinaryHeap<ApcCandidate<P>> capped at max_cache
-    fractional_knapsack(
+    let res = fractional_knapsack(
         blocks.into_par_iter().enumerate().filter_map(|(i, block)| {
             ApcCandidate::try_create(
```
qwang98, line 639: So this should save all ApcCandidates ever computed to the files instead of just the ones that we are left with (after max cache/skip/max column constraints) right?

openvm/src/customize_exe.rs
```diff
@@ -610,12 +655,24 @@ fn create_apcs_with_cell_pgo(
...
+
+    // Write the APC candidates JSON to disk if the directory is specified.
+    if let Some(apc_candidates_dir_path) = &config.apc_candidates_folder {
+        let apc_candidates_json_file = apc_candidates.lock().unwrap();
+        let json_path = apc_candidates_dir_path.join("apc_candidates.json");
+        let file = std::fs::File::create(&json_path)
+            .expect("Failed to create file for APC candidates JSON");
+        serde_json::to_writer(BufWriter::new(file), &*apc_candidates_json_file)
+            .expect("Failed to write APC candidates JSON to file");
+    }
```
qwang98: So we should have one (summary) json file of all ApcCandidate's and one cbor file for each ApcCandidate right?

qwang98:
I like this PR :) as it fills the void that's needed in this other PR: https://github.com/powdr-labs/powdr/pull/2962#discussion_r2193996754

openvm/src/customize_exe.rs
Schaeff: Yes, basically we want to intercept them before they get selected, inspect them, so that we can optimise them better and/or tweak the selection strategy.

qwang98: Cool :)

openvm/src/customize_exe.rs
```diff
@@ -481,11 +482,29 @@ pub fn openvm_bus_interaction_to_powdr<F: PrimeField32, P: FieldElement>(
...
+    total_width_before: usize,
+    // total width after optimisation
+    total_width_after: usize,
+    // path to the apc candidate file
+    apc_candidate_file: String,
 }
 
 impl ApcCandidate<BabyBearField, OpenVmField<BabyBearField>> {
+    /// Try to create an autoprecompile candidate from a block.
+    #[allow(clippy::too_many_arguments)]
```
pacheco: number of args doesn't change, why is this needed?

openvm/src/customize_exe.rs
```diff
@@ -510,43 +529,65 @@ impl ApcCandidate<BabyBearField, OpenVmField<BabyBearField>> {
...
         Some(candidate)
     }
 
     /// Save the candidate to disk.
-    fn save_to_disk(&self, apc_candidates_dir_path: &Path) {
-        let path = apc_candidates_dir_path
+    fn save_to_disk(
+        &self,
+        apc_candidates_dir_path: &Path,
+        apc_candidates_json_file: Arc<Mutex<Vec<ApcCandidateJsonExport>>>,
```
pacheco: maybe this function should instead return the `ApcCandidateJsonExport`?

openvm/src/customize_exe.rs
Schaeff: leftover from before, good catch

openvm/src/customize_exe.rs
Schaeff: done

pacheco (approved):

openvm/src/customize_exe.rs
Schaeff: done

PR #3017, opened 2025-07-09
Display powdr opcode for debug.pil

qwang98:
Currently all Powdr symbolic machines in `debug.pil` appear as `PowdrAir<BabyBear>`, which gives no way of distinguishing one from another.

This PR fixes this by using the existing `air_name` API of powdr executor and plonk executor, which gives names like: `powdr_air_for_opcode_4381`. This makes it easier to debug and in the future, cross reference to `ApcCandidates` cbor and json files written in #3008 and #3015.

### DIFF
--- a/openvm/src/extraction_utils.rs
+++ b/openvm/src/extraction_utils.rs
@@ -3,7 +3,7 @@ use std::sync::{Arc, Mutex};
 
 use crate::air_builder::AirKeygenBuilder;
 use crate::{opcode::instruction_allowlist, BabyBearSC, SpecializedConfig};
-use crate::{AirMetrics, IntoOpenVm};
+use crate::{AirMetrics, IntoOpenVm, SpecializedExecutor};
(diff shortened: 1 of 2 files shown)


openvm/src/extraction_utils.rs
```diff
@@ -299,7 +299,12 @@ pub fn export_pil(writer: &mut impl std::io::Write, vm_config: &SpecializedConfi
...
         let air = executor.air();
-        let name = air.name();
+        let name = match executor {
+            SpecializedExecutor::PowdrExecutor(powdr_executor) => {
+                powdr_executor.air_name() // name with opcode
+            }
+            _ => air.name(),
+        };
 
         if blacklist.contains(&name.as_str()) {
```
chriseth, line 309: Is that a problem after the change, or is it a feature?

qwang98: it's a feature :)

PR #3020, opened 2025-07-09
Move concept of block to APC crate

Schaeff:
Simplify and make the concept of block less openvm specific.
The next step will be to move the saving the apcs to disk to the apc crate.


openvm/src/customize_exe.rs
```diff
@@ -181,13 +174,14 @@ pub fn customize(
     let extensions = blocks_with_apcs
```
qwang98: Can just call this apcs instead of `blocks_with_apcs`.

qwang98 (dismissed):
I inspected the full code and functionally they all make sense to me. Just one small comment above.

On the high level, it also makes sense that block is an Apc concept rather than an OVM concept.

I wonder whether you have any use case to enable in mind or is it more of a clean up? :)

openvm/src/customize_exe.rs
Schaeff: good catch

qwang98 (approved):
Good with change :)

PR #3026, opened 2025-07-10
Improve OVM compilation tests

qwang98:
In general, it makes more sense to always test:
- The full `AirMetrics` (including log up and preprocessed columns), instead of just the main columns.
- The sum of `AirMetrics` for all APCs used, instead of just the first APC in a testing profile (because of wasted data, and we can always just create one APC if we only want to test one).

These won't add any cost to the testing CI run, because all data are already there but we just aren't testing them fully yet.

### DIFF
--- a/openvm/src/lib.rs
+++ b/openvm/src/lib.rs
@@ -538,6 +538,8 @@ pub struct AirMetrics {
     pub bus_interactions: usize,
 }
 
+const APP_LOG_BLOWUP: usize = 2;
+
 impl Add for AirMetrics {
     type Output = AirMetrics;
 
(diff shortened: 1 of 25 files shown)


openvm/src/lib.rs
```diff
@@ -1299,32 +1300,39 @@ mod tests {
         guest: &'a str,
         guest_apc: u64,
         guest_skip: u64,
-        width: usize,
-        constraints: usize,
-        bus_interactions: usize,
-        machine_length: usize,
+        metrics_type: Vec<AirMetricsType>,
+        metrics: Vec<AirMetrics>,
+        machine_count: Vec<usize>,
```
Schaeff: should be a single vector of some struct that has all three

qwang98: Fixed :)

qwang98:
Removed distinction between Powdr and non Powdr metrics for maximum cleanness :)

PR #3032, opened 2025-07-10
Print function name for basic blocks.

chriseth:


Schaeff:
Nice, after #3020 we have a notion of block in the `autoprecompile` crate (`SymbolicBlock`), so I wonder if we should add this metadata in there so we get this for any front end.

openvm/src/customize_exe.rs
```diff
@@ -138,8 +140,13 @@ pub fn customize(
...
+            let name = debug_info
+                .symbols
+                .try_get_one_or_preceding(OPENVM_INIT_PC + block.start_idx as u32)
+                .map(|(symbol, offset)| format!("{} + {offset}", rustc_demangle::demangle(symbol)))
+                .unwrap_or_default();
             tracing::debug!(
-                "Basic block (executed {count} times):\n{}",
+                "Basic block (executed {count} times), {name}:\n{}",
                 block.pretty_print(openvm_instruction_formatter)
             );
```
qwang98: Not sure if I ran it properly, but this is what I got from running the guest program:

```
 Basic block (executed 1023 times), <openvm::serde::err::Error as core::fmt::Debug>::fmt::he2c55f737aaf0e2f + 25:
BasicBlock(start_idx: 97, statements: [
   instr   0:   ADD rd_ptr = 44, rs1_ptr = 40, rs2 = 0, rs2_as = 0
   instr   1:   ADD rd_ptr = 48, rs1_ptr = 48, rs2 = 16777215, rs2_as = 0
   instr   2:   ADD rd_ptr = 40, rs1_ptr = 40, rs2 = 52, rs2_as = 1
   instr   3:   ADD rd_ptr = 52, rs1_ptr = 44, rs2 = 0, rs2_as = 0
   instr   4:   BNE 48 0 2013265905 1 1
])
```

qwang98: A couple questions:
1. I'm not sure if my function name `he2c55f737aaf0e2f` is properly demangled, as it still looks hex gibberish, or are elf program instruction names just like this in the DebugInfo?
2. Are we trying to print the name of the first elf instruction of a basic block? If that's the case, why we also take into account of the first preceding symbol (and its offset to the block start index)?
3. On the high level, what would be the major use case of cross referencing to the elf program?

qwang98: I think I answered my question 1 by looking at this gist: https://gist.github.com/chriseth/7e0e658a3b2197b5f78ef99a979eca9d#file-basic_blocks-txt-L388

I might also answer my question 2. The offset exists because the basic block belongs to the first preceding label, but might not start immediately after the label?

chriseth: The hex suffix is some thing cargo adds to make names unique. The name is always a function name and then there is some offset (number of instructions) into the function. The routine looks for the first symbol in front of the start of the basic block to look for the function the basic block is from. Note that not all labels are also symbols. Most labels in assembly are labels inside a function to jump to, but you would normally not reference that label from outside the function, that's why there is no need for a symbol.

Schaeff (approved):

PR #3033, opened 2025-07-10
Move block detection to APC crate

Schaeff:
Moving block detection to the APC crate. 
After this PR, a larger chunk of the `customize` function is openvm-agnostic.

PR #3036, opened 2025-07-10
Move pgo to apc crate

Schaeff:
Larger PR, but after this basically everything that can be made agnostic is agnostic.
`PowdrConfig` is split in two, as the `autoprecompile` crate does not care about the arithmetization method. This could be cleaned up a bit though.

Need to merge https://github.com/powdr-labs/openvm-reth-benchmark/pull/24/ first and update the hashes.

TODO:
- [x] merge #3033 
- [x] fix reth
- [x] remove POWDR_OPCODE constant in apc and pass it instead

### DIFF
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -168,7 +168,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: powdr-labs/openvm-reth-benchmark
-          ref: 29dc4c3581acbc3818543bef360e17a5938ff19b
+          ref: main
           path: openvm-reth-benchmark
 
       - name: Patch openvm-reth-benchmark to use local powdr


.github/workflows/pr-tests.yml
```diff
@@ -168,7 +168,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: powdr-labs/openvm-reth-benchmark
-          ref: main
+          ref: 29dc4c3581acbc3818543bef360e17a5938ff19b
```
qwang98: This is just to run the CI?

Not sure about the order of update here. It should be this PR and then Reth (so that we have the merge commit of this PR for Reth)?

autoprecompiles/src/blocks/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+            ..self
+        }
+    }
+
+    pub fn with_degree_bound(self, degree_bound: DegreeBound) -> Self {
+        Self {
+            degree_bound,
+            ..self
+        }
+    }
```
qwang98: What would be the use cases of these if the `new()` function requires a value for them already?

autoprecompiles/src/blocks/pgo.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// cost = cells saved per apc * times executed
+    /// max total columns
+    Cell(HashMap<u32, u32>, Option<usize>),
+    /// cost = instruction per apc * times executed
+    Instruction(HashMap<u32, u32>),
+    /// cost = instruction per apc
+    /// cost = instruction per apc
+    #[default]
+    None,
+}
```
qwang98: A few edits (probably should have done at some point before but might as well just fix here):
1. I think at some point we decide to call this `value` (or some other name I forgot) not `cost`.
2. `Pgo::Cell` should also care about the cost (as in value divided by the cost of # of columns).
3. `None` mode's doc is duplicated.

autoprecompiles/src/blocks/pgo.rs
```diff
@@ -0,0 +1,281 @@
...
+            }
+            PgoConfig::None => None,
+        }
+    }
+}
+
+/// Trait for autoprecompile candidates.
+/// Implementors of this trait wrap an APC with additional data used by the `KnapsackItem` trait to select the most cost-effective APCs.
+pub trait Candidate<P, I, B>: Sized + KnapsackItem {
+    type JsonExport: Serialize + for<'de> Deserialize<'de> + Send;
```
qwang98, line 49: What is `for<'de>` here?

autoprecompiles/src/blocks/pgo.rs
```diff
@@ -0,0 +1,281 @@
...
+                pc_index_count.get(&pc_offset).copied()
+            }
+            PgoConfig::None => None,
+        }
+    }
+}
+
+/// Trait for autoprecompile candidates.
+/// Implementors of this trait wrap an APC with additional data used by the `KnapsackItem` trait to select the most cost-effective APCs.
+pub trait Candidate<P, I, B>: Sized + KnapsackItem {
```
qwang98, line 48: I noticed around the crate that we use the single letter type bounds a lot. While with one or two of them, it can be quite understandable (e.g. `F` for field and `M` for machine), having a few of them can be kind of confusing.

By inferring from the trait functions, `P` is the field type, `I` is the machine handler, and `B` is the bus handler?

As a starter fix, does it make sense to try make letters some what consistent? For example, we have `Apc<T>` in its type definition, so here `P` can be `T`? Or what's the naming convention here (why we call it `P`)?

qwang98: Also we have `VmConfig<M, B>` in its definition, so maybe `M` can become `I` or the reverse.

autoprecompiles/src/blocks/pgo.rs
```diff
@@ -0,0 +1,281 @@
...
+    P: FieldElement,
+    I: InstructionMachineHandler<P> + Clone + Sync,
+    B: BusInteractionHandler<P> + Clone + IsBusStateful<P> + Sync,
+>(
+    mut blocks: Vec<BasicBlock<P>>,
+    pgo_program_idx_count: HashMap<u32, u32>,
+    config: &PowdrConfig,
+    max_total_apc_columns: Option<usize>,
+    vm_config: VmConfig<I, B>,
+) -> Vec<Apc<P>> {
```
qwang98, line 77: I have a feeling that with the logics of cells saved and execution count and etc. out sourced to the `value` and `cost` functions of `KnapsackItem`, this is function maybe should be renamed from `cell_pgo` to something like `knapsack_pgo`?

To be honest not sure about the best name, `knapsack_pgo` can actually be more confusing but also more "proper".

autoprecompiles/src/blocks/pgo.rs
Schaeff: It's the lifetime of the thing we're deserializing from. It's the most generic way to require Deserialize.

autoprecompiles/src/blocks/pgo.rs
Schaeff: #3040 mostly solves this by grouping all these traits as associated traits of a single Adapter trait.

So far in the autoprecompile crate T is the powdr field (the only one we use there) while in the openvm crate P is the powdr field and F is the openvm field.

I would try to keep this PR consistent with the existing names. How about adding some docs where it's not obvious?

autoprecompiles/src/blocks/pgo.rs
Schaeff: Yeah I don't necessarily disagree, but this PR being already large I'd keep renamings to another one if possible

autoprecompiles/src/blocks/pgo.rs
```diff
@@ -0,0 +1,281 @@
...
+    B: BusInteractionHandler<P> + Clone + IsBusStateful<P> + Sync,
+>(
+    blocks: Vec<BasicBlock<P>>,
+    config: &PowdrConfig,
+    max_total_apc_columns: Option<usize>,
+    pgo_config: PgoConfig,
+    vm_config: VmConfig<I, B>,
+) -> Vec<Apc<P>> {
+    // sort basic blocks by:
+    // 1. if PgoConfig::Cell, cost = frequency * cells_saved_per_row
```
qwang98, line 221: Should update the comments here for Cell likewise :)

cli-openvm/src/main.rs
```diff
@@ -129,13 +134,19 @@ fn run_command(command: Commands) {
...
             let pgo_config = pgo_config(guest.clone(), guest_opts.clone(), pgo, max_columns, input);
-            let program =
-                powdr_openvm::compile_guest(&guest, guest_opts, powdr_config, pgo_config).unwrap();
+            let program = powdr_openvm::compile_guest(
+                &guest,
+                guest_opts,
+                powdr_config,
+                IMPLEMENTATION,
+                pgo_config,
+            )
```
qwang98, line 148: So we previously default to non Plonk implementation in `compile_guest` and never use the Plonk one?

autoprecompiles/src/blocks/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// Number of basic blocks to skip for autoprecompiles.
+    /// This is either the largest N if no PGO, or the costliest N with PGO.
+    pub skip_autoprecompiles: u64,
+    /// Max degree of constraints.
+    pub degree_bound: DegreeBound,
+    /// The path to the APC candidates dir, if any.
+    pub apc_candidates_dir_path: Option<PathBuf>,
+    /// The opcode id of the first APC instruction. Other APC instructions will have consecutive ids.
+    pub first_apc_opcode: usize,
+}
```
qwang98: When relooking at the structure, does it make more sense to move `PowdrConfig` to `autoprecompiles/src/lib.rs` (one level up)? The `Apc` struct lives there and I think `PowdrConfig` should be the highest level within the crate if possible.

openvm/src/customize_exe.rs
```diff
@@ -131,6 +98,18 @@ pub fn customize(
...
+            let total_non_apc_columns = original_config
+                .chip_inventory_air_metrics()
+                .values()
+                .map(|m| m.total_width())
+                .sum::<usize>();
+            max_total_columns - total_non_apc_columns
+        }),
+        PgoConfig::Instruction(_) | PgoConfig::None => None,
+    };
+
```
qwang98, line 112: Cool to put here as columns calculation depend on OVM.

openvm/src/customize_exe.rs
```diff
@@ -248,7 +233,7 @@ pub fn customize(
 
     CompiledProgram {
         exe,
-        vm_config: SpecializedConfig::new(original_config, extensions, config.implementation),
+        vm_config: SpecializedConfig::new(original_config, extensions, implementation),
```
qwang98, line 236: I seem to answer my question earlier as implementation (for example, the single row chip) is an OVM concept, and therefore not a part of PowdrConfig any more.

autoprecompiles/src/blocks/pgo.rs
qwang98: I see, `P` for `Powdr` lol

qwang98: Will take a look at #3040.

qwang98 (changes requested):
I think mostly boils down to some smaller comments but the whole structure looks good to me.

autoprecompiles/src/blocks/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+pub use pgo::PgoConfig;
+pub use pgo::{generate_apcs_with_pgo, Candidate};
+pub use selection::KnapsackItem;
+
+#[derive(Clone)]
+pub struct PowdrConfig {
+    /// Number of autoprecompiles to generate.
+    pub autoprecompiles: u64,
+    /// Number of basic blocks to skip for autoprecompiles.
+    /// This is either the largest N if no PGO, or the costliest N with PGO.
```
leonardoalt: This comment seems out of place?

autoprecompiles/src/blocks/mod.rs
leonardoalt: I think primarily if you don't want to set all of them besides default, you could do default and then one of these. But here we don't have default so could maybe remove these, but I do think they're used? @Schaeff

autoprecompiles/src/blocks/selection.rs
```diff
@@ -0,0 +1,203 @@
...
+};
+
+use rayon::iter::{IntoParallelIterator, ParallelIterator};
+
+pub trait KnapsackItem {
+    /// Cost of the item, used for sorting and knapsack algorithm.
+    fn cost(&self) -> usize;
+    /// Value of the item, used for sorting and knapsack algorithm. Should be much larger than `cost` to avoid ties.
+    fn value(&self) -> usize;
+    /// Tie breaker for the case when two candidates have the same cost and value. When a tie occurs, the item with higher value of this function is chosen.
```
leonardoalt, line 14: Shouldn't this have the other item for comparison then? I don't understand what exactly one would return here without a direct comparison

.github/workflows/nightly-tests.yml
```diff
@@ -166,7 +166,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: powdr-labs/openvm-reth-benchmark
-          ref: main
+          ref: 29dc4c3581acbc3818543bef360e17a5938ff19b
```
leonardoalt, line 169: reminder: change this and the other one back to `main`

autoprecompiles/src/blocks/selection.rs
Schaeff: It's a global ordering of APCs, in practice we use the opcode id.

autoprecompiles/src/blocks/mod.rs
Schaeff: Done

autoprecompiles/src/blocks/mod.rs
Schaeff: Removed

autoprecompiles/src/blocks/pgo.rs
Schaeff: fixed

openvm/src/customize_exe.rs
Schaeff: Yes, we don't want to pass all this ovm stuff to apc when the max column bound is all we need

cli-openvm/src/main.rs
Schaeff: Yes, it used to be using the Default trait iirc, now it's explicit

.github/workflows/pr-tests.yml
Schaeff: yes, Leo said order is reverting this, merging it and then update reth

.github/workflows/nightly-tests.yml
Schaeff: reverted

leonardoalt (approved):

PR #3039, opened 2025-07-11
Apc stats and columns saved tests

qwang98:
Key design choices revolve around passing `ApcCandidate` stats out from `create_apcs_with_cell_pgo`, `create_apcs_with_pgo`, etc. I explored a few ways:
1. [Current solution] Add an `ApcStats` struct that wraps stats we want to pass to `PowdrPrecompile`. I believe this is the most versatile solution and avoids defects from the two solutions below.
2. Return `ApcCandidate` for `create_apcs_with_[x]_pgo` and `create_apcs_with_pgo`. This solution requires filling dummy zeros for `execution_frequency`, `width_before`, and `width_after`, which I think aren't very clean. Besides, passing these values still require computing the stats we care about later, so I think it's cleaner to explicitly declare what we care about in `ApcStats` instead.
3. Use `Either::Left(ApcCandidate)` for cell PGO while `Either::Right(Apc)` for other PGO modes. This isn't very clean, and also has the disadvantage of requiring match clauses when constructing elements of `PowdrExtension`.

```
struct ApcCandidate<P> {
    apc: Apc<P>,
    execution_frequency: usize,
    width_before: AirWidths,
    width_after: AirWidths,
}
```


openvm/src/powdr_extension/vm.rs
```diff
@@ -76,6 +77,7 @@ pub struct PowdrPrecompile<P: IntoOpenVm> {
     pub machine: SymbolicMachine<P>,
     pub original_instructions: Vec<OriginalInstruction<OpenVmField<P>>>,
     pub is_valid_column: AlgebraicReference,
+    pub apc_stats: Option<ApcStats>,
```
qwang98: Also, we would need an `ApcStats` type as a wrapper for `PowdrPrecompile` anyways.

openvm/src/lib.rs
```diff
@@ -1337,14 +1349,27 @@ mod tests {
...
         assert_eq!(non_powdr_sum, *non_powdr_expected_sum);
 
+        // Test cells saved in Pgo::Cell
+        if is_cell_pgo {
+            let columns_saved = powdr_air_metrics
+                .into_iter()
+                .map(|(_, columns_saved)| columns_saved.unwrap())
+                .sum::<AirWidths>();
+            assert_eq!(columns_saved, params.expected_columns_saved.unwrap());
+            // println!("Columns saved\n: {:?}", columns_saved);
```
Schaeff: remove

openvm/src/customize_exe.rs
```diff
@@ -288,7 +292,7 @@ fn create_apcs_for_all_blocks<P: FieldElement>(
     blocks: Vec<BasicBlock<P>>,
     powdr_config: &PowdrConfig,
     vm_config: VmConfig<OriginalAirs<P>, OpenVmBusInteractionHandler<P>>,
-) -> Vec<Apc<P>> {
+) -> Vec<(Apc<P>, Option<ApcStats>)> {
```
Schaeff: This always returns None so let's keep it as it was before?

openvm/src/customize_exe.rs
```diff
@@ -398,7 +416,7 @@ impl ApcCandidate<BabyBearField> {
                     .get_instruction_metrics(instr.opcode)
                     .unwrap()
                     .widths
-                    .total()
+                    .clone()
```
Schaeff: Let's implement Copy on AirWidths and remove this clone

openvm/src/customize_exe.rs
```diff
@@ -471,7 +489,7 @@ fn create_apcs_with_cell_pgo(
     config: &PowdrConfig,
     original_config: &OriginalVmConfig,
     vm_config: VmConfig<OriginalAirs<BabyBearField>, OpenVmBusInteractionHandler<BabyBearField>>,
-) -> Vec<Apc<BabyBearField>> {
+) -> Vec<(Apc<BabyBearField>, Option<ApcStats>)> {
```
Schaeff: This always returns Some so let's unwrap the Option

openvm/src/customize_exe.rs
```diff
@@ -556,7 +578,7 @@ fn create_apcs_with_instruction_pgo<P: FieldElement>(
     pgo_program_idx_count: HashMap<u32, u32>,
     config: &PowdrConfig,
     vm_config: VmConfig<OriginalAirs<P>, OpenVmBusInteractionHandler<P>>,
-) -> Vec<Apc<P>> {
+) -> Vec<(Apc<P>, Option<ApcStats>)> {
```
Schaeff: Same here, this is always None, let's remove it

openvm/src/customize_exe.rs
```diff
@@ -595,7 +617,7 @@ fn create_apcs_with_no_pgo<P: IntoOpenVm>(
     mut blocks: Vec<BasicBlock<P>>,
     config: &PowdrConfig,
     vm_config: VmConfig<OriginalAirs<P>, OpenVmBusInteractionHandler<P>>,
-) -> Vec<Apc<P>> {
+) -> Vec<(Apc<P>, Option<ApcStats>)> {
```
Schaeff: Same here

openvm/src/customize_exe.rs
```diff
@@ -59,7 +59,7 @@ fn generate_apcs_with_pgo(
     original_config: &OriginalVmConfig,
     pgo_config: PgoConfig,
     vm_config: VmConfig<OriginalAirs<BabyBearField>, OpenVmBusInteractionHandler<BabyBearField>>,
-) -> Vec<Apc<BabyBearField>> {
+) -> Vec<(Apc<BabyBearField>, Option<ApcStats>)> {
```
Schaeff: This one does return Some or None so it should stay as is and the body should be fixed

openvm/src/customize_exe.rs
```diff
@@ -343,8 +350,19 @@ pub fn openvm_bus_interaction_to_powdr<F: PrimeField32, P: FieldElement>(
...
-    width_before: usize,
-    width_after: usize,
+    width_before: AirWidths,
+    width_after: AirWidths,
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct ApcStats {
+    pub columns_saved: AirWidths,
+}
```
Schaeff: Let's have a `AirWidthDiff` struct which has both before and after as fields, and have the columns saved as a method. Then it can be used everywhere we need both, such as in the ApcCandidate.

openvm/src/lib.rs
```diff
@@ -567,11 +567,19 @@ impl AirMetrics {
 #[cfg(test)]
 impl CompiledProgram {
     // Return a tuple of (powdr AirMetrics, non-powdr AirMetrics)
```
Schaeff, line 510: This comment is outdated

openvm/src/lib.rs
```diff
@@ -567,11 +567,19 @@ impl AirMetrics {
...
-    fn air_metrics(&self) -> (Vec<AirMetrics>, Vec<AirMetrics>) {
+    fn air_metrics(&self) -> (Vec<(AirMetrics, Option<AirWidths>)>, Vec<AirMetrics>) {
```
Schaeff: Why is it optional here? Could it be a single option instead of many?

openvm/src/lib.rs
```diff
@@ -1304,6 +1315,7 @@ mod tests {
         guest_apc: u64,
         guest_skip: u64,
         expected_metrics: &'a MachineTestMetrics,
+        expected_columns_saved: Option<AirWidths>, // only available in Pgo::Cell
```
Schaeff: Could use `Option<AirWidthDelta>` here

openvm/src/powdr_extension/vm.rs
Schaeff: What do you mean?

openvm/src/lib.rs
qwang98: Done.

openvm/src/powdr_extension/vm.rs
qwang98: Just saying that having an `ApcStats` field is cleaner than having the entire `Candidate` in `PowdrPrecompile`.

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs

openvm/src/customize_exe.rs

openvm/src/lib.rs
qwang98: I think it's still valid:

The return value is `(Vec<(AirMetrics, Option<AirWidthsDiff>)>, Vec<AirMetrics>)`, where
- `Vec<(AirMetrics, Option<AirWidthsDiff>)>` is Powdr (the Option is only avail in Cell mode)
-  `Vec<AirMetrics>` is Non-Powdr

openvm/src/lib.rs
qwang98: I think it should still be many options. `Vec<(AirMetrics, Option<AirWidths>)>` is basically a Vec of Apc stats, where the first tuple element are regular column/constraint counts while the second tuple element are optional columns saved produced by Cell PGO.

qwang98: Basically here we have a columns saved (`AirWidthsDiff`) stats for each Apc.

openvm/src/lib.rs
qwang98: Done.

openvm/src/customize_exe.rs

Schaeff (approved):

PR #3040, opened 2025-07-11
Trait based apc

Schaeff:
A more extreme attempt than #3036 based on a single `Adapter` trait.


openvm/src/instruction_formatter.rs
```diff
@@ -1,36 +1,42 @@
...
+use openvm_instructions::{instruction::Instruction, VmOpcode};
+use openvm_stark_backend::p3_field::PrimeField32;
 
-pub fn openvm_instruction_formatter<P: FieldElement>(
-    instruction: &SymbolicInstructionStatement<P>,
-) -> String {
-    let SymbolicInstructionStatement { opcode, args } = instruction;
-    let [a, b, c, d, e, f, g] = args.as_slice().try_into().unwrap();
-    let opcode_name = openvm_opcode_formatter(&VmOpcode::from_usize(*opcode));
+pub fn openvm_instruction_formatter<F: PrimeField32>(instruction: &Instruction<F>) -> String {
```
Schaeff, line 5: This is reverted back to what it was before #3036

qwang98: So is this change (that's reverted back) still necessary to make #3036 work?

autoprecompiles/src/adapter.rs
```diff
@@ -0,0 +1,26 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    blocks::{Candidate, Instruction, Program},
+    constraint_optimizer::IsBusStateful,
+    InstructionMachineHandler,
+};
+
+pub trait Adapter: Sized {
+    type Field;
```
qwang98, line 12: This is the OVM field?

autoprecompiles/src/adapter.rs
```diff
@@ -0,0 +1,26 @@
...
+    type PowdrField: FieldElement;
+    type InstructionMachineHandler: InstructionMachineHandler<Self::PowdrField> + Clone + Sync;
+    type BusInteractionHandler: BusInteractionHandler<Self::PowdrField>
+        + Clone
+        + IsBusStateful<Self::PowdrField>
+        + Sync;
+    type Candidate: Candidate<Self> + Send;
+    type Program: Program<Self::Field, Self::Instruction> + Send;
```
qwang98: Why we attach field to `Program` when `Instruction` is attached to `Program` and field is attached to `Instruction`?

qwang98: The `Program` trait definition itself doesn't use the field either.

autoprecompiles/src/blocks/mod.rs
```diff
@@ -70,17 +70,14 @@ impl PowdrConfig {
...
 }
 
-impl<T> BasicBlock<T> {
-    pub fn pretty_print(
-        &self,
-        instr_formatter: impl Fn(&SymbolicInstructionStatement<T>) -> String,
-    ) -> String {
+impl<I> BasicBlock<I> {
+    pub fn pretty_print(&self, instr_formatter: impl Fn(&I) -> String) -> String {
         format!("BasicBlock(start_idx: {}, statements: [\n", self.start_idx)
```
qwang98, line 27: This is probably a design choice question, but how do we determine what to generalize for the `Adapter`. For example, why don't we also generalize `BasicBlock` similar to `Program` and `Instruction`?

autoprecompiles/src/blocks/pgo.rs
```diff
@@ -45,36 +43,31 @@ impl PgoConfig {
...
     type JsonExport: Serialize + for<'de> Deserialize<'de> + Send;
 
     /// Try to create an autoprecompile candidate from a block.
     fn create(
-        apc: Apc<P>,
+        apc: Apc<A::PowdrField, A::Instruction>,
         pgo_program_idx_count: &HashMap<u32, u32>,
-        vm_config: VmConfig<I, B>,
+        vm_config: VmConfig<A::InstructionMachineHandler, A::BusInteractionHandler>,
     ) -> Self;
```
qwang98: Honestly this already looks much better ;)

autoprecompiles/src/blocks/pgo.rs
```diff
@@ -45,36 +43,31 @@ impl PgoConfig {
...
 
 // Note: This function can lead to OOM since it generates the apc for many blocks.
-fn create_apcs_with_cell_pgo<
-    C: Candidate<P, I, B> + Send,
-    P: FieldElement,
-    I: InstructionMachineHandler<P> + Clone + Sync,
-    B: BusInteractionHandler<P> + Clone + IsBusStateful<P> + Sync,
->(
-    mut blocks: Vec<BasicBlock<P>>,
+fn create_apcs_with_cell_pgo<A: Adapter>(
```
qwang98, line 67: Much cleaner.

autoprecompiles/src/lib.rs
```diff
@@ -257,40 +258,43 @@ pub trait InstructionMachineHandler<T> {
...
 
 #[derive(Debug, Serialize, Deserialize)]
-pub struct Apc<T> {
-    pub block: BasicBlock<T>,
+pub struct Apc<T, I> {
+    pub block: BasicBlock<I>,
     pub opcode: u32,
     pub machine: SymbolicMachine<T>,
     pub subs: Vec<Vec<u64>>,
 }
```
qwang98, line 303: Echoing the question about `Program`, I think attaching both `T` and `I` here at least makes sense, because they are actually needed as two separate abstractions across two struct fields.

autoprecompiles/src/lib.rs
```diff
@@ -257,40 +258,43 @@ pub trait InstructionMachineHandler<T> {
...
+    let statements: Vec<_> = block
+        .statements
+        .clone()
+        .into_iter()
+        .map(|instr| instr.into_symbolic_instruction())
+        .map(|instr| SymbolicInstructionStatement {
+            opcode: instr.opcode,
+            args: instr.args.into_iter().map(A::from_field).collect(),
+        })
+        .collect();
```
qwang98, line 331: So `statements` are original ones. I wonder where this conversion happened before if not in `build`?

autoprecompiles/src/blocks/mod.rs
```diff
@@ -70,17 +70,14 @@ impl PowdrConfig {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
-pub struct BasicBlock<T> {
+pub struct BasicBlock<I> {
     /// The index of the first instruction in this block in the original program.
     pub start_idx: usize,
-    pub statements: Vec<SymbolicInstructionStatement<T>>,
+    pub statements: Vec<I>,
```
qwang98, line 22: Previously these the `statements` are Powdr type already right? Now it looks like the original instructions, because `Adapter::Instruction` uses the original `Field`. I wonder why we made this design choice.

openvm/src/customize_exe.rs
```diff
@@ -50,6 +49,74 @@ impl From<powdr_autoprecompiles::constraint_optimizer::Error> for Error {
     }
 }
 
+/// An adapter for the BabyBear OpenVM precompiles.
+/// Note: This could be made generic over the field, but the implementation of `Candidate` is BabyBear-specific.
```
qwang98, line 55: Probably another PR, but does `Candidate` implementation have to be BabyBear specific? My understanding is that the BabyBear specific part are width stuffs, which I believe can be make generic eventually? Probably will involve quite deep refactorings though.

openvm/src/customize_exe.rs
```diff
@@ -50,6 +49,74 @@ impl From<powdr_autoprecompiles::constraint_optimizer::Error> for Error {
...
+}
+
+impl<'a> Adapter for BabyBearOpenVmApcAdapter<'a> {
+    type PowdrField = BabyBearField;
+    type Field = OpenVmField<BabyBearField>;
+    type InstructionMachineHandler = OriginalAirs<BabyBearField>;
+    type BusInteractionHandler = OpenVmBusInteractionHandler<BabyBearField>;
+    type Candidate = OpenVmApcCandidate<BabyBearField, Instr<OpenVmField<BabyBearField>>>;
+    type Program = Prog<'a, OpenVmField<BabyBearField>>;
+    type Instruction = Instr<OpenVmField<BabyBearField>>;
```
qwang98: This is great!

openvm/src/customize_exe.rs
```diff
@@ -50,6 +49,74 @@ impl From<powdr_autoprecompiles::constraint_optimizer::Error> for Error {
...
+    }
+
+    fn from_field(e: Self::Field) -> BabyBearField {
+        BabyBearField::from_openvm_field(e)
+    }
+}
+
+/// A newtype wrapper around `OpenVmProgram` to implement the `Program` trait.
+/// This is necessary because we cannot implement a foreign trait for a foreign type.
+pub struct Prog<'a, F>(&'a OpenVmProgram<F>);
```
qwang98, line 81: Probably dummy question but `Program` trait is in this crate so it's not foreign?

openvm/src/customize_exe.rs
```diff
@@ -73,24 +140,7 @@ pub fn customize(
...
-                ]
-                .iter()
-                .map(|f| BabyBearField::from_openvm_field(*f))
-                .collect(),
-            })
-            .collect_vec(),
-        exe.program.pc_base,
-        exe.program.step,
-    );
+    let program = Prog(&exe.program);
```
qwang98, line 146: Ok I think I'm answering my own questions again, but looks like instead of converting all program instructions at the very start of customize, we now delay it to when actually building the Apc, which means that some instructions that aren't used for Apc are no longer converted and we get some savings.

Is this the correct understanding?

openvm/src/customize_exe.rs
```diff
@@ -284,22 +339,25 @@ pub fn openvm_bus_interaction_to_powdr<F: PrimeField32, P: FieldElement>(
...
+    apc: Apc<P, I>,
     execution_frequency: usize,
     width_before: usize,
     width_after: usize,
 }
 
-impl<B: BusInteractionHandler<BabyBearField> + Clone + Sync + IsBusStateful<BabyBearField>>
-    Candidate<BabyBearField, OriginalAirs<BabyBearField>, B> for OpenVmApcCandidate<BabyBearField>
+impl<'a> Candidate<BabyBearOpenVmApcAdapter<'a>>
+    for OpenVmApcCandidate<BabyBearField, Instr<OpenVmField<BabyBearField>>>
```
qwang98, line 374: Much better ;)

openvm/src/customize_exe.rs
```diff
@@ -284,22 +339,25 @@ pub fn openvm_bus_interaction_to_powdr<F: PrimeField32, P: FieldElement>(
...
 
     fn create(
-        apc: Apc<BabyBearField>,
+        apc: Apc<BabyBearField, Instr<OpenVmField<BabyBearField>>>,
         pgo_program_idx_count: &HashMap<u32, u32>,
-        vm_config: VmConfig<OriginalAirs<BabyBearField>, B>,
+        vm_config: VmConfig<
+            OriginalAirs<BabyBearField>,
+            OpenVmBusInteractionHandler<BabyBearField>,
+        >,
```
qwang98, line 385: Can we somehow make these use the adapter as well? e.g. `BabyBearOpenVmApcAdapter::BusInteractionHandler` and etc.?

openvm/src/customize_exe.rs
```diff
@@ -348,19 +406,19 @@ impl<B: BusInteractionHandler<BabyBearField> + Clone + Sync + IsBusStateful<Baby
         }
     }
 
-    fn into_apc(self) -> Apc<BabyBearField> {
+    fn into_apc(self) -> Apc<BabyBearField, Instr<OpenVmField<BabyBearField>>> {
         self.apc
```
qwang98: Use the adapter as well?

autoprecompiles/src/blocks/mod.rs
qwang98: How about making `Apc` part of the adapter as well ;)

openvm/src/customize_exe.rs
```diff
@@ -369,14 +427,14 @@ struct OpenVmApcCandidateJsonExport<P> {
...
-impl<P> OpenVmApcCandidate<P> {
+impl<P, A> OpenVmApcCandidate<P, A> {
     fn cells_saved_per_row(&self) -> usize {
         // The number of cells saved per row is the difference between the width before and after the APC.
         self.width_before - self.width_after
     }
 }
 
-impl<P> KnapsackItem for OpenVmApcCandidate<P> {
+impl<P, A> KnapsackItem for OpenVmApcCandidate<P, A> {
```
qwang98: Can we make these `<P, I>` to be somewhat consistent?

openvm/src/symbolic_instruction_builder.rs
```diff
@@ -1,7 +1,7 @@
 //! Builds SymbolicInstructionStatement to create input program for testing powdr_autoprecompile::build
 use crate::opcode::*;
-use powdr_autoprecompiles::SymbolicInstructionStatement;
-use powdr_number::FieldElement;
+use openvm_instructions::{instruction::Instruction, VmOpcode};
+use openvm_stark_backend::p3_field::PrimeField32;
```
qwang98, line 4: Makes sense that this is now specific to OVM rather than mixing up `SymbolicInstruction` (Powdr) with OVM.

qwang98 (changes requested):
This looks much cleaner and I think we should definitely merge eventually!

Left many questions/comments.

openvm/src/instruction_formatter.rs
Schaeff: Yeah, in #3036 we change the instructions to being `SymbolicInstruction` to move them into the apc crate, so the formatter needs to work on P. Here we allow the apc crate to work with instructions over the original field, so we're back to the original formatter.

autoprecompiles/src/adapter.rs
Schaeff: Correct

autoprecompiles/src/adapter.rs
Schaeff: Good catch!

openvm/src/customize_exe.rs
Schaeff: Done, bit more verbose but sure.

openvm/src/customize_exe.rs
Schaeff: No the Program trait is in the autoprecompile crate

openvm/src/customize_exe.rs
Schaeff: The main motivation here is to remove code from customize and move it to the apc crate. Performance is probably irrelevant. We basically replace `foo(y: Y) { // use y }` by `foo<X: Into<Y>>(x: X) { y = x.into(); // use y}`

openvm/src/customize_exe.rs
Schaeff: Yes, could be made generic by making all the `get_constraints` etc stuff generic over the field and stark config, now it's generic over neither.

autoprecompiles/src/blocks/mod.rs
Schaeff: Yeah good question, here I tried to do the minimal stuff needed.
Apc doesn't need to be added in my opinion: when it gets returned, we turn it into a client specific precompile which is a different process than Apc generation.

autoprecompiles/src/lib.rs
Schaeff: We used to convert the whole program into symbolic instructions in customize

autoprecompiles/src/lib.rs
Schaeff: true

autoprecompiles/src/blocks/pgo.rs
Schaeff: I think so too!

openvm/src/customize_exe.rs
Schaeff: I tried but it's quite verbose

openvm/src/customize_exe.rs
Schaeff: It's quite verbose as well. Hopefully we can simplify this later.

autoprecompiles/src/blocks/mod.rs
Schaeff: ovm needs the instructions in its own format so that it can execute them. Now the APC generation returns them already in that format, as opposed to as `SymbolicInstructionStatement`s

openvm/src/customize_exe.rs
```diff
@@ -50,6 +49,74 @@ impl From<powdr_autoprecompiles::constraint_optimizer::Error> for Error {
...
+pub struct BabyBearOpenVmApcAdapter<'a> {
+    _marker: std::marker::PhantomData<&'a ()>,
+}
+
+impl<'a> Adapter for BabyBearOpenVmApcAdapter<'a> {
+    type PowdrField = BabyBearField;
+    type Field = OpenVmField<BabyBearField>;
+    type InstructionMachineHandler = OriginalAirs<BabyBearField>;
+    type BusInteractionHandler = OpenVmBusInteractionHandler<BabyBearField>;
+    type Candidate = OpenVmApcCandidate<BabyBearField, Instr<OpenVmField<BabyBearField>>>;
```
leonardoalt:
```suggestion
    type Candidate = OpenVmApcCandidate<BabyBearField, Instr<Self::Field>>;
```

Would it make sense to further enforce this?

Schaeff: Replaced where applicable

autoprecompiles/src/adapter.rs
qwang98: > Why we attach field to `Program` when `Instruction` is attached to `Program` and field is attached to `Instruction`?

Have we attempted fixing this?

qwang98 (changes requested):
I think all fixes where we've agree on are done except this one: https://github.com/powdr-labs/powdr/pull/3040#discussion_r2204204516

leonardoalt:
has conflicts

PR #3054, opened 2025-07-17
Partially remove opcode concept from apc crate, use `VmOpcode` in openvm crate

Schaeff:
This PR does two things:
- it partially removes the notion of opcode in the autoprecompile crate, and instead relies on user provided functions to know if an instruction is allowed/branching
- it uses VmOpcode as much as possible in the openvm crate instead of usize

After this PR, the only remaining notion of opcode in the apc crate is in the `SymbolicInstructionStatement` where it is still a single field element. @georgwiese shared ideas to remove that too, or at least shift it to the client, see #3055

### DIFF
--- a/autoprecompiles/src/lib.rs
+++ b/autoprecompiles/src/lib.rs
@@ -288,7 +288,7 @@ impl<T: FieldElement> PcLookupBusInteraction<T> {
 /// A configuration of a VM in which execution is happening.
 pub struct VmConfig<'a, M, B> {
     /// Maps an opcode to its AIR.
-    pub instruction_machine_handler: &'a M,
+    pub instruction_handler: &'a M,
     /// The bus interaction handler, used by the constraint solver to reason about bus interactions.
     pub bus_interaction_handler: B,
     /// The bus map that maps bus id to bus type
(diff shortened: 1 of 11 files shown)


autoprecompiles/src/lib.rs
```diff
@@ -306,9 +306,15 @@ impl<'a, M, B: Clone> Clone for VmConfig<'a, M, B> {
     }
 }
 
-pub trait InstructionMachineHandler<T, I> {
+pub trait InstructionHandler<T, I> {
     /// Returns the AIR for the given opcode.
     fn get_instruction_air(&self, instruction: &I) -> Option<&SymbolicMachine<T>>;
+
```
Schaeff, line 312: I added this to this trait, technically it could be a separate trait but I tried to keep it simple

openvm/src/customize_exe.rs
```diff
@@ -205,15 +194,6 @@ pub fn customize(
         }
     }
 
-    let blocks = blocks
```
Schaeff, line 208: I think this is already guaranteed by the block extraction

PR #3056, opened 2025-07-18
Support 64 bit elf

qwang98:
WIP

### DIFF
--- a/riscv-elf/src/lib.rs
+++ b/riscv-elf/src/lib.rs
@@ -7,9 +7,9 @@ use std::{
 };
 
 use goblin::elf::{
-    header::{EI_CLASS, EI_DATA, ELFCLASS32, ELFDATA2LSB, EM_RISCV, ET_DYN},
+    header::{EI_CLASS, EI_DATA, ELFCLASS32, ELFCLASS64, ELFDATA2LSB, EM_RISCV, ET_DYN},
     program_header::PT_LOAD,
-    reloc::{R_RISCV_32, R_RISCV_HI20, R_RISCV_RELATIVE},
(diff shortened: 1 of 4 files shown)


riscv-elf/src/lib.rs
```diff
@@ -50,10 +50,10 @@ pub fn load_elf(file_name: &Path) -> ElfProgram {
...
     let elf = Elf::parse(file_buffer).unwrap();
 
-    // Assert the file is 32 bits.
+    // Assert the file is 64 bits.
     assert_eq!(
-        elf.header.e_ident[EI_CLASS], ELFCLASS32,
-        "Only 32-bit ELF files are supported!"
+        elf.header.e_ident[EI_CLASS], ELFCLASS64,
+        "Only 64-bit ELF files are supported!"
     );
```
qwang98, line 57: This passes on fibo example

riscv-elf/src/lib.rs
```diff
@@ -242,21 +241,21 @@ fn pie_relocate_data_sections(
...
+        let addr = r.r_offset;
         if !address_map.is_in_data_section(addr) {
             // Relocation of the text section has already been handled in instruction lifting.
             continue;
         }
 
-        // We only support the R_RISCV_32 relocation type for the data section:
-        assert_eq!(r.r_type, R_RISCV_32, "Unsupported relocation type!");
+        // We only support the R_RISCV_64 relocation type for the data section:
+        assert_eq!(r.r_type, R_RISCV_64, "Unsupported relocation type!");
```
qwang98, line 256: This doesn't pass on fibo example, which is quite unexpected. `r.r_type` should be directly parsed from `goblin::Elf` bytes, so I'm not sure why the elf's class is RV64 but relocation type isn't...

PR #3057, opened 2025-07-18
Move pc base and step to `Adapter` state

Schaeff:
In #3055 we need to access the pc step and base from within `build`.
This PR adds them to the Adapter trait and passes an instance of `A: Adapter` so that they are available everywhere.
As a result, an Adapter is now program-specific, but this seems fine.

### DIFF
--- a/autoprecompiles/src/lib.rs
+++ b/autoprecompiles/src/lib.rs
@@ -336,6 +336,7 @@ impl<T, I> Apc<T, I> {
 }
 
 pub fn build<A: Adapter>(
+    _adapter: &A,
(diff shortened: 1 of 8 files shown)


autoprecompiles/src/lib.rs
```diff
@@ -336,6 +336,7 @@ impl<T, I> Apc<T, I> {
 }
 
 pub fn build<A: Adapter>(
+    _adapter: &A,
```
Schaeff, line 339: To be used in #3055

Schaeff:
Ok good point, if it's enough to store the pc address in the block I'm fine with that.

